PR #11, opened 2026-01-24
chore: import node process explicitly

iiio2:
I think in nuxt, we've explicitly import node:process. That is why I thought we can do it here maybe.


danielroe:
looks good! can you see if you can configure oxlint to require this? 🙏

iiio2:
@danielroe , Yes, I think i've done it.

danielroe:
thank you

iiio2:
Thanks. :heart:

PR #12, opened 2026-01-24
chore: fix spacing between packages

iiio2:
I've used ai (claude pro) to fix it. previously it looks like yarnadd, denoadd, bunadd like that.


danielroe:
I think i already fixed this

danielroe:
can you hard reload and add a screenshot?

iiio2:
Actaully I've seen like that. Suppose you come to the page. the default is npm. Now you switch to pnpm. Then come back to npm, then you see npminstall. also i have seen yarnadd, bunadd like that. Let me check again this.

iiio2:
@danielroe , Yes. It is actually fixed. Thanks.

PR #15, opened 2026-01-24
fix: show directory listing instead of error when opening folder path

azat-io:
When you open a directory while viewing code using breadcrumbs, you get an error.

Before:

image

After:

image


danielroe:
thank you! ❤️

PR #17, opened 2026-01-24
chore: Add warning message on connector modal

neutrino2211:
This adds a little warning to the connector modal. Makes it clear to the user that performing the action will give npmx access to the npmcli and connected contexts as well.

Screenshot 2026-01-24 at 14 15 08

Messaging might need some work though.


danielroe:
this looks great! we should also add a similar message that the connect prints out when starting up

PR #20, opened 2026-01-24
chore: Add warning message from #17 to the cli as well

neutrino2211:
Added the warning message in the UI to the cli logs. Currently done in separate lines.

image

### DIFF
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -2,6 +2,7 @@
 import process from 'node:process'
 import { spawn } from 'node:child_process'
 import * as p from '@clack/prompts'
+import pc from 'picocolors'
 import { defineCommand, runMain } from 'citty'
 import { listen } from 'listhen'
 import { toNodeListener } from 'h3'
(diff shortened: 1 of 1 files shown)


cli/src/cli.ts
```diff
@@ -51,6 +51,7 @@ const main = defineCommand({
 
     if (!npmUser) {
       logWarning('Not logged in to npm. Starting npm login...')
+      logWarning('This allows npmx to access your npm cli and any authenticated contexts.')
```
danielroe: we should display the warning even if user is already logged into npm

neutrino2211: Okay. Should the message be before or after the "Checking npm authentication..." Message?

danielroe: maybe it should be on startup, and require you to confirm you still want to proceed?

neutrino2211: That makes sense

neutrino2211:
It now
- Shows the message on startup
- Exits on selecting 'no'/cancelling
- Shows an exit message using logError <- Is this okay?

PR #25, opened 2026-01-24
fix: wrap package node engines to next line

jhroemer:
So I was looking at fixing the node version layout as a simple contribution: 

Screenshot 2026-01-24 at 17 48 45

But I realize you already fixed it in the meantime by truncating :-)
Anyway - I don't thinking wrapping to two lines is that bad of an idea, if you make sure the div doesn't center align, and right-align the packages text. Truncating can be a little bit annoying sometimes. Just a proposal of course.

Screenshot 2026-01-24 at 21 38 17


danielroe:
thank you! ❤️

PR #26, opened 2026-01-24
chore: improve noise during tests

devdumpling:
## Overview

I've been spending some time trying to better understand the repo and like to start by running tests to get a sense of where testing is at. 

However, I was hitting quite a bit of noise, and noticed it seems to be mostly two issues:

### double `sharp` dep

We have two versions of sharp being installed by dependencies, which is then causing `objc` to complain down the line. I'd think it's safe to resolve this by overriding the `sharp` dep. It's only a single minor diff between the two if you look at `pnpm why sharp -r`. Doing this both resolves the issue and cleans up the double dep. 

Additionally, I've pinned the `onlyBuiltDeps` to specific versions. This is a recent pnpm feature that I've been implementing at work since there have been a lot of accidental builds (e.g. shai hulud). `pnpm` already catches a lot, but this makes it even more explicit. 

### vitest / rolldown unable to parse issues

Currently if you run `pnpm test:unit` you'll get a slew of noise around parsing errors. I believe what's happening is that they're trying to parse coverage for some top level async await patterns in specific files. This might be fixed by moving from `v8` as the provider to `istanbul`, but didn't want to shake it up too much. 

I thought for now it might be reasonable to simply ignore the offending files from coverage to limit the noise, but if people feel this loss of visibility is dangerous (it only affects coverage) or illusory, then we can keep it off. 

## Before / After

This is what you see before (note all the parsing noise that makes TDD hard if you're into that)

https://github.com/user-attachments/assets/223e51ff-93f6-4809-89a2-bc417bd604ca

aaaaand after!

https://github.com/user-attachments/assets/fc90881f-9973-44e3-83ae-5c5f2ac3fe9f

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -40,6 +40,18 @@ export default defineConfig({
     coverage: {
       enabled: true,
       provider: 'v8',
+      // Exclude files that cause parse errors during coverage remapping.
+      // The V8 coverage provider uses rolldown to parse source files, but
+      // rolldown seems to currently fail on Vite's SSR transform output (`await __vite_ssr_import__`).
+      exclude: [
+        '**/node_modules/**',
+        'cli/**',
+        'server/utils/readme.ts',
+        'server/utils/code-highlight.ts',
+        'server/utils/npm.ts',
+        'server/utils/shiki.ts',
+        'server/utils/jsr.ts',
+      ],
     },
   },
 })
--- a/package.json
+++ b/package.json
@@ -78,6 +78,9 @@
     "vite": "^8.0.0-beta.10"
   },
   "pnpm": {
+    "overrides": {
+      "sharp": "0.34.5"
+    },
     "packageExtensions": {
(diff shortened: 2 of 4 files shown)


vitest.config.ts
```diff
@@ -40,6 +40,18 @@ export default defineConfig({
...
+      // rolldown seems to currently fail on Vite's SSR transform output (`await __vite_ssr_import__`).
+      exclude: [
+        '**/node_modules/**',
+        'cli/**',
+        'server/utils/readme.ts',
+        'server/utils/code-highlight.ts',
+        'server/utils/npm.ts',
+        'server/utils/shiki.ts',
+        'server/utils/jsr.ts',
+      ],
```
AriPerkkio, line 54: Sorry to ping on closed PR @devdumpling - happy to discuss this else where too.

How do I reproduce these errors? Even after removing the `exclude` array here I don't see any parsing errors 🤔

PR #56, opened 2026-01-25
chore: redirect chat.npmx.dev

patak-cat:
@danielroe I think this should work, but maybe you want to configure in another way.


danielroe:
it worked!

PR #57, opened 2026-01-25
Fix: updated column is balanced with package links

jhroemer:
This PR balances the license/downloads/deps etc. row somewhat with the package links row beneath it, by right-aligning the last column. Also, currently if there are no dependencies, the deps column is hidden, causing `install size` and `updated` to shift a position to the left. I think it's worth showing `deps` even if the value is 0 - a user might be looking for the section, not finding it, and not realizing that this is because the package has none. Showing it, even if the value is 0, makes this very clear. The npmgraph link should probably be hidden though.

**Before**:
Screenshot 2026-01-25 at 08 13 31

Screenshot 2026-01-25 at 08 13 03

---

**After**: 
Screenshot 2026-01-25 at 08 12 30

Again, just a proposal :-)

### DIFF
--- a/app/pages/[...package].vue
+++ b/app/pages/[...package].vue
@@ -407,11 +407,12 @@ defineOgImageComponent('Package', {
             </dd>
           </div>
 
-          <div v-if="getDependencyCount(displayVersion) > 0" class="space-y-1">
+          <div class="space-y-1">
             <dt class="text-xs text-fg-subtle uppercase tracking-wider">Deps</dt>
             <dd class="font-mono text-sm text-fg flex items-baseline justify-start gap-2">
               {{ getDependencyCount(displayVersion) }}
               <a
+                v-if="getDependencyCount(displayVersion) > 0"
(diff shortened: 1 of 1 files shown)


app/pages/[...package].vue
```diff
@@ -407,11 +407,12 @@ defineOgImageComponent('Package', {
             </dd>
           </div>
 
-          <div v-if="getDependencyCount(displayVersion) > 0" class="space-y-1">
+          <div class="space-y-1">
```
Comment (jhroemer) [inline app/pages/[...package].vue:410]: Arguably this could be its own change, as a precursor for right-aligning the last column (right aligning without showing deps looks pretty bad). Happy to break it up into separate PR's if you want.

jhroemer:
Maybe this is a dumb idea, but we should almost add some kind of extra highlighting in the `deps` column when there's none. It's a good thing!

danielroe:
oops - need to fix mobile alignment

image

jhroemer:
> oops - need to fix mobile alignment
> 
> image

Ah shit, that’s my bad, sorry!

PR #66, opened 2026-01-25
chore: drop for power users

patak-cat:
Removed 'for power users' from the description.


README.md
```diff
@@ -1,6 +1,6 @@
 # npmx.dev
 
-> A fast, modern browser for the npm registry for power users.
+> A fast, modern browser for the npm registry.
```
danielroe, line 3: we mention it in a couple of other places too

patak-cat: ah, let me check

patak-cat: done, consolidated also the message in the manifest

Issue #59, opened 2026-01-25
scroll to top button


danielroe:
on mobile for long readmes we should make it easier to get back to the top of the page

Flo0806:
This thing was already in my ticket :-) https://github.com/npmx-dev/npmx.dev/pull/69
+ 2 other things - I overlooked that here @danielroe 
FYI because merge conflicts

PR #70, opened 2026-01-25
chore: scroll to top

iiio2:
resolves https://github.com/npmx-dev/npmx.dev/issues/59

### DIFF
--- a/app/components/ScrollToTop.vue
+++ b/app/components/ScrollToTop.vue
@@ -1,0 +1,43 @@
+<script setup lang="ts">
+// Track scroll position
+const showButton = ref(false)
+
+// Show button after scrolling down 300px
+function handleScroll() {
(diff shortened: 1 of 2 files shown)


rshigg:
I think it would be good to use [CSS scrollable queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Conditional_rules/Container_scroll-state_queries#using_scrollable_queries) for this with javascript as a fallback. That way once the feature becomes baseline we can get rid of the js entirely.

danielroe:
this and also https://github.com/npmx-dev/npmx.dev/pull/71 overlap with https://github.com/npmx-dev/npmx.dev/pull/69 (which came first by just a few minutes!) so I'm merging it in with credit to you 🙏 

also added support for scrollable queries if there's browser support - but would welcome a review from @rshigg if you want 🙏

PR #73, opened 2026-01-25
feat: add link to node modules inspector

### LINKED ISSUE #52: add link to node-modules-inspector
evaluate linking out or integrating for deep dependency exploration

Gugustinette:
Resolves #52 by adding a link to [Node Modules Inspector](https://node-modules.dev/) on packages.

Not sure if that's the best icon we could get from available sets, but a classic "file tree" icon might conflict with other part of the UI, which displays dependencies etc...

Also considering issues with having too much links displayed ? Maybe we could find another way to display all of these.

### DIFF
--- a/app/pages/[...package].vue
+++ b/app/pages/[...package].vue
@@ -154,6 +154,8 @@ const repositoryUrl = computed(() => {
   return url
 })
 
+const { meta: repoMeta, stars, forks, forksLink } = useRepoMeta(repositoryUrl)
+
 const homepageUrl = computed(() => {
   return displayVersion.value?.homepage ?? null
 })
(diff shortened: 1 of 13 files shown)


danielroe:
in the interests of keeping it tiday, I think this could go next to the 'dependencies' heading in the right pane, with no icon

so:
```
Dependencies         Inspect
```

app/pages/[...package].vue
```diff
@@ -540,6 +540,17 @@ defineOgImageComponent('Package', {
                 socket.dev
               </a>
             </li>
+            <li class="flex-grow">
+              <a
+                :href="`https://node-modules.dev/grid/depth#install=${pkg.name}${displayVersion?.version ? `@${displayVersion.version}` : ''}`"
+                target="_blank"
+                rel="noopener noreferrer"
+                class="link-subtle font-mono text-sm inline-flex items-center gap-1.5"
+              >
```
Comment (danielroe) [inline app/pages/[...package].vue]: maybe:
```suggestion
              title="Inspect dependency tree on node-modules.dev"
              >
```

serhalp:
Works great, @Gugustinette!

> in the interests of keeping it tiday, I think this could go next to the 'dependencies' heading in the right pane, with no icon

Screenshot 2026-01-25 at 12 18 33

@danielroe alternatively or in addition, it could make sense as a second icon here? https://npmgraph.js.org/ + node-modules.dev (+ any future ones we may add... though maybe we should keep the tool selection opinionated)

Gugustinette:
So here ?
Capture d’écran 2026-01-25 à 18 37 26

Here ?
Capture d’écran 2026-01-25 à 18 37 46

Or both ?

danielroe:
I was thinking the second one, but I think @serhalp is right - maybe a second icon next to the module graph

Gugustinette:
Ye I also like that second icon solution.

PR #75, opened 2026-01-25
fix: layout shift during code nav from root to any file/dir

devdumpling:
## Overview

Simple layout shift fix for breadcrumbs. Slightly less visual pleasing for root, but stabilizes. If we'd prefer, we could space it instead but I think the `root` breadcrumb is consistent. Just my 2c. 

https://github.com/user-attachments/assets/39ea72b7-9134-4b62-bb7e-5b3f69bcd8af


devdumpling:
PSA this is the first vue i've like basically ever written.

app/pages/code/[...path].vue
```diff
@@ -339,16 +339,17 @@ useSeoMeta({
...
           class="flex items-center gap-1 font-mono text-sm overflow-x-auto"
         >
           <NuxtLink
+            v-if="filePath"
             :to="getCodeUrl()"
             class="text-fg-muted hover:text-fg transition-colors shrink-0"
           >
             root
           </NuxtLink>
+          <span v-else class="text-fg shrink-0">root</span>
```
Comment (devdumpling) [inline app/pages/code/[...path].vue:352]: Idea here was to only have the link outside of the root, but can simplify if this is overly complex for breadcrumb linkage.

PR #85, opened 2026-01-25
Add lighthouse accessibility testing job to CI

rshigg:
## Summary

Adds automated Lighthouse accessibility testing on PR push/merge. For now it will only produce warnings. 

If we want to see the LH job status in the PRs, we would need the Lighthouse CI Github app installed: https://github.com/apps/lighthouse-ci

## Changes
- Create lighthouse ci config
- Add CI job that sets up pnpm, builds the app, then runs lhci


danielroe:
I've now installed the LH GH app - so we probably can use the action without needing to install the cli within the project

rshigg:
> I've now installed the LH GH app - so we probably can use the action without needing to install the cli within the project

Thank you! The only reason I can think of to install the CLI now is if someone wants to create a report locally. But you can always use `npx @lhci/cli` in that case. I'll remove the dep

.github/workflows/ci.yml
```diff
@@ -73,3 +73,23 @@ jobs:
...
+          cache: pnpm
+
+      - name: 📦 Install dependencies
+        run: pnpm install
+
+      - name: 🏗️ Build project
+        run: pnpm build
+
+      - name: ♿ Accessibility audit (Lighthouse)
+        run: pnpm lhci
```
danielroe: shouldn't we use the lhci action which has this preinstalled?

danielroe: I've used this workflow in the past to trigger based on real deployments:

```yml
  lighthouse:
    environment:
      name: ${{ github.event.deployment.environment || 'Production' }}
      url: ${{ github.event.inputs.url || github.event.deployment_status.environment_url }}
    if: github.event.deployment_status.state == 'success' || github.event_name == 'workflow_dispatch'
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v6
      - name: Audit URLs using Lighthouse
        uses: treosh/lighthouse-ci-action@v12
        with:
          urls: ${{ github.event.inputs.url || github.event.deployment_status.environment_url }}
          uploadArtifacts: true
          temporaryPublicStorage: true
        env:
          LHCI_GITHUB_APP_TOKEN: ${{ secrets.LHCI_GITHUB_APP_TOKEN }}
```

rshigg: Sorry forgot to update the workflow. It should work as expected now 👍

PR #90, opened 2026-01-26
fix: pass expected `@scope:team` format to backend

serhalp:
### Bug

1. `pnpm dev`
2. `pnpm npmx-connector`
3. navigate to `http://localhost:3001/org/<some-org>` (or directly to destination `http://localhost:3001/@org`), where `<some-org>` is an org your logged in user is a part of
4. page hangs for ~30s, eventually loads and the `npmx-connector` prints an error

Screenshot 2026-01-25 at 19 07 17

This is really two separate fixes:

### fix(cli): fail gracefully on invalid `scope:team` format

The `/team/:scopeTeam/users` endpoint was letting validation errors from `validateScopeTeam` bubble up as an unhandled 500 response with an unhandled rejection logged to the console. Now it catches and returns a proper 400 with a helpful message, and logs the full error.

### fix(app): add `@` prefix when constructing `scope:team` for npm operations

npm team/access operations require `@scope:team` format but the frontend was sending `scope:team`
without the `@` prefix.

This was obscured by the fact that the admin operations aren't quite functional yet for most orgs as OTP isn't supported, and by the fact that `listTeamUsers` only fails with this incorrect format when it goes through the CLI, which is only the case for the authenticated user's own orgs.

### DIFF
--- a/app/utils/npm.ts
+++ b/app/utils/npm.ts
@@ -6,6 +6,6 @@
  * @param teamName - The team name
  * @returns The scope:team string in @scope:team format
  */
-export function buildScopeTeam(orgName: string, teamName: string): `@{string}:{string}` {
+export function buildScopeTeam(orgName: string, teamName: string): `@${string}:${string}` {
   return `@${orgName.replace(/^@/, '')}:${teamName}`
 }


cli/src/logger.ts
```diff
@@ -47,6 +48,11 @@ export function logInfo(message: string): void {
   p.log.info(message)
 }
 
+/**
+ * Log a debug message with `obug` (minimal fork of `debug`)
+ */
+export const logDebug = createDebug('npmx-connector')
```
serhalp, line 54: I might be missing something, but debugging was difficult without stack traces and support for non-string objects in general. Given that clack doesn't have affordances for this, debug/obug was the least intrusive approach I could think of.

cli/src/server.ts
```diff
@@ -447,6 +449,17 @@ export function createConnectorApp(expectedToken: string) {
...
 
+    try {
+      validateScopeTeam(scopeTeam)
+    } catch (err) {
+      logError('scope:team validation failed')
+      logDebug(err, { scopeTeamRaw, scopeTeam })
+      throw new HTTPError({
+        statusCode: 400,
+        message: `Invalid scope:team format: ${scopeTeam}. Expected @scope:team`,
+      })
```
serhalp, line 460: This may look a little funny at a glance, but the `logError` is the backend-user-facing pretty clack error output, the `logDebug` is for a developer troubleshooting, and the `throw` is for the response back to the browser for this invalid request.

app/utils/npm.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Constructs a scope:team string in the format expected by npm.
+ * npm operations require the format @scope:team (with @ prefix).
+ *
+ * @param orgName - The organization name (without @)
+ * @param teamName - The team name
+ * @returns The scope:team string in @scope:team format
+ */
+export function buildScopeTeam(orgName: string, teamName: string): string {
```
mxdvl: note: if you like narrower types you could do this…

```suggestion
export function buildScopeTeam(orgName: string, teamName: string): `@{string}:{string}` {
```

cli/src/server.ts
mxdvl: Would it help putting quotes around the invalid input for clarity?

app/utils/npm.ts
mxdvl: thanks for fixing in 361adc3b7a902b4eb10d25e61f8207ae0d2210a6 @danielroe

PR #92, opened 2026-01-26
fix: update org and package container layout

IHIutch:
Update the layouts of the org and package views to be full width, regardless of their contents' width

Also fixes the alignment of the search and sort carent icons

Before:
Layout contents width is constrained to a resembling a mobile layout, even on desktop

After:
Layout contents grow to fill the container element


IHIutch:
Hm, looking at the `unocss.config.ts`, maybe the `.container` shortcut should be updated directly?

danielroe:
this looks so much nicer!

yes, we could try updating the container class directly - does it work just as well?

danielroe:
merging to iterate but feel free to come back to updating the `.container` class

PR #96, opened 2026-01-26
feat: Add binary run scripts to package page

vinnymac:
Adds binary run scripts to package page.

- No binary, no change
- Single binary, similar UI to install
- Many binaries, press maximize to see all options (open to recs 🤗)
- Auto updates with package manager selection

### DIFF
--- a/app/pages/[...package].vue
+++ b/app/pages/[...package].vue
@@ -838,11 +838,11 @@ defineOgImageComponent('Package', {
         </div>
         <div class="relative group">
           <!-- Terminal-style execute command -->
-          <div class="bg-[#0d0d0d] border border-border rounded-lg overflow-hidden">
+          <div class="bg-bg-subtle border border-border rounded-lg overflow-hidden">
             <div class="flex gap-1.5 px-3 pt-2 sm:px-4 sm:pt-3">
-              <span class="w-2.5 h-2.5 rounded-full bg-[#333]" />
-              <span class="w-2.5 h-2.5 rounded-full bg-[#333]" />
-              <span class="w-2.5 h-2.5 rounded-full bg-[#333]" />
+              <span class="w-2.5 h-2.5 rounded-full bg-fg-subtle" />
+              <span class="w-2.5 h-2.5 rounded-full bg-fg-subtle" />
+              <span class="w-2.5 h-2.5 rounded-full bg-fg-subtle" />
             </div>
             <div class="px-3 pt-2 pb-3 sm:px-4 sm:pt-3 sm:pb-4 space-y-1">
               <!-- Execute command -->


danielroe:
image

this doesn't feel quite right...

let's keep the dots as decorative only - what about making install/run as tabs? so you can swap to the 'run' command?

or maybe 'inline' within the same box, under a comment starting with `#`?

vinnymac:
> image
> this doesn't feel quite right...
> 
> let's keep the dots as decorative only - what about making install/run as tabs? so you can swap to the 'run' command?
> 
> or maybe 'inline' within the same box, under a comment starting with `#`?

I like the inline in the same box approach, that seems cleaner and clearer, not sure how to best support showing more than one binary if we did that, but I'll ponder it.

vinnymac:
Here is the updated UI with inline commands and comment, and hover to reveal copy action. Let me know what you think. I am not sure if the title of this section should be `INSTALL` anymore, what do you think about renaming it to `COMMANDS` instead?

Note: I just fixed the highlighting issue on `dlx` in the images below [here](https://github.com/npmx-dev/npmx.dev/pull/96/commits/82919d99ef87901244acf68c28f78c41779a2eb9).

<details><summary>Old proposed UI changes</summary>
<p>


## Single binary

image

## Multiple binaries (collapsed)

image

## Multiple binaries (expanded)

image

## No binaries

image

</p>
</details>

vinnymac:
@bluwy You requested I add some of these features for `create-x` here, curious what you think so far of this.

<details><summary>General use</summary>
<p>

## Express with install, types, and create

image

## Exec locally (degit)

image

## Create command only (create-turbo)
image

## But turbo itself has install, exec, and create
image

## Bun run only

image

</p>
</details> 


<details><summary>Create Command Validations</summary>
<p>

1. /vite - should show create command (same maintainers)
image

 2. /next - should show create command (shared maintainer, same repo)
image

 3. /nuxt - should show create command (same maintainer, same org)
image

 4. /color - should NOT show create command (different maintainers, different org)
image

 5. /lodash - should NOT show create command (no create-lodash exists)
image

</p>
</details>

bluwy:
In my opinion, this looks really nice and well thought-out 😃 The degit case is interesting I think since it's mainly used as a CLI with supplementary API, so `pnpm dlx degit` would be preferred instead, but I don't think there's a way to semantically detect that. (I think it's fine as is)

vinnymac:
~~Looks like light mode was added since I opened this, and some of the views like this create react one are borked~~

<details><summary>Broken</summary>
<p>

image

</p>
</details> 

<details><summary>Fixed</summary>
<p>

image


</p>
</details> 

[`2335356` (this PR)](https://github.com/npmx-dev/npmx.dev/pull/96/commits/233535675fd4589db2b203741c9206924ee0e7d2)

PR #97, opened 2026-01-26
feat: Add weekly downloads to PackageCard

IHIutch:
Adds weekly downloads to Package card. Since you can sort by downloads, its useful feedback to see the download counts

Makes a few additional changes:
- Uses a "stretched link" to make the whole card clickable rather than wrapping full card contents in the anchor tag
- Adds a focus state to PackageCard
- Adjusts card layout to account for the addition of weekly downloads
- This layout breaks the header/footer structure, which I dont _think_ are necessary, replacing them with divs to avoid confusing the semantic structure

Package list view showing updated card layout, including number of weekly downloads


IHIutch:
Im not 100% sure on the layout, would love feedback

One important note: `ProvenanceBadge` includes a link element, which I think is technically accessible since the link isn't semantically nested, but still feels a bit weird because it _appears_ nested because of the stretched link on the card

Screen shot of the Provenance Badge link appearing inside the package card link

danielroe:
we should probably disable that as a link and make it purely visual.

might need to pass a prop down

PR #100, opened 2026-01-26
feat: navigate to package code view on "dot" key

jhroemer:
Uses same styling as the search `/` element, only difference is I resized to 4 w/h like the other inline icons (looks nicer also). This version also hides the icon on mobile - again similar to `/`.

I'm not too familiar with Vue, but as far as I understand [VueUse](https://vueuse.org/) is very standard, and already installed in the project. Let me know if there's something I need to change though.

Screenshot 2026-01-26 at 06 49 09

### DIFF
--- a/app/pages/[...package].vue
+++ b/app/pages/[...package].vue
@@ -672,7 +672,6 @@ defineOgImageComponent('Package', {
                 class="link-subtle font-mono text-sm inline-flex items-center gap-1.5"
                 aria-keyshortcuts="."
               >
-                <span class="i-carbon-code w-4 h-4" aria-hidden="true" />
                 code
                 <kbd
                   class="hidden sm:inline-flex items-center justify-center w-4 h-4 text-xs bg-bg-muted border border-border rounded"


danielroe:
this is great! added [`aria-keyshortcuts`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-keyshortcuts) which we should also have used in the header 

and dropped the icon as it felt a bit much - but happy to revisit if folk think it should have an icon too

jhroemer:
> this is great! added [`aria-keyshortcuts`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-keyshortcuts) which we should also have used in the header

Yes, awesome! Thanks!

> and dropped the icon as it felt a bit much - but happy to revisit if folk think it should have an icon too

Very good call, it's one thing I was unsure of regarding the change and adding the keyboard shortcut icon. My general stance on keyboard shortcuts in web apps is the same as with TUI's, they should be discoverable directly in the UI with minimal effort. So I think the keyboard shortcut icon is really needed, unless we use a tooltip to show info.

PR #106, opened 2026-01-26
chore: remove unused cli deps

jhroemer:
Identified w. [Knip](https://knip.dev/) + double-checked manually and ran the connector locally.


jhroemer:
@danielroe thoughts on running [Knip](https://knip.dev/) in CI? Would need some configuration I believe, there seems to be a bunch of false-positives at least to look into.

danielroe:
absolutely!

PR #110, opened 2026-01-26
fix: improve package card component for mobile version

iiio2:


iiio2:
@danielroe , Thanks. :blush:

PR #114, opened 2026-01-26
feat: version can scroll when version is too many

YFAnt:


serhalp:
hey @YFAnt is this PR still relevant? does this issue still exist on main?

PR #113, opened 2026-01-26
feat: add OpenSearch support

BYK:
## Summary

Add OpenSearch support so browsers can use npmx.dev as a search engine directly from the address bar.

## Changes

- **`public/opensearch.xml`** - OpenSearch description file with search URL template and suggestions endpoint
- **`nuxt.config.ts`** - Autodiscovery `<link>` tag in HTML head
- **`server/api/opensearch/suggestions.get.ts`** - Search suggestions API for browser autocomplete (proxies npm registry search, cached for 60s)

## How to Test

1. Visit `https://npmx.dev/opensearch.xml` to verify the XML
2. Test suggestions: `https://npmx.dev/api/opensearch/suggestions?q=react`
3. **Firefox**: Visit site → click address bar → add "npm" search engine
4. **Chrome**: Settings → Search engines → "Site search" should list npmx.dev

PR #115, opened 2026-01-26
fix: set correct Content-Type for opensearch.xml

BYK:
## Summary

Set the correct `Content-Type` header for the OpenSearch XML response to prevent Vercel from injecting preview toolbar scripts.

Follow up to #113.

## Problem

Vercel's preview toolbar only injects scripts into responses with `Content-Type: text/html`. Without an explicit Content-Type, the XML response was being treated as HTML, causing:
- Feedback script injection at the end of the XML
- Invalid XML that browsers couldn't parse as an OpenSearch description

## Solution

Add `setHeader(event, 'Content-Type', 'application/opensearchdescription+xml')` to the server route handler.

### DIFF
--- a/server/routes/opensearch.xml.get.ts
+++ b/server/routes/opensearch.xml.get.ts
@@ -1,6 +1,9 @@
-export default defineEventHandler(async event => {
+export default defineEventHandler(event => {
   const url = getRequestURL(event)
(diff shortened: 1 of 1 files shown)


danielroe:
thank you! and sorry, it was my mistake!

Issue #94, opened 2026-01-26
No graceful 404 page when navigating directly to nonexistent org


serhalp:
Repro steps:
1. Navigate directly to https://npmx.dev/@react-bot (this org does not exist)
2. See this error:

<img width="501" height="382" alt="Image" src="https://github.com/user-attachments/assets/d3eee39d-7136-48b4-8d0b-aba32b4f0a49" />

For reference, the package (https://npmx.dev/nfjdasnfksdfksd) and user (https://npmx.dev/~nfjdasnfksdfksd) 404 pages are correctly implemented:

<img width="673" height="272" alt="Image" src="https://github.com/user-attachments/assets/095bff68-9481-4c70-8e16-249b3bfdc796" />

<img width="627" height="379" alt="Image" src="https://github.com/user-attachments/assets/075d3afc-3aec-457e-8fc1-ff96a1e1494e" />

parion:
I think #124 closes this issue, but leaves a concerning case open that @MatteoGabriele [was finding](https://github.com/npmx-dev/npmx.dev/pull/124#issuecomment-3799658667).

Now the https://npmx.dev/@react-bot doesn't fail but instead shows the react-bot user's profile data, but as an org page.

<img width="1060" height="594" alt="Image" src="https://github.com/user-attachments/assets/4fc8fab1-6c19-4860-bf92-87c86b9806e4" />

Interestingly, this page shows one less result than the actual user page: https://npmx.dev/~react-bot. I'm thinking this is because, for some reason, the org page includes deprecated packages and the user page doesn't.

And for any other org that doesn't exist, the 404 page seems to be working fine.

<img width="1023" height="416" alt="Image" src="https://github.com/user-attachments/assets/0633a271-4988-433e-961a-b04bd488b7ac" />

PR #124, opened 2026-01-26
fix: skip displaying packages with undefined dist tags in org route

jonathanyeong:
## Why

In production, we get a JS error when accessing this url https://npmx.dev/@react-bot. This is due to a package not having `dist-tags`: `react-server-dom-vite`. 

Interestingly, https://www.npmjs.com/~react-bot?activeTab=packages shows that there are 21 packages but doing a manual count of the package list only shows 20. So I guess their UI is also incorrect.

## What changed

- Filter out any package that has undefined `dist-tags`

## Testing
I manually tested going to http://localhost:3000/@react-bot.

The page now shows a package list.

CleanShot 2026-01-26 at 07 41 58@2x

### DIFF
--- a/app/composables/useNpmRegistry.ts
+++ b/app/composables/useNpmRegistry.ts
@@ -386,7 +386,7 @@ export function useOrgPackages(orgName: MaybeRefOrGetter<string>) {
         const packuments = await Promise.all(batch.map(name => fetchMinimalPackument(name)))
 
         for (const pkg of packuments) {
-          // Filter out any unpublished packages (non-existant dist-tags)
+          // Filter out any unpublished packages (missing dist-tags)
           if (pkg && pkg['dist-tags']) {
             results.push(packumentToSearchResult(pkg))
           }


jonathanyeong:
This issue is related to https://github.com/npmx-dev/npmx.dev/issues/94. But it doesn't implement a graceful 404 page @serhalp am I missing something?

app/composables/useNpmRegistry.ts
```diff
@@ -382,7 +382,7 @@ export function useOrgPackages(orgName: MaybeRefOrGetter<string>) {
         const packuments = await Promise.all(batch.map(name => fetchMinimalPackument(name)))
 
         for (const pkg of packuments) {
-          if (pkg) {
+          if (pkg && pkg['dist-tags']) {
```
43081j, line 390: can you leave a comment here so future people understand whats going on?

npm actually defines `dist-tags` as always being present it seems, so this package must be some weird edge case. its worth noting that, and one day we can strongly type `packumentToSearchResult` to require a `dist-tags`

MatteoGabriele:
I was having a look at this issue as well, and while making sure `dist-tags` is defined fixes the error, in the case of "react-bot" as an organization, which doesn't exist, it seems to start fetching packages for "react-bot" the user.

jonathanyeong:
@MatteoGabriele I was trying to dig into this as well. Looking at npm the org does exist https://www.npmjs.com/~react-bot

Or am I missing something?

43081j:
@jonathanyeong FYI here's how i would fix it longer term:

https://github.com/npmx-dev/npmx.dev/compare/main...optional-dist-tags

i think it'd be better to do it like this at the root rather than at the callsite

in future we should probably detect these as `unpublished` (i.e. npm removed them or the user did)

MatteoGabriele:
@jonathanyeong I'm not 100% sure because the UI doesn't give many hints, but usually when a package is from an organization, such as react-bot, it shows as `@react-bot/react-dom`, for example.
In this case, they show up as `react-dom` and are listed as a collaborator.
I could be wrong, but I thought it was worth mentioning because I have the feeling the response was somehow falling back to an existing matching user instead of an organization.

jonathanyeong:
Oh I like your changes @43081j but what if a package is unpublished from the org. Do we really want to show it? for reference searching for `react-server-dom-vite` doesn't show up in npmx either

CleanShot 2026-01-26 at 08 22 26@2x

43081j:
i think you're probably right we should still filter them out 👍 

but i would still make the change to the type, so we're not blindly assuming it is always set. now and in future there may be other places we make the same incorrect assumption thanks to the types. so it'd be better to fix them up

jonathanyeong:
Updated @43081j, thanks for the comments!

I think @MatteoGabriele your question is unresolved. But I'm not entirely sure what we'd do here 🤔. I would assume we should match npmjs as a "source of truth". But let me know if we should throw out this PR in favour of another one!

> I could be wrong, but I thought it was worth mentioning because I have the feeling the response was somehow falling back to an existing matching user instead of an organization.

It looks like it's falling back to `@[org].vue`. Is that what you'd expect?

MatteoGabriele:
@jonathanyeong this PR still fixes the error, so my concern can be considered out of scope. 

To make another example: https://npmx.dev/@matteo_gabriele 
This should return 404 since I do have an npm account as a user, but it's not an organization.
I will do some research and see if it's worth creating a separate issue.

Issue #117, opened 2026-01-26
Incorrect readme pulled by default


ematipico:
Great tool! I am checking the rendering of various packages, and I noticed that for `@biomejs/biome`, it doesn't pull `README.md`, but it does pull/display a translated version

https://npmx.dev/@biomejs/biome (in my case, I see the chinese one)

PR #130, opened 2026-01-26
fix: prefer classic readme if npm picks up a different filename

jonathanyeong:
Fixes: #117

## Why

When fetching from registry npmjs, the incorrect readme filename is returned:

```
❯ curl -s "https://registry.npmjs.org/@biomejs%2Fbiome" | jq -r '.readmeFilename'
README.zh-TW.md
```

Doing some digging it might be because biomejs stores all its [readmes in a single folder](https://github.com/biomejs/biome/tree/HEAD/packages/@biomejs/biome). Someone had a similar issue [here](https://github.com/npm/npm/issues/5583).

## Changes made
- Check if readmeFilename matches set of standard Readme filenames. If not, fetch the standard filename from jsDelivr.

## Testing

Visiting preview link of biomejs shows english README by default (https://npmxdev-git-fork-jonathanyeong-jy-fix-incorrect-readme-poetry.vercel.app/@biomejs/biome)

CleanShot 2026-01-26 at 11 00 13@2x

### DIFF
--- a/server/api/registry/readme/[...pkg].get.ts
+++ b/server/api/registry/readme/[...pkg].get.ts
@@ -6,6 +6,20 @@ import {
   ERROR_NPM_FETCH_FAILED,
 } from '#shared/utils/constants'
 
+/** Standard README filenames to try when fetching from jsdelivr (case-sensitive CDN) */
+const standardReadmeFilenames = [
+  'README.md',
+  'readme.md',
+  'Readme.md',
+  'README',
+  'readme',
+  'README.markdown',
+  'readme.markdown',
+]
+
+/** Matches standard README filenames (case-insensitive, for checking registry metadata) */
+const standardReadmePattern = /^readme(\.md|\.markdown)?$/i
+
 /**
  * Fetch README from jsdelivr CDN for a specific package version.
  * Falls back through common README filenames.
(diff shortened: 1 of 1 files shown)


server/api/registry/readme/[...pkg].get.ts
```diff
@@ -108,3 +112,21 @@ export default defineCachedEventHandler(
     },
   },
 )
+
+function readmeFilenameMatchesLocale(filename: string | undefined): boolean {
+  if (!filename) {
+    return false
+  }
+
+  const filenames = [
```
Comment (jonathanyeong) [inline server/api/registry/readme/[...pkg].get.ts]: @danielroe if this approach makes sense, I can extract this out into a shared const since it's duplicated in `fetchReadmeFromJsdelivr`

server/api/registry/readme/[...pkg].get.ts
```diff
@@ -79,9 +71,29 @@ export default defineCachedEventHandler(
...
+
       // If no README in packument, try fetching from jsdelivr (package tarball)
-      if (!readmeContent || readmeContent === NPM_MISSING_README_SENTINEL) {
-        readmeContent = (await fetchReadmeFromJsdelivr(packageName, version)) ?? undefined
+      if (
+        !readmeContent ||
+        readmeContent === NPM_MISSING_README_SENTINEL ||
+        !isPreferredReadme(packageData.readmeFilename, standardReadmeFilenames)
+      ) {
+        const locale = navigator.language
```
Comment (danielroe) [inline server/api/registry/readme/[...pkg].get.ts]: this will be undefined as it runs on the server

Comment (jonathanyeong) [reply server/api/registry/readme/[...pkg].get.ts]: Oh great catch. I removed the locale, but will keep the check on the filename so we always return the default README.md. I'm assuming we want to avoid calling `fetchReadmeFromJsdelivr` if possible. Let me know if there's other changes I should make 🙏 !

jonathanyeong:
Of course! Thank you for making the PR better!

PR #135, opened 2026-01-26
feat: generate package docs using `@deno/doc`

devdumpling:
## Overview

This PR is one of two variants adding **autogenerated documentation pages**, accessible via a nice little link and routing to `/docs/{package}/v/{version}`. 

If a package provides a `homepage` in its manifest, the link will default to that. Otherwise, we attempt to generate docs via [`deno docs`](https://docs.deno.com/runtime/reference/cli/doc/) using a WASM binary around `@deno/docs`. More on this below. 

> Note: in the other variant, we use a microservice instead of WASM. This has tradeoffs. See this [PR](https://github.com/npmx-dev/npmx.dev/pull/134).

## Problem

Okay, so `npmx.dev` currently shows all kinds of fun package data (metadata, READMEs, file contents, etc). But it lacks a feature I love from JSR/crates/other modern package browsers... autogenerated docs off type definitions. 

Generating these kind of docs is challenging for npm packages because:
- npm pkgs don't ship in a standard doc format
- TS types may be inline, in `.d.ts`, `@types/*`, jsdocs, or any combination
- extracting and parsing that requires complex ASTs and compute overhead

## Solution

Luckily, there's tons of prior art here. In particular, Deno and the JSR folks basically already solved this problem with [deno doc](https://github.com/denoland/deno_doc). 

> Note: we should make sure they're cool with us using this if we end up going this direction... since we're sort of competing in a sense. 

### Architecture

Uses `@deno/doc` (official Deno doc generator package) compiled to WASM, running directly in the Node.js/Vercel serverless env. No subprocess shelling or microservice required. 

It's a bit more complex than the microservice under the hood though, and has some thorns. It works like this:

1. Fetches the types URL from `esh.sh` with a `x-typescript-types` header
2. Loads the TS definitions using custom fetch-based loader
3. Parses them with `@deno/doc` WASM to extract `DocNode[]`
4. Render as HTML with syntax highlighting

Parts of this are the same as the microservice one, just calling it out here. 

#### Arch Diagram

Screenshot 2026-01-26 at 12 36 31 PM

### Implementation

This was quite a ride for me, as much of it is transforming the format from `deno docs --json`, but there's some key points. 

Most of the heavy lifting is done in `/server/utils/docs/` and is broken down into a few submodules:

- index | entry point, calls `generateDocsWithDeno(pkg, version)`
- client | WASM integration
- processing | flattens namespaces, merge overloads, build symbol lookup, etc
- render | generate HTML with Shiki syntax highlighting server side to not bloat the JS payload
- text | escape HTML, parse JSDocs links, render markdown
- format | format typescript signatures
- types | internal types

**WASM integration**

The WASM integration required a Node.js compatability patch since `@deno/doc` is designed for Deno. This is a bit concerning since we'd have to maintain this or upstream it to the deno repo (possible, but not sure what they'd think). The issues are:

1. It has `.d.ts` exports in JS, which node won't natively resolve (see `export * from './types.d.ts'` lines in it)
2. WASM loader doesn't support Node.js fs, it tries to use the deno native one, so we needed to monkeypatch it to use `node:fs` and `node:url` for `file://` URLs

I managed the patch with pnpm and is auto-applied on install. 

**Frontend implementation**

The frontend implementation simply fetches from the api endpoint (`/api/registry/docs/{pkg}/v/{version}`) and renders the pre-generated HTML in `v-html`. I'm still learning vue/nuxt, so let me know if I'm doing anything dumb here. 

The UI could probably be improved / iterated on. I left that out of scope for this POC / initial implementation. 

**A note on `esm.sh`**

I needed to use `esm.sh` here instead of our existing calls to `jsDeliver`. Why? 

In my testing, I couldn't get `jsDeliver` to resolve `@types/*` packages for untyped packages properly and kept hitting module resolution issues with varying `.js|.ts` extension 😢. I'm betting there's a way to make it work, but `esm.sh` worked on the first try, so I went with that. It also serves the typescript types natively with a `x-typescript-types` header which I use here. 

**If we're concerned about having two underlying providers here (jsDeliver and esm.sh) I can take this back to the drawing board / iterate.**

**Note on bundle size**

This does add some WASM weight to our bundle. It should load once and then cache and only impact server-side, but worth calling out as downside. 

## Alternative approaches

I personally prefer this to the microservice as it works without more infra, but the caveats in the WASM bindings are worth exploring fixes / alternatives to. 

As in the microservice approach, here are the alternatives:

1. [A microservice variant](https://github.com/npmx-dev/npmx.dev/pull/134) -- run `deno doc --json` out of a microservice 
4. Build our own parsing / AST solution natively, likely in Rust (I tried doing this with TypeDoc and it was clumsy and slow). We don't currently have any Rust stuff in the codebase so big decision there. Might be better solve long term.

Option 2 may be interesting, but also comes with the steep cost of diverging from the incredible work the JSR team already put into this. If we want to build out other rust stuff here though, might make sense. We could also theoretically do more interesting stuff than is possible with just the json data we get from `deno doc --json`. 
 
closes https://github.com/npmx-dev/npmx.dev/pull/134

### DIFF
--- a/app/components/DocsVersionSelector.vue
+++ b/app/components/DocsVersionSelector.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import { onClickOutside } from '@vueuse/core'
-import { compareVersions } from '~/utils/versions'
+import { compare } from 'semver'
 
 const props = defineProps<{
   packageName: string
@@ -24,7 +24,7 @@ const MAX_VERSIONS = 10
 /** Safe version comparison that falls back to string comparison on error */
 function safeCompareVersions(a: string, b: string): number {
   try {
-    return compareVersions(a, b)
+    return compare(a, b)
   } catch {
     return a.localeCompare(b)
   }
--- a/app/utils/versions.ts
+++ b/app/utils/versions.ts
@@ -5,14 +5,6 @@ import { compare, valid } from 'semver'
  */
 
 /**
- * Compare two version strings for sorting
- * @param a - First version string
- * @param b - Second version string
- * @returns -1 if a < b, 0 if a == b, 1 if a > b
- */
-export const compareVersions = compare
-
-/**
  * Check if a version string is an exact semver version.
  * Returns true for "1.2.3", "1.0.0-beta.1", etc.
  * Returns false for ranges like "^1.2.3", ">=1.0.0", tags like "latest", etc.


danielroe:
deployed, if you want to test ✅

devdumpling:
> deployed, if you want to test ✅

Thanks! Just finished writing up the description in the microservice variant. https://github.com/npmx-dev/npmx.dev/pull/134

app/pages/[...package].vue
```diff
@@ -672,6 +686,26 @@ defineOgImageComponent('Package', {
...
             </li>
 
+            <li v-if="docsLink">
+              <a
+                v-if="docsLink.isExternal"
+                :href="docsLink.href"
+                target="_blank"
+                rel="noopener noreferrer"
+                class="link-subtle font-mono text-sm inline-flex items-center gap-1.5"
+              >
```
Comment (devdumpling) [inline app/pages/[...package].vue]: Moved this around as the spacing was bothering me before I asked what was going on here. Will drop this change before we merge anything in and fix properly upstream.

patches/@jsr__deno__doc@0.189.1.patch
devdumpling, line 1: i hate this but... it does work

devdumpling: although i realize now this isn't going to work in serverless envs 😭 

back to the drawing board with this part

server/api/registry/docs/[...pkg].get.ts
```diff
@@ -0,0 +1,56 @@
...
+      toc: generated.toc,
+      status: 'ok',
+    } satisfies DocsResponse
+  },
+  {
+    maxAge: 60 * 60, // 1 hour cache
+    swr: true,
+    getKey: event => {
+      const pkg = getRouterParam(event, 'pkg') ?? ''
+      return `docs:v5:${pkg}`
```
Comment (devdumpling) [inline server/api/registry/docs/[...pkg].get.ts]: was rolling multiple versions of this locally during rapid prototyping -- needs to be dropped back before moving from poc

server/utils/code-highlight.ts
```diff
@@ -268,6 +268,9 @@ export async function highlightCode(
         theme: 'github-dark',
       })
 
+      // Shiki doesn't encode > in text content (e.g., arrow functions)
+      html = escapeRawGt(html)
```
devdumpling, line 272: this also fixes a lot of random html validation warnings we're getting

shared/utils/npm.ts
devdumpling, line 1: not sure if this is better defined / housed elsewhere (i didn't see anything for it at first glance flipping through the repo)

devdumpling:
Still trying to debug this, but tough as whatever is failing seems to be specific to the vercel env. I'm guessing I need to configure something for nitro / vercel to make it happy w wasm. 

Test urls are here: 
- https://npmx-9jpr7mbt9-poetry.vercel.app/docs/ufo/v/1.5.0
- https://npmx-9jpr7mbt9-poetry.vercel.app/docs/react/v/19.2.3

I ran the vercel build locally and peeked through the build artifacts. I can see the wasm in the vercel function bundle. It's 4.7mb at the correct path. The patch is there as well and module importing / doc gen works in it. It's under the 25mb vercel limit too so...?

Only things I can think of without having access to the logs are:

- cold start time is too high for this
- function timeout default is too low for this (maybe it's taking longer than 10s or something?)
- memory? i can't imagine this would hit a memory limit but maybe

Or that there's just something completely different wrong I'm missing here (likely).

app/components/DocsVersionSelector.vue
devdumpling, line 1: Handles the docs specific version toggle. 

Might want to align these components eventually (since there's another one we use on the package code page, but this one is specific to the docs page.

app/pages/docs/[...path].vue
Comment (devdumpling) [inline app/pages/docs/[...path].vue:1]: Main UI entrypoint. 

Note we only ever actually get to this page if there isn't a `homepage` listed in the manifest for a package. If we wanted, we could add another way to link to these pages? What do you think @danielroe? 

I feel like respecting the `homepage` link as the default makes sense, just not sure if we want packages to also have this listed here or if it should be an easter egg or what.

app/pages/docs/[...path].vue
```diff
@@ -0,0 +1,448 @@
...
+      <aside
+        v-if="docsData?.toc && !showEmptyState"
+        class="hidden lg:block w-64 xl:w-72 shrink-0 border-r border-border"
+      >
+        <div class="docs-sidebar sticky overflow-y-auto p-4">
+          <h2 class="text-xs font-semibold text-fg-subtle uppercase tracking-wider mb-4">
+            Contents
+          </h2>
+          <!-- eslint-disable vue/no-v-html -->
+          <div class="toc-content" v-html="docsData.toc" />
```
Comment (devdumpling) [inline app/pages/docs/[...path].vue:143]: We're injecting from our rendered markdown data from the server here, same below for the actual content. 

A bit complex admittedly. Open to refactoring this if someone has ideas about better ways to handle.

app/pages/docs/[...path].vue
```diff
@@ -0,0 +1,448 @@
...
+
+/*
+ * Fenced code blocks in descriptions use a subtle left-border style.
+ *
+ * Design rationale: We use two visual styles for code examples:
+ * 1. Boxed style (bg + border + padding) - for formal @example JSDoc tags
+ *    and function signatures. These are intentional, structured sections.
+ * 2. Left-border style (blockquote-like) - for inline code in descriptions.
+ *    These are illustrative/casual and shouldn't compete with the signature.
+ */
```
Comment (devdumpling) [inline app/pages/docs/[...path].vue:334]: LMK if anyone has questions on this. My rationale was that the examples should feel more prominent. For a comparison look at the svelte docs page vs the react one. 

React one uses a lot of `@example` while svelte has inline `@description` with casual examples.

app/pages/[...package].vue
```diff
@@ -676,17 +700,7 @@ defineOgImageComponent('Package', {
...
-                :href="homepageUrl"
-                target="_blank"
-                rel="noopener noreferrer"
-                class="link-subtle font-mono text-sm inline-flex items-center gap-1.5"
-              >
-                <span class="i-carbon-link w-4 h-4" aria-hidden="true" />
-                {{ t('package.links.homepage') }}
-              </a>
-            </li>
+
```
Comment (devdumpling) [inline app/pages/[...package].vue]: I dropped the homepage url here in favor of the docs link, but as I'm looking back I wonder if maybe we just add both instead?

i18n/locales/en.json
```diff
@@ -90,7 +90,9 @@
       "issues": "issues",
       "forks": "fork | forks",
       "jsr": "jsr",
-      "code": "code"
+      "code": "code",
+      "docs": "docs",
+      "socket": "socket"
```
devdumpling: addresses this on main

server/utils/docs/format.ts
```diff
@@ -0,0 +1,96 @@
...
+  return type ? `${param.name}${optional}: ${type}` : `${param.name}${optional}`
+}
+
+/**
+ * Format a TypeScript type.
+ */
+export function formatType(type?: TsType): string {
+  if (!type) return ''
+
+  // Strip ANSI codes from repr (deno doc may include terminal colors since it's built for that)
```
devdumpling, line 75: this is hacky

package.json
```diff
@@ -82,6 +83,24 @@
...
+    "overrides": {
+      "sharp": "0.34.5"
+    },
+    "packageExtensions": {
+      "@nuxt/scripts": {
+        "dependencies": {
+          "estree-walker": "^3.0.3"
+        }
+      }
+    },
```
devdumpling: Did these get dropped since I started working on this? Taking a look now...

devdumpling: ah i see https://github.com/npmx-dev/npmx.dev/commit/14ccc37a03b11580591c1b016065618ea974c985

PR #143, opened 2026-01-26
fix: semver sorting

### LINKED ISSUE #139: versions aren't quite sorted per SemVer spec
Check out https://semver.org/#spec-item-11.

Example: https://npmx.dev/next (expand "Other versions")

Image

Numeric segments should be compared numerically, not alphanumerically.

MatteoGabriele:
Uses the built-in compare method from the `semver` package.

before
Screenshot 2026-01-26 at 20 46 20

after
Screenshot 2026-01-26 at 20 45 44

closes #139

### DIFF
--- a/app/components/AppFooter.vue
+++ b/app/components/AppFooter.vue
@@ -1,13 +1,15 @@
 <script setup lang="ts">
-const isMounted = ref(false)
-const isVisible = ref(false)
-const isScrollable = ref(true)
-const lastScrollY = ref(0)
-const footerRef = ref<HTMLElement>()
+const isMounted = shallowRef(false)
+const isVisible = shallowRef(false)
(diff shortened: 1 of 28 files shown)


app/utils/versions.ts
```diff
@@ -47,19 +47,7 @@ export function parseVersion(version: string): ParsedVersion {
  * @returns Comparison result for sorting
  */
 export function compareVersions(a: string, b: string): number {
```
danielroe: can we just use compare wherever compareVersions is used instead of wrapping it?

MatteoGabriele: Definitely! I thought about keeping the wrapper for future changes, but it's indeed cleaner to use the package's method where needed, which also eliminates the need for tests.

PR #147, opened 2026-01-26
feat: manage link in package description

jycouet:
With https://npmx.dev/@clack/prompts

Before
image

After
image

### DIFF
--- a/app/components/MarkdownText.vue
+++ b/app/components/MarkdownText.vue
@@ -44,9 +44,9 @@ function parseMarkdown(text: string): string {
     }
     const decodedUrl = url.replace(/&amp;/g, '&')
     try {
-      const parsed = new URL(decodedUrl)
-      if (['https:', 'mailto:'].includes(parsed.protocol)) {
-        const safeUrl = parsed.toString().replace(/"/g, '&quot;')
+      const { protocol, href } = new URL(decodedUrl)
+      if (['https:', 'mailto:'].includes(protocol)) {
+        const safeUrl = href.replace(/"/g, '&quot;')
         return `<a href="${safeUrl}" rel="nofollow noreferrer noopener" target="_blank">${text}</a>`
       }
     } catch {}


app/components/MarkdownText.vue
```diff
@@ -34,6 +34,15 @@ function parseMarkdown(text: string): string {
   // Strikethrough: ~~text~~
   html = html.replace(/~~(.+?)~~/g, '<del>$1</del>')
 
+  // Links: [text](url) - only allow http, https, mailto
+  html = html.replace(/\[([^\]]+)\]\(([^)]+)\)/g, (_match, text, url) => {
+    const decodedUrl = url.replace(/&amp;/g, '&')
+    if (/^(https?:|mailto:)/i.test(decodedUrl)) {
```
danielroe: this is probably not sufficient from a security point of view

jycouet: I did an update with
- `URL` class
- manage only `https` & `mailto`

Let me know what you have in mind, I can do more researches 👍

danielroe: looks good to me!

app/components/MarkdownText.vue
```diff
@@ -34,6 +34,19 @@ function parseMarkdown(text: string): string {
   // Strikethrough: ~~text~~
   html = html.replace(/~~(.+?)~~/g, '<del>$1</del>')
 
+  // Links: [text](url) - only allow https, mailto
+  html = html.replace(/\[([^\]]+)\]\(([^)]+)\)/g, (_match, text, url) => {
+    const decodedUrl = url.replace(/&amp;/g, '&')
+    try {
+      const parsed = new URL(decodedUrl)
+      if (['https:', 'mailto:'].includes(parsed.protocol)) {
+        const safeUrl = decodedUrl.replace(/"/g, '&quot;')
```
danielroe: we might need to use the parsed url - `parsed.toString()`

```suggestion
        const safeUrl = parsed.toString().replace(/"/g, '&quot;')
```

danielroe:
oh - one more issue - we have to avoid rendering a link when it's in a `<PackageCard>` e.g. in a search result.

jycouet:
> oh - one more issue - we have to avoid rendering a link when it's in a `<PackageCard>` e.g. in a search result.

You mean here 
image

Even if it's not clickable, it's nicer no?

jycouet:
Or having the text underlined, but not a link in a "semi-plain" mode? :D

danielroe:
I'd be up for it being underlined - but ... that does make the user think they can click it, right? 🤔

jycouet:
Or bold? Or different underline?
Just to not show `[    ]   (     )`

danielroe:
this is what it currently displays (no special chars):

SCR-20260126-uigf

jycouet:
What about ?
`A`
```
if (props.plain) {
  return `<span class="color-fg">${text}</span>`;
}
```
image

`B`
```
<strong>${text}</span>`;
```
image

---

I think I like `A`, but I have to say that I'm a very be designer :D

app/components/MarkdownText.vue
```diff
@@ -34,6 +36,23 @@ function parseMarkdown(text: string): string {
...
 
+  // Links: [text](url) - only allow https, mailto
+  html = html.replace(/\[([^\]]+)\]\(([^)]+)\)/g, (_match, text, url) => {
+    // In plain mode, just render the link text without the anchor
+    if (props.plain) {
+      return text
+    }
+    const decodedUrl = url.replace(/&amp;/g, '&')
+    try {
+      const parsed = new URL(decodedUrl)
```
mxdvl: **note:** you could destructure here, I believe, and avoid the `toString` later

```ts
      const { protocol, href } = new URL(decodedUrl)
```

danielroe:
the problem is - this is only meant to be a description, not to have its own CTA. we shouldn't highlight that CTA for that reason. it should be as muted as the rest of the description. that's a bit strange, but that's because it is being misused a bit by that package, imo...

jycouet:
> the problem is - this is only meant to be a description, not to have its own CTA. we shouldn't highlight that CTA for that reason. it should be as muted as the rest of the description. that's a bit strange, but that's because it is being misused a bit by that package, imo...

I get it, let's keep it like this 👍

PR #151, opened 2026-01-27
feat: add markdown output support for package pages

BYK:
## Summary

Add support for getting package information in Markdown format, useful for AI/LLM consumption and command-line tools.

### Access Methods
- **URL suffix**: `/<package>.md` (e.g., `/vue.md`, `/@nuxt/kit.md`)
- **Accept header**: `Accept: text/markdown`

### Output Includes
- Package metadata (name, version, license, last updated)
- Stats table with weekly downloads (with sparkline trend) and dependency count
- Install command
- Links (npm, repository, homepage, issues)
- Keywords and maintainers
- Full README content

### Security Hardening
- URL validation for homepage/bugs links (prevents `javascript:` protocol injection)
- README size limit (500KB) to prevent resource exhaustion
- Path exclusion alignment between Vercel rewrites and middleware

### DIFF
--- a/server/routes/raw/[...slug].md.get.ts
+++ b/server/routes/raw/[...slug].md.get.ts
@@ -1,0 +1,169 @@
+import { generatePackageMarkdown } from '../../utils/markdown'
+import { isStandardReadme, fetchReadmeFromJsdelivr } from '../../utils/readme-loaders'
+import * as v from 'valibot'
+import { PackageRouteParamsSchema } from '#shared/schemas/package'
+import { NPM_MISSING_README_SENTINEL, ERROR_NPM_FETCH_FAILED } from '#shared/utils/constants'
+
+// Cache TTL matches the ISR config for /raw/** routes (60 seconds)
+const CACHE_MAX_AGE = 60
+
+const NPM_API = 'https://api.npmjs.org'
+
+const standardReadmeFilenames = [
+  'README.md',
+  'readme.md',
+  'Readme.md',
+  'README',
+  'readme',
+  'README.markdown',
+  'readme.markdown',
+]
+
+function encodePackageName(name: string): string {
+  if (name.startsWith('@')) {
+    return `@${encodeURIComponent(name.slice(1))}`
+  }
+  return encodeURIComponent(name)
+}
+
+async function fetchWeeklyDownloads(packageName: string): Promise<{ downloads: number } | null> {
+  try {
+    const encodedName = encodePackageName(packageName)
+    return await $fetch<{ downloads: number }>(
+      `${NPM_API}/downloads/point/last-week/${encodedName}`,
+    )
+  } catch {
+    return null
+  }
+}
+
+function parsePackageParamsFromSlug(slug: string): {
+  rawPackageName: string
+  rawVersion: string | undefined
+} {
+  const segments = slug.split('/').filter(Boolean)
+
+  if (segments.length === 0) {
+    return { rawPackageName: '', rawVersion: undefined }
+  }
+
+  const vIndex = segments.indexOf('v')
+
+  if (vIndex !== -1 && vIndex < segments.length - 1) {
+    return {
+      rawPackageName: segments.slice(0, vIndex).join('/'),
+      rawVersion: segments.slice(vIndex + 1).join('/'),
+    }
+  }
+
+  const fullPath = segments.join('/')
+  const versionMatch = fullPath.match(/^(@[^/]+\/[^@]+|[^@]+)@(.+)$/)
+  if (versionMatch) {
+    const [, packageName, version] = versionMatch as [string, string, string]
+    return {
+      rawPackageName: packageName,
+      rawVersion: version,
+    }
+  }
+
+  return {
+    rawPackageName: fullPath,
+    rawVersion: undefined,
+  }
+}
+
+export default defineEventHandler(async event => {
+  // Get the slug parameter - Nitro captures it as "slug.md" due to the route pattern
+  const params = getRouterParams(event)
+  const slugParam = params['slug.md'] || params.slug
+
+  if (!slugParam) {
+    throw createError({
+      statusCode: 404,
+      statusMessage: 'Package not found',
+    })
+  }
+
+  // Remove .md suffix if present (it will be there from the route)
+  const slug = slugParam.endsWith('.md') ? slugParam.slice(0, -3) : slugParam
+
+  const { rawPackageName, rawVersion } = parsePackageParamsFromSlug(slug)
+
+  if (!rawPackageName) {
+    throw createError({
+      statusCode: 404,
+      statusMessage: 'Package not found',
+    })
+  }
+
+  const { packageName, version } = v.parse(PackageRouteParamsSchema, {
+    packageName: rawPackageName,
+    version: rawVersion,
+  })
+
+  let packageData
+  try {
+    packageData = await fetchNpmPackage(packageName)
+  } catch {
+    throw createError({
+      statusCode: 502,
+      statusMessage: ERROR_NPM_FETCH_FAILED,
+    })
+  }
+
+  let targetVersion = version
+  if (!targetVersion) {
+    targetVersion = packageData['dist-tags']?.latest
+  }
+
+  if (!targetVersion) {
+    throw createError({
+      statusCode: 404,
+      statusMessage: 'Package version not found',
+    })
+  }
+
+  const versionData = packageData.versions[targetVersion]
+  if (!versionData) {
+    throw createError({
+      statusCode: 404,
+      statusMessage: 'Package version not found',
+    })
+  }
+
+  let readmeContent: string | undefined
+
+  if (version) {
+    readmeContent = versionData.readme
+  } else {
+    readmeContent = packageData.readme
+  }
+
+  const readmeFilename = version ? versionData.readmeFilename : packageData.readmeFilename
+  const hasValidNpmReadme = readmeContent && readmeContent !== NPM_MISSING_README_SENTINEL
+
+  if (!hasValidNpmReadme || !isStandardReadme(readmeFilename)) {
+    const jsdelivrReadme = await fetchReadmeFromJsdelivr(
+      packageName,
+      standardReadmeFilenames,
+      targetVersion,
+    )
+    if (jsdelivrReadme) {
+      readmeContent = jsdelivrReadme
+    }
+  }
+
+  const weeklyDownloadsData = await fetchWeeklyDownloads(packageName)
+
+  const markdown = generatePackageMarkdown({
+    pkg: packageData,
+    version: versionData,
+    readme: readmeContent && readmeContent !== NPM_MISSING_README_SENTINEL ? readmeContent : null,
+    weeklyDownloads: weeklyDownloadsData?.downloads,
+  })
+
+  setHeader(event, 'Content-Type', 'text/markdown; charset=utf-8')
+  setHeader(event, 'Cache-Control', `public, max-age=${CACHE_MAX_AGE}, stale-while-revalidate`)
+
+  return markdown
+})
--- a/server/utils/markdown.ts
+++ b/server/utils/markdown.ts
@@ -1,0 +1,225 @@
+import type { Packument, PackumentVersion } from '#shared/types'
+import { normalizeGitUrl } from '#shared/utils/git-providers'
+import { joinURL } from 'ufo'
+
+const MAX_README_SIZE = 500 * 1024 // 500KB, matching MAX_FILE_SIZE in file API
+
+function formatNumber(num: number): string {
+  return new Intl.NumberFormat('en-US').format(num)
+}
+
+function formatBytes(bytes: number): string {
+  if (bytes < 1024) return `${bytes} B`
+  if (bytes < 1024 * 1024) return `${(bytes / 1024).toFixed(1)} kB`
+  return `${(bytes / (1024 * 1024)).toFixed(1)} MB`
+}
+
+function escapeMarkdown(text: string): string {
+  return text.replace(/([*_`[\]\\])/g, '\\$1')
+}
+
+function isHttpUrl(url: string): boolean {
+  const parsed = URL.parse(url)
+  if (!parsed) return false
+  return parsed.protocol === 'http:' || parsed.protocol === 'https:'
+}
+
+function getRepositoryUrl(
+  repository?:
+    | {
+        type?: string
+        url?: string
+        directory?: string
+      }
+    | string,
+): string | null {
+  if (!repository) return null
+  // Handle both string and object forms of repository field
+  const repoUrl = typeof repository === 'string' ? repository : repository.url
+  if (!repoUrl) return null
+  const normalized = normalizeGitUrl(repoUrl)
+  // normalizeGitUrl returns null for empty/invalid URLs, and may return non-HTTP URLs
+  if (!normalized || !isHttpUrl(normalized)) return null
+  // Strip .git suffix for cleaner display URLs
+  const cleanUrl = normalized.replace(/\.git$/, '')
+  // Append directory for monorepo packages (only available in object form)
+  if (typeof repository !== 'string' && repository.directory) {
+    return joinURL(`${cleanUrl}/tree/HEAD`, repository.directory)
+  }
+  return cleanUrl
+}
+
+export interface PackageMarkdownOptions {
+  pkg: Packument
+  version: PackumentVersion
+  readme?: string | null
+  weeklyDownloads?: number
+  installSize?: number
+}
+
+export function generatePackageMarkdown(options: PackageMarkdownOptions): string {
+  const { pkg, version, readme, weeklyDownloads, installSize } = options
+
+  const lines: string[] = []
+
+  // Title
+  lines.push(`# ${pkg.name}`)
+  lines.push('')
+
+  // Description
+  if (pkg.description) {
+    lines.push(`> ${escapeMarkdown(pkg.description)}`)
+    lines.push('')
+  }
+
+  // Version and metadata line
+  const metaParts: string[] = []
+  metaParts.push(`**Version:** ${version.version}`)
+
+  if (pkg.license) {
+    const licenseText = typeof pkg.license === 'string' ? pkg.license : pkg.license.type
+    metaParts.push(`**License:** ${escapeMarkdown(licenseText)}`)
+  }
+
+  if (pkg.time?.modified) {
+    const date = new Date(pkg.time.modified)
+    metaParts.push(`**Updated:** ${date.toLocaleDateString('en-US', { dateStyle: 'medium' })}`)
+  }
+
+  lines.push(metaParts.join(' | '))
+  lines.push('')
+
+  // Stats section
+  lines.push('## Stats')
+  lines.push('')
+
+  // Build stats table
+  const statsHeaders: string[] = []
+  const statsSeparators: string[] = []
+  const statsValues: string[] = []
+
+  // Weekly downloads
+  if (weeklyDownloads !== undefined) {
+    statsHeaders.push('Downloads (weekly)')
+    statsSeparators.push('---')
+    statsValues.push(formatNumber(weeklyDownloads))
+  }
+
+  // Dependencies count
+  const depCount = version.dependencies ? Object.keys(version.dependencies).length : 0
+  statsHeaders.push('Dependencies')
+  statsSeparators.push('---')
+  statsValues.push(String(depCount))
+
+  // Install size
+  if (installSize !== undefined) {
+    statsHeaders.push('Install Size')
+    statsSeparators.push('---')
+    statsValues.push(formatBytes(installSize))
+  } else if (version.dist?.unpackedSize) {
+    statsHeaders.push('Package Size')
+    statsSeparators.push('---')
+    statsValues.push(formatBytes(version.dist.unpackedSize))
+  }
+
+  if (statsHeaders.length > 0) {
+    lines.push(`| ${statsHeaders.join(' | ')} |`)
+    lines.push(`| ${statsSeparators.join(' | ')} |`)
+    lines.push(`| ${statsValues.join(' | ')} |`)
+    lines.push('')
+  }
+
+  // Links section
+  const links: Array<{ label: string; url: string }> = []
+
+  links.push({ label: 'npmx', url: `https://npmx.dev/package/${pkg.name}` })
+  links.push({ label: 'npm', url: `https://www.npmjs.com/package/${pkg.name}` })
+
+  const repoUrl = getRepositoryUrl(pkg.repository)
+  if (repoUrl) {
+    links.push({ label: 'Repository', url: repoUrl })
+  }
+
+  if (version.homepage && version.homepage !== repoUrl && isHttpUrl(version.homepage)) {
+    links.push({ label: 'Homepage', url: version.homepage })
+  }
+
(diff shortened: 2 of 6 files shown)


server/utils/markdown.ts
```diff
@@ -0,0 +1,280 @@
...
+  // Maintainers
+  if (pkg.maintainers && pkg.maintainers.length > 0) {
+    lines.push('## Maintainers')
+    lines.push('')
+    for (const maintainer of pkg.maintainers.slice(0, 10)) {
+      // npm API returns username but @npm/types Contact doesn't include it
+      const username = (maintainer as { username?: string }).username
+      const name = maintainer.name || username || 'Unknown'
+      if (username) {
+        lines.push(`- [${name}](https://www.npmjs.com/~${username})`)
```
okineadev:
```suggestion
        lines.push(`- [${name}](https://npmx.com/~${username})`)
```

server/utils/markdown.ts
```diff
@@ -0,0 +1,280 @@
...
+  }
+
+  return data
+    .map(val => {
+      const normalized = (val - min) / range
+      const index = Math.round(normalized * (SPARKLINE_CHARS.length - 1))
+      return SPARKLINE_CHARS[index]
+    })
+    .join('')
+}
```
okineadev: I don't know if this is true, but I think that LLM will have better understanding with a graph in the form of numbers instead of symbols, it's easier for us humans to perceive something visually

server/utils/markdown.ts
```diff
@@ -0,0 +1,280 @@
...
+      lines.push('')
+      lines.push('*[README truncated due to size]*')
+    } else {
+      lines.push(trimmedReadme)
+    }
+    lines.push('')
+  }
+
+  return lines.join('\n')
+}
```
okineadev, line 225: Maybe it would be better to use [Handlebars](handlebarsjs.com) template for Markdown generation?

okineadev (changes requested):

server/utils/markdown.ts
BYK: How about both? This entire patch is LLM generated so I have a feeling that they are pretty competent at this point :D

server/utils/markdown.ts
BYK: I'd hesitate adding a template engine for this but I can look into switching to template strings. I like the array-based construction here as much as it looks old school. Simple, clear, effective. What's the problem you want to solve with handlebars? To have a better understanding of the final output?

danielroe:
ultimately it might make sense to share some logic with the client (see https://github.com/npmx-dev/npmx.dev/pull/169) so we can click 'copy as markdown'

...although that button could also fetch the markdown from the server endpoint, which might be better from a JS bundle size

so... just linking so you're aware

BYK:
@atinux @okineadev are you waiting on me for something? (just making sure I'm not the blocker here)

atinux:
Sorry I was quite busy.

I suggest to not use a middleware here but instead having a server route with `/raw` prefix similar to how we have here: https://github.com/unjs/undocs/blob/main/app/server/routes/raw/%5B...slug%5D.md.get.ts

Then like on https://github.com/unjs/undocs/blob/main/app/modules/md-rewrite.ts, we can add the rewrite in the generated `vercel.json` to it happens at the CDN level

Would you be up to update your PR with this approach @BYK ?

BYK:
@atinux no worries. Updated :)

BYK:
Addressed the sparkline feedback - now showing both numeric trend data and visual sparkline:
```
Downloads (weekly): 1,234,567 [1.2M 1.3M 1.4M 1.5M] ▁▂▃▅
```
This provides both LLM-friendly numeric data and human-readable visual trend.

BYK:
@atinux ready for another look?

vercel.json
```diff
@@ -1,6 +1,18 @@
...
+      "source": "/:path((?!api|_nuxt|_v|__nuxt|search|code|raw/).*)",
+      "has": [{ "type": "header", "key": "accept", "value": "(.*?)text/markdown(.*)" }],
+      "destination": "/raw/:path.md"
+    },
+    {
+      "source": "/:path((?!api|_nuxt|_v|__nuxt|search|code|raw/).*)",
+      "has": [{ "type": "header", "key": "user-agent", "value": "curl/.*" }],
+      "destination": "/raw/:path.md"
+    }
+  ],
```
okineadev, line 27: We also need support for `package/*` paths

```plaintext
$ curl -L https://npmxdev-git-fork-byk-byk-featmd-poetry.vercel.app/package/vite
A server error has occurred

FUNCTION_INVOCATION_FAILED

arn1::r7r5t-1770279708442-3169599c59e0

$ curl -L https://npmxdev-git-fork-byk-byk-featmd-poetry.vercel.app/vite
# vite

> Native-ESM powered web dev build tool

...
```

server/utils/markdown.ts
```diff
@@ -0,0 +1,289 @@
...
+    lines.push(version.keywords.slice(0, 20).join(', '))
+    lines.push('')
+  }
+
+  // Maintainers
+  if (pkg.maintainers && pkg.maintainers.length > 0) {
+    lines.push('## Maintainers')
+    lines.push('')
+    for (const maintainer of pkg.maintainers.slice(0, 10)) {
+      // npm API returns username but @npm/types Contact doesn't include it
```
okineadev:
```suggestion
      // npm API returns username but `@npm/types` `Contact` doesn't include it
```

okineadev (changes requested):

ghostdevv:
ah! we do now have a copy as markdown button on the frontend #1020 - if you rebase you could make any change you need to use your new logic? It would be nice if that button also fetched it dynamically

server/utils/markdown.ts
```diff
@@ -0,0 +1,289 @@
...
+      const username = (maintainer as { username?: string }).username
+      const name = maintainer.name || username || 'Unknown'
+      if (username) {
+        lines.push(`- [${name}](https://npmx.dev/~${username})`)
```
ghostdevv: this may not be correct as maintainers is a user supplied field

ghostdevv (changes requested):

vercel.json
atinux: Agree, are you happy to add this alias @BYK ?

okineadev:
I think we should also take a look at https://github.com/npmx-dev/npmx.dev/pull/1382
It would be nice to somehow integrate the agent instruction fetching functionality as well

vercel.json
BYK: Added `/package/:path` rewrite rules in `vercel.json` that strip the `package/` prefix before routing to `/raw/:path.md`. These are placed before the general catch-all rewrites so they take priority.

Both `Accept: text/markdown` header and `curl` user-agent cases are handled.

server/utils/markdown.ts
BYK: Good catch! Fixed — now escaping the maintainer name with `escapeMarkdown()` since it's user-supplied, and also `encodeURIComponent()` on the username in the link href to prevent URL injection.

BYK:
Thanks for the pointer to #1382! That PR adds `llms.txt`/`llms_full.txt` support with agent instruction file fetching (CLAUDE.md, .cursorrules, etc.) — it's a substantial feature with its own middleware, types, and test suite.

I think it makes sense to keep these as separate PRs that can be coordinated:
- **This PR (#151)**: Markdown output via `/raw/` routes + content negotiation (curl/Accept header)
- **PR #1382**: `llms.txt` standard routes with agent instruction files

They share some underlying infrastructure (markdown generation, package metadata), so once both land, there could be follow-up work to consolidate shared utilities. But merging them now would significantly expand scope.

BYK:
@okineadev @ghostdevv sorry for the long delay. Should be good to go now?

okineadev:
> @okineadev @ghostdevv sorry for the long delay. Should be good to go now?

Personally, I abstain from voting.

BYK:
@okineadev who can vote so we can avoid bitrot? :)

server/routes/raw/[...slug].md.get.ts
```diff
@@ -0,0 +1,226 @@
...
+      }
+    } catch {
+      // Try next filename
+    }
+  }
+
+  return null
+}
+
+async function fetchWeeklyDownloads(packageName: string): Promise<{ downloads: number } | null> {
```
Comment (43081j) [inline server/routes/raw/[...slug].md.get.ts:29]: can we possibly use `usePackageDownloads` instead of this? iirc it does the same fetch.

we could also expand it to support ranges. instead of it being specific to this file

server/utils/markdown.ts
```diff
@@ -0,0 +1,290 @@
...
+  if (bytes < 1024) return `${bytes} B`
+  if (bytes < 1024 * 1024) return `${(bytes / 1024).toFixed(1)} kB`
+  return `${(bytes / (1024 * 1024)).toFixed(1)} MB`
+}
+
+function escapeMarkdown(text: string): string {
+  return text.replace(/([*_`[\]\\])/g, '\\$1')
+}
+
+function normalizeGitUrl(url: string): string {
```
43081j: git-providers does have a `normalizeGitUrl` too. is that doing the same job as this?

server/utils/markdown.ts
```diff
@@ -0,0 +1,290 @@
...
+    lines.push(`> ${escapeMarkdown(pkg.description)}`)
+    lines.push('')
+  }
+
+  // Version and metadata line
+  const metaParts: string[] = []
+  metaParts.push(`**Version:** ${version.version}`)
+
+  if (pkg.license) {
+    metaParts.push(`**License:** ${pkg.license}`)
```
43081j: licenses are also arbitrary user input, so unless this has already been validated, maybe we should `escapeMarkdown(pkg.license)`

vercel.json
```diff
@@ -1,6 +1,52 @@
...
+      ],
+      "destination": "/raw/:path.md"
+    },
+    {
+      "source": "/package/:path(.*)",
+      "has": [
+        {
+          "type": "header",
+          "key": "user-agent",
+          "value": "curl/.*"
```
43081j: are we sure about this? i might want to curl the HTML. i feel like we should just be testing the accept types rather than user agents

43081j (commented):
left a few comments but otherwise looks good

server/routes/raw/[...slug].md.get.ts
Comment (BYK) [reply server/routes/raw/[...slug].md.get.ts:29]: `usePackageDownloads` is a client-side Vue composable using `useLazyAsyncData` and `$npmApi` — it can't be used in a server-side route handler. The route handler needs direct `$fetch` calls to the npm API. These are separate concerns (client composable vs server route), so I'll keep the direct fetch here.

server/utils/markdown.ts
BYK: Good call! Switched to the shared `normalizeGitUrl` from `#shared/utils/git-providers`. Added `.git` suffix stripping after the call since the shared version doesn't do that.

server/utils/markdown.ts
BYK: Good catch — now escaping with `escapeMarkdown(pkg.license)`.

vercel.json
BYK: Agreed — removed the curl user-agent rewrite. Now only `Accept: text/markdown` header triggers the markdown response. Users can explicitly request markdown format without being forced into it based on their tool.

server/routes/raw/[...slug].md.get.ts
Comment (43081j) [reply server/routes/raw/[...slug].md.get.ts:29]: ah you are correct, i was thinking it was a utility rather than an app-specific composable.

anyhow, my point was more that we are certainly duplicating logic. we should at least have a shared utility i think, even if it just constructs the URL based on the parameters etc. not necessarily in this PR but at some point.

43081j (approved):

mbtools:
Thanks for this feature! 👍 

Here's some feedback to make this more useful (for a follow-up PR):

### Stats

- Historical download data isn't useful for AI (and hard to digest for humans). I suggest removing those numbers

image

- Add installed size
- Add number of vulnerabilities

### Install

- Shows `npm` command. This might contradict the package manager preferred by consumers or what's defined in `package.json` or what's in the readme. I suggest removing the section.

### Links

- Has link to `npm` but where's the link to `npmx.dev` ;-)

### Enhancements

- Add compatibility (engines) section
- Add available dist-tags

BYK:
Thanks for the detailed feedback @mbtools! Great suggestions. These are all solid improvements for follow-up work:

- **Stats**: Agreed the sparkline trend data is more noise than signal for most consumers. Install size is already included when available (from the install-size API). Adding vulnerability count would be valuable.
- **Install section**: Good point about package manager assumptions — could show a generic `<package-manager> install <pkg>` or remove it entirely.
- **Links**: Ha, good catch — adding an npmx.dev self-link makes sense!
- **Enhancements**: Engines/compatibility and dist-tags would both be useful additions.

I'll track these as follow-up items once this PR lands.

BYK:
Well, just went ahead and implemented them @mbtools 😅 

Hope we can get this merged soon

codecov.yml
```diff
@@ -13,6 +13,11 @@ coverage:
         only_pulls: true
         informational: true
 
+# Ignore files that are covered by browser tests (Playwright) rather than unit tests
+ignore:
+  - 'app/pages/**/*'
+  - 'app/layouts/**/*'
+
```
ghostdevv: Is this still needed?

server/utils/markdown.ts
```diff
@@ -0,0 +1,228 @@
...
+  return text.replace(/([*_`[\]\\])/g, '\\$1')
+}
+
+function isHttpUrl(url: string): boolean {
+  try {
+    const parsed = new URL(url)
+    return parsed.protocol === 'http:' || parsed.protocol === 'https:'
+  } catch {
+    return false
+  }
```
ghostdevv:
```suggestion
  const parsed = URL.parse(url)
  if (!parsed) return false
  return parsed.protocol === 'http:' || parsed.protocol === 'https:'
```

server/routes/raw/[...slug].md.get.ts
```diff
@@ -0,0 +1,196 @@
...
+    })
+  }
+
+  let readmeContent: string | undefined
+
+  if (version) {
+    readmeContent = versionData.readme
+  } else {
+    readmeContent = packageData.readme
+  }
```
Comment (ghostdevv) [inline server/routes/raw/[...slug].md.get.ts:140]: we've updated the readme rendering a few times now, can you check to make sure there isn't any normalisation or utils that exist somewhere we can use?

server/routes/raw/[...slug].md.get.ts
```diff
@@ -0,0 +1,196 @@
+import { generatePackageMarkdown } from '../../utils/markdown'
```
Comment (ghostdevv) [inline server/routes/raw/[...slug].md.get.ts:1]: do we need this file anymore? i.e. does something exist already as I know this PR is fairly old now

ghostdevv (changes requested):

server/utils/markdown.ts
mbtools: URL.parse is deprecated 😉

ghostdevv: I don't think so? It's only a couple years old...

codecov.yml
BYK: Yes — `origin/main` has no `ignore` rules in codecov.yml. Without this, the patch coverage check fails on the 1-line `<link rel=alternate>` change in the Vue page file, which is tested by Playwright browser tests rather than unit tests.

server/utils/markdown.ts
BYK: Done — switched to `URL.parse`.

server/routes/raw/[...slug].md.get.ts
Comment (BYK) [reply server/routes/raw/[...slug].md.get.ts:140]: Good call — `readme-loaders.ts` has `isStandardReadme` and `fetchReadmeFromJsdelivr` already exported. Refactored to import those instead of duplicating them. The `standardReadmeFilenames` constant is kept local since it's not exported from `readme-loaders.ts` (and the route handler needs to pass it to `fetchReadmeFromJsdelivr`).

server/routes/raw/[...slug].md.get.ts
Comment (BYK) [reply server/routes/raw/[...slug].md.get.ts:1]: Checked — `formatBytes`, `formatNumber`, and `escapeMarkdown` each exist in other places but with different behavior (e.g. badge handler uses compact notation and different unit labels, app composables are locale-aware). The markdown generation logic (`generatePackageMarkdown`) is unique to this feature. So yes, this file is still needed, but I've refactored the route handler to import `isStandardReadme` and `fetchReadmeFromJsdelivr` from `readme-loaders.ts` instead of duplicating them.

server/utils/markdown.ts
mbtools: `new URL` with try/catch was correct 

Fyi @ghostdevv https://nodejs.org/api/deprecations.html#dep0169-insecure-urlparse

codecov.yml
serhalp: The coverage checks are informative, not required/blocking. Can you please revert this config change? 🙏🏼

server/utils/markdown.ts
serhalp: It's a bit of a confusing design decision by node.js. `url.parse` is deprecated; the newer [`URL.parse`](https://nodejs.org/api/url.html#urlparseinput-base) is not. They are different!

mbtools: That is confusing indeed. Thx. The current code is good. You can mark this comment resolved 👍

codecov.yml
BYK: Fair point — reverted. The coverage checks are informational so the failing patch check won't block anything.

43081j:
@BYK feel free to not bother catching up until we have a 2nd approval (ideally from @ghostdevv since she reviewed last time).

otherwise you'll forever be playing catch up. we can just do it at the end 👍

mbtools:
Tested and it works great with "Accept: text/markdown". 👍 

It does not work with URL suffix: `/<package>.md. I'm just getting an empty HTML file:

```log
curl https://npmx-mtytt72bl-npmx.vercel.app/@registrytest/package.md
<!DOCTYPE html><html><head><meta http-equiv="refresh" content="0; url=/package/@registrytest/package.md"></head></html>
```

BYK:
@mbtools Good catch — the `.md` suffix URLs were missing from the Vercel rewrite rules. Only the `Accept: text/markdown` header path was wired up.

Fixed by adding unconditional rewrites for `.md` suffix:
- `/package/<pkg>.md` → `/raw/<pkg>.md`
- `/<pkg>.md` → `/raw/<pkg>.md`

Also rebased onto latest main. Should work on the next preview deploy.

vercel.json
```diff
@@ -1,6 +1,38 @@
...
   "$schema": "https://openapi.vercel.sh/vercel.json",
   "trailingSlash": false,
+  "rewrites": [
+    {
+      "source": "/package/:path(.*)\\.md",
+      "destination": "/raw/:path.md"
+    },
+    {
+      "source": "/:path((?!api|_nuxt|_v|__nuxt|search|code|raw/).*)\\.md",
+      "destination": "/raw/:path.md"
```
mbtools: I'm not familiar with the `skills` feature here but it looks like that includes markdown. Can you please check that the rewrites don't collide with this?

https://github.com/BYK/npmx.dev/blob/741d4aa0575c1fd6c0dde6bd4c7183984dc4ede6/server/routes/skills/%5B...pkg%5D.get.ts#L23

mbtools:
Tested again and it all works well. Just see my comment above

vercel.json
BYK: Good catch — the skills endpoint does serve `.md` files at paths like `/skills/<pkg>/v/<ver>/<skill>/refs/guide.md`, and the broad `.md` catch-all would have clobbered those.

Audited all top-level routes and expanded the exclusion list. The exclusions now cover:
- Framework: `api`, `_nuxt`, `_v`, `__nuxt`
- Our own: `raw/`
- Routes that serve markdown paths: `skills/`, `package-code/`, `package-docs/`, `blog/`
- All other top-level pages: `diff/`, `compare`, `search`, `org/`, `profile/`, `brand`, `about`, `accessibility`, `pds`, `privacy`, `recharging`, `settings`, `translation-status`, `oauth-client-metadata`, `opensearch`, `.well-known/`

Only bare package names (e.g. `/vue.md`, `/@scope/pkg.md`) and explicit `/package/<pkg>.md` trigger the markdown rewrite now.

gameroman (commented):
Technically, a package name can have `.md` at the end of it, for example https://npmx.dev/package/package.md

BYK:
@gameroman Great catch! `package.md` is indeed a real npm package (https://www.npmjs.com/package/package.md) and our top-level `/:path*.md` rewrite would have intercepted `/package.md` and served it as markdown output for a package named `package` instead of showing the page for `package.md`.

Fixed by removing the bare `/:path*.md` catch-all rewrite entirely. The remaining rewrites are:
- `/package/:path*.md` → safe, our explicit namespace (packages can't start with `package/`)
- `Accept: text/markdown` header rewrites → also safe since browsers never send this header

So the `.md` URL shorthand (e.g. `/vue.md`) no longer works — but it was inherently ambiguous. The explicit `/package/vue.md` form and `Accept: text/markdown` header both continue to work.

BYK:
@43081j @ghostdevv — could you take another look when you get a chance? The main change since last review is the addition of `.md` URL suffix rewrites in `vercel.json`, and subsequent fixes to avoid collisions with:
- the `skills/` endpoint (which serves `.md` files)
- npm packages with names ending in `.md` (e.g. `package.md`) — thanks @gameroman for catching this!

gameroman:
I've just checked again and https://npmxdev-git-fork-byk-byk-featmd-npmx.vercel.app/package/package.md still loads the markdown output when https://npmx.dev/package/package.md loads the package, I'm not sure what the best solution would be in this case

mbtools:
How about adding `?format=markdown` or `?md` instead of the `.md`?

gameroman:
> How about adding `?format=markdown` or `?md` instead of the `.md`?

That could work probably

Or something like `https://npmx.dev/package-markdown/package-name`

BYK:
@gameroman @mbtools I think the simplest solution is to drop the `.md` suffix approach entirely. The fundamental issue is that it's inherently ambiguous — any suffix-based convention will collide with real package names.

Instead, we already have two unambiguous mechanisms:

1. **Direct URL**: `/raw/<pkg>.md` — this is the actual route handler, no rewrite needed. E.g. `curl https://npmx.dev/raw/vue.md`
2. **Content negotiation**: `Accept: text/markdown` header on any package URL. E.g. `curl -H 'Accept: text/markdown' https://npmx.dev/package/vue`

Both are machine-discoverable via the `<link rel="alternate" type="text/markdown" href="/raw/..."> ` tag in the HTML page head.

No query params, no new URL namespaces, no collision risk. Just removed the `.md` suffix rewrites.

mbtools:
Sounds good. Please update the PR description at the top to match the implementation 😉 👍

gameroman:
Hi @BYK should I reopen this?

Could you also update the PR description please

BYK:
@gameroman, I'm okay if you guys want me to resurrect this. I chatted with some folks and was told heavy traffic was an issue and this would probably make it worse so I let it go :D

gameroman:
I see, then let's keep it as "not planned" probably 👍

ghostdevv:
I would like to figure this out, I've also kept #1382 open, but yea it'll take some time to figure out and traffic will hurt pending some changes

BYK:
Sounds good, LMK when we are ready and I'm happy to give this another shot.

Issue #164, opened 2026-01-27
i18n support


MatteoGabriele:
Since npmjs has never supported multiple languages, this could have been a barrier for many non-English-speaking developers. 
We should begin paving the way for additional translations, starting with English first, and add more as the app matures.

MatteoGabriele:
Even though we will add more languages later, we can close this for now. 
see #166

PR #166, opened 2026-01-27
feat: add i18n support

MatteoGabriele:
Adds i18n support and translates the homepage UI. 
Translating the entire app at once is impractical due to rapid changes; I prefer starting with a few basics and gradually translating bit by bit. 
Let me know if you agree or suggest an alternative approach.

see #164


danielroe:
thank you! ❤️ 

I'm extracting the remaining strings and will push shortly

danielroe:
that was a lot of string extraction

Issue #47, opened 2026-01-24
show vulnerability warnings for direct dependencies


danielroe:
batch query osv api, check if semver constraints intersect affected ranges

PR #167, opened 2026-01-27
feat: show vulnerability warnings for direct dependencies

Flo0806:
Resolves: #47 

---
## Summary

### Features:                                                                                                                                         
- New server API endpoint `/api/osv/vulnerabilities` that batch-queries the OSV API for npm packages                                                
- Vulnerability banner in `PackageDependencies.vue` showing total count and severity breakdown                                                      
- Inline security icons next to affected dependencies with tooltip details                                                                        
- Collapsible vulnerability details in `PackageVulnerabilities.vue` with expandable list                                                            
                                                                                                                                                    
### Refactoring:                                                                                                                                      
- Created `shared/utils/severity.ts` with global severity color constants and helpers                                         
- Removed duplicate severity definitions across components
- `PackageVulnerabilities.vue` now uses the shared OSV endpoint instead of direct API calls                                                         
                                                                                                                                                    
### Tests:                                                                                                                                            
- tests for severity utils                                                                             
- tests for tooltip and severity class helpers

## Live testing
> search for `package-json` then in detail choose version `5.0.0` you will see the `GET` package has **vulnerabilities**. klick on the shield icon jumps to the detail page of `GET` and shows details.


server/api/osv/vulnerabilities.post.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    // Query OSV for each package in parallel (with concurrency limit)
+    const results: Record<string, PackageVulnerabilities> = {}
+    const batchSize = 10
+
+    for (let i = 0; i < validPackages.length; i += batchSize) {
+      const batch = validPackages.slice(i, i + batchSize)
+      const batchResults = await Promise.all(
+        batch.map(async pkg => {
+          const info = await queryOsvForPackage(pkg.name, pkg.version)
```
danielroe: I'm pretty sure osv already has a batch api we can directly query from the frontend without needing to create a server endpoint

Flo0806: It's because I love backends :-) I change it!

danielroe: the one reason we might use the server api is if we want to query _all_ transitive dependencies - which would be a nice feature.

Flo0806: Then let me do this! I love challenges 😎

PR #177, opened 2026-01-27
feat(a11y): animate only using motion-safe

MatteoGabriele:
Improves accessibility for users who experience motion sensitivity or motion sickness.


MatteoGabriele:
Fixing a couple of tests that were relying on the classes I have prefixed with `motion-safe:`

PR #172, opened 2026-01-27
feat: light theme

jellydeck:
closes #108 

theme is derived from css variables at `:root`, then using `data-theme` prop on root (html) tag we toggle between themes.

tasks:
- [x] move all hardcoded colors to css variables
- [x] use oklch colors
- [x] theme preference persistence (local storage?)

PR #189, opened 2026-01-27
feat: use accent color on charts

graphieros:
Uses the accent color on :

- sparkline chart
- enlarged line chart

The pulse color is slightly lighter when an accent is used.


danielroe:
could we make it a bit more muted? maybe only the pulse has accent colour?

graphieros:
I think we need to wait  for #172 to be merged, because adaptations on charts will be required. Good idea for the accent on the pulse only. You can reject this PR ? I'll redo everything nicely after light mode is settled.

PR #199, opened 2026-01-27
feat: add package manager tabs to local connector install command

jonathanyeong:
## Current behaviour
We keep track of the selected package manager when we change it in the package. But it's not clear that I had selected that package manager in the first place when trying to install local connector.

https://github.com/user-attachments/assets/4c9a76a2-4536-4a8b-8f72-d6c89f428494

## New behaviour

https://github.com/user-attachments/assets/47fef397-18ca-4aa0-8db5-78b34f86ea03

### DIFF
--- a/app/components/ConnectorModal.vue
+++ b/app/components/ConnectorModal.vue
@@ -125,6 +125,14 @@ watch(open, isOpen => {
               </p>
 
               <div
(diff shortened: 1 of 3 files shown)


jonathanyeong:
Thanks for the review @serhalp! I'm going to close this PR for now. From Discord discussions, we don't want to show the tabs in the connector modal and maybe putting it in the new settings page is better.

PR #209, opened 2026-01-28
feat: add binary run scripts to package page

vinnymac:
Fixed issues with https://github.com/npmx-dev/npmx.dev/pull/96 and moved to a new pr after messing up my git history. Rather than fixing it and force pushing, just cherry picked the work I liked out of it.

You can see some examples of much of the changes here:

<details><summary>General use</summary>
<p>

## Express with install, types, and create

image

## Exec locally (degit)

image

## Create command only (create-turbo)
image

## But turbo itself has install, exec, and create
image

## Bun run only

image

</p>
</details> 

<details><summary>Create Command Validations</summary>
<p>

1. /vite - should show create command (same maintainers)
image

 2. /next - should show create command (shared maintainer, same repo)
image

 3. /nuxt - should show create command (same maintainer, same org)
image

 4. /color - should NOT show create command (different maintainers, different org)
image

 5. /lodash - should NOT show create command (no create-lodash exists)
image

</p>
</details>


app/utils/run-command.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  // Check create-* patterns
+  if (baseName?.startsWith('create-') || pkg.name.includes('/create-')) {
+    return true
+  }
+
+  // Has bin but no entry points
+  const hasBin =
+    pkg.bin !== undefined && (typeof pkg.bin === 'string' || Object.keys(pkg.bin).length > 0)
+  const hasEntryPoint = !!pkg.main || !!pkg.module || !!pkg.exports
```
43081j, line 36: this isn't always true. some CLIs do have a `main` which exports a programmatic API for running the same process the CLI runs

vinnymac: Do you think it's safer to just omit the main check here, and create false positives in scenarios where main exists but it isn't a programmatic API? (could be the majority)

Or perhaps it is feasible to detect this scenario some other way I am not thinking of? 🤔

danielroe: I think avoiding false positives is better (we just don't show a command that does exist - no harm, no foul)

danielroe:
I'll try to merge in main and address the failing browser test unless you manage to do it first

apart from that it's ready to go imo

vinnymac:
@danielroe thanks for pushing up these improvements I'll take a look at that test 🫡

danielroe:
cc: @yanthomasdev - getting this from the 'generate lunario overview' action:
```
Error: fatal: bad revision '04e5fa92a9218ed8d0f14757cc659555d223f537^...'
```

Issue #32, opened 2026-01-24
improve search relevance and discovery


danielroe:
better ranking, category browsing, similar packages suggestions

Haroenv:
If you want, you can use the index used for yarn if you put a "search powered by Algolia" somewhere. Shoot me an email if you're interested :)

We can also tweak the relevance and rules a bit if they're interesting for everyone

danielroe:
@Haroenv want to come join the discord server? https://chat.npmx.dev

Haroenv:
I've made a prototype in #174, does this look like what we want? Which way should we want to be able to configure which type of search to use?

PR #210, opened 2026-01-28
fix: normalize search queries to lowercase

sacrosanctic:


app/composables/useNpmRegistry.ts
```diff
@@ -247,7 +247,7 @@ export function useNpmSearch(
       }
 
       const params = new URLSearchParams()
-      params.set('text', q)
+      params.set('text', q.toLowerCase())
```
serhalp: 🤔 I'll defer to @danielroe here but I think ideally we would normalize this a bit further upstream so that the caching (line 241–242) can collapse to a normalize key.

danielroe: yes, that makes sense.

also - if the API returns different results based on case, is this really something we should obscure?

danielroe:
discussed in discord - we'll track instead improvements in https://github.com/npmx-dev/npmx.dev/issues/32, and prevent browsers autocapitalising home page input (which was the issue) in 33aa4dacc5517201e3b70c3952d0ebccfe2ab4b2

PR #211, opened 2026-01-28
feat: enhance package list UX with table view, filters, and sorts

serhalp:
https://github.com/user-attachments/assets/2ad8445f-ec06-46e1-8882-1c7a8f282d64

Add advanced filtering, sorting, and view options to package list pages.

- Add card/table view mode toggle with Local Storage persistence
- Add table view with sortable columns (name, downloads, last updated), with asc/desc toggle
- Add structured filtering: text search, download range, keywords (AND), updated since
- Add a few search operators for power users: `name:`, `desc:`, `kw:` (e.g. `name:react kw:hooks`)
- Add column picker for table view with disabled "coming soon" columns
- Add pagination options (infinite scroll vs paginated with page size)
- When "Filters" panel is collapsed, show a summary of active filters in operator syntax!
- Make keywords in table clickable to add to filters
- Make maintainers clickable links to user pages

### DIFF
--- a/app/composables/useNpmRegistry.ts
+++ b/app/composables/useNpmRegistry.ts
@@ -268,7 +268,7 @@ export interface NpmSearchOptions {
   size?: number
 }
 
-/** public */
+/** @public */
 export function useNpmSearch(
   query: MaybeRefOrGetter<string>,
   options: MaybeRefOrGetter<NpmSearchOptions> = {},


danielroe:
notes:
- infinite scroll on table view (or pagination)
- 'updated at' dates all seem to be 'today'

serhalp:
> 'updated at' dates all seem to be 'today'

🤔 can't repro. all seems to work for me on both table/card view and with/without relative dates setting.

serhalp:
- on the main search results page (not the /@:org page) it defaults sort to "Score" which is disabled. I think this should be Relevance and not disabled?
- on the main search results page... it's a fairly confusing feature, as it applies filters and sort only to the visible results on the page at the time. we should maybe just hide filters and sort entirely on this page (at least for now), but keep the two views?

danielroe:
agreed 👍

danielroe:
here's what I'm talking about:

https://github.com/user-attachments/assets/37236160-ecf1-4b03-abb7-53ca0e507734

PR #214, opened 2026-01-28
feat: improve charts

graphieros:
- bump vue-data-ui to 3.13.4 with maxWidth support for the large chart
- set higher height for larger chart in mobile bp
- add a close button close to thumb in chart modal in mobile

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ The aim of [npmx.dev](https://npmx.dev) is to provide a better browser for the n
 ## Shortcuts
 
 > [!IMPORTANT]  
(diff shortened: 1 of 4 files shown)


app/components/PackageDownloadAnalytics.vue
9romise, line 1: Hi! 👋 Thanks for your work! ❤️
I think we can use ~~[useResizeObserver](https://vueuse.org/core/useResizeObserver/)~~ [useElementSize](https://vueuse.org/core/useElementSize/) here.

graphieros: Thanks :)
So much better indeed!

PR #216, opened 2026-01-28
feat: add icons in install command

yuyinws:
image

image


patak-cat:
nice! would you grayscale them? I think it is better to avoid making these icons pop so much compared to other places in the UI

yuyinws:
Replaced with simple-icons.

PR #224, opened 2026-01-28
feat: allow linking to individual headers on package page

okineadev:
This will allow jumping straight to the readme view just like on GitHub by appending `#readme` to the URL - `https://github.com/<user>/<repo>#readme`

On npmx it will be like this - `https://npmx.com/<package>#readme`

---

The only problem I see right now is that when adding `#readme` to the URL it jumps incorrectly to the section

image

It should jump so that the "README" heading is visible:

image

There is probably some problem with the `<header>`.


okineadev:
Great addition!

This PR need to be renamed and the description changed.

PR #225, opened 2026-01-28
refactor: move settings to a page

### LINKED ISSUE #222: Settings page
add /settings route, move current functionality, plus future options

we will need to start gating some features with preferences soon

MatteoGabriele:
Moves the settings panel to an actual page

closes #222 

#### Notes
I didn't pay attention to a redesign, but we need to check it out, especially regarding the low visibility of toggles.

### DIFF
--- a/server/utils/readme.ts
+++ b/server/utils/readme.ts
@@ -151,7 +151,7 @@ const ALLOWED_ATTR: Record<string, string[]> = {
   code: ['class'],
   pre: ['class', 'style'],
   span: ['class', 'style'],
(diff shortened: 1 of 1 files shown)


MatteoGabriele:
@danielroe, it definitely has less contrast because there's no background. Do we have designs that I can look at, otherwise I can play with it a little

I do feel that toggles should probably change their background color when toggled off. They will never work when placed on a plain background

MatteoGabriele:
@danielroe made some tweaks. Feel free to change anything you want to change

Issue #223, opened 2026-01-28
downloads chart improvements


dominikg:
- [ ] the filename for generated chart downloads should contain package name and date range
- [ ] the minimap start and end date show up as 1 week ranges, leading to a confusing 4 dates showing on hover, should just show actual dates, or no dates at all given the huge pickers above
- [ ] changing the start date one day back leads to a steep fall down a cliff at the end (assuming that week ranges don't match up anymore and the last one is just one day) - so better to sync end date to the closest end of nex interval?

graphieros:
In `PackageDownloadAnalytics.vue`:
 
- minimap tooltips: I think we can leverage `zoom.customFormat` to display end/start dates only for the weekly granularity. I'll make the process simple by updating vue-data-ui to locate the tooltip (left  or right) to parse the date properly, and keep weekStart and weekEnd accordingly.
- I wonder if we need better date pickers that adapt to the granularity

graphieros:
I also noticed: 
- translations of the granularity are not used on the yAxis label

btea:
> the filename for generated chart downloads should contain package name and date range

#212 Part of it is completed, but the name does not include a specific time. Should we add it?

graphieros:
> [#212](https://github.com/npmx-dev/npmx.dev/pull/212) Part of it is completed, but the name does not include a specific time. Should we add it?

I have a PR open #230 on which I can do that, as I made some changes to fix the minimap tooltip issue

Edit: pushed in 027f5f3

graphieros:
✅ filenames: #846
✅ minimap date labels: #691 
✅ downloads steep falls: #1177

PR #230, opened 2026-01-28
fix: chart improvements

graphieros:
Related to #223

- bump vue-data-ui from 3.13.4 to 3.13.5
- expose the zoom tooltip sides to format weekly dates accordingly with a shorter label (left tooltip with weekStart; right tooltip with weekEnd):
image

- add start & end dates in filenames

### DIFF
--- a/app/components/PackageDownloadAnalytics.vue
+++ b/app/components/PackageDownloadAnalytics.vue
@@ -458,17 +458,26 @@ const config = computed(() => ({
       },
       callbacks: {
         img: ({ imageUri }: { imageUri: string }) => {
-          loadFile(imageUri, `${packageName}-${selectedGranularity.value}.png`)
+          loadFile(
+            imageUri,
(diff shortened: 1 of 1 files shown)


graphieros:
One item from #223 is not fixed yet, but these improvements could probably be included in the meanwhile.
I don't really have a clue on how to solve the date selection thing. Hoping for someone with a better brain than mine to have a go at it ^^

dominikg:
maybe calculating the other end of date range from interval is too much to ask for. better labeling and filename support are great. thank you so much

patak-dev (approved):
nice! let's merge and iterate

graphieros:
> maybe calculating the other end of date range from interval is too much to ask for. better labeling and filename support are great. thank you so much

I do believe a cool date picker component which limits selections based on granularity could provide ease of use (if such a component exists)

PR #233, opened 2026-01-28
feat(package): add fund button

### LINKED ISSUE #226: Sponsor button?
I suggest adding the same sponsoring button as on https://npmjs.com:

Image

9romise:
Resolves #226 
I'm not sure if there's enough space in this area. I'd like to get a better approach
image

### DIFF
--- a/app/pages/[...package].vue
+++ b/app/pages/[...package].vue
@@ -183,6 +183,15 @@ const docsLink = computed(() => {
   }
 })
 
+const fundingUrl = computed(() => {
+  let funding = displayVersion.value?.funding
+  if (Array.isArray(funding)) funding = funding[0]
+
+  if (!funding) return null
+
+  return typeof funding === 'string' ? funding : funding.url
+})
+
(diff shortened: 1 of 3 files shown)


okineadev:
Great!

I wanted to do this too but couldn't figure out where to get the `package.json` data

I think it would be better to make the same big button on the right side as on https://npmjs.com

43081j (approved):
seems good to me 👍 

this is probably the most sensible place to render it for now at least too. if that bar gets too busy eventually, we can split it up a bit

i do agree in future maybe we want a sidebar link instead, but this seems good for now

okineadev:
> seems good to me 👍
> 
> this is probably the most sensible place to render it for now at least too. if that bar gets too busy eventually, we can split it up a bit
> 
> i do agree in future maybe we want a sidebar link instead, but this seems good for now

It's just such a small link that hardly anyone will notice

okineadev:
I noticed a new bug - when there are too many links, they wrap incorrectly to the right instead of the left:

image

PR #236, opened 2026-01-28
chore: fix responsiveness on narrow viewports

iiio2:
In narrow viewports like (iphone se 374 width, galaxy s10 360px width), the website breaks its responsiveness, horizontal scrolling comes. I've tested in mozilla developer edition browser.


iiio2:
Thanks @43081j , @patak-dev  :blush:

PR #241, opened 2026-01-28
fix: y axis label translation

graphieros:
This fixes the 'granularity' part of the translation that was not applied

### DIFF
--- a/app/components/PackageDownloadAnalytics.vue
+++ b/app/components/PackageDownloadAnalytics.vue
@@ -16,7 +16,6 @@ const {
   createdIso: string | null
 }>()
 
-const { t } = useI18n()
 const { accentColors, selectedAccentColor } = useAccentColor()
 const colorMode = useColorMode()
 const resolvedMode = ref<'light' | 'dark'>('light')


app/components/PackageDownloadAnalytics.vue
```diff
@@ -487,7 +488,9 @@ const config = computed(() => ({
       stroke: isDarkMode.value ? '#4A4A4A' : '#a3a3a3',
       labels: {
         axis: {
-          yLabel: $t('package.downloads.y_axis_label', { granularity: selectedGranularity.value }),
+          yLabel: t('package.downloads.y_axis_label', {
+            granularity: t(`package.downloads.granularity_${selectedGranularity.value}`),
```
danielroe:
```suggestion
          yLabel: $t('package.downloads.y_axis_label', {
            granularity: $t(`package.downloads.granularity_${selectedGranularity.value}`),
```

app/components/PackageDownloadAnalytics.vue
```diff
@@ -16,6 +16,7 @@ const {
   createdIso: string | null
 }>()
 
+const { t } = useI18n()
```
danielroe:
```suggestion
```

PR #242, opened 2026-01-28
feat(i18n): add Lunaria for tracking

yanthomasdev:
This PR adds https://lunaria.dev (v1 prerelease) to add translation tracking with key completion support. It also generates a dashboard (its the ejected pre-v1 version, since v1 still doesn't have the new dashboard implemented yet) you can build with `build:lunaria` and access in `dist/lunaria/index.html`.

I've also added the Lunaria GitHub Action, that will comment on any PRs modifying/adding/removing tracked files (`i18n/locales/*.json`) with a summary of changes and warn if the translation/source was changed in another PR and it is now outdated, missing keys, etc. 

Here's how it looks:
image

TODO:

- Decide how we will deploy the dashboard - we can inject inside the Nuxt app by changing its output directory, rebuild the dashboard in Nuxt/Vue, or deploy separately and sync with the app's deployment (e.g. making it https://i18n.npmx.dev).
- Add new languages that are added/update Lunaria config with any languages or content structure that are done before this PR is merged.

Here's an example of the final dashboard:
image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf
--- a/.github/workflows/semantic-pull-requests.yml
+++ b/.github/workflows/semantic-pull-requests.yml
@@ -1,0 +1,35 @@
+name: chore
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+permissions: {}
+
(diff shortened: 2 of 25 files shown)


danielroe:
let's gooooo! 🚀

PR #245, opened 2026-01-28
feat: change i18n configuration

userquin:
This PR includes:
- moves i18n configuration to config folder
- adds support for multiple json files (country variants)
- allows configure `rtl`, `numberFormats`, `dateTimeFormats` and `pluralRule`
- setups initial language from local storage (added to the settings, you may need to clear `npmx-settings` entry + page refresh) via client plugin => now language is restored on page refrresh
- restores stored language from local storage: changed setting page to just bind the settings.language
- added some composables and helpers functions: counting kilo/mega for example (intl compact style => check i18n composable, we may need to add some missing entries if required => time ago for example)
- included `isHydrated` and `wrapI18n` hack at elk.zone
- added script to prepare translation json files for lunaria

I have left the same languages used at elk.zone, keeping just the current used here.

### DIFF
--- a/app/pages/about.vue
+++ b/app/pages/about.vue
@@ -163,42 +163,6 @@ const { data: contributors, status: contributorsStatus } = useFetch<GitHubContri
 
         <div>
           <h2 class="text-xs text-fg-subtle uppercase tracking-wider mb-4">
-            {{ $t('about.vision.title') }}
-          </h2>
-          <ul class="space-y-3 text-fg-muted list-none p-0">
-            <li class="flex items-start gap-3">
-              <span class="text-accent shrink-0 mt-1.5 w-1.5 h-1.5 rounded-full bg-current" />
-              <span>
-                <strong class="text-fg">{{ $t('about.vision.speed_first') }}</strong> &mdash;
-                {{ $t('about.vision.speed_first_description') }}
-              </span>
-            </li>
-            <li class="flex items-start gap-3">
-              <span class="text-accent shrink-0 mt-1.5 w-1.5 h-1.5 rounded-full bg-current" />
-              <span>
-                <strong class="text-fg">{{ $t('about.vision.url_compatible') }}</strong> &mdash;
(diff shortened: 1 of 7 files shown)


userquin:
I'm going to resolve conflicts and merge messages for lunaria 🤞

userquin:
> I would maybe make so `lunaria-json-files` are `lunaria/files` or something so everything concerning Lunaria outside config is in the same folder, but definitely not blocking.

I added that folder to `.gitignore` then removed, we can generate  json files inside the lunaria/files folder 👍

lunaria/prepare-json-files.ts
```diff
@@ -0,0 +1,47 @@
...
+  await fs.mkdir(destFolder)
+  await Promise.all(currentLocales.map(l => mergeLocale(l)))
+}
+
+async function loadJsonFile(name: string) {
+  return JSON.parse(await fs.readFile(path.resolve(`${localesFolder}/${name}`), 'utf8'))
+}
+
+async function mergeLocale(locale: LocaleObject) {
+  console.log(locale)
```
userquin: remove this

app/utils/i18n.ts
```diff
@@ -1,7 +1,9 @@
 import { useI18n as useOriginalI18n } from 'vue-i18n'
 
 export function useI18n() {
-  const { t, d, n, ...rest } = useOriginalI18n()
+  const i18n = useOriginalI18n()
+
+  const { t, d, n, ...rest } = i18n
```
userquin: argg, undo this

userquin:
I'm going to split this PR to include i18n configuration and handle hydration missmatch issues.

skaldebane:
storing the language in local storage works but it means there's always a flash of english UI before the js kicks in
using a cookie would solve this, as it would be rendered in the correct language on the server

this is what i usually use in sveltekit, could take a stab at it, tho i'm not super familiar with nuxt just yet :P

userquin:
closing this, we'll try with another approach: i18n configuration and new JSON files already moved at main

Issue #248, opened 2026-01-28
Unable to navigate using anchor fragments in READMEs


examosa:
On the package detail page, it would appear that markdown headings are not properly converted to anchors which makes navigation within the README nonfunctional.

See [`css-loader`](https://npmx.dev/css-loader#css-loader) as an example compared to its listing on [npm](https://www.npmjs.com/package/css-loader#css-loader).

PR #254, opened 2026-01-28
fix: anchor navigation in details README

Flo0806:
fixes: #248

---

## Summary                                                                                                                                                                                                                                   
- Generate GitHub-style `id` attributes for README headings (enables anchor links like `#installation`)                                                                                                                                                                                                                                                                                           
- Handle duplicate headings with suffixes (`foo`, `foo-1`, `foo-2`)                                                                                                                                                                                                                                                                
- Add smooth scrolling with `prefers-reduced-motion` support                                                                                                                                                                                                                                                                            
- Add `scroll-padding-top` to prevent fixed header from covering anchors

### DIFF
--- a/server/utils/readme.ts
+++ b/server/utils/readme.ts
@@ -184,7 +184,8 @@ function slugify(text: string): string {
 function resolveUrl(url: string, packageName: string, repoInfo?: RepositoryInfo): string {
   if (!url) return url
   if (url.startsWith('#')) {
-    return url
+    // Prefix anchor links to match heading IDs (avoids collision with page IDs)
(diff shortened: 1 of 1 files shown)


danielroe:
that seems good - but what if they collide with our own, like `#getting-started` or `#dependencies`?

Flo0806:
I can check ours and handle possible duplicates... moment

Flo0806:
> that seems good - but what if they collide with our own, like `#getting-started` or `#dependencies`?

Thx. to good old stackoverflow :-) I found a easy to maintain solution

PR #267, opened 2026-01-29
refactor: prefer `VueUse` where possible

9romise:


danielroe (approved):

danielroe:
would you take a look at the a11y regression? 🙏

PR #273, opened 2026-01-29
feat: atproto oauth

### LINKED ISSUE #77: feat: atproto oauth
foundation for social features

zeucapua:
Implements AT Protocol OAuth using `@atproto/api` and `@atproto/oauth-client-node`. The form allows users to login with a handle from the app, create an account through [selfhosted.social](https://selfhosted.social), or go through Bluesky.

Closes #77 

https://github.com/user-attachments/assets/3e6afecf-7326-435a-85ed-9a981f9940d4

Notes:
- requires `NUXT_SESSION_PASSWORD` env variable to encrypt cookies
- login will fail on preview due to the redirect URIs not matching the generated preview URL

### DIFF
--- a/lunaria/files/en-US.json
+++ b/lunaria/files/en-US.json
@@ -359,7 +359,7 @@
       "title": "Local Connector",
       "close_modal": "Close modal",
       "connected": "Connected",
-      "logged_in_as": "Logged in as ~{user}",
+      "connected_as_user": "Connected as ~{user}",
       "connected_hint": "You can now manage packages and organizations from the web UI.",
       "disconnect": "Disconnect",
       "run_hint": "Run the connector on your machine to enable admin features.",
@@ -739,7 +739,24 @@
     "connect_npm_cli": "Connect to npm CLI",
     "connect_atmosphere": "Connect to Atmosphere",
     "connecting": "Connecting...",
-    "ops": "ops"
+    "ops": "ops",
+    "disconnect": "Disconnect"
+  },
+  "auth": {
+    "modal": {
+      "title": "Atmosphere",
+      "close": "Close",
+      "connected_as": "Connected as {'@'}{handle}",
+      "disconnect": "Disconnect",
+      "connect_prompt": "Connect with your Atmosphere account",
+      "handle_label": "Handle",
+      "handle_placeholder": "alice.bsky.social",
+      "connect": "Connect",
+      "create_account": "Create a new account",
+      "connect_bluesky": "Connect with Bluesky",
+      "what_is_atmosphere": "What is an Atmosphere account?",
+      "atmosphere_explanation": "{npmx} uses the {atproto} to power many of its social features, allowing users to own their data and use one account for all compatible applications. Once you create an account, you can use other apps like {bluesky} and {tangled} with the same account."
+    }
   },
   "header": {
     "home": "npmx home",


app/components/AuthModal.vue
```diff
@@ -0,0 +1,134 @@
...
+                    spellcheck="false"
+                    class="w-full px-3 py-2 font-mono text-sm bg-bg-subtle border border-border rounded-md text-fg placeholder:text-fg-subtle transition-colors duration-200 focus:border-border-hover focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-fg/50"
+                  />
+                </div>
+
+                <details class="text-sm">
+                  <summary
+                    class="text-fg-subtle cursor-pointer hover:text-fg-muted transition-colors duration-200"
+                  >
+                    What is an Atmosphere account?
```
43081j: i think we need to figure out a nice way to make account creation very clear, too.

currently if someone doesn't have a bsky account, and has no idea what atproto is, this info doesn't give them much. they need to now dig through the atproto docs and do some googling to find out how to make an account

so we should probably at least mention bsky as an example of how you might have such an account

app/components/AuthButton.vue
```diff
@@ -0,0 +1,19 @@
...
+
+<template>
+  <div class="relative">
+    <button
+      type="button"
+      class="relative font-mono text-sm flex items-center justify-center w-fit rounded-md transition-colors duration-200 hover:bg-bg-subtle focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-fg/50"
+      :aria-label="ariaLabel"
+      @click="showModal = true"
+    >
+      {{ user?.miniDoc?.handle || 'login' }}
```
43081j: `user` itself is a "minidoc", no? so this and the other usages can just be `user?.handle`?

server/api/auth/atproto.get.ts
```diff
@@ -0,0 +1,100 @@
...
+  async set(key: string, val: NodeSavedState) {
+    setCookie(this.event, this.stateKey, btoa(JSON.stringify(val)))
+  }
+
+  async del() {
+    deleteCookie(this.event, this.stateKey)
+  }
+}
+
+export class SessionStore implements NodeSavedSessionStore {
```
43081j: can we move this into its own file somewhere? possibly in `utils`. same for `StateStore`

server/api/auth/atproto.get.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  const { session: authSession } = await atclient.callback(
+    new URLSearchParams(query as Record<string, string>),
+  )
+  const agent = new Agent(authSession)
+  event.context.agent = agent
+
+  const session = await useSession(event, {
+    password: process.env.NUXT_SESSION_PASSWORD as string,
```
43081j: this is being cast because it is `string | undefined` - so i think we should actually just throw at the start of this function if `NUXT_SESSION_PASSWORD` isn't set.

then by this point it'll be guaranteed to be a string.

we should probably do the same in the other handlers too

43081j:
just a couple of minor comments but looks good to me, pretty straightforward! nice work 🎉

app/components/AuthModal.vue
fatfingers23: I'm good if we want to go ahead use [selfhosted.social](https://selfhosted.social) (it's mine and zeu's and I'm the admin of it) PDS for account creations. You start the OAuth flow to the server with the PDS's url and a create flag. With that the user can sign up for an account on the PDS and be redirected back to npmx authenticated. 
This will give users handles ending in `.selfhosted.social`

I am also good if y'all want me to manage a PDS server with an alternate handle ending for npmx, like npmx.town. atproto oauth doesn't like serving apps if the account logging in shares a domain with the PDS. This is also just more branding if we want them to have that ending in the handle

server/api/auth/session.delete.ts
```diff
@@ -0,0 +1,9 @@
+export default defineEventHandler(async event => {
+  const session = await useSession(event, {
+    password: process.env.NUXT_SESSION_PASSWORD as string,
```
jonathanyeong: Should we create an example `.env` file that we can commit so people know that they can set `NUXT_SESSION_PASSWORD` without digging into the code? 

There's also [runtime config](https://nuxt.com/docs/4.x/guide/going-further/runtime-config#environment-variables) (not sure if that's useful for this case)

server/api/auth/atproto.get.ts
```diff
@@ -0,0 +1,100 @@
...
+  })
+
+  const response = await fetch(
+    `https://slingshot.microcosm.blue/xrpc/com.bad-example.identity.resolveMiniDoc?identifier=${agent.did}`,
+  )
```
fatfingers23, line 58: Might be a good idea to add a user agent here. Helps fig out

app/components/AuthButton.vue
zeucapua: Don't know why, but that minidoc prop is required, otherwise it'll be undefined and always say 'login' 🤷‍♀️

43081j: pinged you on discord with the solution

server/api/auth/atproto.get.ts
```diff
@@ -0,0 +1,62 @@
...
+  const agent = new Agent(authSession)
+  event.context.agent = agent
+
+  const session = await useSession(event, {
+    password: process.env.NUXT_SESSION_PASSWORD,
+  })
+
+  const response = await fetch(
+    `${SLINGSHOT_ENDPOINT}/xrpc/com.bad-example.identity.resolveMiniDoc?identifier=${agent.did}`,
```
jonathanyeong, line 56: For my knowledge, why does the endpoint contain `bad-example`?

fatfingers23: That's [fig](https://bsky.app/profile/bad-example.com) lol. XRPC's need a nsid to be complete and they used their domain for it 

https://atproto.com/specs/xrpc#lexicon-http-endpoints

shared/utils/fetch-cache-config.ts
```diff
@@ -24,6 +26,9 @@ export const FETCH_CACHE_ALLOWED_DOMAINS = [
   'api.bitbucket.org', // Bitbucket API
   'codeberg.org', // Codeberg (Gitea-based)
   'gitee.com', // Gitee API
+  //microcosm endpoints for atproto data
```
Kai-ros: NIT:
```suggestion
  // microcosm endpoints for atproto data
```

server/utils/atproto/storage.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+}
+
+/**
+ * Storage key prefix for oauth session storage.
+ */
+export const OAUTH_SESSION_CACHE_STORAGE_BASE = 'oauth-atproto-session'
+
+export class OAuthSessionStore implements NodeSavedSessionStore {
+  //TODO not sure if we will support multi accounts, but if we do in the future will need to change this around
```
Kai-ros: NIT:
```suggestion
  // TODO: not sure if we will support multi accounts, but if we do in the future will need to change this around
```

server/utils/atproto/storage.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  async del() {
+    let sessionKey = getCookie(this.event, this.cookieKey)
+    if (sessionKey) {
+      await this.storage.del(sessionKey)
+    }
+    deleteCookie(this.event, this.cookieKey)
+  }
+}
```
Kai-ros, line 82: Could we add a factory at the bottom to clean up the API a bit?
```
export const useOAuthStorage = (event: H3Event) => {
  return {
    states: new OAuthStateStore(event),
    sessions: new OAuthSessionStore(event),
  }
}
```
That way when it gets invoked elsewhere, like you do above in `server/api/auth/atproto.get.ts`, it could be as simple as `const storage = useOAuthStorage(event)`.

server/api/auth/session.delete.ts
```diff
@@ -0,0 +1,17 @@
+export default eventHandlerWithOAuthSession(async (event, oAuthSession) => {
+  if (!process.env.NUXT_SESSION_PASSWORD) {
+    throw createError({
+      status: 500,
+      message: 'NUXT_SESSION_PASSWORD not set',
```
Kai-ros: NIT: Extract to constants file
```suggestion
      message: UNSET_NUXT_SESSION_PASSWORD,
```

server/api/auth/session.get.ts
```diff
@@ -0,0 +1,14 @@
+export default defineEventHandler(async event => {
+  if (!process.env.NUXT_SESSION_PASSWORD) {
+    throw createError({
+      status: 500,
+      message: 'NUXT_SESSION_PASSWORD not set',
```
Kai-ros: NIT: Extract to constants file
```suggestion
      message: UNSET_NUXT_SESSION_PASSWORD,
```

server/api/auth/atproto.get.ts
```diff
@@ -0,0 +1,62 @@
...
+import { NodeOAuthClient } from '@atproto/oauth-client-node'
+import { createError, getQuery, sendRedirect } from 'h3'
+import { OAuthSessionStore, OAuthStateStore } from '#server/utils/atproto/storage'
+import { SLINGSHOT_ENDPOINT } from '#shared/utils/constants'
+
+export default defineEventHandler(async event => {
+  if (!process.env.NUXT_SESSION_PASSWORD) {
+    throw createError({
+      status: 500,
+      message: 'NUXT_SESSION_PASSWORD not set',
```
Kai-ros, line 13: NIT: Extract to constants file
```suggestion
      message: UNSET_NUXT_SESSION_PASSWORD,
```

server/api/auth/atproto.get.ts
Kai-ros, line 1: We should add validation here since we're stepping outside of the system, the query params and a third-party API. Casting these with `as` could hide potential runtime crashes if for whatever reason the PDS or Slingshot return unexpected shapes.

1. `AuthRequestSchema` or `AuthQuerySchema` - For the validate `getQuery(event)`
2. MiniDocSchema - For the fetch response from the Slingshot endpoint

server/routes/oauth-client-metadata.json.get.ts
Kai-ros, line 1: As I understand it, this will be exposed to the public via the ATProto network? Would be nice to validate the output  with a schema here as well. Just to guard against breakage for the OAuth handshake with an invalid URL or missing field.

server/api/auth/session.get.ts
Kai-ros, line 1: If we add a validation schema here we get the benefit of not having to perform repeat manual checks for `NUXT_SESSION_PASSWORD` and having a `SessionDataSchema` parse the` session.data` ensures the FE gets a reliable shape. Bonus is we won't accidentally leak any internal session metadata added down the line.

danielroe:
we can ignore the provenance warnings btw

fatfingers23:
Dropped a pr going to @zeucapua's branch to address some of the items to get them into this pr
- Comments formatting
- useOAuthStorage factory
- moved the session stuff to `eventHandlerWithOAuthSession` to address those and give a shared area

I did not address
- schema/query validation will probably do that in another PR. I got something to add to oauth called a confidential client where logins last longer than 2 weeks. This initial pr is probably getting a bit long in the tooth with how fast the repo is moving and I have a feat branch i'm basing off of it as well. would like to merge sooner than later
- Was some debate about the login screen wording and explaining the idea that npmx is now a full atproto app, but "connects" users. I can take a shot at that but thought someone else may like to

https://github.com/zeucapua/npmx.dev/pull/3

fatfingers23:
I think we are good to get this merged when someone has a chance to take a final look! 🥳

danielroe (approved):

PR #275, opened 2026-01-29
feat: save locale in cookies, set `lang` and `dir` html attrs

skaldebane:
Saves locale in cookies using `detectBrowserLanguage`, and sets `lang` and `dir` html attributes.


danielroe:
we won't use cookies as this isn't compatible with caching the pages.

thank you for the PR, though ❤️

PR #282, opened 2026-01-29
refactor(i18n): externalize and organize hardcoded strings

shuuji3:
I found strings and ARIA labels that were hardcoded and not translatable. 

This PR replaces hardcoded text and labels in Vue components and adds new keys to `en.json`.


danielroe (approved):
thank you ❤️

PR #292, opened 2026-01-29
chore: update "Help translate" link

shuuji3:
Now we have lunaria page that includes the current translation status and the contribution guide.


danielroe:
see https://github.com/npmx-dev/npmx.dev/pull/289 - what do you think?

shuuji3:
@danielroe this is awesome and very informative! I want to implement this to Elk too 😀 

image

Issue #287, opened 2026-01-29
When typing in to the root search box  during the transtion from / to /search focus is lost


orta:
- Open up `https://npmx.dev`
- start typing 'hello' 
- get redirected with animation to `https://npmx.dev/search?q=hel` 
- lose 'lo' due to focus being lost

Cool project!

garthdw:
Seeing the same on Firefox on MacOS

https://github.com/user-attachments/assets/f11b801d-cc7e-4049-b412-713388246374

Trying to type `projen`

PR #293, opened 2026-01-29
fix: debounce navigation to the search page

iiio2:
resolves https://github.com/npmx-dev/npmx.dev/issues/287


iiio2:
Thanks. :blush:

PR #295, opened 2026-01-29
feat: use accent color for checkbox and text inputs

bluwy:
Added accent colors in these places which I think would be nice to have:

image

image
image


danielroe:
I think we should use it almost everywhere we have focus rings

bluwy:
There's the `focus-ring` unocss shortcut that we can use more I think.

PR #257, opened 2026-01-28
feat: Blog Posts (WIP)

Kai-ros:
## Overview

This PR is an omnibus of all the tasks being tackled to resolve [ISSUE 176](https://github.com/npmx-dev/npmx.dev/issues/176).

## Changes

- [x] Lexicons
- [x] Prose-styling using unocss typography
- [x] anthony's markdown [solution](https://github.com/unplugin/unplugin-vue-markdown) 
- [x] Standard site push - (Currently mocking the PDS push) 
- [x] Redesign emails templates w/npmx visual language
- [ ] TBD...

## See Also

- Daily Project Updates - [HackMD](https://hackmd.io/@e7pGbT7hTrONe4H59uAIPg/ByMdCwoIbg)

PR #297, opened 2026-01-29
feat: add standard site lexicons to project

jonathanyeong:
This PR pulls out the standard site lexicons generated in the [setup blog PR](https://github.com/npmx-dev/npmx.dev/pull/257/changes#diff-3c789d8cdf22446066f39297c2c23857bf389c17cabd1cd55ca2272b34b62ca5) just in case this was blocking @fatfingers23 and @zeucapua work.

### DIFF
--- a/modules/lunaria.ts
+++ b/modules/lunaria.ts
@@ -27,9 +27,17 @@ export default defineNuxtModule({
     if (!isCI || !existsSync(lunariaDistPath)) {
       mkdirSync(lunariaDistPath, { recursive: true })
       nuxt.hook('nitro:build:before', async () => {
-        execSync('node --experimental-transform-types ./lunaria/lunaria.ts', {
-          cwd: nuxt.options.rootDir,
-        })
+        try {
+          execSync('node --experimental-transform-types ./lunaria/lunaria.ts', {
+            cwd: nuxt.options.rootDir,
+          })
+        } catch (e) {
+          // do not throw when building for staging
+          const env = process.env.VERCEL_ENV
+          if (!isCI || (env && env === 'production')) {
+            throw e
+          }
+        }
       })
     }
   },


shared/types/lexicons/site.ts
```diff
@@ -0,0 +1,5 @@
+/*
+ * THIS FILE WAS GENERATED BY "@atproto/lex". DO NOT EDIT.
+ */
+
+export * as standard from './site/standard.js'
```
danielroe: do we want to gitignore the generated files?

jonathanyeong: @fatfingers23 mentioned that it was okay to commit the generated types and we can generate new ones as we go. Might let Bailey chime in. I'm easy either way!

fatfingers23: We're good to. The lexicon schema files should be the source of truth and can generate types/clients from it in any language. Double checked the atproto repo and they're doing the same

https://github.com/bluesky-social/atproto/blob/main/packages/api/src/client/types/com/atproto/sync/getRepo.ts

fatfingers23:
LGTM! And thank you for breaking those out. That does help.

shared/types/lexicons/site.ts
fatfingers23: whoops sorry. I should of been more specific. We're good to leave the generated files in the project, it's what the Bluesky team and others do. But I also don't think it's bad if we remove the either and depend on the lex generation tool

lunaria/files/en-US.json
```diff
@@ -573,7 +573,7 @@
     },
     "contributors": {
       "title": "Contributors",
-      "description": "npmx is fully open source, built by an amazing community of contributors.",
+      "description": "npmx is fully open source, built by an amazing community of contributors. Join us and let's build the npm browsing experience we always wanted, together.",
```
jonathanyeong, line 576: @danielroe hmm not sure how this was pulled in 🤔.

PR #298, opened 2026-01-29
fix(i18n): add all missing translations for German

ameetmadan:


ameetmadan:
@danielroe the site still shows missing translations, do have any idea why? 

image

PR #299, opened 2026-01-29
chore: ignore debounce for empty input

iiio2:
follow-up https://github.com/npmx-dev/npmx.dev/pull/293 
There is one condition I did not notice well. I think when the input field is empty, if we click on search button we can directly navigate to "/search". On that time we do not need debounce.

### DIFF
--- a/lunaria/files/en-US.json
+++ b/lunaria/files/en-US.json
@@ -569,7 +569,7 @@
     },
     "contributors": {
       "title": "Contributors",
-      "description": "npmx is fully open source, built by an amazing community of contributors.",
+      "description": "npmx is fully open source, built by an amazing community of contributors. Join us and let's build the npm browsing experience we always wanted, together.",
       "loading": "Loading contributors...",
       "error": "Failed to load contributors",
       "view_profile": "View {name}'s GitHub profile"


danielroe (commented):
what's the bug? doesn't debouncedNavigate still navigate even if it's empty?

iiio2:
@danielroe,  I think it takes slightly more time after clicking on the search button. It may affect the UX. If we click on Search button, it should navigate immediately to "/search". The debounce may need to typing something in the input field. It should not take time when we click on the search button when input is empty.

app/pages/index.vue
```diff
@@ -13,7 +13,16 @@ const debouncedNavigate = debounce(() => {
...
+  // If triggered by button click (not input), navigate immediately when empty
+  const isButtonClick = event?.type === 'click' || event?.type === 'submit'
+  const isEmpty = !searchQuery.value.trim()
+
+  if (isButtonClick && isEmpty) {
+    router.push('/search')
+    return
+  }
+
   debouncedNavigate()
```
danielroe: what about just:

```suggestion
  return searchQuery.value.trim() ? debouncedNavigate() : router.push('/search')
```

not sure why we need to care about button click vs enter....

iiio2:
Thanks. :blush:

PR #308, opened 2026-01-29
chore: pin github actions versions to commit hash

43081j:
Also turns off install scripts entirely.

Minor nice to have for security.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
           cache: pnpm
 
       - name: 📦 Install dependencies
-        run: pnpm install --ignore-scripts
+        run: pnpm install
 
       - name: 🔠 Lint project
         run: pnpm lint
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -103,15 +103,15 @@ jobs:
...
       - run: corepack enable
-      - uses: actions/setup-node@v6
+      - uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0
         with:
           node-version: lts/*
           cache: pnpm
 
       - name: 📦 Install dependencies
-        run: pnpm install
+        run: pnpm install --ignore-scripts
```
danielroe: should we set this in `pnpm-workspace.yaml` instead, with `ignoreDepScripts`?

(ignore-scripts just ignores scripts from _us_ not our dependencies)

43081j: we could do, ill update 👍

okineadev:
We need to configure Renovate too, so it will do it automatically, right?

43081j:
it'll do it automatically apparently 👍

PR #311, opened 2026-01-29
fix: improve labelling of landmarks

knowler:
I did a very quick pass of the markup structure and made a few changes that immediately stood out:

## Removes unnecssary `aria-label` for `banner`/`contentinfo` landmarks

Removed the label from the body-scoped `<header>` element. When scoped to the body and not the `<main>` element or other sectioning content (e.g. `<article>`, `<header>`, `<footer>`, `<nav>`, `<aside>`, `<section>`), a `<header>` element will implicitly have the role of `banner` which is a landmark role. According to [the ARIA spec for the `banner` role](https://w3c.github.io/aria/#banner):

> The author _SHOULD_ mark no more than one element on a page with the banner role.

Accessible names are typically used to disambiguate elements with the same role and since there is only one element with this role, then the label is unnecessary.

The same applies to the body-scoped `<footer>` element which has [the `contentinfo` landmark role](https://w3c.github.io/aria/#contentinfo).

I should add that it is totally fine to use `<header>` and `<footer>` elements within the `<main>` element like you are.

## Removes redundant “navigation” label from navigation landmarks

When assistive technology presents an element with the `navigation` landmark role (e.g. `<nav>` elements) it includes the role in the presentation. For example, a screen reader will read “navigation” or, for a labelled landmark, “Main navigation.” So, adding “navigation” to the label is redundant, since a screen reader would read “Main navigation navigation” (and include it that way in its rotor/landmark list).

### DIFF
--- a/app/components/AppHeader.vue
+++ b/app/components/AppHeader.vue
@@ -51,9 +51,7 @@ onKeyStroke(',', e => {
 </script>
 
 <template>
-  <header
-    class="sticky top-0 z-50 bg-bg/80 backdrop-blur-md border-b border-border"
-  >
+  <header class="sticky top-0 z-50 bg-bg/80 backdrop-blur-md border-b border-border">
     <nav :aria-label="$t('nav.main_navigation')" class="container h-14 flex items-center">
       <!-- Left: Logo -->
       <div class="flex-shrink-0">


knowler:
@danielroe Oops, I didn’t realize syncing with the base branch would cancel that automerge.

danielroe:
no worries - just fyi, the merge queue will sync automatically and make sure tests pass when branch is up-to-date, so no need to sync unless there are conflicts

PR #313, opened 2026-01-29
feat: keep search box in header

### LINKED ISSUE #291: Stable search bar on top navbar
The new always visible search bar makes such a big difference, amazing how these details changes the way we use the app.

We should take it one step further. Imagine if the search bar was always in the navbar instead of losing all the height space for the current Search block.

Image

We would then not only have more space for results, but there will also not be any view transition movement when searching for new packages while looking at a package page. The search bar will stay up there, in the same way the url bar works for search in the browser. The effect will be more of results fading in
and if we do that, we could do the search in the navbar a tiny bit bigger if needed, but to be honest, looks great to me as is.

garthdw:
Avoid taking up space for searchbox on search page.

Mainly moved the query string watching from search.vue to AppHeader.vue

Closes #291 

https://github.com/user-attachments/assets/ea5ef4f6-b235-4379-a769-f22d9bf2734d

### DIFF
--- a/tests/url-compatibility.spec.ts
+++ b/tests/url-compatibility.spec.ts
@@ -76,7 +76,7 @@ test.describe('npmjs.com URL Compatibility', () => {
       await goto('/~sindresorhus', { waitUntil: 'domcontentloaded' })
 
       // Should show username
-      await expect(page.locator('h1')).toContainText('@sindresorhus')
+      await expect(page.locator('h1')).toContainText('~sindresorhus')
       // Should show packages heading (user has packages)
       await expect(page.getByRole('heading', { name: 'Packages' })).toBeVisible()
     })


danielroe:
might need to update the current search behaviour tests 🙏

tests/interactions.spec.ts
```diff
@@ -6,12 +6,13 @@ test.describe('Search Pages', () => {
...
 
-    const searchInput = page.locator('input[type="search"]')
-    await expect(searchInput).toBeFocused()
-
     const firstResult = page.locator('[data-result-index="0"]').first()
     await expect(firstResult).toBeVisible()
 
+    const searchInput = page.locator('input[type="search"]')
+    // manually focus search
+    await searchInput.focus()
```
danielroe: I think we want it to autofocus when the search page is opened.

garthdw: I have gone with autofocus on the search input.

I did start down a hacky approach to only focus when AppHeader is mounted and route is search. I can switch to that approach if we want it to only happen on the search page instead of any "first load" page.

danielroe (approved):
❤️

danielroe:
we should think about mobile view too....

currently there's no search input on mobile

might want to have a small input that goes full width when focused?

garthdw:
This is where I got to for mobile

https://github.com/user-attachments/assets/fc8b9360-afd6-4ebe-85da-c1de7c354b8e

danielroe:
looks perfect to me!

PR #314, opened 2026-01-29
fix: remove unnecessary `search` landmark nested in `<search>`

knowler:
The [`<search>` element’s implicit ARIA role is `search`](https://w3c.github.io/html-aam/#el-search), so setting `role=search` on the nested `<form>` is redundant and adds noise (i.e. it will show up twice).

If you want to [increase support](https://caniuse.com/mdn-html_elements_search), then the role should be added to the `<search>` element instead.


knowler:
This might need to be amended when/if #313 merges since it’s modifying the same bit of code.

knowler:
Oh no, looks like some weird autoformatting happened after I installed the Git hooks…

UPDATE: I force pushed a `--no-verify` commit to only reflect my changes. The CI can fix the formatting.

PR #317, opened 2026-01-29
feat(ui): replace all left/right css styles for rtl support

userquin:
This PR includes:
- replaces all margins and paddings:  `ml-`, `mr-`, `pl-` and `pr-` with `ms-`, `me-`, `ps-` and `pe-` respectivelly
- replaces all texts alignments: `text-left` and `text-right` with `text-start` and `text-end` respectivelly
- replaces all abosulte positions: `left-` and `right-` with `inset-is-` and `inset-ie` respectivelly
- replaces all borders and rounded borders: `border-l`, `border-l-`,  `border-r`, `border-r-`, `rounded-l`and `rounded-r` with `border-is`, `border-is-`,  `border-ie`, `border-ie-`, `rounded-is` and `rounded-ie` respectivelly
- adds `rtl-flip` to icons that have left or right direction: will require a deep review to check every icon
- changes buttons styles for role switch at setting page: using RTL have wrong translate-x values => just added scoped style using `html[dir="rtl"]` to change the span inside the last span containing the check; replaced also justify-between with justify-start style adding a new span between the text and the check to allow swap positions on RTL
- changed AppFooter component to add the same hack with justify-between for the links container (docs, source, social... links).
- changes all icons found replacing previous styles to use `i-<collection>:` when using `i-<collection>-` (use `:` instead `-` for separator)

I still need to review a few entries:`left-full` and `right-full` and review all the icons that have left or right direction t add `rtl-flip` class.

You can check this PR on dev server adding `dir="rtl"` in the html element, we need to add `ar` and `ar-EG` locales.

Maybe this PR can be used as guide if you don't want that ammount of changes.

### DIFF
--- a/app/assets/main.css
+++ b/app/assets/main.css
@@ -80,6 +80,10 @@ html {
   }
 }
 
+html[dir='rtl'] .rtl-flip {
+  transform: scale(-1, 1);
+}
+
 body {
   margin: 0;
   background-color: var(--bg);
--- a/app/components/AppFooter.vue
+++ b/app/components/AppFooter.vue
@@ -1,54 +1,56 @@
+<script setup lang="ts"></script>
 <template>
   <footer class="border-t border-border mt-auto" :aria-label="$t('footer.site_footer')">
     <div class="container py-3 sm:py-8 flex flex-col gap-2 sm:gap-4 text-fg-subtle text-sm">
-      <div class="flex flex-col sm:flex-row items-center justify-between gap-2 sm:gap-4">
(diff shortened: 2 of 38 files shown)


userquin:
I'm going to send small PR for every page and component.

danielroe:
I'm happy if you want to batch these prs up rather than per component.

userquin:
> I'm happy if you want to batch these prs up rather than per component.

I want to check if there is conflict, will send pr per page/component

userquin:
closing, all inidivual PR merged (PackageCard queued)

PR #334, opened 2026-01-29
feat: add RTL support to `OrgTeamsPanel`

userquin:
releated to #317

### DIFF
--- a/app/components/OrgTeamsPanel.vue
+++ b/app/components/OrgTeamsPanel.vue
@@ -261,12 +261,13 @@ watch(lastExecutionTime, () => {
     class="bg-bg-subtle border border-border rounded-lg overflow-hidden"
   >
     <!-- Header -->
-    <div class="flex items-center justify-between p-4 border-b border-border">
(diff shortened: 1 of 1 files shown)


app/components/OrgTeamsPanel.vue
```diff
@@ -261,12 +261,13 @@ watch(lastExecutionTime, () => {
     class="bg-bg-subtle border border-border rounded-lg overflow-hidden"
   >
     <!-- Header -->
-    <div class="flex items-center justify-between p-4 border-b border-border">
+    <div class="flex items-center justify-start p-4 border-b border-border">
```
danielroe, line 264: @knowler is this the same issue as #340?

danielroe (approved):

app/components/OrgTeamsPanel.vue
knowler: Yep, I’ve noticed a few other places where it can be dropped as well. I’ll make a PR after I get Vue devtools installed (or whatever you all use to see components).

PR #339, opened 2026-01-29
fix(i18n): pass scope to components

BobbieGoede:
This prevents these warning logs:
image

### DIFF
--- a/app/pages/about.vue
+++ b/app/pages/about.vue
@@ -56,7 +56,7 @@ const { data: contributors, status: contributorsStatus } = useFetch<GitHubContri
             {{ $t('about.what_we_are.title') }}
           </h2>
           <p class="text-fg-muted leading-relaxed mb-4">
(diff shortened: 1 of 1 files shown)


userquin:
shouldn't be global if not specified? at elk.zone we don't have any `i18n-t`  with scope global 🤔

BobbieGoede:
@userquin yeah it falls back to global by default, this is just to get rid of the warnings

cernymatej:
@BobbieGoede potentially a suggestion for a breaking change, but why isn't `global` the default (and not just a fallback) - i.e. no wraning?
I personally never use local scopes at all and I think that most users don't either. (might be mistaken, though...)

BobbieGoede:
@cernymatej this behavior is from vue-i18n, by default the scope is inherited from its parent which could be global or local.

I think the warning was added because this behavior was confusing, without making a breaking change. I agree that a static default scope would probably be more intuitive.

PR #340, opened 2026-01-29
refactor: use `justify-content: space-between` in `AppFooter`

knowler:
In #318, `justify-content: space-between` was removed unnecessary and a hack was introduced in its place.

[Flexbox layout already is directionality/writing-mode aware](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Flexible_box_layout/Relationship_with_other_layout_methods#writing_modes) and `justify-content` has no bearing on the order of items, [just their placement within the “main” axis](https://css-tricks.com/snippets/css/a-guide-to-flexbox/#aa-justify-content) (which is what gets flipped for `dir=rtl`; for a visual see: https://rtlstyling.com/posts/rtl-styling/#flexbox-layout-module).

This switches back to what it was before which is simpler as it doesn’t add extra elements or presentational, spacer elements that are hidden with ARIA.

To test, set `dir=rtl` on the `<html>` element and compare with the live site. Visually there should be no change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -242,7 +242,7 @@ npmx.dev uses [@nuxtjs/i18n](https://i18n.nuxtjs.org/) for internationalization.
 ### Approach
 
 - All user-facing strings should use translation keys via `$t()` in templates and script
-- Translation files live in `i18n/locales/` (e.g., `en-US.json`)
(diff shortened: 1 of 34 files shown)


knowler:
Merge conflicts between this and #342 are resolved in [31e1c78](https://github.com/npmx-dev/npmx.dev/pull/340/commits/31e1c78257db7c4d33f5cfeda8a058507508af42).

PR #360, opened 2026-01-29
feat(i18n): add cs support 

VentyCZ:
Adds support for Czech language. Also fixes a description of an option on the settings page - used the same key for caption and description.


userquin:
run `pnpm build:lunaria` from root, you will need to add  `lunaria/files/cs-CZ.json` to git

VentyCZ:
It's not perfect, but close enough 😄

PR #363, opened 2026-01-29
fix: tighten code display for ascii art

posva:
I noticed that some ascii art wasn't dispalying correctly. I didn't test this throughout but it seemed like a quick fix.

Before the fix:

CleanShot 2026-01-29 at 22 42 07@2x

After the fix:

CleanShot 2026-01-29 at 22 42 14@2x

red lines are for debug

Checked on https://npmx.dev/table

### DIFF
--- a/app/assets/main.css
+++ b/app/assets/main.css
@@ -281,6 +281,9 @@ html.light .shiki span {
   white-space: pre;
   word-break: normal;
   overflow-wrap: normal;
+  /* Makes unicode and ascii art work properly */
+  line-height: 1.25;
+  display: inline-block;
 }
 
 .readme-content ul,


okineadev:
I see this on my phone right now:

| | | |
|:-:|:-:|:-:|
||||

posva:
I can't repro on ios, they do appear slightly off on both pc and phone
CleanShot 2026-01-30 at 09 29 00

I think this is because of the mono font itself, maybe something about fallbacks on your device?

Using `monospace`, it looks better:

CleanShot 2026-01-30 at 09 30 58

okineadev:
> I can't repro on ios, they do appear slightly off on both pc and phone
> CleanShot 2026-01-30 at 09 29 00
> 
> I think this is because of the mono font itself, maybe something about fallbacks on your device?
> 
> Using `monospace`, it looks better:
> 
> CleanShot 2026-01-30 at 09 30 58
> 

So there may be a problem with the fonts

We have to look into this

PR #375, opened 2026-01-30
feat(i18n): improve translations in `settings` section and add `i18n` section for `zh-CN`

HoshinoSuzumi:


9romise:
Oops, looks like we're working on the same thing. I'll remove the overlapping changes we’ve both made. Before that, I think we could review each other’s changes to produce a better final translation.
See #377

i18n/locales/zh-CN.json
```diff
@@ -44,15 +44,37 @@
...
+    "title": "设置",
+    "tagline": "定制你的 npmx 体验",
+    "meta_description": "通过主题、语言和显示偏好，定制你的 npmx.dev 体验。",
+    "sections": {
+      "appearance": "外观",
+      "display": "显示",
+      "language": "语言"
+    },
     "relative_dates": "相对时间",
+    "relative_dates_description": "显示 \"三天前\" 而不是完整日期",
```
9romise:
```suggestion
    "relative_dates_description": "显示 \"3天前\" 而不是完整日期",
```
align with the actual display.

PR #376, opened 2026-01-30
fix(i18n): move keys to the correct scope

9romise:
related to #374
/cc @userquin


userquin:
arggg, I use Gemini PRO for both sorry 🙏  and didn't check the structure for es-419.json 🥺, looks like we also need to change "rw" and "ro" (just remove them, vue-i18n will use the English version), tomorrow will do a deep review to es.json.

userquin (approved):
If the structure I added was wrong... 

LGTM

PR #380, opened 2026-01-30
fix: styling issues in light mode

### LINKED ISSUE #191: Filter dropdowns in sparkline have the wrong coloring
I don't know how this works, just reporting my experience. Feel free to close if I'm being dumb!

Image

RYGRIT:
Fixes #191.
However, I think this solution isn't elegant enough, and I haven't found a better one. I would appreciate some guidance.

At the same time, dynamic color scheme support has been added to the date input field.

Before:
75be22a9-a9ef-4a3e-9067-79ccd8af067e

After:
72c72d20-c0a5-4e2a-a54b-7c0213c5cd54

Verified on Windows (Light/Dark mode).

### DIFF
--- a/app/assets/main.css
+++ b/app/assets/main.css
@@ -154,7 +154,7 @@ button {
 
 /* Skip link */
 .skip-link {
-  position: absolute;
+  position: fixed;
   top: -100%;
   left: 0;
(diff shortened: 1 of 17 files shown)


danielroe (approved):
thank you!

PR #390, opened 2026-01-30
fix: horizontal scrollbar for table view + columns picker

iiio2:
In the table view mode, we cannot see all the data. A horizontal scrollbar is given so that we can see all the data in mobile version. Also we did not see data in columns picker in mobile version.


iiio2:
Thanks @serhalp , @danielroe

Issue #394, opened 2026-01-30
Update the screenshots in the readme


btea:
|current|latest|
|---|---|
|<img width="1435" height="890" alt="Image" src="https://github.com/user-attachments/assets/4a4edb9b-64f4-4771-89a3-c5762bdaac4c" />|<img width="1997" height="1198" alt="Image" src="https://github.com/user-attachments/assets/4e7b7493-9bac-44db-8d39-6025acdcbf6b" />|

The latest page displays richer content, and some display issues existed in the earlier screenshots, such as the area circled in red in the image above.

Flo0806:
<img width="2105" height="1514" alt="Image" src="https://github.com/user-attachments/assets/1a2a3205-0227-46dc-b1f9-48f9a65691d3" />

Thx.

PR #398, opened 2026-01-30
chore: update screenshot

Flo0806:
resolves: #394 

---

update screenshot to current version


Flo0806:
That's the magic of Windows... not.. Now it should be much better.

Issue #397, opened 2026-01-30
Search input loses focus after transition


gunnnnii:
When using the search on the front page on mobile safari, the transition to the search screen causes the focus to be lost. This disrupts your typing a bit.

https://github.com/user-attachments/assets/14fa5e64-48af-4834-858c-08770aef9cbf

MatteoGabriele:
This may be related to #286 as well, so once this fix is merged, I'll check again to confirm the issue is resolved.

PR #399, opened 2026-01-30
fix: search field losing focus after transition

Flo0806:
fixes: #397 

---

add a watcher to handle visible state after transition correctly and set focus


danielroe (approved):

danielroe:
we already test focus in the browser tests - would you see why it's failing, and add a test case if applicable? 🙏

Flo0806:
The test check the focus on the settings page only. The issue happened while the route changing from home - settings + running transition. I added a new test which is failing for me without the fix.

PR #408, opened 2026-01-30
refactor: remove redirect while typing

MatteoGabriele:
Personally, I think that redirecting while typing creates different issues:
- The input loses focus, which might interrupt the typing experience
- The search input animation is a little jarring, in my opinion, even if I do not have motion sensitivity (or maybe I do)
- It's more predictable


app/pages/index.vue
```diff
@@ -1,23 +1,15 @@
...
-
+const isMobile = useIsMobile()
 const router = useRouter()
 const searchQuery = shallowRef('')
-const searchInputRef = useTemplateRef('searchInputRef')
-const { focused: isSearchFocused } = useFocus(searchInputRef)
 
-const isMobile = useIsMobile()
-
-const debouncedNavigate = debounce(() => {
```
danielroe: what about instead using a trailing rather than leading debounce? it would likely have the same effect....

MatteoGabriele: But we don't actually redirect unless the user presses Enter or clicks "Search".
This makes the experience less _unexpected_ in my opinion, since people might be trying to remember things while typing. 

PS: For the trim, I just remembered that v-model has a `.trim` modifier built in.

MatteoGabriele: @danielroe, do you want me to try anything in particular for this one? Maybe I didn't fully understand your comment

danielroe: no, I fixed that in a separate PR: https://github.com/npmx-dev/npmx.dev/pull/526

does that fix any of the issues you were trying to fix here?

MatteoGabriele: The focus issue seems to be resolved with your PR. 👍 
There’s also a UX preference question around whether the animation should exist at all, which is more subjective. If you’d prefer, we can close this PR and keep the animation as-is.

MatteoGabriele: Update: I have encountered the missing letter again when typing. It's hard to reproduce, but it's out there somewhere. just FYI

MatteoGabriele:
There are too many updates, and my changes no longer make sense in this PR. 
I will close it for now. Maybe we can discuss whether to keep the transition, particularly if it worsens the typing experience, such as missing letters or other issues.

Issue #407, opened 2026-01-30
"This Serverless Function has crashed." when visiting URL with typo


SivaramPg:
Example - https://npmx.dev/packages/pwa-icons

Used `/packages` instead of `/package`. 

Ideally should throw 404 error if path is not found.

<img width="1200" height="690" alt="Image" src="https://github.com/user-attachments/assets/d53902a7-2adf-4522-b97a-aec2ee1630fd" />

PR #418, opened 2026-01-30
fix: throw 404 statusCode instead of 400

Flo0806:
fixes: #407 

---

This should handle the error more graceful for vercel and should show the error.vue instead of vercel's server function crashed page.


Flo0806:
And I thought: should I change all to 404 😅

PR #420, opened 2026-01-30
chore(deps): upgrade dependencies

danielroe:


pnpm-workspace.yaml
```diff
@@ -7,8 +7,8 @@ savePrefix: ''
 
 overrides:
   sharp: 0.34.5
-  vite: npm:@voidzero-dev/vite-plus-core@latest
-  vitest: npm:@voidzero-dev/vite-plus-test@latest
+  vite: npm:@voidzero-dev/vite-plus-core@0.0.0-833c515fa25cef20905a7f9affb156dfa6f151ab
+  vitest: npm:@voidzero-dev/vite-plus-test@0.0.0-833c515fa25cef20905a7f9affb156dfa6f151ab
```
hacknug: We have these overrides but `package.json` already points to the same dep+version?

hacknug (approved):
I can finally run the project on my Intel MacBook :tada:

pnpm-workspace.yaml
danielroe: without overrides, things like `@nuxt/vite-builder` or `@nuxt/test-utils` would complain about missing peerDeps, or just pull in the wrong version.

hacknug: Got it. Thanks for explaining 💖

Issue #410, opened 2026-01-30
Layout spacing breaks for long package names


MatteoGabriele:
When the package name is longer than most one-word packages, the spacing between the title, tabs, and description gets a bit off.

<img width="872" height="275" alt="Image" src="https://github.com/user-attachments/assets/ff9fc325-e5f0-4a6a-b57e-73b76bb83d80" />
<img width="847" height="222" alt="Image" src="https://github.com/user-attachments/assets/3bfdf627-b64a-45e6-9264-6d548f04fd55" />

Flo0806:
I guess this is good as it is, isn't it? it change the design on breakpoints. I can read it on each size. I understand what you mean - But the tabs are placed on right side, and the chips/badges are placed behind the name on smaller screens.

MatteoGabriele:
I think the layout could be improved. The issue can be closed if we are fine with how it behaves now.

Maybe container query could help with this one, repositioning elements when they don't fit anymore. 
I've seen longer names where the layout changes further.

<img width="879" height="286" alt="Image" src="https://github.com/user-attachments/assets/f13786b5-1c83-4e5f-9c31-19206112c68b" />

<img width="1477" height="465" alt="Image" src="https://github.com/user-attachments/assets/e440d5c9-c184-47f7-98f4-6dff1e4c192a" />

serhalp:
I wonder if this helped? https://github.com/npmx-dev/npmx.dev/pull/481
Would someone like to take another look?

MatteoGabriele:
@serhalp Yes, I think your changes fix this. The distribution of elements looks more intentional.
We can close it if you guys agree. There's an open PR from @Flo0806, but I'm not sure of its current status.

I leave here a video to compare

https://github.com/user-attachments/assets/e50138ae-5c27-48e5-9265-98ef7214a6c4

MatteoGabriele:
This issue has been addressed multiple times in different PRs. I believe it has already been resolved, so we can close it.

Issue #201, opened 2026-01-28
compare two or more packages across a choice of facets(!)


serhalp:
Say I'm trying to decide between two alternative packages to add to my project. It would be great to be able to pull up a quick side-by-side comparison, and choose the facets I want to include in the comparison (downloads, security score, last updated... whatever we add in the future).

serhalp:
not sure yet if this would benefit from code reuse with https://github.com/npmx-dev/npmx.dev/issues/49 or if it's different enough

serhalp:
closed by #383

PR #422, opened 2026-01-30
fix: improved responsiveness for smaller screens

Flo0806:
fixes: #410 

---

try new positions for chips and tabs for better responsiveness on smaller screens (detail page)


garthdw:
Personally it feels like the docs/code buttons are too close to the chips.

I wonder if keeping the version next to the package name would separate them a bit more?

image

I guess there's only a ton of space because I am looking at vue and not something with a longer name.

Flo0806:
the thing is: what to do with longer package names? handle it separately or use one way. What is your idea?

serhalp:
👁️ I'm adding a 'compare to... `c`' to the docs/code panel over in #201 and running into similar issues (or rather... I'm making it worse). it might be worth stepping back and rethinking this design in general? 🤔 not sure, but I might noodle on it...

MatteoGabriele:
My 2 cents is that, since the package name has such a variable length, it might be worth letting it stand alone and re-arranging other pieces of the UI away from the name

MatteoGabriele:
There are a couple of things to check as well, in my opinion.
- Remove the spacing on the left side of the package name for better alignment
- Center align the version and chips; otherwise, it looks a bit off

Screenshot 2026-02-01 at 13 38 05

Flo0806:
@danielroe Are you on it just now? Otherwise I'll fix this finally.

danielroe:
IMO the version should remain where it is, and these 'chips' should behave as a unit, and either all go to the next line, or all stay on the previous line

MatteoGabriele:
@danielroe, as you can see from this comment on the issue, after a recent update, they already kinda behave like that https://github.com/npmx-dev/npmx.dev/issues/410#issuecomment-3828815263

danielroe:
perfect - I think that works.

danielroe:
thank you as always @Flo0806 - sorry a fix got merged first 🙏

PR #426, opened 2026-01-30
fix: focus search inputs only on non-touch devices

danielroe:


app/pages/index.vue
```diff
@@ -78,7 +84,7 @@ defineOgImageComponent('Default')
                 name="q"
                 :placeholder="$t('search.placeholder')"
                 v-bind="noCorrect"
-                :autofocus="!isMobile"
+                :autofocus="autofocus"
```
garthdw, line 87: Do we still need the autofocus if we are manually focusing on mount?

app/utils/responsive.ts
```diff
@@ -0,0 +1,12 @@
+/** @public */
+export function isTouchDevice() {
+  if (import.meta.server) {
+    return false
+  }
+  return (
+    'ontouchstart' in window ||
+    navigator.maxTouchPoints > 0 ||
+    // @ts-ignore
```
garthdw, line 9: Should this rather be `@ts-expect-error`?

app/pages/index.vue
danielroe: for client side navigation - without it it doesn't focus when navigating from index to search page

but I'm very up for other ideas!

garthdw: A dodgy mechanism I looked at implementing was watching the route. Then if the new route.name was search but old one was not, it would focus the searchbox. Might be able to do the same thing here just include the touch part? Although if you are starting search on home screen on mobile you likely want to be auto focused to search to complete it. 

I don't know vue/vue router enough to know if on direct nav to search if oldRoute has a value or not.

PR #434, opened 2026-01-30
fix(i18n): update i18n logic

userquin:
This PR includes:
- simplified  logic for locales at setting page including lang in the select options
- change locales type at config/i18n.ts
- fix arabic locale, it is wrong, we need `ar` at locales entry in config/i18n.ts module
- config/i18n.ts now exports correct translation files (I'm checking it, since we cannot buuild lunaria status.json without files tracked => ar.json removed and ar-EG.json added
- update lunaria/files


i18n/locales/en.json
```diff
@@ -329,7 +329,7 @@
       "size": "Size",
       "deps": "Deps",
       "updated": "Updated",
-      "get_started": "Get started",
+      "install": "Install",
```
userquin: uhmmm I didin't change this on my local 🤔

userquin:
REOPENED: ~~I'm going to send a new PR...~~

danielroe (approved):

PR #440, opened 2026-01-30
feat: add keyboard shortcut highlight on "?" press

jhroemer:
Damn this repo is on 🔥 !

Sooo.. Finally had a bit of time to look at this idea. I went a bit back and forward on it, and for now I believe the original idea works best, which is simply to trigger the highlight on "?". This goes well with the `<kbd>` element usage, and is a really simple approach. I think it works really well in dark mode, less well in light mode (see the video). 

https://github.com/user-attachments/assets/6ef5c626-3253-484e-be9f-c6c8796d624d

@danielroe you also proposed to trigger the highlight on any key-press that isn't already bound to something. What gives me pause about that feature is that it reminds me about clicking in the wrong spot in a Figma prototype. Enough for me to prefer the original over it. What works well with that idea is that it makes it possible to "stumble over" an otherwise somewhat hidden feature - something my original idea doesn't really solve.

I also played around with a variant that opens a small section at the bottom, and lists the keyboard shortcuts available in the current view. Ultimately that idea falls on its face for two reasons: it collides with the footer and we just don't have that many shortcuts right now (4 at most) - though you could add other tips to the section "did you know.." style or whatever. 

Screenshot 2026-01-30 at 21 09 01

**I'm happy to get feedback on this, and I'm also happy pursue another direction, or put it on ice for now 😄** 

closes https://github.com/npmx-dev/npmx.dev/pull/461


app/app.vue
```diff
@@ -81,3 +95,24 @@ if (import.meta.client) {
     <ScrollToTop />
   </div>
 </template>
+
+<style>
+@keyframes kbd-highlight {
```
jhroemer: I don't have much Vue/Nuxt experience (well none), but my instinct is to co-locate unless it's some global util/component/etc used in many places.

garthdw:
https://github.com/user-attachments/assets/77bb7bb8-87a1-4c7c-9b19-3537a265a8d6

If you hold down `?` it goes a bit funky, I think it should stay highlighted while holding down.

jhroemer:
> Screen.Recording.2026-01-30.at.22.21.41.mov 
> If you hold down `?` it goes a bit funky, I think it should stay highlighted while holding down.

Feature, not a bug...? 😀
Jokes aside, that indeed looks funky! I'll have a look, thanks!

jhroemer:
@garthdw I found two ways of dealing with this, one being the [dom reflow](https://www.harrytheo.com/blog/2021/02/restart-a-css-animation-with-javascript/) hack, and another one to use the web animations API. Since we're already using a js script and not pure css for this, I think the web animations API works really well here. Also, we have a single source of truth for the 700ms animation length (had to be used for the settimeout in addition to the css transition). Pushed a fix w. web animations API - wdyt?

app/app.vue
```diff
@@ -59,6 +61,29 @@ function handleGlobalKeydown(e: KeyboardEvent) {
 
     router.push('/search')
   }
+
+  if (e.key === '?') {
+    e.preventDefault()
+
+    for (const anim of kbdAnimations) {
+      anim.cancel()
+    }
```
garthdw: What about something like

```
    let isRunning = false
    for (const anim of kbdAnimations) {
      // set animation time back if still holding key down
      if (anim.playState === 'running') {
        anim.currentTime = 100
        isRunning = true
      }
    }
    if (isRunning) return
    kbdAnimations = []
```

Weirdly it plays the first animation fully then actually makes it work properly.

Although, it's probably better to rather do a keydown/keyup in that case and just get css to animation the in/out.

knowler:
Wouldn’t it just be easier to set a single class name on the `<html>` element (i.e. one DOM mutation) and then use a CSS transition for each element?

```css
kbd {
  transition: box-shadow /* whatever duration + easing */;
}

.show-shortcuts kbd {
  box-shadow: 0 0 4px 2px var(--accent);
}
```

Also, it might be worth looking into the perf of `box-shadow` animations (CSS triggers is quite outdated, but FWIW: https://css-triggers.com/#:~:text=C-,box%2Dshadow,-Initial%20render). If they’re costly (i.e. trigger layout or lots of paint), it might be better to use a pseudo element with the box-shadow already applied and `opacity: 0` then transition that instead.

```css
kbd {
	position: relative;
}

kbd::before {
	position: absolute;
	inset: 0;
	content: "";
	border-radius: inherit;
	box-shadow: 0 0 4px 2px var(--accent);
	opacity: 0;
	transition: opacity 150ms ease-in-out;
}

.show-shortcuts kbd::before {
	opacity: 1;
}
```

I don’t know Vue super well, but I’d just add an event listener for `keydown` to add the class, then add a once-called event listener for `keyup` to remove the class. Might wanna check that there aren’t any text fields focused when you check for the key.

jellydeck:
as per suggestion of @knowler 
we can set a data-attribute on `<html>` tag, which will turn true on press of `?` key.
so that'd be something like this

```js
const showKbdHints = ref(false)

useHead({
  htmlAttrs: {
    'data-kbd-hints': () => showKbdHints.value,
  },
})


function handleGlobalKeydown(e: KeyboardEvent) {
  if (e.key === '?') {
    e.preventDefault()
    showKbdHints.value = true
  }
}

function handleGlobalKeyup(e: KeyboardEvent) {
  if (e.key === '?') {
    showKbdHints.value = false
  }
}

```

```css
kbd {
  position: relative;
}

kbd::before {
  content: '';
  position: absolute;
  inset: 0;
  border-radius: inherit;
  box-shadow: 0 0 4px 2px var(--accent);
  opacity: 0;
  transition: opacity 200ms ease-out;
  pointer-events: none;
}

html[data-kbd-hints='true'] kbd::before {
  opacity: 1;
}
```

jhroemer:
> Wouldn’t it just be easier to set a single class name on the `<html>` element (i.e. one DOM mutation) and then use a CSS transition for each element?
> 
> ```css
> kbd {
>   transition: box-shadow /* whatever duration + easing */;
> }
> 
> .show-shortcuts kbd {
>   box-shadow: 0 0 4px 2px var(--accent);
> }
> ```
> 
> Also, it might be worth looking into the perf of `box-shadow` animations (CSS triggers is quite outdated, but FWIW: https://css-triggers.com/#:~:text=C-,box%2Dshadow,-Initial%20render). If they’re costly (i.e. trigger layout or lots of paint), it might be better to use a pseudo element with the box-shadow already applied and `opacity: 0` then transition that instead.
> 
> ```css
> kbd {
> 	position: relative;
> }
> 
> kbd::before {
> 	position: absolute;
> 	inset: 0;
> 	content: "";
> 	border-radius: inherit;
> 	box-shadow: 0 0 4px 2px var(--accent);
> 	opacity: 0;
> 	transition: opacity 150ms ease-in-out;
> }
> 
> .show-shortcuts kbd::before {
> 	opacity: 1;
> }
> ```
> 
> I don’t know Vue super well, but I’d just add an event listener for `keydown` to add the class, then add a once-called event listener for `keyup` to remove the class. Might wanna check that there aren’t any text fields focused when you check for the key.

Ah this is awesome, thanks! First time I see the CSS triggers article, really cool.
A few observations:
- The end result is a bit different, since what I had in my head was a slightly longer animation (I'm usually a fan of snappier animations but in this case the idea was something that lights up, holds for a bit, and then goes out). Your approach (and @garthdw I believe too) is more centered around holding down the key. And that might be a nicer approach - so the animation is really snappy by default, but you can hold if you need longer to identify the elements. I tested it locally and it works well! 
- The only thing is that the animation can end up getting stuck (and staying), if you release the shift key slightly before "?". This is pretty easy to solve by simply removing the data attribute on any key release.

Pushed a new change (tagged you as co-author, hope you don't mind). Mind having an extra look?

PR #441, opened 2026-01-30
style: make versions list scrollable if there are too many elements

DDeenis:
Makes the versions list scrollable, so the user doesn't have to scroll the whole page to see a long list of versions.

Before:
image

After:
image

"Other versions" list is now also scrollable:
image


garthdw (approved):

danielroe:
I'm not sure about this...

for a start, maybe skip this behaviour on mobile?

and I guess, on desktop we also need some kind of indication that there's content within to scroll?

DDeenis:
@danielroe sure, the mobile change makes sense. About indication though, isn't the scrollbar enough? We can make the scroll container a bit bigger/smaller so a subtle cropped item could be visible, indicating that there is more to scroll
image

I won't push hard on this change though, it just was really frustrating for me that I have to scroll the whole page to see the full list of versions

danielroe:
you won't see the scrollbar on MacOS/iOS unless the user starts to scroll it

DDeenis:
I see. Good to know - that is a problem.
I've made the requested change (mobile), and slightly increased the container height to partly show the last element.
Will leave a decision up to you @danielroe - feel free to close if you feel that this change is unnecessary

serhalp:
hey @DDeenis is this PR still needed? it looks like nowadays the right pane scrolls independently from the main content pane. does that solve this?

serhalp:
@DDeenis I'm going to close this for now, but feel free to let me know if you think this is still needed. Thank you for the contribution!

PR #447, opened 2026-01-30
fix: improve large chart readability on mobile

graphieros:
## Simplified UI on mobile:
-  no x axis labels
- increased scale labels & axis names font sizes
- slight increase of the chart height
- slightly narrower minimap

| Before | After |
|-|-|
|image| image|


serhalp (approved):
🎉

PR #448, opened 2026-01-30
feat: call out latest version in version list

tom-sherman:
Calls out the latest version in the list, to make it clear that latest has special significance.

Opted to preserve the existing order for now, although there are arguments for always putting the latest version at the top.

I also ensure that the latest tag always appears in the visible rows, forcing it to the bottom if it would otherwise be hidden. This ensures the latest tag is always visible and prominent.

image

### DIFF
--- a/app/components/PackageVersions.vue
+++ b/app/components/PackageVersions.vue
@@ -330,7 +330,7 @@ function getTagVersions(tag: string): VersionDisplay[] {
       <!-- Dist-tag rows (limited to MAX_VISIBLE_TAGS) -->
       <div v-for="row in visibleTagRows" :key="row.id">
         <div
-          class="flex items-center gap-2 pr-2"
+          class="flex items-center gap-2 pe-2"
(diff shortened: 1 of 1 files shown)


app/components/PackageVersions.vue
```diff
@@ -354,65 +364,67 @@ function getTagVersions(tag: string): VersionDisplay[] {
...
+              />
             </div>
           </div>
         </div>
 
         <!-- Expanded versions -->
         <div
           v-if="expandedTags.has(row.tag) && getTagVersions(row.tag).length > 1"
-          class="ms-4 ps-2 border-is border-border space-y-0.5"
+          class="ms-4 ps-2 border-is border-border space-y-0.5 pr-2"
```
danielroe: it's possible with rtl support we want `pe-2`?

see https://github.com/npmx-dev/npmx.dev/blob/main/CONTRIBUTING.md#rtl-support

tom-sherman: Good point, fixed.

serhalp (approved):
Nice, thank you!

PR #453, opened 2026-01-31
chore(deps): remove happy-dom

wojtekmaj:
Vitest Browser Mode is used (yay!) so happy-dom serves no purpose.


wojtekmaj:
> thanks! I wonder why knip didn't catch this 🤔

My best guess is that you almost never use happy-dom directly so this must be hacked around. At best it could have been derived from Vitest config... But wasn't.

PR #455, opened 2026-01-31
chore: replace `corepack enable` with pnpm/action-setup

wojtekmaj:
> [!NOTE]
> This PR has an alternative: #452

Following Node.js decision not to include Corepack in future versions of Node.js, I suggest replacing it with pnpm/action-setup, which not only removes dependency on Corepack, but also fixes dependency caching which, if you study CI logs, despite being setup correctly, simply does not seem to work.

Without cache: https://github.com/npmx-dev/npmx.dev/actions/runs/21535589549/job/62060513946

With cache: https://github.com/npmx-dev/npmx.dev/actions/runs/21535614560/job/62060587748

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -18,11 +18,15 @@ jobs:
 
     steps:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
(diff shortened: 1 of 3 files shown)


serhalp (approved):
👍🏼 This makes sense to me. In a few months we won't have built-in corepack anyway.

PR #464, opened 2026-01-31
feat: auto-detect `package.json` skills

onmax:
Add tabbed UI to switch between [skills-npm](https://github.com/antfu/skills-npm) and [skills CLI](https://github.com/vercel-labs/skills) installation methods.

## Sidebar trigger

Agent Skills in sidebar

## Modal tabs

| skills-npm | skills CLI |
|------------|------------|
| skills-npm | skills-cli |

## Changes

- Agent Skills card moved to sidebar (after keywords) - removes CLS
- Custom hexagon icon matching Agent Skills spec
- Tabbed install methods: skills-npm vs skills CLI
- "Click to expand" hint for skill list
- "Compatible with skills-npm" positive framing

### Fragmented skill repositories

Not many module authors store skills in the same repository as the source code. For example, the repositories for VueUse are [vueuse/vueuse](https://github.com/vueuse/vueuse) and [vueuse/skills](https://github.com/vueuse/skills), which makes it hard for us to link the source code with the skills.

I like how [@antfu](https://github.com/antfu) with [skills-npm](https://github.com/antfu/skills-npm) (and [npm-agentskills](https://github.com/onmax/npm-agentskills)) tries to solve this by "forcing" module authors to add the skills to the same repository as the source code (the docs usually live in the same repo already!), as these should not be in different repositories.

### Non-standard .well-known path

[RFC 8615](https://datatracker.ietf.org/doc/html/rfc8615) specifies `.well-known` URIs must be at the domain root (`/.well-known/...`). This implementation uses `/{package}/.well-known/skills/...` (or `/@org/pkg/.well-known/skills/...` for scoped packages).

Tradeoff: per-package namespacing works better with the skills CLI (`npx skills add <url>/{package}`), but technically doesn't follow the RFC.

## Open questions

- [ ] Is the architecture and the new paths acceptable?
- [ ] Is the UI/UX good?

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "module-replacements": "2.11.0",
     "nuxt": "4.3.0",
     "nuxt-og-image": "5.1.13",
+    "ofetch": "1.5.1",
     "perfect-debounce": "2.1.0",
     "sanitize-html": "2.17.0",
     "semver": "7.7.3",


danielroe:
very cool!

the $ should probably be inline with the first part of the line, when wrapping:

SCR-20260131-jtgd

do we need to make it clearer that the number of skills is clickable? (or the entire Skills / 4 metric should be a button)

open questions:
 - should this go in the sidebar under keywords, etc.?
 - should it instead be in 'get started'?
 - I expect to have a command to run with `skills-npm` - if it just works then we should say something like 💡 these skills are compatible with `skills-npm` - just set it up and install

onmax:
I also thought about put it in the sidebar, on the right. I think I like it more there and also removes the issue with CLS.

regarding to the last suggestion, I agree with you, I will implement it

I will also change the icon from robot to the agent spec svg

onmax:
@danielroe ready!

danielroe:
@onmax did you forget to push?

onmax:
yes 🤦 , done!

app/pages/code/[...path].vue
```diff
@@ -99,9 +99,18 @@ const fileContentUrl = computed(() => {
...
+  execute: fetchFileContent,
+} = useFetch<PackageFileContentResponse>(() => fileContentUrl.value!, { immediate: false })
+
+watch(
+  fileContentUrl,
+  url => {
+    if (url) fetchFileContent()
+  },
+  { immediate: true },
 )
```
Comment (danielroe) [inline app/pages/code/[...path].vue:114]: is ths fixing another issue?

danielroe (approved):
made a few refactors to align with playgrounds display - I'd be up for updating the style of both though in future! and I think we need to think about how to reduce CLS when doing client-side loading for this kind of info....

app/pages/code/[...path].vue
Comment (onmax) [reply app/pages/code/[...path].vue:114]: yes, let me test if this has been fixed in main. Otherwise opening another pr now

Comment (onmax) [reply app/pages/code/[...path].vue:114]: https://github.com/npmx-dev/npmx.dev/pull/630

PR #471, opened 2026-01-31
feat(i18n): strip unused i18n keys

trueberryless:
I find it pretty useful when translating to be able to check locally if I have updated everything as sometimes it get's chaotic with a lot of outdated strings...

Therefore I made a little script that can easily be called with `pnpm` for all the languages, which outputs the missing keys and also which have been removed in the original.

In the future, if we keep the single JSON file, we could integrate this into CI, checking for all i18n PRs if everything has been updated automatically.

Not sure, if this goes into the desired direction of the project, as I saw some other issues and PRs improving i18n in other ways. So just close if this is not desired 👍

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -34,14 +34,14 @@ jobs:
       - name: 📦 Install browsers
         run: pnpm playwright install
 
+      - name: 🌐 Compare translations
+        run: pnpm i18n:check
+
       - name: 🌍 Update lunaria data
         run: pnpm build:lunaria
 
       - name: 🔠 Fix lint errors
         run: pnpm lint:fix
-
-      - name: 🌐 Compare translations
-        run: pnpm i18n:check
 
       - name: 🧪 Update unit test snapshots
         run: pnpm test:unit -u


trueberryless:
> I would love to run this in CI and automatically remove unused keys in other languages
> 
> in autofix.yml, it would ensure when removing keys from one lang, they are automatically removed from every other one

Should I create the github workflow for that directly here? (Would be later, as I got work to do now until evening or so)

package.json
```diff
@@ -13,6 +13,7 @@
     "build:lunaria": "node --experimental-transform-types ./lunaria/lunaria.ts",
     "dev": "nuxt dev",
     "dev:docs": "pnpm run --filter npmx-docs dev --port=3001",
+    "i18n:check": "node scripts/compare-translations.js",
```
ghostdevv: we're targeting node 24 right? if so this could be a ts file :p

scripts/compare-translations.js
```diff
@@ -0,0 +1,89 @@
+import { existsSync, readFileSync } from 'fs'
+import { join } from 'path'
```
ghostdevv:
```suggestion
import { existsSync, readFileSync } from 'node:fs'
import { join } from 'node:path'
```

ghostdevv (changes requested):

package.json
danielroe: yes!

trueberryless:
> hope you don't mind - I pushed the refactor 🙏 

Awesome, I'm thankful you like the idea 💡

PR #476, opened 2026-01-31
perf: use `shallowRef` and `useTemplateRef`

OrbisK:
Replacing some refs with shallowRefs

### DIFF
--- a/app/components/DependencyPathPopup.vue
+++ b/app/components/DependencyPathPopup.vue
@@ -5,13 +5,8 @@ defineProps<{
 }>()
 
 const isOpen = shallowRef(false)
-const popupEl = ref<HTMLElement | null>(null)
+const popupEl = useTemplateRef('popupEl')
 const popupPosition = shallowRef<{ top: number; left: number } | null>(null)
-
-// Function ref - captures the element when popup mounts
-function setPopupRef(el: unknown) {
-  popupEl.value = (el as HTMLElement) || null
-}
 
 function closePopup() {
   isOpen.value = false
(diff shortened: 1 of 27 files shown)


app/components/DependencyPathPopup.vue
```diff
@@ -5,14 +5,9 @@ defineProps<{
...
 const isOpen = shallowRef(false)
-const popupEl = ref<HTMLElement | null>(null)
+const popupEl = useTemplateRef('popupEl')
 const popupPosition = shallowRef<{ top: number; left: number } | null>(null)
 
-// Function ref - captures the element when popup mounts
-function setPopupRef(el: unknown) {
-  popupEl.value = (el as HTMLElement) || null
-}
-
```
OrbisK, line 15: @Flo0806 was there a reason to use a setter here? from my tests conditional useTemplateRef works fine with onClickOutside 😅

danielroe (approved):

danielroe:
we should create an eslint rule

OrbisK:
> we should create an eslint rule

I also thought about this. I think we have an lint rule at vueuse that `ref` has to be imported as `deepRef`. I think this might help raise awareness.

OrbisK:
Not sure how to migrate this to oxlint 

https://github.com/vueuse/vueuse/blob/bbd185d8f83699e5735273dc2296a4ff8358a715/eslint.config.js#L175-L190

app/components/DependencyPathPopup.vue
Flo0806: I had some problems, yes. But if it works for you - perfect. Your way is the better

OrbisK: If you can still reproduce the bug we might be able to fix it upstream (vueuse or vue)

Flo0806: No... I think I had just to restart the dev server...

PR #485, opened 2026-01-31
perf: use `$fetch.raw` instead fetch with separate controller + timeout

OrbisK:
I know these are micro-optimisations. Initially I was looking to see if every timeout was cleaned up properly.


server/api/contributors.get.ts
```diff
@@ -15,7 +15,7 @@ export default defineCachedEventHandler(
       {
         headers: {
           'Accept': 'application/vnd.github.v3+json',
-          'User-Agent': 'npmx.dev',
+          'User-Agent': 'npmx',
```
OrbisK, line 18: every other user-agent is `npmx` not `npmx.dev`

OrbisK:
Tests are failing. I will take a look

OrbisK:
@danielroe Okay. I was super confused becaused it worked locally. But it was cached all the time. 😅 

I think this might be a ofetch bug (or at least a type issue). The problem is here

https://github.com/OrbisK/npmx.dev/blob/397c8a521c9ec4d91ea179cad5845d1ac8c80fb6/server/utils/docs/client.ts#L102

`response.text()` throws because the body is already read. I wonder if `.raw` should not expose `.text()` and `.json()`

PR #488, opened 2026-01-31
feat: add collapsible section component

garthdw:
Component assumes it's a sidebar collapsible section.

Collapsed state is stored in localStorage.

Allowing user preferences for whether to collapse sections by default.

Applied collapsible section to sidebar sections from downloads down.

https://github.com/user-attachments/assets/e7054e2a-f836-4851-bfb0-48031f81d89d

I tried to make it as CSS driven as I could so that ideally the prehydrate stuff can take effect. However, I think the fact that I'm injecting the styles dynamically means it's not aware.

Would love some help here.


garthdw:
Will sort the tests out in a bit.

danielroe (approved):

PR #495, opened 2026-01-31
feat(i18n): improve update process

jycouet:
- [x] adding a `--fix` option to the script to simplify the update process (doc updated)
- [x] update `fr` doc at the same time

### DIFF
--- a/lunaria/files/fr-FR.json
+++ b/lunaria/files/fr-FR.json
@@ -123,7 +123,11 @@
       "vulns": "Vulnérabilités",
       "updated": "Mis à jour",
       "view_dependency_graph": "Voir le graphe de dépendances",
-      "inspect_dependency_tree": "Inspecter l'arbre de dépendances"
+      "inspect_dependency_tree": "Inspecter l'arbre de dépendances",
+      "size_tooltip": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -15,6 +15,7 @@
     "dev": "nuxt dev",
     "dev:docs": "pnpm run --filter npmx-docs dev --port=3001",
     "i18n:check": "node --experimental-transform-types scripts/compare-translations.ts",
+    "i18n:check:fix": "node --experimental-transform-types scripts/compare-translations.ts --fix",
```
danielroe, line 18: we should run this in `autofix.yml` then 🙏 

it already automatically strips unused keys, and it would be convenient if it also added new ones. maybe we don't even need the `--fix` suffix?

jycouet: With this PR, the translated text would be:
```
"create_account": "EN TEXT TO REPLACE: Create a new account"
```
To me, this is something you run manually when you want to update translations:
- Fill in all missing keys
- Have the English text by default, so you can immediately see what needs to be translated without switching files or hunting for keys

danielroe: ahhh, makes sense!

PR #505, opened 2026-01-31
fix: search input looses focus when navigating to /search

cullophid:
The main input on the home page has a debounce timer of 250ms.  This means that it fails if your typing speed drops below 200 WPM. 
This PR sets it to 500ms which still feels fast without breaking typing flow
The auto focus attribute on SerachBox was not working in this case, so focus is not set on url search query change.


OrbisK:
This is not a mobile only issue. ~~I think we just need to exlude the search query from the page key~~

nvm this does not work because we are navigating to search from index...

OrbisK:
Perhaps the index page could be aliased to the search page, but display the hero when accessed via `/`. This would allow us to exclude the search q from the page key, thus preventing the page from being rerendered/losing focus.

danielroe:
wait! I have an idea.... we _teleport_ the search bar into the index.vue

cullophid:
This is just fixing the immediate issue, but it does not make it go away entirely. When switching input there is a split second where user input is ignored. 

It is much less likely with 500ms, but it can still happen.

I would suggest not doing the debounced search from the index page.🤷‍♂️

cullophid:
> Perhaps the index page could be aliased to the search page, but display the hero when accessed via `/`. This would allow us to exclude the search q from the page key, thus preventing the page from being rerendered/losing focus.

But they are still two different inputs, so the focus would still have to move to a different element no?

cullophid:
Closing this, @danielroe is on it.

Issue #513, opened 2026-01-31
Sort dropdown too wide on mobile 


wojtekmaj:
![Screenshot_20260131_170109_Samsung Internet.jpg](https://github.com/user-attachments/assets/bc61bc0e-9d5b-4fc7-b05f-e0d4ed1c2fca)

Search results sort dropdown takes too much space (likely a mix of margin + 100% width)

whitep4nth3r:
What device/browser are you using? Struggling to recreate the issue.

Here's what I see:

<img width="467" height="565" alt="Image" src="https://github.com/user-attachments/assets/e7ac5c68-3905-4c17-975b-c952918602d1" />

serhalp:
I think it's because its width is set to the max width of all the values in the select, and that max width is dependent on the locale.

We could maybe remove those "Coming soon"s I added 😶

wojtekmaj:
> What device/browser are you using? Struggling to recreate the issue.

I'm on a Galaxy S23 Plus, fully up to date, using stock Samsung Internet.

It doesn't seem to be browser related but cannot be reproduced on English locale apparently!

whitep4nth3r:
That would make sense @serhalp!

PR #518, opened 2026-01-31
fix: remove disabledReason to avoid UI issues when not in english

whitep4nth3r:
This could potentially fix https://github.com/npmx-dev/npmx.dev/issues/513#issuecomment-3828783367

I cleaned up everywhere else that used `disabledReason`.

Instead of the "coming soon" in the dropdown, we could consider adding a message in the UI that more options are coming soon.

That being said, is it wrong to consider that the people using this tool will most likely understand why the options are disabled anyway?


whitep4nth3r:
I'm an idiot. Disregard this.

PR #522, opened 2026-01-31
refactor(a11y): use `<dialog>` tag for modal components

nulfrost:
**[Still a work in progress, will update as I go]**

## Background

Did some exploring in the codebase and on the website and found that we could make the modals a bit better from an accessibility and user experience point of view. The first thing being that we can use the ``<dialog>`` tag since it has been [baseline since March 2022](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog). This allows us to take advantage of the features it offers out of the box such as:

- Focus trapping
- Escape to dismiss
- An easy way to style the backdrop
- Implicit ARIA roles

I've started making a general ``Modal``component that can be used in places where there are modals. Right now it only expects you to pass in a title for the modal and it handles the close button interaction but this could change if we want it to handle more.

Some resources I'm using as I make these changes:

- MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog
- W3 ARIA Authoring Practices Guide: https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/
- Kevin Powell's video about dialogs/modals: https://www.youtube.com/watch?v=TAB_v6yBXIE

## Showcase

https://github.com/user-attachments/assets/a179124e-136f-46d7-85ab-cddd8dc4c5b2

This video shows that the focus is trapped, we can only interact with the elements within the modal and that the scroll is locked when the modal is open

### DIFF
--- a/app/components/PackageWeeklyDownloadStats.vue
+++ b/app/components/PackageWeeklyDownloadStats.vue
@@ -9,6 +9,13 @@ const props = defineProps<{
 
 const chartModal = useModal('chart-modal')
 
+const isChartModalOpen = shallowRef(false)
+function openChartModal() {
+  isChartModalOpen.value = true
+  // ensure the component renders before opening the dialog
+  nextTick(() => chartModal.open())
+}
+
 const { data: packument } = usePackage(() => props.packageName)
 const createdIso = computed(() => packument.value?.time?.created ?? null)
 
@@ -194,7 +201,7 @@ const config = computed(() => {
       <template #actions>
         <button
           type="button"
-          @click="chartModal.open()"
+          @click="openChartModal"
           class="link-subtle font-mono text-sm inline-flex items-center gap-1.5 ms-auto shrink-0 self-center focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-fg/50 rounded"
           :title="$t('package.downloads.analyze')"
         >
@@ -241,11 +248,11 @@ const config = computed(() => {
     </CollapsibleSection>
   </div>
 
-  <ChartModal>
+  <ChartModal v-if="isChartModalOpen" @close="isChartModalOpen = false">
     <PackageDownloadAnalytics
       :weeklyDownloads="weeklyDownloads"
       :inModal="true"
-      :packageName="packageName"
+      :packageName="props.packageName"
       :createdIso="createdIso"
     />
   </ChartModal>
--- a/app/components/PackageDownloadAnalytics.vue
+++ b/app/components/PackageDownloadAnalytics.vue
@@ -5,12 +5,7 @@ import { useDebounceFn, useElementSize } from '@vueuse/core'
 import { useCssVariables } from '../composables/useColors'
 import { OKLCH_NEUTRAL_FALLBACK, transparentizeOklch } from '../utils/colors'
 
-const {
-  weeklyDownloads,
-  inModal = false,
-  packageName,
-  createdIso,
-} = defineProps<{
+const props = defineProps<{
   weeklyDownloads: WeeklyDownloadPoint[]
   inModal?: boolean
   packageName: string
@@ -131,7 +126,7 @@ function formatXyDataset(
     return {
       dataset: [
         {
-          name: packageName,
+          name: props.packageName,
           type: 'line',
           series: dataset.map(d => d.downloads),
           color: accent.value,
@@ -149,7 +144,7 @@ function formatXyDataset(
     return {
       dataset: [
         {
-          name: packageName,
+          name: props.packageName,
           type: 'line',
           series: dataset.map(d => d.downloads),
           color: accent.value,
@@ -162,7 +157,7 @@ function formatXyDataset(
     return {
       dataset: [
         {
-          name: packageName,
+          name: props.packageName,
           type: 'line',
           series: dataset.map(d => d.downloads),
           color: accent.value,
@@ -175,7 +170,7 @@ function formatXyDataset(
     return {
       dataset: [
         {
-          name: packageName,
+          name: props.packageName,
           type: 'line',
           series: dataset.map(d => d.downloads),
           color: accent.value,
@@ -235,10 +230,10 @@ const hasUserEditedDates = shallowRef(false)
 
 function initDateRangeFromWeekly() {
   if (hasUserEditedDates.value) return
-  if (!weeklyDownloads?.length) return
+  if (!props.weeklyDownloads?.length) return
 
-  const first = weeklyDownloads[0]
-  const last = weeklyDownloads[weeklyDownloads.length - 1]
+  const first = props.weeklyDownloads[0]
+  const last = props.weeklyDownloads[props.weeklyDownloads.length - 1]
   const start = first?.weekStart ? toIsoDateOnly(first.weekStart) : ''
   const end = last?.weekEnd ? toIsoDateOnly(last.weekEnd) : ''
   if (isValidIsoDateOnly(start)) startDate.value = start
@@ -265,7 +260,7 @@ function initDateRangeFallbackClient() {
 }
 
 watch(
-  () => weeklyDownloads?.length,
+  () => props.weeklyDownloads?.length,
   () => {
     initDateRangeFromWeekly()
     initDateRangeFallbackClient()
@@ -342,7 +337,7 @@ watch(
 
 const { fetchPackageDownloadEvolution } = useCharts()
 
-const evolution = shallowRef<EvolutionData>(weeklyDownloads)
+const evolution = shallowRef<EvolutionData>(props.weeklyDownloads)
 const pending = shallowRef(false)
 
 let lastRequestKey = ''
@@ -354,7 +349,7 @@ const debouncedLoad = useDebounceFn(() => {
 
 async function load() {
   if (!import.meta.client) return
-  if (!inModal) return
+  if (!props.inModal) return
 
   const o = options.value
   const extraBase =
@@ -366,14 +361,14 @@ async function load() {
 
   const startKey = (o as any).startDate ?? ''
   const endKey = (o as any).endDate ?? ''
-  const requestKey = `${packageName}|${createdIso ?? ''}|${o.granularity}|${extraBase}|${startKey}|${endKey}`
+  const requestKey = `${props.packageName}|${props.createdIso ?? ''}|${o.granularity}|${extraBase}|${startKey}|${endKey}`
 
   if (requestKey === lastRequestKey) return
   lastRequestKey = requestKey
 
   const hasExplicitRange = Boolean((o as any).startDate || (o as any).endDate)
-  if (o.granularity === 'week' && weeklyDownloads?.length && !hasExplicitRange) {
-    evolution.value = weeklyDownloads
+  if (o.granularity === 'week' && props.weeklyDownloads?.length && !hasExplicitRange) {
+    evolution.value = props.weeklyDownloads
     pending.value = false
     displayedGranularity.value = 'weekly'
     return
@@ -384,8 +379,8 @@ async function load() {
 
   try {
     const result = await fetchPackageDownloadEvolution(
-      () => packageName,
-      () => createdIso,
+      () => props.packageName,
+      () => props.createdIso,
       () => o as any, // FIXME: any
     )
 
@@ -404,7 +399,7 @@ async function load() {
 }
 
 watch(
-  () => inModal,
+  () => props.inModal,
   () => {
     // modal open/close should be immediate
     load()
@@ -414,8 +409,8 @@ watch(
 
 watch(
   () => [
-    packageName,
-    createdIso,
+    props.packageName,
+    props.createdIso,
     options.value.granularity,
     (options.value as any).weeks,
     (options.value as any).months,
@@ -437,9 +432,9 @@ watch(
 )
 
 const effectiveData = computed<EvolutionData>(() => {
-  if (displayedGranularity.value === 'weekly' && weeklyDownloads?.length) {
+  if (displayedGranularity.value === 'weekly' && props.weeklyDownloads?.length) {
     if (isWeeklyDataset(evolution.value) && evolution.value.length) return evolution.value
-    return weeklyDownloads
+    return props.weeklyDownloads
   }
   return evolution.value
 })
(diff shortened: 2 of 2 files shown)


app/app.vue
```diff
@@ -93,26 +81,7 @@ if (import.meta.client) {
...
-  transition: opacity 200ms ease-out;
-  pointer-events: none;
-}
-
-html[data-kbd-hints='true'] kbd::before {
-  opacity: 1;
-}
-</style>
+  <!-- <ConnectorModal /> -->
+  <AuthModal />
```
nulfrost: I felt like it didn't make sense for these modals to be inside of the ``HeaderAccountMenu.vue`` file? Though I'm learning with ``<Teleport>`` I could put them back there...might play with it again

app/assets/main.css
```diff
@@ -554,3 +554,32 @@ input[type='search']::-webkit-search-results-decoration {
...
+  opacity: 1;
+  transition: opacity 200ms ease;
+  transition-behavior: allow-discrete;
+}
+
+@starting-style {
+  dialog:open {
+    opacity: 0;
+  }
+}
```
knowler: Safari doesn’t seem to be liking something (i.e. modals are open but not visible and seem to have `opacity: 0` applied). ~~If I recall, you might need to order `@starting-style` before the prior `dialog:open` statement from lines 575 to 579. I believe order of appearance is causing the `@starting-style` styles to apply after for some reason (probably a Safari bug). I could be wrong.~~

app/assets/main.css
```diff
@@ -554,3 +554,32 @@ input[type='search']::-webkit-search-results-decoration {
   animation-duration: 0.3s;
   animation-timing-function: cubic-bezier(0.22, 1, 0.36, 1);
 }
+
+/* Locking the scroll whenever any of the modals are open */
+html:has(dialog:open) {
+  overflow: hidden;
+}
```
knowler, line 241: Might need a different scroll locking mechanism for Safari, because I don’t think this works for it.

app/assets/main.css
knowler: Oh! Safari doesn’t support `:open` yet. So, you’ll need to use `[open]` instead.

knowler: You could also use `:modal` instead which Safari does support (and there is wide support for overall). That’s a bit more specific to what’s happening (i.e. the dialog is in a modal state which means that its open). A closed `<dialog>` is not `:modal`, so it’d work.

app/assets/main.css
knowler: I might actually be wrong…

app/assets/main.css
nulfrost: Yeah noticed these issues when I went to test stuff in Safari :( will make these changes. MDN had ``:open`` which is why I went with that initially

app/assets/main.css
nulfrost: Ugh yeah, another thing I noticed. Might need to re-think this bit

nulfrost: My assumption is that it might be what you mentioned before with ``:open``

app/assets/main.css
nulfrost: > You could also use `:modal` instead which Safari does support (and there is wide support for overall). That’s a bit more specific to what’s happening (i.e. the dialog is in a modal state which means that its open). A closed `<dialog>` is not `:modal`, so it’d work.

wait I didn't know there was an actual ``:modal`` attribute wow https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:modal. Okay I'll use that instead

app/components/Modal.client.vue
```diff
@@ -0,0 +1,38 @@
...
+  }
+}
+</script>
+
+<template>
+  <Teleport to="body">
+    <dialog
+      class="w-full bg-bg border border-border rounded-lg shadow-xl max-h-[90vh] overflow-y-auto overscroll-contain m-0 m-auto p-6 text-white"
+      v-bind="$attrs"
+    >
```
knowler, line 31: It might vary depending on the specific modal, but we can get the backdrop light dismiss using [the new `closedby=any` attribute value combo](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#closedby) for the `<dialog>` element. Safari doesn’t support it yet unfortunately, but it could be used as a progressive enhancement. There are a few options for polyfilling this behaviour if we think we need it.

knowler:
```suggestion
    <dialog
      closedby="any"
      class="w-full bg-bg border border-border rounded-lg shadow-xl max-h-[90vh] overflow-y-auto overscroll-contain m-0 m-auto p-6 text-white"
      v-bind="$attrs"
    >
```

nulfrost: oh that's awesome! I'm not sure if/how we're doing polyfills for this project today but if it's easy to add I'm totally game 👍🏿

knowler: I imagine we’d probably want to add it in a lightweight way that doesn’t completely cover all that `closedby` can do.

There’s a bit of a hack here: https://codepen.io/paramagicdev/pen/gbYompq

The idea is to set `pointer-events: none` on for `::backdrop`, then listen for clicks that don’t include the `dialog` element. I haven’t fully checked if there are any issues with this approach (`pointer-events` can be problematic).

nulfrost: > I imagine we’d probably want to add it in a lightweight way that doesn’t completely cover all that `closedby` can do.
> 
> There’s a bit of a hack here: https://codepen.io/paramagicdev/pen/gbYompq
> 
> The idea is to set `pointer-events: none` on for `::backdrop`, then listen for clicks that don’t include the `dialog` element. I haven’t fully checked if there are any issues with this approach (`pointer-events` can be problematic).

Will try this out now

app/app.vue
```diff
@@ -72,9 +72,20 @@ function handleGlobalKeyup() {
...
+/* A hack to get light dismiss to work in safari because it does not support closedby="any" yet */
+// https://codepen.io/paramagicdev/pen/gbYompq
+// see: https://github.com/npmx-dev/npmx.dev/pull/522#discussion_r2749978022
+function handleModalLightDismiss(e: any) {
+  const modal = document.querySelector<HTMLDialogElement>('dialog:modal')
+  if (modal && modal.open && !e.composedPath().includes(modal)) {
+    modal.close()
+  }
+}
+
```
nulfrost, line 80: trying to figure out how to get this to work...added some debug logs and I can see it's being triggered but it runs when I try to open the modal so it closes the modal immediately. I have an idea of _why_ that happens but going to take a break and come back to this. I think this is the last remaining thing before this is ready for a second once over

nulfrost: Ah, this was the trick! Thought I didn't need it but I do

```ts
 setTimeout(() => {
      authModal.showModal()
    })
```

app/components/ClaimPackageModal.vue
```diff
@@ -38,6 +38,13 @@ async function checkAvailability() {
   }
 }
 
+function openConnectorModal() {
+  const connectorModal = document.querySelector<HTMLDialogElement>('#connector-modal')
+  if (connectorModal) {
+    connectorModal.showModal()
+  }
+}
+
```
nulfrost: I actually have to re-define this functions quite a bit, any tips on refactoring?

app/composables/useModal.ts
```diff
@@ -0,0 +1,22 @@
...
+    if (modal) {
+      modal.close()
+    }
+  }
+
+  return {
+    open,
+    close,
+  }
+}
```
nulfrost, line 24: How does this look?

app/components/PackageWeeklyDownloadStats.vue
```diff
@@ -242,28 +247,13 @@ const config = computed(() => {
...
-  <ChartModal v-model:open="showModal">
-    <template #title>{{ $t('package.downloads.modal_title') }}</template>
-
+  <ChartModal>
     <PackageDownloadAnalytics
       :weeklyDownloads="weeklyDownloads"
       :inModal="true"
       :packageName="packageName"
       :createdIso="createdIso"
     />
```
danielroe, line 257: this isn't 100% working right now - I think because this is rendered at a different time, we're getting much short period of data - maybe a problem with reactivity in vue-data-ui?

app/components/Modal.client.vue
```diff
@@ -1,26 +1,37 @@
...
+})
 </script>
 
 <template>
   <Teleport to="body">
     <dialog
+      ref="dialogRef"
       closedby="any"
       class="w-full bg-bg border border-border rounded-lg shadow-xl max-h-[90vh] overflow-y-auto overscroll-contain m-0 m-auto p-6 text-white"
+      :aria-labelledby="modalTitleId"
```
knowler, line 29: Nice catch!

nulfrost:
Thanks @danielroe I was definitely struggling for a bit there 😅 was trying to figure out how the ``ref`` stuff worked but I see how you've done it and I think I get it now

app/components/PackageWeeklyDownloadStats.vue
nulfrost: Yeah I noticed that too, wasn't sure if it was something I caused but didn't dig into it at all

app/app.vue
knowler: I’m not sure if this is working when I test in Safari. TBH if it’s too difficult to get working I think it’s fine to use `closedby=any` as a progressive enhancement and just get rid of the hack. There are still ways for people to dismiss the modal (i.e. by shortcut or the dismiss button).

app/assets/main.css
```diff
@@ -554,3 +554,8 @@ input[type='search']::-webkit-search-results-decoration {
   animation-duration: 0.3s;
   animation-timing-function: cubic-bezier(0.22, 1, 0.36, 1);
 }
+
+/* Locking the scroll whenever any of the modals are open */
+html:has(dialog:modal) {
```
knowler, line 238: Related to the effect of this: when the modal opens the scrollbar for the page disappears which causes a bit of a shift. It might be worth adding [`scrollbar-gutter: stable`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/scrollbar-gutter#stable) to avoid that (might need to add it to `html` without the `:has(dialog:modal)`).

nulfrost: Ah right, good call

knowler: Very odd, it seems like Firefox is working exactly as expected, but I’m not seeing it work in Safari or Chrome.

knowler:
I’m consistently seeing an issue where if you’re on the settings page and press escape, the app navigates back to the previous page. I think there’s a conflict with the keyboard shortcuts. We should add a check if a modal is open somewhere here: https://github.com/npmx-dev/npmx.dev/blob/368e2d09e951fa62a0bcd913900c9cbe90ccfa50/app/pages/settings.vue#L9-L19

app/assets/main.css
nulfrost: Oh weird, I'm seeing it work in Firefox and Chrome but not Safari.. 🤔

nulfrost:
> I’m consistently seeing an issue where if you’re on the settings page and press escape, the app navigates back to the previous page. I think there’s a conflict with the keyboard shortcuts. We should add a check if a modal is open somewhere here:
> 
> https://github.com/npmx-dev/npmx.dev/blob/368e2d09e951fa62a0bcd913900c9cbe90ccfa50/app/pages/settings.vue#L9-L19

Hmm this seems intentional, it looks like if you're focused on any form elements it does not navigate backwards. There's already the back link to the homepage so I'm not sure if this is needed at all? Or maybe it was just a nice to have?

app/pages/settings.vue
```diff
@@ -10,7 +10,8 @@ onKeyStroke(
   'Escape',
   e => {
     const target = e.target as HTMLElement
-    if (!['INPUT', 'SELECT', 'TEXTAREA'].includes(target?.tagName)) {
+    console.log(target)
+    if (!['INPUT', 'SELECT', 'TEXTAREA', 'BUTTON'].includes(target?.tagName)) {
```
nulfrost: This seems to work, I opened a modal and pressing escape once closes it. Pressing it a 2nd time navigates backwards

knowler: This would only cover if the dismiss button was still initially focused when the modal opens, but that could lose focus if a user clicks around inside the modal. It’d be better to check if the `html` element matches `html:has(:modal)`:

```js
// add to the condition
!document.documentElement.matches(':has(:modal)')
```

nulfrost: yeah that worked so much better

app/assets/main.css
```diff
@@ -554,3 +554,9 @@ input[type='search']::-webkit-search-results-decoration {
   animation-duration: 0.3s;
   animation-timing-function: cubic-bezier(0.22, 1, 0.36, 1);
 }
+
+/* Locking the scroll whenever any of the modals are open */
```
knowler, line 237: This works reliably to scroll lock including Safari macOS and iOS:

```css
html:has(dialog:modal),
html:has(dialog:modal) body {
	overflow: hidden;
  height: 100%;
}
```

nulfrost: hmm no difference on my end

https://github.com/user-attachments/assets/f7efe7c1-3c20-4100-80f6-5234cc310afd

knowler: Ok, I seemed to have a cached version on my iPhone—seems to work fine on the preview.

knowler: Actually, it seems to be inconsistent. Like the first time you open it, it’s fine, but then you do something else and open it again and suddenly the scroll isn’t locked. In any case, that seems like an edge case and I think we can ignore.

nulfrost: Removing the explicit scrollbar width seemed to do it, wonder if there's a way to keep it and have no shift

knowler (approved):
From my perspective this is good. I’ll defer to @danielroe and others with regard to code (as I only have a basic understanding of Vue/Nuxt).

nulfrost:
> From my perspective this is good. I’ll defer to @danielroe and others with regard to code (as I only have a basic understanding of Vue/Nuxt).

Appreciate you for taking the time to review this and make corrections 🙏🏿

danielroe (approved):

PR #524, opened 2026-01-31
refactor: use an `<article>` element for for package readme

knowler:
As discussed in https://github.com/npmx-dev/npmx.dev/pull/503#discussion_r2749739592 and https://github.com/npmx-dev/npmx.dev/pull/503#discussion_r2749740907, @rshigg and I noticed that a package’s readme would be better wrapped with an `<article>` element as it’s a standalone document. It’s alright that it’s a descendant of another `<article>` element (which wraps the page’s main contents).

Worth noting that an `<article>` element and its corresponding `article` ARIA role is not a navigational landmark, so this change will only impact users when they are traversing through the page.

### DIFF
--- a/app/pages/[...package].vue
+++ b/app/pages/[...package].vue
@@ -923,7 +923,7 @@ function handleClick(event: MouseEvent) {
 
       <div class="area-sidebar">
         <!-- Sidebar -->
-        <aside class="sticky top-20 space-y-6 sm:space-y-8 min-w-0 overflow-hidden">
+        <div class="sticky top-20 space-y-6 sm:space-y-8 min-w-0 overflow-hidden">
           <!-- Maintainers (with admin actions when connected) -->
           <PackageMaintainers :package-name="pkg.name" :maintainers="pkg.maintainers" />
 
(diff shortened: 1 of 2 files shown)


rshigg:
Any reason to withhold removing the aside and ancestor article? I think this is a good opportunity to get that all sorted at once

app/pages/[...package].vue
Comment (knowler) [inline app/pages/[...package].vue]: I’m having second thoughts. I’m thinking this `<div>` that contains the README contents should be the `<article>` element, since our “README” heading isn’t a part of that document.

Comment (knowler) [reply app/pages/[...package].vue]: If we do that, then I’ll probably remove the top-level `<article>` for sure since they’ll probably have the same name which might be confusing.

Comment (knowler) [reply app/pages/[...package].vue]: b671b9a I made the change here. I’m still considering removing the top-level `<article>` and `<aside>` elements. Taking sometime to think on it.

Comment (knowler) [reply app/pages/[...package].vue]: @rshigg So, I think I’m done here and these are the changes:

- Instead of replacing the `<section>` with the `README` heading with an `<article>` element, I’ve used it to wrap the readme contents itself, which were wrapped in a `<div>`. This is because our heading is not a part of that document.
- I considered removing `main > article` element because it’d likely have a similar name to the README one, but I’ve left it because it really doesn’t have a whole lot of impact. The `article` ARIA role is really only exposed in macOS VoiceOver which has very low usage in comparison to other screen readers. So, you could say this whole PR was negligible, but we are using more semantic HTML now, so that’s at least a plus.
- I did remove the `<aside>` element used for the (presentational) sidebar.

PR #527, opened 2026-01-31
fix: persist user's locale

### LINKED ISSUE #509: Language selection doesn't persist after refresh
I have noticed that on both mobile Samsung Internet and desktop Microsoft Edge that after I refresh the page it goes back to English. Noticed that on localhost too, so I don't think it's related to e.g. ad blocking extensions.

Navigating through the page using UI persists selection.

DDeenis:
Resolves #509 

Haven't worked with `@nuxt/i18n` before, feedback welcome 🙏

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -236,11 +236,7 @@ export default defineNuxtConfig({
     locales: currentLocales,
     defaultLocale: 'en-US',
     strategy: 'no_prefix',
-    detectBrowserLanguage: {
-      useCookie: true,
-      cookieKey: 'npmx-locale',
-      fallbackLocale: 'en-US',
-    },
+    detectBrowserLanguage: false,
     langDir: 'locales',
   },
 })
--- a/app/composables/useSettings.ts
+++ b/app/composables/useSettings.ts
@@ -1,6 +1,7 @@
 import type { RemovableRef } from '@vueuse/core'
 import { useLocalStorage } from '@vueuse/core'
 import { ACCENT_COLORS } from '#shared/utils/constants'
+import type { LocaleObject } from '@nuxtjs/i18n'
 
 type AccentColorId = keyof typeof ACCENT_COLORS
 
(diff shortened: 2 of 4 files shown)


nuxt.config.ts
```diff
@@ -236,7 +236,11 @@ export default defineNuxtConfig({
     locales: currentLocales,
     defaultLocale: 'en-US',
     strategy: 'no_prefix',
-    detectBrowserLanguage: false,
+    detectBrowserLanguage: {
+      useCookie: true,
```
stephfh: Since the app generally seems to utilise localStorage to persist user settings, it could be an idea to store the currently selected lang in `useSettings` alongside it instead of using cookies? That would allow for using the settings object to get the language in the app, whenever it's needed.

danielroe: let's try that.

DDeenis: That was my first thought, but after reading docs and asking AI I decided to go with the cookie solution because:
1. Cookies are available server-side (attached to requests). Thought it was important, since the app uses SSR.
2. This is the built-in persistence method, so I guess it's encouraged?

MatteoGabriele: Wouldn't you have hydration issues using localstorage?

danielroe: useLocalStorage I think only hydrates itself after mounted, meaning there _shouldn't_ be a hydration mismatch

cookie is worse because we actually statically render much of the site, so we can't SSR based on a user's cookie

wojtekmaj: At least having locale in a cookie would allow us to maybe, one day, fix that and SSR the correct locale. Why close the doors that could stay open 🤷🏻‍♂️

DDeenis: Tbh considering that setting up cookie-based persistence is just a matter of changing the config, it shouldn't be too hard to switch between localStorage and cookies. If the localStorage is the best option at this time, and we can tolerate the one time language settings reset in case of switching to cookies, I'm happy to go with the localStorage.

danielroe: @wojtekmaj even so, I have a plan to fix the rendering of a SSG site... I just need to create the PR...

thanks @DDeenis - let's do localStorage for now

DDeenis:
Updated to use settings (`localStorage`) instead of cookies.
Used plugin to cleanly separate this functionality from the rest of the code. Would be happy to receive a code review!

PR #539, opened 2026-01-31
chore: add CSS RTL detector

userquin:
This PR adds a new UnoCSS Preset to detect, replace and warn wrong left/right usages.

There are some edge cases where we cannot do anything, for example `pl8`, if the code contains for example `const pl8 =...` the preset rule for padding will warn...

The rules covered with the preset are:
- padding left and right with dash: `pl-` and `pr-` (the rule will replace the css with `ps-` and `pe-` content respetivelly)
- marging left and right with dash: `ml-` and `mr-` (the rule will replace the css with `ms-` and `me-` content respetivelly)
- text left and right: `text-left` and `text-right` (the rule will replace the css with `text-start` and `text-end` content respetivelly)
- border left and right:  `border-l` and `border-r` (the rule will replace the css with `border-is` and `border-ie` content respetivelly)
- rounded border left and right: `rounded-l` and `rounded-r` (the rule will replace the css with `rounded-is` and `rounded-ie` content respetivelly)

### DIFF
--- a/uno.config.ts
+++ b/uno.config.ts
@@ -25,8 +25,8 @@ export default defineConfig({
       },
     }),
     // keep this preset last
-    presetRtl(),
-  ],
+    process.env.CI ? undefined : presetRtl(),
+  ].filter(Boolean),
   transformers: [transformerDirectives(), transformerVariantGroup()],
   theme: {
     font: {


danielroe (approved):

danielroe:
maybe skip this in ci to save time? not sure of the cost

userquin:
> maybe skip this in ci to save time? not sure of the cost

using `std-env` or `process.env.CI`?

userquin:
> maybe skip this in ci to save time? not sure of the cost

The cost is the same, the rules added are the same at preset-wind4, just the preset registration

PR #541, opened 2026-01-31
fix: add skeleton for pre-hydration toggles

MatteoGabriele:
Fixes the mismatch in the saved value of "Relative dates" toggle.
It also adds a skeleton placeholder for all toggles on the settings page.

https://github.com/user-attachments/assets/f12f4cc2-fa06-413c-bc39-bebb00a658d3


danielroe (approved):
perfect 👌

app/pages/settings.vue
```diff
@@ -101,27 +101,32 @@ defineOgImageComponent('Default', {
...
+                    aria-hidden="true"
+                  >
+                    <span
+                      class="pointer-events-none inline-block h-5 w-5 rounded-full shadow-sm ring-0 transition-transform duration-200 ease-in-out motion-reduce:transition-none"
+                      :class="settings.relativeDates ? 'bg-bg' : 'bg-fg-muted'"
+                    />
+                  </span>
+                </button>
+                <template #fallback>
+                  <ToggleSkeleton :label="$t('settings.relative_dates')" />
```
danielroe, line 127: btw, you can also do this with Toggle.server.vue and Toggle.client.vue (with the .server.vue component being the skeleton)

MatteoGabriele: oh wow! I've never thought about that. Pretty cool.
I can improve this later with a proper toggle component or handle it here in the same PR. Whatever you think is best.

danielroe: let's merge and iterate 🚀

PR #544, opened 2026-01-31
fix: local execute for yarn v1 and 2+ compatibility

### LINKED ISSUE #515: yarn command for npm-connector is wrong
It's currently `yarn npmx-connector` which only works (? - also not sure about that) on legacy versions of Yarn.

If the user selects their pm as Yarn we may need to display two commands, not one.

vinnymac:
For when we publish `npmx-connector` to `npm` enable package manager preferences, this will make the command for `yarn` package manager become `npx npmx-connector`.

It also means `npx` will be used anywhere `yarn` local execution is accessed. Which I think is fine. Interestingly, "execute remote" is currently set to [yarn dlx](https://yarnpkg.com/cli/dlx) still, which is only supported by Yarn 2+.

I might get shade for saying it, but we may want to consider only supporting Yarn 2+ and dropping Yarn 1 users given how old and outdated it is and so many better alternatives around. Just my two cents. As-is the current UI doesn't really support v1 of yarn, since `dlx` is not an option.

Closes #515

### DIFF
--- a/app/utils/install-command.ts
+++ b/app/utils/install-command.ts
@@ -25,6 +25,8 @@ export const packageManagers = [
     id: 'yarn',
     label: 'yarn',
     action: 'add',
+    // For both yarn v1 and v2+ support
+    // local exec defers to npx instead
     executeLocal: 'npx',
     executeRemote: 'yarn dlx',
     create: 'yarn create',


serhalp:
I think it's common enough to treat yarn v1 and yarn "berry" v2+ as effectively two different package managers. We could just have them be separate options. Or not support v1...

and technically for yarn v1 we could use `yarn global add npmx-connector && npmx-connector`

wojtekmaj:
I mean yeah but for the sake of what? It is important for devs to have install commands crafted specially for their package manager, but for one-time command no one could possibly care. I'd just scrap the whole thing and just hardcode this to npx command for every package manager :D

danielroe:
`npx` would honestly be fine for me for this one.

wojtekmaj (approved):

PR #545, opened 2026-01-31
feat: add stars to open graph images for package pages

whitep4nth3r:
I used the same star as the i-carbon:star used on the package page template, but obviously had to inline it as an SVG in this file.

Here's a selection showing the number formatting:

image

image

image

### DIFF
--- a/app/components/OgImage/Package.vue
+++ b/app/components/OgImage/Package.vue
@@ -1,8 +1,11 @@
 <script setup lang="ts">
-withDefaults(
+import { computed, toRefs } from 'vue'
+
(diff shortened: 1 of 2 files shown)


danielroe (approved):

serhalp:
@danielroe we were thinking this one was maybe superseded? https://discord.com/channels/1464542801676206113/1464544758020968448/1467302164895367280

danielroe:
I agree - but I think we could merge both.

trueberryless:
Nice, love this addition!

Issue #531, opened 2026-01-31
Period and "d" keys are intercepted by shortcuts when text box is focused


LofiTurtle:
On the main page for a package, pressing "." or "d" when a text box is focused (the search box, the ATproto login account handle popup) causes the navigation to the docs or source code, instead of typing the character.

Steps:

1. Open https://npmx.dev/@atproto/lex
2. Connect > Connect to Atmosphere
3. Type `example.com`
4. See that `examplecom` was typed, and the source code is now open in the background

PR #547, opened 2026-01-31
fix: ignore shortcuts on text input

shamilkotta:
fixes: #531

### DIFF
--- a/app/pages/[...package].vue
+++ b/app/pages/[...package].vue
@@ -4,6 +4,7 @@ import type { JsrPackageInfo } from '#shared/types/jsr'
 import { assertValidPackageName } from '#shared/utils/npm'
 import { joinURL } from 'ufo'
 import { areUrlsEquivalent } from '#shared/utils/url'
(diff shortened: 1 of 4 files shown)


serhalp:
oh, awesome @shamilkotta, thank you! I was about to open a PR to fix this but I saw yours 😄.

let me just merge https://github.com/npmx-dev/npmx.dev/pull/571 first and then I'll merge main into your branch and add similar test coverage for this bug ⏳

app/pages/[...package].vue
```diff
@@ -353,7 +361,11 @@ onKeyStroke(
   { dedupe: true },
 )
 
-onKeyStroke('c', () => {
+onKeyStroke('c', e => {
+  const target = e.target as HTMLElement
+  if (target.tagName === 'INPUT' || target.tagName === 'TEXTAREA' || target.isContentEditable) {
```
Comment (43081j) [inline app/pages/[...package].vue]: Instead of repeating this 3 times, can you extract this to a function? So we end up with `if (isTextInput(e))`

PR #549, opened 2026-01-31
fix: improve layout of peer dependencies list

### LINKED ISSUE #532: Peer dependency versions cut off on mobile

### LINKED ISSUE #550: peer package versions not being displayed with long peer
When version on peer is huge, the rest of versions not displayed (maybe related to #532), for example, using vite-plugin-pwa:

vite-plugin-pwa peer versions not beig displayed

brenoepics:
Fixes #532, fixes #550
                                                                                                             
Reworked the peer dependency layout to use justify-between for better space distribution.
Replaced the optional text badge with an information icon to make the UI more compact, freeing up space to display versions. Even on desktop, dependencies were hard/impossible to read, this improves that significantly.

This change doesn’t strictly depend on replacing the “optional” badge to the icon version, if you prefer keeping the text, we can do so, with slightly less space for versions, but still a big improvement over the current layout.

### DIFF
--- a/app/composables/useNpmRegistry.ts
+++ b/app/composables/useNpmRegistry.ts
@@ -382,9 +382,11 @@ export function useNpmSearch(
 
       // Update cache
       if (cache.value && cache.value.query === q) {
+        const existingNames = new Set(cache.value.objects.map(obj => obj.package.name))
+        const newObjects = response.objects.filter(obj => !existingNames.has(obj.package.name))
         cache.value = {
           query: q,
-          objects: [...cache.value.objects, ...response.objects],
+          objects: [...cache.value.objects, ...newObjects],
           total: response.total,
         }
       } else {
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -219,6 +219,10 @@ export default defineNuxtConfig({
     },
   },
 
+  debug: {
+    hydration: true,
+  },
+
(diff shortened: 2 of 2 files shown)


userquin:
We need to find a solution, on mobile cannot see all peer dependencies:

image

brenoepics:
hmm, yeah, I was thinking about adding a scroll, but It didn't make much sense since it could conflict with click events, the thing is that there are 2 unfixable issues: versions list can be enormous and that will add up with package name (e.g. `@organization/package-name` and versions `0.0.0 || 1.0.0 || 2.0.0 || 3.0.0...` (also why I made optional as icon instead of text) unless you guys are ok with multi line dependencies so we can show versions on 2 lines if needed.

wojtekmaj:
I think multiline would make the most sense.

danielroe:
can we adopt a similar approach to https://github.com/npmx-dev/npmx.dev/pull/585 ?

> I'm happy to have a more consistent format. justify between seems fine but some people have said it's difficult to trace across so it might be harder work for the eye
> truncation seems essential, perhaps with the ability to expand? some peer dep constraints are very long.
(I guess we could wrap ... but often it's not that important to see at a glance every single version for long version strings)
> https://discord.com/channels/1464542801676206113/1464926468751753269/1467311813191274627

PR #551, opened 2026-01-31
fix: use `useFetch` for `useDependencyAnalysis`

OrbisK:
This fixes a bug where the state key was not updated when the version or package changed.

Now, requests are deduped correctly as well. We could remove `useDependencyAnalysis()` entirely in favour of just having ``useFetch(() =>`/api/registry/vulnerabilities/${encodePackageName(toValue(packageName))}/v/${toValue(version)}`)``

------------

There will be two followup issues:

1. nuxt (upstream): deduped fetches cause theses warnings
image

2. the vulnerabilies endpoint does not handle namespaced packages correctly (https://github.com/npmx-dev/npmx.dev/issues/552)

image

### DIFF
--- a/app/components/PackageDependencies.vue
+++ b/app/components/PackageDependencies.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { useDependencyAnalysis } from '~/composables/useDependencyAnalysis'
 import { SEVERITY_TEXT_COLORS, getHighestSeverity } from '#shared/utils/severity'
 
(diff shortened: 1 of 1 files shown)


OrbisK:
@danielroe I think ci/knip is lying. Can I help you with: https://github.com/webpro-nl/knip/pull/1265 ? (I removed some unnecessary imports)

danielroe:
please! do you want permissions for my fork?

PR #556, opened 2026-01-31
feat: display author profile picture

### LINKED ISSUE #435: Use author profile pictures
Currently, to keep it minimal and avoid privacy issues with Gravatar, the username pages only contain an author placeholder with the first letter of the author name.

This feature request proposes adding the actual author image from Gravatar and gathers feedback and opinions regarding this feature (feel free to leave your comment below or upvote (👍) / downvote (😕))

Gugustinette:
Resolves #435 

How it works :
- Find username email address by exploring its public packages metadata
- If found, construct the Gravatar URL using it
- Else, fallback to the previous solution with the username's initial

I cached the data for one day, I consider it way enough for the profile picture which realistically doesn't change many times a day.

### DIFF
--- a/app/components/User/Avatar.vue
+++ b/app/components/User/Avatar.vue
@@ -5,6 +5,9 @@ const props = defineProps<{
 
 const { data: gravatarUrl } = useLazyFetch(() => `/api/gravatar/${props.username}`, {
   transform: res => (res.hash ? `/_avatar/${res.hash}?s=128&d=404` : null),
+  getCachedData(key, nuxtApp) {
+    return nuxtApp.static.data[key] ?? nuxtApp.payload.data[key]
+  },
 })
 </script>
 
--- a/server/api/gravatar/[username].get.ts
+++ b/server/api/gravatar/[username].get.ts
@@ -1,15 +1,8 @@
-import type { H3Event } from 'h3'
-import { createError, getQuery } from 'h3'
+import { createError } from 'h3'
 import * as v from 'valibot'
 import { GravatarQuerySchema } from '#shared/schemas/user'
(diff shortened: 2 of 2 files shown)


Gugustinette:
Oh ye I missed [that](https://github.com/npmx-dev/npmx.dev/blob/8825b0aa96a5e90a160e6823fe40fd44e8a33895/cli/src/npm-client.ts#L180).

The CLI has access to the user email which _**could**_ be private, so I guess we should ignore this and keep the public-data-only solution.
Now for the image itself, I think it is better to use the Gravatar url directly, so its overall management is up to the Gravatar platform and the Gravatar user account.

But maybe we need the expertise of someone more qualified in this ? However I don't know who would be a good fit.

danielroe:
a quick fix would be to add a server endpoint to proxy to gravatar, so we don't expose user IPs to a third party without consent

Gugustinette:
Oooh ok I did not think about it that way, that's on me 👀

Then we would add cache to it, which means if the user wants to change or delete its profile picture, it'll still be live on npmx for as long as the cache lives.
Is that ok ?

danielroe:
seems fine to me 👍

Gugustinette:
Made the changes : the API endpoint now returns a data url with the image in base 64.
Thus the server is the one taking care of the request to Gravatar.

trivikr:
While browsing [preview](https://npmx-96p22pcgx-poetry.vercel.app/) in three different browsers, the Gravatar of `antfu` seems to be getting cached for other users. It's not consistent, but I noticed it while viewing maintainers of vite.

danielroe:
thanks for the spot! a bit of code I hadn't tidied up

PR #561, opened 2026-01-31
fix: don't rely on pkg.updated for update date

### LINKED ISSUE #535: Recently updated sorting doesn't work properly
Pictured is the list of packages published by me (wojtekmaj). The 2nd most recently updated package appears to be @wojtekmaj/enzyme-adapter-utils which I haven't touched for 4 years.

wojtekmaj:
Fixes #535

It appears that the dates in `updated` field are not reliable at all (see #535 for details). Therefore I removed it and left only `package.date`, and now I get the exact results that I expected.

### DIFF
--- a/app/components/PackageTableRow.vue
+++ b/app/components/PackageTableRow.vue
@@ -17,8 +17,7 @@ const emit = defineEmits<{
 const pkg = computed(() => props.result.package)
 const score = computed(() => props.result.score)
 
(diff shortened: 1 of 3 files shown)


wojtekmaj:
CC @serhalp, original author of #211, would love to get your approval on this before proceeding.

danielroe:
yes, I spotted this as well - thank you ❤️

serhalp:
Thank you for fixing this @wojtekmaj! 🤝

PR #566, opened 2026-01-31
feat: pin package header 📌

thasmo:
Pins the package header during scrolling.

https://github.com/user-attachments/assets/cb1a9de4-94a2-4fc4-8935-375ed27eaf12

The template changes look a bit wild, but I basically just kept the now-pinned-content in the already existing `<header>` element and just moved the rest into a new element below the `<header>` element.

Eager to get feedback, especially on the `checkHeaderPosition` function.

❤️

### DIFF
--- a/app/pages/[...package].vue
+++ b/app/pages/[...package].vue
@@ -19,6 +19,27 @@ definePageMeta({
 
 const router = useRouter()
 
+const header = useTemplateRef('header')
+const isHeaderPinned = shallowRef(false)
+
+function checkHeaderPosition() {
+  const el = header.value
+  if (!el) return
+
+  const style = getComputedStyle(el)
+  const top = parseFloat(style.top) || 0
+  const rect = el.getBoundingClientRect()
+
+  isHeaderPinned.value = Math.abs(rect.top - top) < 1
+}
+
+useEventListener('scroll', checkHeaderPosition, { passive: true })
+useEventListener('resize', checkHeaderPosition)
+
+onMounted(() => {
+  checkHeaderPosition()
+})
+
 const { packageName, requestedVersion, orgName } = usePackageRoute()
 const selectedPM = useSelectedPackageManager()
 const activePmId = computed(() => selectedPM.value ?? 'npm')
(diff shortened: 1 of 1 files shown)


app/pages/[...package].vue
```diff
@@ -12,6 +12,32 @@ definePageMeta({
...
+  window.addEventListener('scroll', checkHeaderPosition, { passive: true })
+  window.addEventListener('resize', checkHeaderPosition)
+
+  checkHeaderPosition()
+})
+
+onBeforeUnmount(() => {
+  window.removeEventListener('scroll', checkHeaderPosition)
+  window.removeEventListener('resize', checkHeaderPosition)
+})
```
Comment (OrbisK) [inline app/pages/[...package].vue]: we should use vueuse `useEventListener` instead

app/pages/[...package].vue
```diff
@@ -12,6 +12,32 @@ definePageMeta({
 
 const router = useRouter()
 
+const header = templateRef('header')
+const isHeaderPinned = ref(false)
```
Comment (OrbisK) [inline app/pages/[...package].vue]:
```suggestion
const header = useTemplateRef('header')
const isHeaderPinned = shallowRef(false)
```

app/pages/[...package].vue
Comment (thasmo) [reply app/pages/[...package].vue]: Thanks! :)

app/pages/[...package].vue
```diff
@@ -12,6 +12,27 @@ definePageMeta({
...
+
+  isHeaderPinned.value = Math.abs(rect.top - top) < 1
+}
+
+onMounted(() => {
+  useEventListener('scroll', checkHeaderPosition, { passive: true })
+  useEventListener('resize', checkHeaderPosition)
+
+  checkHeaderPosition()
+})
```
Comment (OrbisK) [inline app/pages/[...package].vue]:
```suggestion
useEventListener('scroll', checkHeaderPosition, { passive: true })
useEventListener('resize', checkHeaderPosition)

onMounted(() => {
  checkHeaderPosition()
})
```

Comment (thasmo) [reply app/pages/[...package].vue]: Thanks! ❤️

thasmo:
Ah, something's messed up. Going to fix this. 🤔

thasmo:
> Ah, something's messed up. Going to fix this. 🤔

Should be fine now.

43081j:
Is it not possible to have a sticky header through CSS rather than manually adding a scroll listener? 

Maybe I'm missing something 🤔

thasmo:
> Is it not possible to have a sticky header through CSS rather than manually adding a scroll listener?

It is! It's "only" needed for the bottom border, which gets added conditionally when the header is in the "pinned state".

43081j:
ooh yeah i understand now! thanks for clarifying.

its a shame CSS doesn't have a pseudo-class for this 😅

PR #570, opened 2026-01-31
feat: customized opengraph images for packages

orta:
image

My first vue PR, but I spent a bunch of time with claude code figuring it out. Feels nice 👍🏻


orta:
You can test some OG images by changing this URL: 

- https://npmx-gd6urg5eq-poetry.vercel.app/__og-image__/image/danger/og.png
- https://npmx-gd6urg5eq-poetry.vercel.app/__og-image__/image/vitest/og.png
- https://npmx-gd6urg5eq-poetry.vercel.app/__og-image__/image/vue/og.png

danielroe:
thank you ❤️ 

my one comment is that I feel it might be ... too much information. (this is an inevitable consequence of showing any arbitrary amount of text, e.g. title/description)

and I'm not totally sure. I'd welcome other points of view on this.

orta:
I guess my argument is that people are just recreating this opengraph manually?

https://bsky.app/profile/jaydip.me/post/3mdzacpfkpc2d

Could do with adding the hearts too, but that'd depend on your caching

trueberryless:
Personally, I think we should maybe truncate the description of the package to avoid much text and keep only the most important stats, e.g. version, downloads/wk, maybe license and stars.

But adapting the aspect ratio makes a lot of sense for platform that do not support `2:1` 👍

IdrisGit:
- I think the box icon for package can be removed as well,  it doesn't add any important context
- The version number can be moved somewhere in the footer as well
- Move the GitHub repo detail under the package name

whitep4nth3r:
Do we need the number of dependencies? I think we want the number of likes so removing the deps would free up some space. 

I don't think we need the description, especially as we don't know the length of the description.

whitep4nth3r:
I do think we need something to identify that this is from npmx potentially. We don't have a proper logo yet (really) but something to that effect could replace the package icon.

alexdln:
It's funny - I like dependencies, but the date seems unnecessary to me.

And of course, likes, yes. You can check where they come from in the main branch

image

And yes, perhaps the description should either be removed (_among other things, sometimes markdown is left in it, and this will make it so that such a package will not be shared via preview_), the preview should have more air

"./" is our distinctive sign. I would prefer using it instead of blue box

I have a feeling the items in the bottom might not fit in some cases, and then something will break. Check those please - https://npmx.dev/package/rescript


Regarding the code:

The content needs to be loaded inside OgImagePackage. You can look at main as well. I can answer about details or jump in and update if needed.

When deciding is it one line or several lines, you need to include the version in the condition

I see that changes were lost because there were too many conflicts with my recent edits. Could you restore them on your own? You need to use your layout, but copy the logic inside the component, just like in main.

orta:
At this point, this PR is pretty dead for me, I've got atproto stuff to care about when I am in this space - if folks want to pick it up and get it in, you all have my permission

orta:
FWIW, I wouldn't recommend just black holing it, it is better than what there is now, fits the design asethetic of the other images, and does represent ~10 hours of real thinking on the problem. 

I just don't want to commit to the back and forth

PR #571, opened 2026-01-31
fix: remove compare icon from header, add `c` shortcut

serhalp:
For consistency with `settings` in the header.

This does not change the mobile menu, which still has only the icon and no keyboard shortcut.

### Before

Screenshot 2026-01-31 at 17 53 53

### After

Screenshot 2026-01-31 at 17 53 36

### DIFF
--- a/test/e2e/interactions.spec.ts
+++ b/test/e2e/interactions.spec.ts
@@ -76,3 +76,63 @@ test.describe('Search Pages', () => {
     await expect(headerSearchInput).toBeFocused()
   })
 })
+
+test.describe('Keyboard Shortcuts', () => {
+  test('"c" navigates to /compare', async ({ page, goto }) => {
+    await goto('/settings', { waitUntil: 'hydration' })
+
+    await page.keyboard.press('c')
+
+    await expect(page).toHaveURL(/\/compare/)
+  })
+
+  test('"c" on package page navigates to /compare with package pre-filled', async ({
+    page,
+    goto,
+  }) => {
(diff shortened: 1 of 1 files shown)


app/components/AppHeader.vue
```diff
@@ -72,6 +72,21 @@ onKeyStroke(
   },
   { dedupe: true },
 )
+
+onKeyStroke(
+  'c',
```
danielroe, line 77: we probably need to remove the similar handler from `[...package].vue`

serhalp: oh damn, good point.
but that one seeds the compare with the current package; they aren't quite the same.

brb putting my nuxt thinking cap on.

app/components/AppHeader.vue
```diff
@@ -72,6 +72,24 @@ onKeyStroke(
...
+  e => {
+    // Allow more specific handlers to take precedence
+    if (e.defaultPrevented) return
+
+    // Don't trigger if user is typing in an input
+    const target = e.target as HTMLElement
+    if (target.tagName === 'INPUT' || target.tagName === 'TEXTAREA' || target.isContentEditable) {
```
43081j, line 84: Are we going to end up needing this logic in most key bindings?

Assuming nuxt doesn't have anything for it, maybe we should extract it to a helper

serhalp: 😅 yeah I almost did that here but then I saw https://github.com/npmx-dev/npmx.dev/pull/547 and decided to wait and push the refactor to that branch instead

PR #573, opened 2026-01-31
feat: add toc to readme

### LINKED ISSUE #542: Readme ToC
GitHub has a very handy way of building a ToC automatically for README files. That would definitely help navigating long READMEs like this: https://npmx.dev/react-pdf#readme

vinnymac:
Adds an outline to the readme so that users can bounce around more easily. Similar to the Github ToC.

image

Closes #542

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,6 +14,43 @@ We want to create 'a fast, modern browser for the npm registry.' This means, amo
 - Layout shift, flakiness, slowness is The Worst. We need to continually iterate to create the most performant, best DX possible.
 - We want to provide information in the best way. We don't want noise, cluttered display, or confusing UI. If in doubt: choose simplicity.
 
+## Table of Contents
+
+- [Getting started](#getting-started)
+  - [Prerequisites](#prerequisites)
+  - [Setup](#setup)
+- [Development workflow](#development-workflow)
+  - [Available commands](#available-commands)
+  - [Project structure](#project-structure)
+  - [Local connector CLI](#local-connector-cli)
+- [Code style](#code-style)
+  - [TypeScript](#typescript)
+  - [Server API patterns](#server-api-patterns)
+  - [Import order](#import-order)
+  - [Naming conventions](#naming-conventions)
+  - [Vue components](#vue-components)
+- [RTL Support](#rtl-support)
+- [Localization (i18n)](#localization-i18n)
+  - [Approach](#approach)
+  - [Adding a new locale](#adding-a-new-locale)
+  - [Update translation](#update-translation)
+  - [Adding translations](#adding-translations)
+  - [Translation key conventions](#translation-key-conventions)
+  - [Using i18n-ally (recommended)](#using-i18n-ally-recommended)
+  - [Formatting numbers and dates](#formatting-numbers-and-dates)
+- [Testing](#testing)
+  - [Unit tests](#unit-tests)
+  - [Component accessibility tests](#component-accessibility-tests)
+  - [End to end tests](#end-to-end-tests)
+- [Submitting changes](#submitting-changes)
+  - [Before submitting](#before-submitting)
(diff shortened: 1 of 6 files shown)


i18n/locales/en.json
```diff
@@ -181,7 +181,8 @@
     "readme": {
       "title": "Readme",
       "no_readme": "No README available.",
-      "view_on_github": "View on GitHub"
+      "view_on_github": "View on GitHub",
+      "toc_title": "Outline"
```
vinnymac, line 206: Would we like me to add the other langs and try my best, or leave that for others to do?

app/utils/scrollToAnchor.ts
```diff
@@ -0,0 +1,33 @@
...
+  // Calculate scroll position with header offset (matches scroll-padding-top in main.css)
+  const HEADER_OFFSET = 80
+  const elementTop = element.getBoundingClientRect().top + window.scrollY
+  const targetScrollY = elementTop - HEADER_OFFSET
+
+  // Use scrollTo for precise control
+  window.scrollTo({
+    top: targetScrollY,
+    behavior: 'smooth',
+  })
```
vinnymac, line 38: I tried a few scrolling implementations here, I found this was the most reliable. Some other ones occasionally were scrolling up every once in a while in Safari, but not in Firefox/Chrome 🤔

app/composables/useActiveTocItem.ts
```diff
@@ -0,0 +1,198 @@
...
+      behavior: 'smooth',
+    })
+
+    // Monitor scroll until it settles, THEN update URL hash
+    let lastScrollY = window.scrollY
+    let stableFrames = 0
+    let rafId: number | null = null
+    const STABLE_THRESHOLD = 5 // Number of frames with no movement to consider settled
+
+    const checkScrollSettled = () => {
```
wojtekmaj: Might be easier to leverage scrollend: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollend_event

i18n/locales/en.json
wojtekmaj: IMO out of scope of this PR

app/composables/useActiveTocItem.ts
vinnymac: I tried scrollend in my second attempt actually, and it still had the Safari issue. I might revisit this though.

EDIT: must've been something else I fixed, added this back

thasmo:
Some thoughts.

Very nice, we need this! :D

If there's only a single item, would we maybe hide or disable the button?
image

IMO the TOC in the sidebar is not needed and/or is kinda misplaced, if I may say so. 😅

vinnymac:
> IMO the TOC in the sidebar is not needed and/or is kinda misplaced, if I may say so. 😅

Good catch, this was from my first attempt, accident

trueberryless:
Looks pretty clean.

I just noticed that the preview seems to affect the margin of the section ("Getting started") above, as there is no whitespace between the horizontal line and the package manager dropdown anymore:

image

garthdw:
@trueberryless this seems to be a similar issue as what happened to prod deploy.

https://discord.com/channels/1464542801676206113/1464926468751753269/1467527828747714762

The sidebar is underneath readme as well.

vinnymac:
Yea, the preview app looks nothing like local 🙃

Also the new sticky package header appears to have broken this, since the scroll offset scrolls past (hidden behind header) the ToC item you've chosen. I'll investigate that.

danielroe:
~~might be nice to support this on mobile as well 🙏~~

oh it does! maybe a cache issue last time I checked

server/api/registry/readme/[...pkg].get.ts
Comment (danielroe) [inline server/api/registry/readme/[...pkg].get.ts:1]: we need to bump the cache key here

danielroe (approved):
🔥

PR #580, opened 2026-01-31
fix: small docs change to trigger redeploy

trueberryless:
> it should be fixed now - just make a commit to change the docs and it should redeploy

[Discord - @danielroe](https://discord.com/channels/1464542801676206113/1465764080207003811/1467300197712396510)


trueberryless:
Redeployed on main, fixed now! 🙌

PR #581, opened 2026-01-31
feat: add "Least downloaded" and "Least recently updated" sort options

wojtekmaj:
This PR adds two new sort options:
- Least downloaded
- Least recently updated

Disclaimer: AI generated translations (GPT 5.2 used)


app/pages/~[username]/index.vue
```diff
@@ -24,27 +24,40 @@ const updateUrl = debounce((updates: { page?: number; filter?: string; sort?: st
...
+  | SortOption
+  | 'downloads'
+  | 'updated'
+  | undefined
+const normalizedSort: SortOption =
+  rawSort === 'downloads'
+    ? 'downloads-desc'
+    : rawSort === 'updated'
+      ? 'updated-desc'
+      : (rawSort ?? 'downloads-desc')
```
Comment (wojtekmaj) [inline app/pages/~[username]/index.vue:53]: Backwards compatibility

serhalp:
@wojtekmaj I was just noticing earlier today that I didn't apply my package list results UX revamp to the `/~:user` page, only to `/search` results and `/@:org`. I wonder if a cleaner and more complete solution here would be to port the `/@:org` changes to the `/~:user` page? you'd get all these sorts, filters, table view, etc.

wojtekmaj:
> @wojtekmaj I was just noticing earlier today that I didn't apply my package list results UX revamp to the `/~:user` page, only to `/search` results and `/@:org`. I wonder if a cleaner and more complete solution here would be to port the `/@:org` changes to the `/~:user` page? you'd get all these sorts, filters, table view, etc.

That sounds even better to me! Feel free to scrap this PR.

i18n/locales/fr-FR.json
```diff
@@ -335,8 +335,10 @@
       "get_started": "Commencer"
     },
     "sort": {
-      "downloads": "Plus téléchargés",
-      "updated": "Récemment mis à jour",
+      "downloads_desc": "Plus téléchargés",
+      "downloads_asc": "Moins téléchargés",
+      "updated_desc": "Récemment mis à jour",
+      "updated_asc": "Les moins récemment mis à jour",
```
mxdvl:
```suggestion
      "updated_asc": "Moins récemment mis à jour",
```

danielroe:
btw, I think we should avoid using AI-generated translations - we have the infrastructure set up for people to do the translation, but they won't get pinged to improve it if we auto-gen translations

(we could look at a stopgap-mechanism if we had a way to flag: 'this key is translated but needs human review', but maybe do that in another PR)

danielroe:
ah - spent some time on this before realising @serhalp was going to port the improvements from other pages to this one.

feel free to do that @wojtekmaj if you have time before he does. 🙏 

I'll close this PR in the mean time.

PR #585, opened 2026-01-31
fix: correct truncate and indents in package versions

### LINKED ISSUE #579: [Bug] [Front] Broken versions layout
Currently, the versions are displayed rather wrong - dates don't fit and the layout is a bit off. I see truncate is included, but it sometimes doesn't work well with grid and inline elements.

Image

Image

[https://npmx.dev/react](https://npmx.dev/react)

There is also a problem with dancing indents (_which does not look like it was planned_)

Image

alexdln:
resolves #579 

A set of minor improvements for the package version block:

* `min-w-0` is important because the list is inside a grid, and this is a standard traversal (_and the grid is for smooth animation_);

* `pe-2` was probably forgotten. There are screenshots in the attached issue showing this;

* `<div class="overflow-hidden">` to fix the width of the left side of the flex row;

* `block` for links so that they are correctly cropped, since inline elements work crookedly (_in general, it is always useful to make links block if they are an independent element_)


danielroe:
maybe related to https://github.com/npmx-dev/npmx.dev/pull/549, #532, #550

danielroe:
(as in, perhaps the same fix will solve those issues 🙏)

PR #588, opened 2026-02-01
feat(i18n): update German translation

trueberryless:


trueberryless:
As a native speaker, I agree with the suggestion from the AI, applied them all - autofix will sync the rest 👍 

Thanks @serhalp!

PR #590, opened 2026-02-01
fix: replace MarkdownText component with useMarkdown composable

### LINKED ISSUE #562: Incorrect markdown parsing in package description
[This PR](https://github.com/npmx-dev/npmx.dev/pull/460) introduced stripping HTML from package descriptions and seems that markdown was not taken into account and we now display empty space.
Before:
Image
After:
Image

If we don't want to display markdown or any other certain information in that spot it would be good to hide the empty div taking up space:
Image

HTML works fine:
Before:
Image
After:
Image

DDeenis:
Resolves #562

 `MarkdownText` component processes and shrinks the passed text, so we can't really know if it's empty or not on the outside, which breaks some conditional logic.
Since the actual rendering logic is quite simple (dumping html into a span), using a composable instead allows to correctly evaluate if the processed string is empty or now.

### DIFF
--- a/app/components/Package/Card.vue
+++ b/app/components/Package/Card.vue
@@ -26,10 +26,6 @@ const pkgDescription = useMarkdown(() => ({
   plain: true,
   packageName: props.result.package.name,
 }))
-
-const emit = defineEmits<{
-  focus: [index: number]
-}>()
 </script>
 
 <template>


danielroe:
maybe a good candidate for some unit tests?

DDeenis:
Apologies, forgot to update tests. Will do in the morning.

DDeenis:
Need some more fixes to do

danielroe:
I this one was introduced by https://github.com/npmx-dev/npmx.dev/commit/57975af14e6a8e73e95c70a0be00490482f47f6f

DDeenis:
Nah, I just missed the a11y test file still referencing `MarkdownText.vue`

DDeenis:
Nothing happened at `MarkdownText.vue` in January 2026

PR #592, opened 2026-02-01
fix: deduplicate packages in user packages page.

### LINKED ISSUE #583: Duplicated packages on user page
How to reproduce:
Scroll down really quick until it tries to load the next page (pagination), then when you scroll bar you see everything duplicated.

Image

brenoepics:
Fixes #583


app/composables/useNpmRegistry.ts
```diff
@@ -382,9 +382,11 @@ export function useNpmSearch(
 
       // Update cache
       if (cache.value && cache.value.query === q) {
+        const existingNames = new Set(cache.value.objects.map(obj => obj.package.name))
+        const newObjects = response.objects.filter(obj => !existingNames.has(obj.package.name))
```
serhalp, line 386: Am I understanding correctly that this is a band-aid that 'fixes' duplicates just in time rather than a fix for the root cause?

brenoepics: I tested a few different profiles with just a few packages and many packages, yeah, the root cause was likely a race condition in `useNpmSearch` itself, I added the dedupe logic in there to make sure we never duplicate packages, but it makes sense to add a fix to line 416:

https://github.com/npmx-dev/npmx.dev/blob/79187be22ee9a335922208ceaa8b0d5bf3a6314a/app/composables/useNpmRegistry.ts#L413-L421

something like:

```ts
if (!newSize || isLoadingMore.value) return
```

That might already cover this edge case that leads to duplication, though I’m not entirely sure if something else is also contributing.

Also, likely unrelated, but when a user or org has a lot of packages, quickly scrolling up seems to cause issues or at least a noticeable delay in rendering.

Let me know if I should:

- keep as is
- just fix this race condition and revert the other change while keeping `isLoadingMore`
or
- Keep the dedupe fix and also add the `isLoadingMore` guard to fix this root cause.

danielroe: it may be so.

PR #597, opened 2026-02-01
fix: hide types badge for binary-only packages

### LINKED ISSUE #458: Ignore missing types for clis
On https://npmx.dev/flint, it says "types ❌". I think that's because of #45 for us, but it made me think about it and I think it's worth hiding the types badge for packages without exports but only bins, similarly to how we (?) ignore cjs/mjs for binary-only packages (or at least should IMO). For example, [oxlint-tsgolint](https://npmx.dev/oxlint-tsgolint). I think if the only file(s) are in `bin` or if `exports` is an empty object would be sufficient. The former would still be a bit coarse for unbundled CLis but I think it'd still be worth doing anyway.

gusa4grr:
- Add `isBinary` prop to PackageMetricsBadges component
- Fix tooltip by replacing :title with AppTooltip

Closes #458

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,3 +1,3 @@
 {
-  "recommendations": ["oxc.oxc-vscode", "Vue.volar", "lokalise.i18n-ally"]
+  "recommendations": ["oxc.oxc-vscode", "Vue.volar", "lokalise.i18n-ally", "antfu.unocss"]
 }


danielroe (approved):

serhalp:
I wonder if we could share reuse this logic in the /compare page?

gusa4grr:
> I wonder if we could share reuse this logic in the /compare page?

Happy to make this change. Created #631

PR #603, opened 2026-02-01
chore: enable `noUnusedLocals` in `tsconfig`

9romise:


serhalp:
since the project uses the experimental oxlint, I wonder if we should ideally just be enabling more oxlint rules? e.g. the `no-unused-...` ones here: https://oxc.rs/docs/guide/usage/linter/rules.html 🤔

app/components/ConnectorModal.vue
9romise, line 1: Follow up to #540

9romise:
The `no-unused-..` rules are categorized under `correctness`, which we've already set to `error` in the project, but I'm not seeing any complaints from it. I suspect it might be related to Vue? 🤔

.oxlintrc.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "./node_modules/oxlint/configuration_schema.json",
+  "$schema": "https://unpkg.com/oxlint/configuration_schema.json",
```
9romise, line 2: Since we are using `vite-plus` now, we can't read the file from this path.

serhalp:
> The no-unused-.. rules are categorized under correctness, which we've already set to error in the project, but I'm not seeing any complaints from it. I suspect it might be related to Vue? 🤔

Good catch @9romise! I found it: https://github.com/oxc-project/oxc/issues/11440. Those rules aren't yet implemented in the oxlint vue plugin!

PR #604, opened 2026-02-01
refactor: remove class shortcuts

essenmitsosse:
This remove a few unnecessary UnoCSS shortcuts as follows:
- Add a `Card` component and remove `card` and `card-interactive` shortcut
- Add `SkeletonInline` and `SkeletonBlock` components and remove `skeleton` shortcut
- Remove these single-use shorts and inline their atomic classes:
    - `code-block`
    - `input-base`
- Remove unused shortcuts:
    - `divider`
    - `section`

There are some more Uno shortcuts that can be removed, but they need a little more care and probably their own PR. Those were the easy wins.

### DIFF
--- a/test/nuxt/a11y.spec.ts
+++ b/test/nuxt/a11y.spec.ts
@@ -57,6 +57,7 @@ afterEach(() => {
 import {
   AppFooter,
   AppHeader,
+  BaseCard,
   UserAvatar,
   BuildEnvironment,
   CallToAction,
@@ -206,6 +207,25 @@ describe('component accessibility audits', () => {
   describe('AppFooter', () => {
     it('should have no accessibility violations', async () => {
       const component = await mountSuspended(AppFooter)
+      const results = await runAxe(component)
+      expect(results.violations).toEqual([])
+    })
+  })
+
+  describe('BaseCard', () => {
+    it('should have no accessibility violations', async () => {
+      const component = await mountSuspended(BaseCard, {
+        slots: { default: '<p>Card content</p>' },
+      })
+      const results = await runAxe(component)
+      expect(results.violations).toEqual([])
+    })
+
+    it('should have no accessibility violations with exact match highlight', async () => {
+      const component = await mountSuspended(BaseCard, {
+        props: { isExactMatch: true },
+        slots: { default: '<p>Exact match content</p>' },
+      })
       const results = await runAxe(component)
       expect(results.violations).toEqual([])
     })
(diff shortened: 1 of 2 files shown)


danielroe:
can I ask what the motivation is for this? does it improve performance or reduce the amount of emitted css or markup?

essenmitsosse:
@danielroe  This relates to [your answer to my question](https://discord.com/channels/1464542801676206113/1464926468751753269/1467325101941985300)

Main benefits:
- Dropping unused classes to clean up code (I assume they would be removed by tree shaking anyway)
- I usually find a mix of atomic utility classes and custom classes harder to maintain, because you loose some benefits of the utility classes. This goes after the [philosophy of Tailwind](https://tailwindcss.com/docs/styling-with-utility-classes#using-components) to model style reuse via components. While there are many benefits, for me the two biggest advantages are:
    - makes it easy to cupple behavior with style in those components
    - improves discoverability and introspection — figuring out where a CSS class is used is messy at best, but every decent IDE should be able to tell you where a component is used
- avoid future naming conflicts

Currently I found it a bit hard to figure out, where some styles where coming from, and simplifying the ways components can be styled across the code base would improve DX imho.

Like I said in the initial PR message: This isn't a complete change, there are some other shortcuts, that can also be moved into components, but I wanted to get the easy ones out of the way and would open individual PRs for the more complex cases.

Should the mixture of atomic utility classes and shortcuts be a deliberate design decision, this PR can probably be closed.

app/components/SkeletonInline.vue
```diff
@@ -0,0 +1,12 @@
+<!-- Vue component that only takes a class as prop and adds `bg-bg-elevated rounded animate-skeleton-pulse` to that class -->
+<script setup lang="ts">
```
danielroe: fyi: you shouldn't create style or class bindings on vue components - vue handles it itself

app/components/BaseCard.vue
danielroe, line 1: we should avoid single-named components as this can cause confusion with native html elements

essenmitsosse: I was thinking about moving this to a folder like `/Base`. This would also solve the naming issue when importing it in `/Packages/Card.vue`.

danielroe (approved):

app/components/BaseCard.vue
danielroe: makes sense - either is good. 👍 

we can do that once we start adding more base components

PR #605, opened 2026-02-01
refactor: scope css class

essenmitsosse:
Removes global CSS for `skip-link` and Readme text and scopes the styles to the component (Readme is extracted to a component).

### DIFF
--- a/app/app.vue
+++ b/app/app.vue
@@ -91,6 +91,29 @@ if (import.meta.client) {
   </div>
 </template>
 
+<style scoped>
+/* Skip link */
+.skip-link {
+  position: fixed;
+  top: -100%;
+  inset-inline-start: 0;
+  padding: 0.5rem 1rem;
+  background: var(--fg);
+  color: var(--bg);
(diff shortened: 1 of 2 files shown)


app/pages/[...package].vue
```diff
@@ -939,12 +939,7 @@ function handleClick(event: MouseEvent) {
           </a>
         </h2>
         <!-- eslint-disable vue/no-v-html -- HTML is sanitized server-side -->
-        <article
-          v-if="readmeData?.html"
-          class="readme-content prose prose-invert max-w-[70ch]"
-          v-html="readmeData.html"
-          @click="handleClick"
-        />
+        <Readme v-if="readmeData?.html" v-html="readmeData.html" @click="handleClick" />
```
Comment (garthdw) [inline app/pages/[...package].vue:942]: Could the handle click be moved into the Readme component?

I believe it's for redirecting npm links to npmx. Which would be useful for the code view as well.

Comment (essenmitsosse) [reply app/pages/[...package].vue:942]: Yeah good point. Will open a separate PR.

Issue #516, opened 2026-01-31
Convert ChartModal to a modal route


graphieros:
`ChartModal.vue` currently sits inside a basic modal.

It is open from `PackageWeeklyDownloadStats.vue` (the sparkline chart), which passes on the weekly download data already loaded to the large chart (`PackageDownloadAnalytics.vue`).

It would be awesome to have ChartModal open in a dedicated modal route :)

PR #615, opened 2026-02-01
chore: convert chart modal to modal route

iiio2:
resolves https://github.com/npmx-dev/npmx.dev/issues/516


danielroe:
this may require updating after https://github.com/npmx-dev/npmx.dev/pull/522 was merged 🙏

iiio2:
@danielroe , I think I've done it.

danielroe:
looking, this isn't a modal route - it's just another page.

the idea of a modal route is to render the modal if you click the button, but to change the url so on hard reload it reloads a full-screen page with the same content.

ghostdevv:
Ah I think this is also resolved unfortunately - let me know if I'm wrong!

PR #621, opened 2026-02-01
fix: polish versions layout

### LINKED ISSUE #532: Peer dependency versions cut off on mobile

alexdln:
A minor revision and improvement of another PR (#549 ) based on what was discussed and implemented in the previous PR.

I've finalized the overall layout. Each version can take up a maximum of 40% of the page. If possible, the version is not cropped, but the title is. If neither fits, both are cropped. Content is always at the edges.

I also removed the span between the name and version. I see it was added during the RTL improvements, but I checked that everything looks great in RTL without it now.

The layout on the mobile version has also improved

Example with optional deps - [vitest](https://npmx.dev/vitest)

Closes #532
closes #549

### DIFF
--- a/app/components/PackageDependencies.vue
+++ b/app/components/PackageDependencies.vue
@@ -187,12 +187,10 @@ const sortedOptionalDependencies = computed(() => {
             </NuxtLink>
             <span
               v-if="peer.optional"
-              class="shrink-0 text-fg-muted hover:text-fg transition-colors duration-200 cursor-help inline-flex items-center"
+              class="px-1 py-0.5 font-mono text-[10px] text-fg-subtle bg-bg-muted border border-border rounded shrink-0"
               :title="$t('package.dependencies.optional')"
-              :aria-label="$t('package.dependencies.optional')"
-              role="img"
             >
-              <span aria-hidden="true" class="i-carbon:information w-3 h-3 block" />
+              {{ $t('package.dependencies.optional') }}
             </span>
           </div>
           <NuxtLink


app/components/PackageDependencies.vue
```diff
@@ -188,19 +187,20 @@ const sortedOptionalDependencies = computed(() => {
...
-              class="px-1 py-0.5 font-mono text-[10px] text-fg-subtle bg-bg-muted border border-border rounded shrink-0"
+              class="shrink-0 text-fg-muted hover:text-fg transition-colors duration-200 cursor-help inline-flex items-center"
               :title="$t('package.dependencies.optional')"
+              :aria-label="$t('package.dependencies.optional')"
+              role="img"
             >
-              {{ $t('package.dependencies.optional') }}
+              <span aria-hidden="true" class="i-carbon:information w-3 h-3 block" />
             </span>
           </div>
```
alexdln, line 195: Hmm, that's not mine 👀

danielroe: wdym?

alexdln: I'm not sure if this was agreed. I can revert it and the author might do an additional MR for this part

Please tell what works best for you

app/components/PackageDependencies.vue
```diff
@@ -188,19 +187,20 @@ const sortedOptionalDependencies = computed(() => {
             </NuxtLink>
             <span
               v-if="peer.optional"
-              class="px-1 py-0.5 font-mono text-[10px] text-fg-subtle bg-bg-muted border border-border rounded shrink-0"
+              class="shrink-0 text-fg-muted hover:text-fg transition-colors duration-200 cursor-help inline-flex items-center"
               :title="$t('package.dependencies.optional')"
+              :aria-label="$t('package.dependencies.optional')"
```
danielroe: this was included in the other PR

I think the current `?` icon isn't sufficient to communicate a peer dep is optional, and we should either choose another icon or keep the badge

app/components/PackageDependencies.vue
danielroe: understood - added a comment

app/components/PackageDependencies.vue
alexdln: Meanwhile, an icon is used for outdated deps, and it seems useful to standardize them. But I don't even know what icon would be appropriate (_if an icon is used in one place to indicate a dependency specifics, it is useful to use icons everywhere in such cases_). I agree that the "?" is indeed read as a tooltip 🤔

alexdln: So, I think I'll revert to the way it was and create an additional issue. I hope we would find a suitable one there

danielroe: agreed. for now, let's revert this change and then we can open a new PR to update/align icons

alexdln: Done, thanks!

(_Just for history - issue for icons - #628_)

Issue #536, opened 2026-01-31
[Bug] Settings UI mismatch


gameroman:
1. Open https://npmx.dev/settings

<img width="703" height="571" alt="Image" src="https://github.com/user-attachments/assets/bce24902-4f39-4a4f-a9ee-cf3028854a26" />

2. Enable `Relative dates`

<img width="722" height="323" alt="Image" src="https://github.com/user-attachments/assets/bfd0da47-7acf-46b1-b54c-e0e2ef247121" />

3. Open for example https://npmx.dev/nuxt in a new tab to verify it was enabled

<img width="353" height="295" alt="Image" src="https://github.com/user-attachments/assets/26e54660-da59-4c4c-ac8e-56932f20fff8" />

4. Reload settings page and you'll see `Relative dates` as disabled

<img width="704" height="326" alt="Image" src="https://github.com/user-attachments/assets/4af8d4c9-36bd-4b85-92ae-38131706433d" />

5. Reload page with https://npmx.dev/nuxt to verify it is still enabled

<img width="343" height="301" alt="Image" src="https://github.com/user-attachments/assets/d6bbaf01-4881-40fb-b376-f0b3c5e16fb1" />

---

> [!WARNING]
> For correct reproduction might be a good idea to use a different browser

danielroe:
this was resolved in https://github.com/npmx-dev/npmx.dev/pull/541

PR #622, opened 2026-02-01
fix: add options to prevent ssr mismatch on settings

damevin:
### What this PR do and why ?

Fixes a settings persistence issue where toggles could reset when opening a package page in a new tab or reloading. We now initialize settings after mount without overwriting existing localStorage, and add an E2E test to ensure settings persist across tabs and reloads.

### How to setup and validate locally

- Go to `/settings`, toggle any setting
- Open `/nuxt` in a new tab
- Reload /settings and `/nuxt` -> toggles stay as set
- Optional: `pnpm test:browser -- tests/settings.spec.ts`

Related to https://github.com/npmx-dev/npmx.dev/issues/536


danielroe:
I can't actually reproduce this on npmx.dev

damevin:
I think the bug was fixed by this #560 since we now have server/client component I've tested yesterday night lol, but it's looks good now

danielroe:
moving too fast! ⚡

PR #623, opened 2026-02-01
fix: wait more before re-rendering the chart

graphieros:
Maybe overkill but hopes it works in preview


graphieros:
Flaky in prod

Issue #564, opened 2026-01-31
a11y: no focus trap when mobile menu is opened


wojtekmaj:
Steps to reproduce:
- Small screen
- Tab tab tab... until you're at the Menu button
- Activate it
- Menu opens
- Tab again

Expected: First menu item focued

Actual: Some link underneath the menu gets focused

PR #624, opened 2026-02-01
fix: a11y mobile menu items

shamilkotta:
fixes: #564 

https://github.com/user-attachments/assets/f4c0538c-9d86-48cb-b7c4-b6567b9c4ceb


app/composables/useFocusTrap.ts
```diff
@@ -0,0 +1,56 @@
...
+  '[tabindex]:not([tabindex="-1"])',
+].join(',')
+
+function getFocusableElements(container: HTMLElement): HTMLElement[] {
+  return Array.from(container.querySelectorAll<HTMLElement>(FOCUSABLE_SELECTOR)).filter(
+    el => el.offsetParent !== null,
+  )
+}
+
+export function useFocusTrap() {
```
OrbisK: cant we use https://vueuse.org/integrations/usefocustrap/ here?

shamilkotta: sure, i'll update it, was not aware of this utility

PR #625, opened 2026-02-01
style: fix alignment for invalid user

iiio2:
When the user is invalid, the username moves on the bottom. I think this should be on the middle like other valid user. 

https://npmx.dev/~danielroe
https://npmx.dev/~danielroe007


danielroe (approved):
nice!

PR #627, opened 2026-02-01
fix: simplify translation sync logic and maintain original order

shuuji3:
Previously, the translation script added new keys at the end of each object, causing key order differences between the original `en.json` and other locale files. This made it difficult to compare translations side-by-side (I usually do).

This PR refactors the sync logic. Instead of constructing a flattened object and applying 'add' and 'remove' operations independently, the new logic traverses the reference object and updates values in-place. This ensures that all locale files strictly maintain the same structure and key order as `en.json`.

This should not change the current behavior and output text (except reordering) but just to confirm, this is my understanding of the script's behavior:
- without `--fix`: only removes extra keys that are not in en.json.
- with `--fix`: fully synchronizes the locale with en.json (adds missing keys and removes extra ones)

### DIFF
--- a/scripts/compare-translations.ts
+++ b/scripts/compare-translations.ts
@@ -18,19 +18,6 @@ const COLORS = {
 
 type NestedObject = { [key: string]: unknown }
 
-const flattenObject = (obj: NestedObject, prefix = ''): Record<string, unknown> => {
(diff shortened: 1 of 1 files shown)


danielroe (approved):
❤️

PR #630, opened 2026-02-01
fix: prevent 404 on code viewer direct URL access

onmax:
## Problem

1. Go to https://npmx.dev/lodash
2. Click code and select a file
3. Reload page

Code viewer 404s on direct URL access: https://npmx.dev/code/lodash/v/4.17.23/_apply.js

## Cause

`server/routes/[...path].get.ts` catch-all intercepts all requests and throws 404 for non `.well-known/skills` paths.

## Fix

Move to middleware pattern - can pass through non-matching requests instead of throwing.

- `server/routes/[...path].get.ts` (104 lines) → `server/middleware/well-known-skills.ts` (55 lines)


onmax:
For explicit ISR config, we could add:

```ts
'/:pkg/.well-known/skills/**': { isr: 3600 },
'/:scope/:pkg/.well-known/skills/**': { isr: 3600 },
```

PR #639, opened 2026-02-01
feat: update all sidebar sections to be collapsible sections

garthdw:
Move top level sections into their own component.

image

Missing sections were:
- keywords
- maintainers
- compatibility
- install scripts
- AI skills

### DIFF
--- a/app/components/Package/Keywords.vue
+++ b/app/components/Package/Keywords.vue
@@ -1,4 +1,6 @@
 <script setup lang="ts">
+import { NuxtLink } from '#components'
+
 defineProps<{
   keywords?: string[]
 }>()


whitep4nth3r:
I made a little screen recording of the experience and encountered a couple of issues when navigating with a keyboard.

1. _Sometimes_ I am unable to use enter to close the section again. This happens most noticeably on the "latest version" section most obviously right at the end of the recording. When the group opens, the toggle button sometimes loses focus.
2. For keyboard navigation, in dark mode, the focus outlines (or underlines) are not very clear in terms of colour contrast. I had trouble seeing where I was tabbed and focussed to. This is not specifically related to this PR but was very noticeable in testing.

https://github.com/user-attachments/assets/3c7e5561-3375-46ff-96a3-795768fd9ece

whitep4nth3r:
I created an issue to address focus outlines so you can ignore that in this PR :)

https://github.com/npmx-dev/npmx.dev/issues/647

garthdw:
Thanks @whitep4nth3r, the collapsibles under versions are done "manually" as opposed to using the new component. They also seem to have a lot more logic than a simple collapse/no collapse.

So likely there's some DOM stuff happening that's shifting focus. Certainly something we will want to look at.

danielroe:
I think we are missing only the playgrounds component

PR #640, opened 2026-02-01
fix: use correct well-known path

danielroe:
closes https://github.com/npmx-dev/npmx.dev/pull/637


onmax:
now this https://npmx.dev/ui/@nuxt/.well-known/skills/index.json crashes

i reverted `<pkgname>` with `<@org>`

PR #641, opened 2026-02-01
feat: unifying npm registry requests with caching

OrbisK:
I think it is not an easy change to refactor all npm requests to `useNpmRegistry` (yet). As a first step, we could migrate all npm requests to `$npmRegistry` and `$npmApi`. Wdyt?

### DIFF
--- a/app/composables/npm/useNpmSearch.ts
+++ b/app/composables/npm/useNpmSearch.ts
@@ -54,6 +54,8 @@ export function useNpmSearch(
   query: MaybeRefOrGetter<string>,
   options: MaybeRefOrGetter<NpmSearchOptions> = {},
 ) {
+  const { $npmRegistry } = useNuxtApp()
+
   // Client-side cache
   const cache = shallowRef<{
     query: string
@@ -144,7 +146,6 @@ export function useNpmSearch(
 
   // Fetch more results incrementally (only used in incremental mode)
   async function fetchMore(targetSize: number): Promise<void> {
-    const { $npmRegistry } = useNuxtApp()
     const q = toValue(query).trim()
     if (!q) {
       cache.value = null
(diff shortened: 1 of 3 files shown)


danielroe:
this looks good!

would you investigate why the browser tests might be failing?

OrbisK:
> this looks good!
> 
> would you investigate why the browser tests might be failing?

Working on it. Maybe you know what the issue is. e.g. the org page is broken.

https://npmxdev-git-fork-orbisk-feat-concept-npm-requests-poetry.vercel.app/@nuxt

image

Here: https://github.com/npmx-dev/npmx.dev/pull/641/changes#diff-93fe589cec6bc2b865b353a91ba90b4407706ff4baf16189a4c2c538351aa26aR523

OrbisK:
I think the problem is that `cachedFetch` from the plugin uses the server side fetch on the client 🤷🏽‍♂️ 

I think I have a fix.

OrbisK:
I am super confused. Are `$npmRegistry` request meant to run client side? Because there are cors issues when requesting the orgs

OrbisK:
Okay. The request returns the data and 200 but cors is still blocking it 🤔

danielroe (approved):

danielroe:
$npmRegistry should be idempotent, yes. (it should be safe to call in either case)

if there's stale data (or missing data) from the server, then it will refetch on the client

OrbisK:
Progress: I have fixed it locally. But i am not sure how...

app/composables/useNpmRegistry.ts
```diff
@@ -33,6 +30,8 @@ async function fetchBulkDownloads(
   const downloads = new Map<string, number>()
   if (packageNames.length === 0) return downloads
 
+  const { $npmApi } = useNuxtApp()
+
```
OrbisK: @danielroe this line seems to be the problem.

replacing it by a simple mock "fixes" the issue. Any ideas 😅 
```ts
 const $npmApi = ()=>{
  return Promise.resolve({data: null,})
 }
```

43081j:
caught it up from main for you @OrbisK (lots of conflicts)

code has moved around a fair bit so can you have a read through again to be sure its all good?

OrbisK:
> caught it up from main for you @OrbisK (lots of conflicts)
> 
> code has moved around a fair bit so can you have a read through again to be sure its all good?

looks good. 👍🏽 Not sure how to fix the Nuxt instance issue though. I think this might be a Nuxt bug.

Issue #617, opened 2026-02-01
Change origin of md relative links


alexdln:
Some projects (like [here](https://npmx.dev/robindoc/v/3.7.9#user-content-advantages)) have relative links to other readme files and other internal md files within the readme.

For everything except md, everything looks correct, but for md - the links start pointing to jsdelivr with a 404 error.

We need to specify the user's Github origin for such links (_just like npm_)

<img width="516" height="157" alt="Image" src="https://github.com/user-attachments/assets/4838e100-a198-41ee-8a7c-a4650c8f84c1" />

Links to https://cdn.jsdelivr.net/npm/robindoc/docs/03-customization/README.md

> [!WARNING]
> If GitHub is unknown, we should leave the links unchanged (npm has the same behavior) ([example pkg](https://npmx.dev/atproxy/v/0.0.0-experimental-26a68cb#:~:text=under%20your%20control.-,Read%20docs,-Usage))

PR #643, opened 2026-02-01
fix: support git protocol in readme

alexdln:
A microfix for link formatting in readmes. Because there was no condition here, the existing logic didn't work in certain scenarios (for example [here](https://npmx.dev/robindoc/v/3.7.9#user-content-advantages)). Judging by the method comment, this logic was originally planned, but apparently got lost.

Partially closes issue #617, but I think I still need to think about where to navigate users. For now, it's more likely to be a fix for what's expected already


serhalp (approved):
thanks!

if you're up for it, a follow-up PR adding a test would for this be great!

PR #644, opened 2026-02-01
fix: remove max-width for readme container on larger screens

thasmo:
Removes the max-width for the README container on larger screens where it's not needed but instead creates a small layout issue.

**bug**
image

**fix**
image


serhalp:
ooh hell yes @thasmo! this has been bugging me! 🎉

PR #645, opened 2026-02-01
feat: allow align and style on img tag

garthdw:
Fixes obsidium/v/2.0.0 package readme

image
Left is old, right is new

### DIFF
--- a/server/utils/readme.ts
+++ b/server/utils/readme.ts
@@ -139,7 +139,7 @@ const ALLOWED_TAGS = [
 
 const ALLOWED_ATTR: Record<string, string[]> = {
   a: ['href', 'title', 'target', 'rel'],
-  img: ['src', 'alt', 'title', 'width', 'height', 'align', 'style'],
+  img: ['src', 'alt', 'title', 'width', 'height', 'align'],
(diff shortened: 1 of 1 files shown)


server/utils/readme.ts
```diff
@@ -139,7 +139,7 @@ const ALLOWED_TAGS = [
 
 const ALLOWED_ATTR: Record<string, string[]> = {
   a: ['href', 'title', 'target', 'rel'],
-  img: ['src', 'alt', 'title', 'width', 'height'],
+  img: ['src', 'alt', 'title', 'width', 'height', 'align', 'style'],
```
43081j: `style` can cause quite a lot of chaos so im not too sure about this

i feel like we should have an allow list of CSS properties within `style` instead of allowing everything

danielroe: agreed! github would be an example to copy here

garthdw: Gonna remove the style attribute for now.

Looks like we allow style on a few elements already, so we likely want to do it across them, which will likely be a bigger PR.

43081j (approved):

PR #646, opened 2026-02-01
fix(i18n): package count translations

kuosandys:
Hi there, I was checking out the project (really love the UX btw!) and noticed a few issues with the package count displays:

- total package count wasn't displayed in pagination view for some languages (e.g. zh-CN)
- number formatting for packages was inconsistent between package count and pagination controls
- missing the count argument to enable pluralization

These seemed like small fixes, so I went ahead and fixed them. This is my first contribution to the project, so if there are any additional changes needed or if it's preferred to open an issue first to discuss, I'd be happy to do that.

### DIFF
--- a/lunaria/files/es-419.json
+++ b/lunaria/files/es-419.json
@@ -157,7 +157,7 @@
       "locally": "Ejecutar localmente"
     },
     "readme": {
-      "title": "Léeme",
+      "title": "Léame",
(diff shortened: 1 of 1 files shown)


danielroe (approved):

danielroe:
thank you ❤️

PR #649, opened 2026-02-01
fix: configure general rule for trailing slash

### LINKED ISSUE #648: Configure trailing slash redirects
TrailingSlash is the "/" symbol at the end of the URL. There's currently no setting for this, so we can open either npmx.dev/robindoc or npmx.dev/robindoc/, but the latter will return a 404.

So, we know the package exists, but it's showing an error.

Showing the same package at two URLs isn't ideal either. Therefore, it's better to configure trailingSlash (_redirect to a page without it to match the current behavior_)

alexdln:
Trailing slash setup. For example I often visit a page with a slash in the end and see a 404 error, even though the package exists. Now it will redirect.

I did this using vercel config. I'm a little concerned that this approach won't work during local development, but it seems better than leaving things as they are. Any better ideas would be appreciated [if needed]

Closes #648

### DIFF
--- a/app/middleware/trailing-slash.global.ts
+++ b/app/middleware/trailing-slash.global.ts
@@ -1,0 +1,23 @@
+/**
+ * Removes trailing slashes from URLs.
+ *
+ * This middleware only runs in development to maintain consistent behavior.
+ * In production, Vercel handles this redirect via vercel.json.
+ *
+ * - /package/vue/ → /package/vue
+ * - /docs/getting-started/?query=value → /docs/getting-started?query=value
+ */
+export default defineNuxtRouteMiddleware(to => {
+  if (!import.meta.dev) return
+
+  if (to.path !== '/' && to.path.endsWith('/')) {
+    return navigateTo(
+      {
+        path: to.path.slice(0, -1),
+        query: to.query,
+        hash: to.hash,
+      },
+      { replace: true },
+    )
+  }
+})


danielroe:
this is good

we can also add a global route middleware, dev only, which redirects

alexdln:
@danielroe 
Does it make sense for us to do different approaches for dev and prod? Or should we always redirect through middleware?

danielroe:
it'll be more performant to do the cdn-level redirect as opposed to middleware

alexdln:
_Done_

serhalp:
@danielroe can this not be accomplished with nitro `routeRules`?

(not a blocker!)
I was separately looking into self-hosting instructions for folks to deploy their own npmx instance and found quite a few Vercelisms to work through. This adds another. Is it a goal of this project to avoid these in the long term?

danielroe:
absolutely it should be self-hostable - I'm not aware of anything at all in the project that's vercel-specific though. 🤔

all the config for vercel cache is only enabled when deployed on vercel. that's why I extracted it into a module. you should be able to build + run locally with no issues.

this shouldn't add another, if done correctly.

alexdln:
It sounds like we should enable middleware not only for dev, but for non-Vercel

danielroe:
i think we have to be careful

there are some providers, like netlify(?) that add trailing slashes for static routes and cause infinite loops

maybe better to sanitise package name to remove slash on package page?

alexdln:
It's not ideal to have one page with two different paths. Plus, it might be a problem for search engines (npmx have canonical URLs, but redirects are still better)

Also, this will be a very targeted fix, but the problem will persist on all other pages and potentially flare up later

danielroe:
we should add a rel=canonical for sure 

and we already have a util to extract thd package name from path. I think we just change that

I was also working on a PR to make Vue Router extract params which is ultimately what we'd want I think, but I'm blocked for now

alexdln:
To clarify: it would be better if we won't redirect, but instead show the correct page for both `/package-name` and `/package-name/`. And this would apply to all pages?

danielroe:
yes

there is a nuxt config option to respect trailing slashes but it has caused so many issues with some providers that I'd prefer to leave it off if we can

alexdln:
I've created a new PR (#673). Should we close this one or leave it for further discussion and a final decision?

danielroe:
I would merge this one too, as a nice-to-have

danielroe (approved):

PR #652, opened 2026-02-01
fix(ui): make all icons inline-block by default

RYGRIT:


Gugustinette:
It would have been nice to add some context in the PR description, even though it is pretty simple 🙏

Before :
Capture d’écran 2026-02-01 à 17 54 18

After :
Capture d’écran 2026-02-01 à 17 54 53

Looks good to me 👍

RYGRIT:
@Gugustinette Thank you very much for your suggestion.

PR #654, opened 2026-02-01
feat: add contributor number to about page and improve legibility

whitep4nth3r:
I didn't want to manually count the avatars when looking at how awesome the amount of contributors is, so I added the contributor number to the contributors title.

The only thing is that this doesn't account for i18n e.g. right to left languages, so I'd love some input on that.

This PR also increases the size of the headings on the about page for legibility.

I think the count is good to show for a quick glance at how active the community is, and will help with outreach stuff in the future.


whitep4nth3r:
~~Is there a way to just rerun the a11y test @patak-dev because it really shouldn't have failed and the error is weird so I think it's a red herring.~~

Fixed the a11y issues that crept in somewhere!

PR #655, opened 2026-02-01
fix: make sure chevron in pagination result select is visible

stephfh:
Making sure the “show amount of results / page“ select has its downwards chevron visible. 

SCR-20260201-opzk

SCR-20260201-oqdt

Repro on production:
Search for a package and swap the result list to list view so that the pagination is visible underneath, e.g. https://npmx.dev/search?q=vue


Gugustinette:
LGTM 👍

stephfh:
Ah, no worries, then this is obsolet any way, good solution to make it the default, so you don't have to think about it on any icon.

I'll close. :D

PR #656, opened 2026-02-01
fix: show "none" if package has no license

Nadjaroemer:
Currently we hide the license column if there's no license. With this PR render "None" instead, similar to what npmjs.com does.

license-none

### DIFF
--- a/lunaria/files/en-US.json
+++ b/lunaria/files/en-US.json
@@ -298,7 +298,8 @@
       "no_types": "No TypeScript types"
     },
     "license": {
-      "view_spdx": "View license text on SPDX"
+      "view_spdx": "View license text on SPDX",
+      "none": "None"
     },
     "vulnerabilities": {
       "no_description": "No description available",


whitep4nth3r:
Great idea!

We'll need translations for the word "None". The files live in i18n/locales.

We may need to mobilise the i18n team in the Discord! Would be good to post in the channel: https://discord.com/channels/1464542801676206113/1465734070817259601

app/pages/[...package].vue
```diff
@@ -735,12 +735,13 @@ function handleClick(event: MouseEvent) {
...
         >
-          <div v-if="pkg.license" class="space-y-1 sm:col-span-2">
+          <div class="space-y-1 sm:col-span-2">
             <dt class="text-xs text-fg-subtle uppercase tracking-wider">
               {{ $t('package.stats.license') }}
             </dt>
             <dd class="font-mono text-sm text-fg">
-              <LicenseDisplay :license="pkg.license" />
+              <LicenseDisplay v-if="pkg.license" :license="pkg.license" />
+              <span v-else>None</span>
```
Comment (stephfh) [inline app/pages/[...package].vue]: Additionally to what @whitep4nth3r said here https://github.com/npmx-dev/npmx.dev/pull/656#issuecomment-3831318397 

You could add the variant for english in en.json, IIRC `$t('package.license.none')` should be according to the naming schema for language keys.

The i18n:check for the language files should then be automagically infer that the key is missing for the respective file.

Comment (Nadjaroemer) [reply app/pages/[...package].vue]: Thank you so much @stephfh! I configured the variant in `en.json`, want to have a second look?

Nadjaroemer:
> Great idea!
> 
> We'll need translations for the word "None". The files live in i18n/locales.
> 
> We may need to mobilise the i18n team in the Discord! Would be good to post in the channel: https://discord.com/channels/1464542801676206113/1465734070817259601

Thank you so much @whitep4nth3r . I've made some changes and will reach out in the i18n channel.

app/pages/[...package].vue
Comment (stephfh) [reply app/pages/[...package].vue]: Looks good on my machine. :D

stephfh (approved):

PR #657, opened 2026-02-01
feat: put env string next to 'npmx'

### LINKED ISSUE #1135: Better design for env styling
#657 adds env name string next to "npmx" name to be able to distinguish deploy environments such as "dev", "preview", and "alpha", but the current styling is not so good.

shuuji3:
How about adding the env name ('dev', 'preview', 'canary', or 'alphe' (for 'release' env)) next to the "npmx" like Elk (https://main.elk.zone vs https://elk.zone) in addition to the footer?

The logo color is already different but sometimes not enough to distinguish (ref. https://discord.com/channels/1464542801676206113/1464544843740217455/1467534525218295889 😄)

We could drop "alpha" for production if it's preferable.

Closes #1135 

---
Screenshot of home page with 'dev' string right next to 'npmx' name

Screenshot of home page with 'dev' string right next to 'npmx' name (light theme)

Screenshot of home page with 'alpha' string right next to 'npmx' name
(manually changed string to 'alpha' here, color is not for 'release')

---
Screenshot of header title with 'dev' string

Screenshot of header title with 'dev' string (light theme)

Screenshot of header title with 'alpha' string
(same)

### DIFF
--- a/app/components/AppHeader.vue
+++ b/app/components/AppHeader.vue
@@ -213,14 +213,16 @@ onKeyStroke(
           :to="{ name: 'index' }"
           :aria-label="$t('header.home')"
           dir="ltr"
-          class="inline-flex items-center gap-1 header-logo font-mono text-lg font-medium text-fg hover:text-fg/90 transition-colors duration-200 rounded"
+          class="relative inline-flex items-center gap-1 header-logo font-mono text-lg font-medium text-fg hover:text-fg/90 transition-colors duration-200 rounded"
         >
-          <AppLogo class="w-8 h-8 rounded-lg" />
-          <span>npmx</span>
-          <sup class="text-sm ms--1.5 italic text-fg-muted">
-            <!-- TODO: improve styling and show 'alpha' until March 3 annoucement -->
-            {{ env === 'release' ? '' : env }}
-          </sup>
+          <AppLogo class="w-7 h-7 rounded-lg" />
+          <span class="pb-0.5">npmx</span>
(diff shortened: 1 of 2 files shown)


serhalp:
I like it. I wouldn't put anything for prod builds. and I wonder if it would be better in the header somewhere? next to the logo? otherwise this env flag only shows on the landing page and not on the search, compare, setting, etc. pages.

danielroe:
this is a nice idea. I think it needs a design tweak as italic monospace isn't the most attractive font - I'll take a look

ghostdevv:
rebased! updated ss (with blue enabled of course):

image

essenmitsosse:
Yeah this is a great idea! Definitely solves a real issue! ❤️

shuuji3:
Maybe we could show only 'dev' and 'preview' for now until we come up with better design so that we can reduce development friction?

shuuji3:
I removed 'alpha' and created the issue for better desgin https://github.com/npmx-dev/npmx.dev/issues/1135.

danielroe (approved):
perfect!

alexdln:
@shuuji3 Thanks for the implementation and the idea ❤️

I adjusted the styles a bit to improve a bit general view on different devices, if you have any other UI ideas, feel free to ping me, I'll be happy to discuss and help

shuuji3:
@alexdln Thanks, it looks good to me! I like colorizing it with accent colors 🎨

PR #660, opened 2026-02-01
fix(a11y): add missing h1 in the search page

MatteoGabriele:
Adds the h1 title to the Search page for a11y.

### DIFF
--- a/lunaria/files/en-US.json
+++ b/lunaria/files/en-US.json
@@ -27,7 +27,7 @@
     "found_packages": "No packages found | Found 1 package | Found {count} packages",
     "updating": "(updating...)",
     "no_results": "No packages found for \"{query}\"",
-    "title": "Search",
+    "title": "search",
     "not_taken": "{name} is not taken",
     "claim_prompt": "Claim this package name on npm",
     "claim_button": "Claim \"{name}\"",
--- a/lunaria/files/it-IT.json
+++ b/lunaria/files/it-IT.json
@@ -24,7 +24,7 @@
     "found_packages": "Trovati {count} pacchetti",
     "updating": "(aggiornando...)",
     "no_results": "Nessun pacchetto trovato per \"{query}\"",
-    "title": "Cerca",
+    "title": "cerca",
     "not_taken": "{name} è disponibile",
     "claim_prompt": "Ottieni il nome di questo pacchetto su npm",
     "claim_button": "Ottieni \"{name}\"",


app/pages/search.vue
```diff
@@ -589,9 +589,12 @@ defineOgImageComponent('Default', {
 
 <template>
   <main class="flex-1" :class="{ 'overflow-x-hidden': viewMode !== 'table' }">
+    <h1 id="search-heading-title" class="sr-only">
+      {{ $t('search.results') }}
+    </h1>
     <!-- Results area with container padding -->
     <div class="container-sm py-6">
-      <section v-if="query" :aria-label="$t('search.results')">
+      <section v-if="query" aria-labelledby="search-heading-title">
```
knowler: I wouldn’t have a region landmark here (see #384 for more details). The results are already the main content of the page.

whitep4nth3r:
Why does it need to be screen reader only? 🙃

MatteoGabriele:
> Why does it need to be screen reader only? 🙃

The current design doesn't have a title, so I left it as is while adding the missing H1.
We can add a title if we want to, but I'm not sure it adds any value, since the focus is on the results.

serhalp:
Pardon my quality drawing, but it does seem like there's a very natural space for an h1 on the search page, so maybe let's give it a try?

Screenshot 2026-02-01 at 12 19 51

stephfh:
Having a visible h1 on the search page would also be consistent to the other sub pages, considering they have one as well. 😉

whitep4nth3r (approved):

garthdw:
image
image

There's some inconsistency in top padding between the pages, I do feel the tighter padding is better. Potentially a separate PR though.

MatteoGabriele:
@garthdw There are several inconsistencies across different pages. Since the Package page is most often accessed from the Search page, meaning you'll switch between them frequently, I adopted the heading style from the Package page to minimize spacing and text size differences. 

I would love to adjust the container size, headings, and padding across all pages to ensure a consistent look. However, this might be better suited for a different PR since this one started with a different scope.

i18n/locales/en.json
```diff
@@ -27,7 +27,7 @@
     "found_packages": "No packages found | Found 1 package | Found {count} packages",
     "updating": "(updating...)",
     "no_results": "No packages found for \"{query}\"",
-    "results": "Search results",
+    "title": "Search",
```
danielroe:
```suggestion
    "title": "search",
```

PR #664, opened 2026-02-01
fix: use fast-npm-meta to get latest versions

danielroe:
adopts https://github.com/antfu/fast-npm-meta

there are some much faster endpoints with quick data...

in addition I've taken the opportunity to significantly slim down the payloads to remove unneeded data on past versions. this moves us from ~50k -> 20k for the payload for `nuxt`, for example.

### DIFF
--- a/app/composables/useNpmRegistry.ts
+++ b/app/composables/useNpmRegistry.ts
@@ -7,13 +7,13 @@ import type {
   NpmPerson,
   PackageVersionInfo,
 } from '#shared/types'
-import type { PackageVersionsInfoWithMetadata, ResolvedPackageVersion } from 'fast-npm-meta'
+import { getVersions } from 'fast-npm-meta'
+import type { ResolvedPackageVersion } from 'fast-npm-meta'
 import type { ReleaseType } from 'semver'
 import { mapWithConcurrency } from '#shared/utils/async'
 import { maxSatisfying, prerelease, major, minor, diff, gt, compare } from 'semver'
 import { extractInstallScriptsInfo } from '~/utils/install-scripts'
 import type { CachedFetchFunction } from '#shared/utils/fetch-cache-config'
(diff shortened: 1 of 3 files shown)


server/utils/npm.ts
```diff
@@ -23,6 +24,53 @@ export const fetchNpmPackage = defineCachedFunction(
...
+export const fetchFastNpmMeta = defineCachedFunction(
+  async (name: string, specifier?: string): Promise<FastNpmMetaResponse> => {
+    const encodedName = encodePackageName(name)
+    const url = specifier
+      ? `${FAST_NPM_META_API}/${encodedName}@${encodeURIComponent(specifier)}`
+      : `${FAST_NPM_META_API}/${encodedName}`
+    return await $fetch<FastNpmMetaResponse>(url)
+  },
+  {
+    maxAge: 60 * 5,
```
Gugustinette: Shouldn't we use something like `CACHE_MAX_AGE_FIVE_MINUTES` ?

shared/types/fast-npm-meta.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * Types for fast-npm-meta API responses
+ * @see https://github.com/antfu/fast-npm-meta
```
serhalp: 😁 👀
Image

can we just import the types instead of duplicating them? https://npmx.dev/code/fast-npm-meta/v/1.0.0/dist/index.d.mts

danielroe: 🤯

danielroe: I should have known Anthony wouldn't publish an untyped api

snewell92 (approved):
Yoooo. lfgo; reading this diff gave me a lot of ideas too.

danielroe:
@snewell92 can't wait to see them!

PR #665, opened 2026-02-01
perf: use osv batch api for dep analysis

danielroe:
this updates osv analysis to use their batch api which _significantly_ speeds up resolution for packages with large dep trees

### DIFF
--- a/server/utils/atproto/oauth.ts
+++ b/server/utils/atproto/oauth.ts
@@ -31,6 +31,7 @@ export function getOauthClientMetadata() {
     application_type: 'web',
     token_endpoint_auth_method: 'none',
     dpop_bound_access_tokens: true,
+    response_types: ['code'],
   }) as OAuthClientMetadataInput
 }
 
--- a/shared/schemas/oauth.ts
+++ b/shared/schemas/oauth.ts
@@ -10,4 +10,5 @@ export const OAuthMetadataSchema = object({
   application_type: string(),
   token_endpoint_auth_method: string(),
   dpop_bound_access_tokens: boolean(),
+  response_types: array(string()),
 })


server/utils/dependency-analysis.ts
```diff
@@ -135,22 +189,27 @@ export const analyzeDependencyTree = defineCachedFunction(
...
+      // Step 2: Fetch full vulnerability details only for packages with vulns
+      // This is typically a small fraction of total packages
+      const vulnerablePackageInfos = vulnerableIndices.map(i => packages[i]!)
 
-    for (let i = 0; i < packages.length; i += batchSize) {
-      const batch = packages.slice(i, i + batchSize)
-      const results = await Promise.all(
-        batch.map(pkg => queryOsv(pkg.name, pkg.version, pkg.depth!, pkg.path || [])),
+      const detailResults = await Promise.all(
+        vulnerablePackageInfos.map(pkg => queryOsvDetails(pkg)),
```
serhalp: In egregious cases, this might fire off too many requests in parallel. Perhaps we could use a util here that limits concurrency to something like 25?

server/utils/dependency-analysis.ts
```diff
@@ -65,11 +105,18 @@ async function queryOsv(
       })
     }
 
-    return { status: 'ok', data: { name, version, depth, path, vulnerabilities, counts } }
+    return {
```
serhalp, line 121: It looks like [the batch endpoint is also paginated](https://google.github.io/osv.dev/post-v1-querybatch/#pagination).

It seems unlikely we'd ever have incomplete results in the first page:

> - An individual query within the queryset returns more than 1,000 vulnerabilities
> - The entire queryset returns more than 3,000 vulnerabilities total

but maybe just check for a non-nil `next_page_token` and log a warning/error for future visibility?

danielroe:
@serhalp are you investigating?

garthdw (approved):
Feels faster on projen and @babel/core packages

Issue #661, opened 2026-02-01
package page 504 or stuck flickering skeleton loader


serhalp:
In this recording you can see I tried to load the same page (https://npmx.dev/skills) three times, with one success, one 504, and one failure with a flickering skeleton loader:

https://github.com/user-attachments/assets/016e0c8f-92f2-437e-a75b-7d4e01565928

In case this is useful, here's the 504 one:
```
504: GATEWAY_TIMEOUT Code: FUNCTION_INVOCATION_TIMEOUT ID: yul1::mmk6b-1769961619497-753e5f7c2715  
```

Function logs should be here but I don't have access: https://vercel.com/poetry/npmx.dev/7f4VTkR9JZg4wh24tLsSXBVVRwra/logs??requestIds=yul1%3A%3Ammk6b-1769961619497-753e5f7c2715

@vinnymac and @jellydeck have also seen this

I don't have reliable repro steps, unfortunately. It happens intermittently when loading an uncached package.

serhalp:
I found what appears to be a reproducible case: https://npmx.dev/gatsby (but if it ever does successfully load, it will be cached and load successfully for everyone for some number of minutes...).

DDeenis:
Looking into it right now. What I found so far:
- `status` is flip-flopping between `pending` and `success` after the package has been fetched
- Thousands of requests are being sent to fetch the package (all hit cache)

My repro steps:
- Re-run the dev server (to clear cache)
- Open http://127.0.0.1:3000/skills

DDeenis:
Replacing `useLazyAsyncData` with `useAsyncData` removed flickering. Package info is still not being rendered sometimes, but that might be a separate issue.
@serhalp I'm not familiar with Nuxt, what's your thoughts?

shamilkotta:
@serhalp @DDeenis i've opened a PR seems like the status flip was because of DownloadStatus component

BoxenOfDonuts:
It appears the issues with the skills package specifically is that  `/skills/_payload.json`  is getting sent to the skills route handler, which then is throwing an error for invalid package name. 

```json
{
  "error": true,
  "url": "http://127.0.0.1:3000/skills/_payload.json?dev",
  "statusCode": 404,
  "statusMessage": "Server Error",
  "message": "Invalid package name format",
  "stack": [
    "Invalid package name format",
[...]

PR #675, opened 2026-02-01
fix: prevent infinite recursive loading of package stats

shamilkotta:
fixes: #661

### DIFF
--- a/app/components/Package/WeeklyDownloadStats.vue
+++ b/app/components/Package/WeeklyDownloadStats.vue
@@ -251,7 +251,7 @@ const config = computed(() => {
       :weeklyDownloads="weeklyDownloads"
       :inModal="true"
       :packageName="props.packageName"
-      :createdIso="props.createdIso"
+      :createdIso="createdIso"
     />
   </PackageChartModal>
 </template>


43081j:
just for learning, if anything, can you explain _why_ this fixes it?

code looks good but it'd be good to understand why moving to a prop solves it

DDeenis:
It looks like the important point here is that the `usePackage` call was removed from a nested component. Would like to hear why it worked too though.

shamilkotta:
for some reason the usePackage call is not caching which leads to WeeklyDownloadStats component re triggering the usePackage call again, this causes status flip on the main screen  and cycle continues

we might want to revisit the caching issue

DDeenis:
Ah, I think I get it now. My theory is this:
1. `usePackage` is called by the route
2. It renders `PackageWeeklyDownloadStats`
3. `PackageWeeklyDownloadStats` also calls `usePackage`
4. It forces the main component (route) to display a skeleton, because the `status` is `pending` again
5. Cached response is being returned
6. `PackageWeeklyDownloadStats` is being rendered again

app/components/Package/WeeklyDownloadStats.vue
```diff
@@ -5,6 +5,7 @@ import { OKLCH_NEUTRAL_FALLBACK, lightenOklch } from '~/utils/colors'
 
 const props = defineProps<{
   packageName: string
+  createdIso: string | null
```
43081j, line 8:
```suggestion
  createdTime: string | null
```

since that's what it is underneath

app/components/Package/WeeklyDownloadStats.vue
```diff
@@ -253,7 +251,7 @@ const config = computed(() => {
       :weeklyDownloads="weeklyDownloads"
       :inModal="true"
       :packageName="props.packageName"
-      :createdIso="createdIso"
+      :createdIso="props.createdIso"
```
danielroe:
```suggestion
      :createdIso="createdIso"
```

danielroe (approved):

PR #676, opened 2026-02-01
test: move nuxt unit tests to `test/nuxt`

43081j:
We need this to resolve types correctly amongst other things.


test/nuxt/composables/useStructuredFilters.spec.ts
```diff
@@ -2,7 +2,7 @@ import { describe, expect, it } from 'vitest'
 import {
   hasSearchOperators,
   parseSearchOperators,
-} from '../../app/composables/useStructuredFilters'
+} from '../../../app/composables/useStructuredFilters'
```
danielroe: you can totally drop the import or simply import from `~/composables/useStructuredFilters` if you want

test/nuxt/server/utils/readme.spec.ts
danielroe, line 28: tests for server utilities should not be in `test/nuxt` - test nuxt emulates the browser and will not work with server utils

43081j:
sad times, @danielroe any idea what we should do here:

> Caused by: SyntaxError: The requested module '/node_modules/.pnpm/sanitize-html@2.17.0/node_modules/sanitize-html/index.js?v=4f273e36' does not provide an export named 'default'

its a commonjs dependency, with an export assignment

this must work at runtime since we bundle everything, so why not in tests? 🤔

danielroe:
it's not bundled by nuxt - the server files are bundled by nitro and run in a different context

test/nuxt is a special dir and shouldn't be used for testing anything in server/

43081j:
we can merge this one and ill do the type check setup in a follow up pr

PR #680, opened 2026-02-01
feat: add border to switch styles

lukewarlow:
Dark Before:

image

Dark After:

image

Light Before:

image

Light After:

image

This is added to improve the contrast. It's still not great but a follow-up can change the actual border colour when in prefers-contrast: more, with this change that's at least *some* contrast.

### DIFF
--- a/app/components/Settings/Toggle.client.vue
+++ b/app/components/Settings/Toggle.client.vue
@@ -22,7 +22,7 @@ const checked = defineModel<boolean>({
     </span>
     <span
       class="relative inline-flex h-6 w-11 shrink-0 items-center rounded-full border-2 transition-colors duration-200 ease-in-out motion-reduce:transition-none cursor-pointer"
-      :class="checked ? 'bg-accent border-transparent' : 'bg-bg border border-border'"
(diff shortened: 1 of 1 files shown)


garthdw:
image

Should active take into account accent color for border?

trueberryless:
I personally do not 100% agree with the design of the toggles with those outlines. Another idea to increase the contrast could be to change the background color of a inactive toggle to something lighter (in dark mode) and respectively darker (in light mode). IMO we should keep the little shadow, it creates a nice little effect. But that's just my humble opinion. Great catch! 🙌

lukewarlow:
> Should active take into account accent color for border?

Good catch this should probably changed so that its border-transparent when checked.

lukewarlow:
Now the border is transparent when checked but shows when it's not. I suspect this was the intent of the original code.

image

lukewarlow:
> IMO we should keep the little shadow, it creates a nice little effect. But that's just my humble opinion. Great catch!

I don't feel too strongly about this, I think it can look weird to have the border and shadow, but like above I've changed this so it's shadowed only when checked. I think that gives the best overall look.

image

PR #683, opened 2026-02-01
fix: allow wrapping in footer container

wojtekmaj:
Before:

localhost_3000_settings

After:

localhost_3000_settings (1)


serhalp (commented):
this looks great, thank you!

PR #685, opened 2026-02-01
fix: prevent modal closing on inner clicks

### LINKED ISSUE #667: Downloads dialog can be closed by clicking within some areas of the dialog itself
Dialog can be closed by clicking within the padding area but also below the title. I assume this shouldn't be the case.

https://github.com/user-attachments/assets/d9771d19-896c-443a-8586-82c410f86812

cnaples79:
## Summary
- only treat clicks outside the dialog bounds as light dismiss
- avoid closing modals when clicking padding or title areas

## Rationale
- addresses npmx-dev/npmx.dev#667: clicking inside modal padding/title closed the dialog

## Changes
- app/app.vue: check click coordinates against dialog bounding box before closing

## Test Plan
- All CI passing on fork

Fixes #667

### DIFF
--- a/app/app.vue
+++ b/app/app.vue
@@ -79,14 +79,6 @@ function handleGlobalKeyup() {
   showKbdHints.value = false
 }
 
-// Feature check for native light dismiss support via closedby="any"
-// https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#closedby
-const supportsClosedBy =
-  import.meta.client &&
-  typeof HTMLDialogElement !== 'undefined' &&
-  typeof HTMLDialogElement.prototype === 'object' &&
-  'closedBy' in HTMLDialogElement.prototype
-
 // Light dismiss fallback for browsers that don't support closedby="any" (Safari + old Chrome/Firefox)
 // https://codepen.io/paramagicdev/pen/gbYompq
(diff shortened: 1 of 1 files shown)


knowler:
So, the particular code that is causing this is a hack to add light dismiss to Safari (see #522 for context). I suggest that we feature check [`closedby=any`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#closedby) which does the light dismiss for us for free and then keep the hack for cases [where `closedby=any` isn’t supported](https://caniuse.com/mdn-html_elements_dialog_closedby) (Safari + old Chrome/Firefox).

You can feature check `closedby=any` using:

```js
typeof HTMLDialogElement !== "undefined" &&
typeof HTMLDialogElement.prototype === "object" &&
"closedBy" in HTMLDialogElement.prototype
```

Feel free to simply that to the final condition if this code will only run in a browser.

I said this in #522, but **I do think it could be worth removing the hack entirely if it’s causing issues or not working correctly**. Users can already dismiss the modal using the Escape key or with the dismiss button; light dismiss is a progressive enhancement. This is likely a better route than keeping it as light dismiss is not easy to implement with `dialog` otherwise.

danielroe:
yes, I think that's a better fix - thank you 🙏

PR #686, opened 2026-02-01
feat: dependencies page

### LINKED ISSUE #29: add package dependencies tab with grouped view
show deps grouped by type (prod, dev, peer, optional) with clickable links to each package

stordahl:
Closes #29 by making the existing "Dependencies" header in the sidebar clickable, and adding a /deps route structured similar to /code.


mikouaji:
Could you move it to `/package-deps/{path}` so it will go nicely with #674 ?

Molnes:
Looking at the vercel builds for this, and between the two last builds it fails to properly get some of the information?
Deployment before f8fe55f: 
https://npmx-5i5hdxeyz-poetry.vercel.app/is-even
image


And after f8fe55f:
https://npmx-3b1seggis-poetry.vercel.app/is-even
image
The info does show up if i explicitly select the version https://npmx-3b1seggis-poetry.vercel.app/is-even/v/1.0.0
I don't see why that commit would stop it from showing up tho. Or might just be me misunderstanding ¯\\_(ツ)_/¯

app/pages/package-deps/[...path].vue
```diff
@@ -0,0 +1,248 @@
...
+  }
+  return 'Browse package dependencies'
+})
+
+const canonicalUrl = computed(() => {
+  if (packageName.value && version.value) {
+    return `https://npmx.dev/package-deps/${packageName.value}/v/${version.value}`
+  }
+  return 'https://npmx.dev/package-deps'
+})
```
Comment (OrbisK) [inline app/pages/package-deps/[...path].vue:99]: I think we could resolve the route here instead of hardcoding it.

PR #689, opened 2026-02-01
feat: set color scheme meta data value based on user settings

### LINKED ISSUE #677: Need to declare color-scheme.
Because npmx.dev doesn't currently use the color-scheme property on html (or the color-scheme meta tag) to advertise its support for color schemes, it fails to play nicely with forced dark mode.

To see examples of the issues load the website in chromium, open dev tools go to the rendering tab and toggle "Enable automatic dark mode" to on.

If your system is set to dark mode or your setting is set to dark mode explicitly, you'll notice that the logo is wrong, and switches lose contrast.

Similarly if your setting is explicitly set to light mode, you'll see the browser forces it to dark mode and again breaks various aspects of the UI.

In system mode the color-scheme property should be set to "light dark", in light mode it should be "light only", and in dark mode it should be "dark only".

thasmo:
Fixes #677.

This sets the `color-scheme` to an appropriate value as described in #677.

App >  | system | light | dark |
-- | -- | -- | --
dark | dark light | only light | only dark
light | light dark | only light | only dark
^ OS

### DIFF
--- a/app/app.vue
+++ b/app/app.vue
@@ -21,6 +21,16 @@ const localeMap = locales.value.reduce(
   {} as Record<string, Directions>,
 )
 
+const darkMode = usePreferredDark()
+const colorMode = useColorMode()
+const colorScheme = computed(() => {
+  return {
+    system: darkMode ? 'dark light' : 'light dark',
+    light: 'only light',
+    dark: 'only dark',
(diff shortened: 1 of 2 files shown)


lukewarlow:
Lgtm!

PR #690, opened 2026-02-01
feat(ui): improve compare types in FacetRow for CLI package

### LINKED ISSUE #631: Adjust "types" presence for binary only packages in `/compare` route
As suggested by @serhalp [here](https://github.com/npmx-dev/npmx.dev/pull/597#issuecomment-3830041975)

DoD:
* Should say N/A in types row for binary only package

gusa4grr:
- fix tooltips in FacetRow
- add 'muted' status option with text-fg-subtle styling
- show types as "N/A" with muted style for binary-only packages
- move binary detection utils to shared/utils for reuse
- use Packument type for registry API responses

Closes #631

### DIFF
--- a/lunaria/files/en-US.json
+++ b/lunaria/files/en-US.json
@@ -852,6 +852,7 @@
       "deselect_all": "Deselect all facets",
       "select_category": "Select all {category} facets",
       "deselect_category": "Deselect all {category} facets",
+      "binary_only_tooltip": "This package exposes binaries and no exports",
       "categories": {
         "performance": "Performance",
         "health": "Health",


gusa4grr:
image

app/composables/usePackageComparison.ts
```diff
@@ -259,6 +268,13 @@ function computeFacetValue(facet: ComparisonFacet, data: PackageComparisonData):
       }
 
     case 'types':
+      if (data.isBinaryOnly) {
```
serhalp, line 276: Could we add a tooltip on the 'N/A' with a line explaining why it's N/A? something like "This package exposes binaries and no exports"

serhalp (approved):
awesome, thank you for this!

Issue #430, opened 2026-01-30
Use locales to format dates in charts


graphieros:
- pass unix timestamps to xAxis values instead of strings 
- use `datetimeFormatter` in vue-data-ui chart configs

⚙️ An update of vue-data-ui is necessary:
- support all i18n locales
- code split only used  translations for days, months etc

wojtekmaj:
In footer relative build date too!

graphieros:
> In footer relative build date too!

I'm sorry, could you elaborate ? I did not understand :)

wojtekmaj:
In the footer we now have a relative build date that seems to suffer from the same issue, it doesn't seem to react to locale changes:

![Screenshot_20260131_170923_Samsung Internet.jpg](https://github.com/user-attachments/assets/207d1a52-d0a5-4897-be6c-198a3d6bb710)

graphieros:
Ah ok I see, but this is not related to the present issue, which targets specifically how the chart configuration is to be plugged with locales :)

serhalp:
Is this still missing?

graphieros:
I think we are ok now, closing

PR #691, opened 2026-02-01
fix: plug locale to large chart config

graphieros:
- bump vue-data-ui to 3.14.1
- plug locale in config
- simplified UI: time labels only visible on hover

image

I had to reluctantly restore the re-rendering of the chart onMounted as it appears it was still flaky. Something with the dialog does not go well with the chart.


danielroe (approved):

graphieros:
I believe this closes #430

PR #692, opened 2026-02-01
fix: correct route query parsing for multiple params

### LINKED ISSUE #688: Server error with multiple search params
In many places, there's an override type for route.query, and it works as if it were a string. But if you pass multiple queries, it will be an array. This results in a 500 error. If the crawler notices such a link, it will save it and complain about it for a long time, so it's always best to avoid server errors.

I'm fixing it to receive the first element if it's an array.

Examples:
- https://npmx.dev/@vitejs?q=1&q=2
- https://npmx.dev/search?q=vue&q=vite

alexdln:
In many places, there was a force override of types on route.query, and then it worked as if it was a string. But if you pass multiple queries, it would be an array. The result was a 500 error (no trim or no toLowerCase).

I added a common method and use it everywhere. The method returns the value, either the first element of the array, or an empty string

Closes #688


43081j (approved):
good catch!

the types would've caught this and we casted it away 😅

PR #695, opened 2026-02-01
fix: provide resolvedVersion to pkg composable

danielroe:


Gugustinette:
Indeed, it was as simple as using the resolved version by default 😆

serhalp:
This is a great candidate for some test coverage, if anyone's up for it!

PR #696, opened 2026-02-01
chore: guard undefined values in `uno-preset-rtl`

43081j:
Part of the test cleanup.

When we eventually type-check this, it'll fail right now as these are
all `T | undefined`. Adding `!` since in all cases, we know it is not
null.


wojtekmaj:
Personally I'm not a fan of non-null assertions. They give you a false sense of security should things ever go south and produce harder-to-debug errors than they need to be. Why not just add error handling and throw if something is unexpectedly missing?

PR #699, opened 2026-02-02
fix: fork `<NuxtTime>` for i18n support

danielroe:
this addresses some hydration mismatches we were encountering, and prevents layout shift/flicker

### DIFF
--- a/app/assets/main.css
+++ b/app/assets/main.css
@@ -27,6 +27,14 @@
   --accent: var(--accent-color, oklch(1 0 0));
   --accent-muted: var(--accent-color, oklch(0.922 0 0));
 
+  /* accent colors */
+  --swatch-coral: oklch(0.704 0.177 14.75);
+  --swatch-amber: oklch(0.828 0.165 84.429);
+  --swatch-emerald: oklch(0.792 0.153 166.95);
+  --swatch-sky: oklch(0.787 0.128 230.318);
+  --swatch-violet: oklch(0.78 0.148 286.067);
+  --swatch-magenta: oklch(0.78 0.15 330);
+
   /* syntax highlighting colors */
   --syntax-fn: oklch(0.727 0.137 299.149);
   --syntax-str: oklch(0.829 0.088 252.458);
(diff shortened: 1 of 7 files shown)


app/components/DateTime.vue
```diff
@@ -1,86 +1,200 @@
...
-  }>(),
-  {
-    title: undefined,
-    dateStyle: undefined,
-    year: undefined,
-    month: undefined,
-    day: undefined,
-  },
-)
+interface NuxtTimeProps {
```
wojtekmaj: Could these be simply imported from `nuxt/dist/app/components/nuxt-time.d.vue.ts`?

app/components/DateTime.vue
```diff
@@ -1,86 +1,177 @@
...
+  const { relativeStyle, ...rest } = props
+  if (relativeDates.value) {
+    return new Intl.RelativeTimeFormat(_locale ?? locale.value, {
+      ...defaults,
+      ...rest,
+      style: relativeStyle,
+    })
+  }
+  return new Intl.DateTimeFormat(_locale ?? locale.value, { ...defaults, ...rest })
 })
```
wojtekmaj, line 63: Maybe we could avoid type assertions by doing:


```suggestion
const relativeFormatter = computed(() => {
  const { relativeStyle, ...rest } = props
  
  return new Intl.RelativeTimeFormat(_locale ?? locale.value, {
    ...defaults,
    ...rest,
    style: relativeStyle,
  })
})

const formatter = computed(() => {
  const { relativeStyle, ...rest } = props
  
  return new Intl.DateTimeFormat(_locale ?? locale.value, { ...defaults, ...rest })
})
```

?

app/components/DateTime.vue
```diff
@@ -1,86 +1,177 @@
...
+        dataset[`data-${propInKebabCase}`] = props?.[prop as keyof typeof props]
+      }
+    }
+  }
+  onPrehydrate(el => {
+    const now = (window._nuxtTimeNow ||= Date.now())
+    // eslint-disable-next-line eslint-plugin-unicorn/consistent-function-scoping
+    const toCamelCase = (name: string, index: number) => {
+      if (index > 0) {
+        return name[0]!.toUpperCase() + name.slice(1)
```
wojtekmaj, line 111:
```suggestion
        return name.charAt(0).toUpperCase() + name.slice(1)
```

will avoid the need of non-null assertion

app/components/DateTime.vue
```diff
@@ -1,86 +1,177 @@
...
+      if (name.startsWith('data-')) {
+        let optionName = name
+          .slice(5)
+          .split('-')
+          .map(toCamelCase)
+          .join('') as keyof (Intl.DateTimeFormatOptions & Intl.RelativeTimeFormatOptions)
+
+        if ((optionName as string) === 'relativeStyle') {
+          optionName = 'style'
+        }
```
wojtekmaj, line 131: Maybe just a tiny bit more accurate - still assertion needed but at least we're "telling the truth" here:

```suggestion
        let optionName = name.slice(5).split('-').map(toCamelCase).join('') as
          | keyof (Intl.DateTimeFormatOptions & Intl.RelativeTimeFormatOptions)
          | 'relativeStyle'

        if (optionName === 'relativeStyle') {
          optionName = 'style'
        }
```

serhalp:
@danielroe I'll close it for now, feel free to reopen if I'm wrong :)

PR #700, opened 2026-02-02
fix(a11y): improve chart tooltip contrast in light mode

### LINKED ISSUE #633: Download chart tooltip in light mode
We cannot see the number of download 

see screenshot 

Image

JarvisInvestInsight:
## Summary

Fixes the download chart tooltip being invisible in light mode as reported in #633.

## Problem

The tooltip background used \g-[var(--bg)]/10\ (10% opacity) which works well in dark mode where the contrast comes from the backdrop blur, but in light mode the low opacity resulted in poor contrast against the light chart background.

## Solution

Dynamically set the background opacity based on the color mode:
- **Dark mode:** 10% opacity (maintains the current frosted glass effect)
- **Light mode:** 90% opacity (ensures readable text contrast)

The change is inside the computed \config\ which already depends on \isDarkMode\, so the tooltip will update correctly when the theme changes.

Fixes #633

---
*Found this helpful? [☕ Buy me a coffee](https://ko-fi.com/jarvisdev)*

### DIFF
--- a/app/components/Package/DownloadAnalytics.vue
+++ b/app/components/Package/DownloadAnalytics.vue
@@ -595,7 +595,11 @@ const config = computed(() => {
         customFormat: ({ datapoint }: { datapoint: Record<string, any> }) => {
           if (!datapoint) return ''
           const displayValue = formatter({ value: datapoint[0]?.value ?? 0 })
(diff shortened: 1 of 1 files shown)


JarvisInvestInsight:
Closing: this PR was incorrectly branched from #650 instead of main, and the issue is also addressed by #626.

PR #711, opened 2026-02-02
feat: avatar for user and org

iiio2:
org
user


danielroe:
thank you, but I think we'll use https://github.com/npmx-dev/npmx.dev/pull/556 to do gravatars instead

PR #712, opened 2026-02-02
feat: package likes

### LINKED ISSUE #79: feat: package likes
users can like packages; consider either using existing `community.lexicon.interaction.like` or creating something `dev.npmx` specific

- [ ] UI for liking in repo pages (and search cards?)

fatfingers23:
Our first social feature 🥳 This closes #79 

Adds
- Likes and unlikes
- Generic cache that uses `useStorage` locally and redis in prod with set expire 
- Caches on the Constellation requests to return total counts faster since Constellation cannot show counts until it indexs it and that takes a few milliseconds to seconds depending on traffic
- Some more helpers for working with atproto for checking for scopes, redirecting to login
- Since we have added a scope the client needs to re-authenticate with the new one if a user was already logged in. Right now if it sees the logged in client does not have the new scope it directs to the PDS to reauth.
- If the user likes without being logged in shows them the auth modal

Could be Improved
- Not sure on the icon. Went with what I could find. I think a filled one may work better and an animation would be nice. 
- I think the caching layer I added for Likes could be abstracted out for better use with constellation. Constellation is the source of truth, but we need caching at times for faster feed back. Will have to think on how that will work
- The redirect for scopes reauth could probably be improved to let the user know?
- Auth always redirects back to main page. I think we could probably add an item to add some state where on redirect it loads the previous page

Concerns
- Race conditions on likes and many users. Interested to see how it holds up with a lot of users at once. And just genreal race condtions of state
- I'm not familiar with how Nuxt handles dependencies and not sure if the way I did the Redis connection will reuse client connections or open new ones? 
- I need to test more on error handling on everything and showing errors from the API. But this PR was getting really long in the tooth and I think some of that can be divided up for others who may not know atproto stuff but are much better with Nuxt than me

I probably won't be able to address feedback till tomorrow around 22:00 UTC

https://github.com/user-attachments/assets/2fab594b-b8bf-4b35-a31b-8091aabab13c

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -152,9 +152,9 @@ export default defineNuxtConfig({
         driver: 'fsLite',
         base: './.cache/fetch',
       },
-      'generic-cache': {
+      'atproto': {
         driver: 'fsLite',
-        base: './.cache/generic',
+        base: './.cache/atproto',
       },
     },
     typescript: {
--- a/modules/cache.ts
+++ b/modules/cache.ts
@@ -1,3 +1,4 @@
+import process from 'node:process'
 import { defineNuxtModule } from 'nuxt/kit'
 import { provider } from 'std-env'
 
@@ -26,6 +27,11 @@ export default defineNuxtModule({
       nitroConfig.storage[FETCH_CACHE_STORAGE_BASE] = {
         ...nitroConfig.storage[FETCH_CACHE_STORAGE_BASE],
         driver: 'vercel-runtime-cache',
+      }
+
+      const env = process.env.VERCEL_ENV
+      nitroConfig.storage.atproto = {
+        driver: env === 'production' ? 'vercel-kv' : 'vercel-runtime-cache',
       }
     })
   },


app/pages/package/[...package].vue
```diff
@@ -334,6 +335,29 @@ const canonicalUrl = computed(() => {
...
+})
+
+const { mutate: likePackage } = useLikePackage(packageName.value)
+const { mutate: unlikePackage } = useUnlikePackage(packageName.value)
+
+const likeAction = async () => {
+  if (user.value?.handle == null) {
+    authModal.open()
+  } else {
+    const result = likesData.value?.userHasLiked ? await unlikePackage() : await likePackage()
```
Comment (Adebesin-Cell) [inline app/pages/package/[...package].vue]: I think this approach could be improved a bit. Right now, the API request is sent immediately on every click.

It might be better to optimistically update the UI and debounce the request, so users can toggle the button freely and only the final state is sent to the API after a short delay. This would also allow us to provide immediate feedback that the action was registered, and handle request failures more gracefully (e.g., by rolling back the UI state or showing an error).

This would avoid unnecessary API calls and result in a smoother UX overall.

server/api/auth/social/like.delete.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  const likesUtil = new PackageLikesUtils()
+
+  const getTheUsersLikedRecord = await likesUtil.getTheUsersLikedRecord(
+    body.packageName,
+    loggedInUsersDid,
+  )
+  if (getTheUsersLikedRecord) {
+    //Checks if the user has a scope to like packages
+    await checkOAuthScope(oAuthSession, LIKES_SCOPE)
```
Adebesin-Cell: I have a couple of suggestions here:

1. Since `checkOAuthScope` is meant as a guard that throws if the user lacks the required scope, it could be moved up to the start of the handler. This makes the intent clearer and prevents unnecessary work if the user isn’t authorized. A short comment noting it’s a guard would improve readability.

2. The delete + unlike operations appear independent, so we could run them in parallel using `Promise.all` to reduce latency. Just need to ensure there are no ordering or race conditions between them.

server/api/social/like.post.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const likesUtil = new PackageLikesUtils()
+
+  const hasLiked = await likesUtil.hasTheUserLikedThePackage(body.packageName, loggedInUsersDid)
+  if (hasLiked) {
+    throw createError({
+      status: 400,
+      message: 'User has already liked the package',
+    })
+  }
```
wojtekmaj: Suggestion: silently succeed here instead of throwing an error. 

Scenario: user likes the package twice with quirky internet connection and/or slower than usual response times.

Potential result on the UI: user see the like added then seeing the error message that liking has failed.

Been there, done that, with an app with more than 500,000 users :D

server/api/auth/social/like.post.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const result = await client.create(dev.npmx.feed.like, like)
+  if (!result) {
+    throw createError({
+      status: 500,
+      message: 'Failed to create a like',
+    })
+  }
+
+  return await likesUtil.likeAPackageAndRetunLikes(body.packageName, loggedInUsersDid, result.uri)
```
wojtekmaj: Typo 

```suggestion
  return await likesUtil.likeAPackageAndReturnLikes(body.packageName, loggedInUsersDid, result.uri)
```

server/api/auth/social/like.post.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  const body = await readBody<{ packageName: string }>(event)
+
+  if (!body.packageName) {
+    throw createError({
+      status: 400,
+      message: 'packageName is required',
+    })
+  }
```
wojtekmaj: Could leverage valibot here, this way you wouldn't have to resort to custom null checks to ensure that body is actually of type you expect it to be.

server/utils/atproto/utils/likes.ts
```diff
@@ -0,0 +1,246 @@
...
+          userHasLiked,
+          CACHE_MAX_AGE,
+        )
+      }
+    }
+
+    return {
+      totalLikes: totalLikes,
+      userHasLiked,
+    } as PackageLikes
```
wojtekmaj: Why is this type assertion needed here?

server/utils/atproto/utils/likes.ts
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  /**
+   * Gets the definite answer if the user has liked a npm package. Either from the cache or the network
+   * @param packageName
+   * @param usersDid
+   * @returns
```
wojtekmaj, line 122: 👀

server/api/social/like.delete.ts
alexdln, line 1: How about making xrpc routes (sth like `/xrpc/npmx.feed.like.create`)?

fatfingers23: Right now we're just using the the server side oauth client so if we made this endpoint an XRPC it wouldn't be quite correct since it uses the cookie for authentication. But there is a plan for some XRPC endpoints for other things just need to make a middleware for it to authenticate the service auth jwt

nuxt.config.ts
```diff
@@ -162,6 +162,10 @@ export default defineNuxtConfig({
         driver: 'fsLite',
         base: './.cache/atproto-oauth/session',
       },
+      'generic-cache': {
```
danielroe: we probably also need to update modules/cache.ts to configure production redis

server/utils/atproto/utils/likes.ts
```diff
@@ -0,0 +1,246 @@
+import { getCacheAdatper } from '../../cache'
+import { $nsid as likeNsid } from '#shared/types/lexicons/dev/npmx/feed/like.defs'
+import type { Backlink } from '~~/shared/utils/constellation'
```
danielroe: here and some other places too:

```suggestion
import type { Backlink } from '#shared/utils/constellation'
```

app/pages/package/[...package].vue
```diff
@@ -334,6 +335,29 @@ const canonicalUrl = computed(() => {
   return requestedVersion.value ? `${base}/v/${requestedVersion.value}` : base
 })
 
+//atproto
+const { user } = useAtproto()
+
+const authModal = useModal('auth-modal')
+
+const { data: likesData } = useFetch(() => `/api/social/likes/${packageName.value}`, {
+  default: () => ({ totalLikes: 0, userHasLiked: false }),
```
Comment (danielroe) [inline app/pages/package/[...package].vue:369]: this should probably be client-side only:


```suggestion
  default: () => ({ totalLikes: 0, userHasLiked: false }),
  server: false,
```

server/api/auth/social/like.delete.ts
fatfingers23: Good shout out on the checkOAuthScope function! I changed the name to `throwOnMissingOAuthScope`. For the delete I feel like the delete record to atproto should succeed first so left that as is for now. But I am going to look more at your first comment and there may be something else with the optimistic updates

server/utils/atproto/utils/likes.ts
fatfingers23: Typescript is still a whole new world to me really and I was thinking I needed the type to be explicit 😅. ty

nuxt.config.ts
fatfingers23: May have to leave that one to you 😅. Do you have something to read up on that? The local generic-cache defined there is used just locally for dev and then uses upstash redis like you did for the oauth session lock when in production if that makes a difference.

app/composables/useAtproto.ts
```diff
@@ -1,4 +1,30 @@
+import { ERROR_NEED_REAUTH } from '#imports'
+import type { FetchError } from 'ofetch'
 import type { UserSession } from '#shared/schemas/userSession'
+import type { PackageLikes } from '#server/utils/atproto/utils/likes'
+
+export async function authRedirect(identifier: string, create: boolean = false) {
+  let query = { handle: identifier } as {}
```
43081j: nit: this should be a type annotation rather than a cast, and shouldn't be `{}`

```suggestion
  let query: LocationQueryRaw = { handle: identifier }
```

or if we can't pull the exact type from vue-router, use `Record<string, string>` or something

app/composables/useAtproto.ts
```diff
@@ -1,4 +1,30 @@
...
+    {
+      path: '/api/auth/atproto',
+      query,
+    },
+    { external: true },
+  )
+}
+
+export async function handleAuthError(e: unknown, userHandle?: string | null): Promise<never> {
+  const fetchError = e as FetchError
```
43081j: given we cast the error anyway at all call sites, can this function just accept a `FetchError`

otherwise we're doing an unsafe cast here, if anyone in future calls it from somewhere that doesn't only do a `$fetch` in the `try` block

app/composables/useAtproto.ts
```diff
@@ -13,3 +39,61 @@ export function useAtproto() {
 
   return { user, pending, logout }
 }
+
+export function useLikePackage(packageName: string) {
```
43081j: this file is becoming a bit of a "catch all" as the atproto features quickly expand. we should split them up i think, and have a `app/composables/atproto/*`, one file per composable.

that'll also help with organising the tests once they exist

its currently named after _one_ of the composables, and the rest are just thrown into the mix

app/pages/package/[...package].vue
Comment (43081j) [reply app/pages/package/[...package].vue]: i agree, optimistic updates would fit nicely here and shouldn't be too difficult 👍

server/api/social/like.delete.ts
43081j, line 1: i think all the `social/` endpoints need to live outside of the `auth/` directory

they're not auth related, they just rely on a session.

- `auth/*` anything to do with actually authenticating (login, logout, session, etc)
- `social/*` anything to do with social, whether it requires a session or not

server/api/social/like.delete.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  const likesUtil = new PackageLikesUtils()
+
+  const getTheUsersLikedRecord = await likesUtil.getTheUsersLikedRecord(
+    body.packageName,
+    loggedInUsersDid,
+  )
+
+  if (getTheUsersLikedRecord) {
```
43081j, line 26: what if it isn't truthy? somehow

feels like something would have to be very wrong for that to happen, but we should log an error if so

server/utils/cache.ts
```diff
@@ -0,0 +1,107 @@
...
+  if (!entry.ttl) return false
+  const now = Date.now()
+  const expiresAt = entry.cachedAt + entry.ttl * 1000
+  return now > expiresAt
+}
+
+/**
+ * Local implmentation of a cache to be used during development
+ */
+export class StorageCacheAdapter implements CacheAdapter {
```
43081j: we should probably organise these into files:

- `server/utils/cache/adapter.ts` - or `shared.ts`, holds the interface and any shared functions
- `server/utils/cache/local.ts` - this `StorageCacheAdapter`
- `server/utils/cache/redis.ts`
- `server/utils/cache/get.ts` - not sure about this one, but just holds the main `getCacheAdapter`. can't live in the shared file since that'd cause circular imports

generally we should have one class per file imo

43081j:
looking good!

i left a bunch of comments, mostly about code organisation though tbh

app/composables/useAtproto.ts
```diff
@@ -13,3 +38,61 @@ export function useAtproto() {
...
+  const { user } = useAtproto()
+  const data = ref<PackageLikes | null>(null)
+  const error = ref<Error | null>(null)
+  const pending = ref(false)
```
OrbisK:
```suggestion
  const data = shallowRef<PackageLikes | null>(null)
  const error = shallowRef<Error | null>(null)
  const pending = shallowRef(false)
```

app/pages/package/[...package].vue
```diff
@@ -334,6 +335,30 @@ const canonicalUrl = computed(() => {
...
+  server: false,
+})
+
+const { mutate: likePackage } = useLikePackage(packageName.value)
```
Comment (OrbisK) [inline app/pages/package/[...package].vue]: passing `packageName.value` means, that when `packageName` updates the `likePackage` will like the initial value.

app/composables/useAtproto.ts
OrbisK: By the way, I dont think we need these as composables. We almost never want the data from the ref itself, but rather the data returned by mutate. Just use `$fetch` where we need it or capsule just the `$fetch` in a util (IMO)

server/api/social/like.delete.ts
fatfingers23: Same. I think some of this like logic is kind of "let's see what all comes of it stage". Do we have something for logging or is it just `console.error`?

43081j: I think just a regular console.error for now

They should show up in the vercel logs

fatfingers23:
Thank you everyone for all the feedback! I think I have addressed everything brought up. About to log off for a few hours and going come back to do some final testing and I have 2 todos I want to hit

- There's a like type I can share with the backend
- Look into the user session being loaded when not needed

danielroe:
it seems that likes aren't persisted, right? as in, when I navigate back to a package, it shows up as 'unliked'

fatfingers23:
> it seems that likes aren't persisted, right? as in, when I navigate back to a package, it shows up as 'unliked'

They should be 😬. Was some heavy refactoring today so may be a bit of a bug. Going give it some more heavy testing when I come back. Is that on the preview URL? Should be loaded into cache for a while and once the cache expire it grabs from constellation if you have liked the packaged and the total count

fatfingers23:
I think I've taken troubleshooting about as far as I can on this side of the fence :/. I think someone with access to the Vercel service will need to look. All works locally both for upstash redis and useStorage

When you like it updates the record in your user repo but does not show, notice after the 10 minutes it shows you liked it. Which is about the time it takes for cache entry to expire. I am passing in a fresh instance of `$fetch` to the likes service I made so it should not be using the cachedFetch. On the cache side I tried removing upstash option and tried  useStorage with both `vercel-kv` and `vercel-runtime-cache`. Same thing there. 

Feels very much so like something in Vercel is caching web requests or that the cache backends may not be writing as expected. Any help would be greatly appreciated 🙏

server/utils/atproto/utils/likes.ts
```diff
@@ -0,0 +1,235 @@
...
+  async likeAPackageAndReturnLikes(
+    packageName: string,
+    usersDid: string,
+    atUri: string,
+  ): Promise<PackageLikes> {
+    const totalLikesKey = CACHE_PACKAGE_TOTAL_KEY(packageName)
+    const subjectRef = PACKAGE_SUBJECT_REF(packageName)
+
+    const splitAtUri = atUri.replace('at://', '').split('/')
+    const backLink = {
```
43081j:
```suggestion
    const backLink: Backlink = {
```

then drop the cast

server/utils/cache/local.ts
```diff
@@ -0,0 +1,45 @@
...
+  if (!entry.ttl) return false
+  const now = Date.now()
+  const expiresAt = entry.cachedAt + entry.ttl * 1000
+  return now > expiresAt
+}
+
+/**
+ * Local implmentation of a cache to be used during development
+ */
+export class StorageCacheAdapter implements CacheAdapter {
```
43081j: it may be worth calling this `LocalCacheAdapter`

43081j:
left some more minor comments but otherwise looks good to me!

very nice work.

fatfingers23:
It's finally done 😭. After testing by deploying to another vercel project I have finally found the answer. The api endpoints for likes was caching and still showing that initial load of zero likes. Added for the `/api/social/**` endpoints to never cache since will most likely be handling our own caching in these instances and are for likes.

@danielroe I think it's ready if you want to give it another test and look.

Gonzo17:
Hello :) I just saw the new feature, nice work! It has a minor issue, the translation `package.links.like` (used as `title` of the button) is missing.

PR #720, opened 2026-02-02
chore: update example env openssl command

mezotv:
The current .env.example openssl command is broken and doesn't work so I updated it to work properly!


mezotv:
> interestingly, both variants work for me on a Mac

Strange, what version of openssl are you on? Im on 3.3.6

danielroe:
3.6.0

SCR-20260202-iwrb

mezotv:
> 3.6.0

Weird well the single dash version should work for both

PR #726, opened 2026-02-02
fix(i18n): update Brazilian Portuguese (pt-BR) translation

victoriamdo:
Added missing translations and fixed translations related to atproto.


i18n/locales/pt-BR.json
```diff
@@ -754,27 +772,27 @@
     "connect": "conectar",
     "account": "Conta",
     "npm_cli": "npm CLI",
-    "atmosphere": "Atmosphere",
+    "atmosphere": "Atmosfera",
```
wojtekmaj, line 775: That's translatable?

victoriamdo: Yes, I'm very involved with atproto enthusiasts from Brazil, and we say Atmosfera (or ATmosfera, but I choose not to capitalize the T for this translation).

PR #729, opened 2026-02-02
fix(ui): fallback to exact package lookup for short queries

### LINKED ISSUE #722: Unable to search 1 character packages
When trying to search packages with a single letter, e.g. `n` (A NodeJS manager), you'll get a package not found:

Image

https://www.npmjs.com/package/n

Should show the exact match

MatteoGabriele:
This is a proposal to fix the query length issue. 
NPM has a query minimum of 2 characters, so it won't return anything even if the package actually exists.

https://github.com/user-attachments/assets/63535205-69d3-4d98-bf29-6e278d2aafad

closes #722

### DIFF
--- a/.github/workflows/provenance.yml
+++ b/.github/workflows/provenance.yml
@@ -14,7 +14,7 @@ permissions:
   contents: read
 jobs:
   check-provenance:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
     steps:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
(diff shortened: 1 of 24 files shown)


Netail:
> maybe instead just query to see if that specific package exists, instead of making a search?
> 
> (but still display it as though it were a search)

Would love that too, a regular list entry with the "exact" label

MatteoGabriele:
@danielroe implemented the change. I'm not entirely sold about how I obtained the data, as it's an exact package search that resembles a search result.
Let me know what you think.

https://github.com/user-attachments/assets/1c092492-db71-4103-b6c5-b47c947c0472

MatteoGabriele:
@danielroe I made some changes. I decided to handle the exception in the search result call directly so that the UI wouldn't need any weird exception handling.
Let me know what you think. I'm not a backend developer per se, so I don't have much experience on that _side_.

https://github.com/user-attachments/assets/efdcbee9-d434-4677-8527-724ba4ac4037

wojtekmaj (approved):
LGTM and works

MatteoGabriele:
> Could we avoid the api call entirely since it will always error?

Good point! I'll make a change

danielroe (approved):

PR #732, opened 2026-02-02
chore: provide avatar across npmx

iiio2:
Currently we have avatar only for user, not for orgs and search option. I think we can improve it.

org1
org2
pooya

### DIFF
--- a/app/components/Org/Avatar.vue
+++ b/app/components/Org/Avatar.vue
@@ -1,0 +1,59 @@
+<script setup lang="ts">
+const props = withDefaults(
+  defineProps<{
+    orgName: string
+    size?: 'sm' | 'md' | 'lg'
+  }>(),
+  {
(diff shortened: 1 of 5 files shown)


iiio2:
@pi0, I've taken your screenshot. I hope you did not mind. :smile:

app/components/Org/Avatar.vue
```diff
@@ -0,0 +1,59 @@
...
+  defineProps<{
+    orgName: string
+    size?: 'sm' | 'md' | 'lg'
+  }>(),
+  {
+    size: 'md',
+  },
+)
+
+const hasError = ref(false)
```
OrbisK, line 12:
```suggestion
const hasError = shallowRef(false)
```

iiio2:
@danielroe, oh I see. Then I close this pr.

PR #741, opened 2026-02-02
feat: add `npmx.dev` branded email templates

Kai-ros:
Email templates in `.html` form for the PDS automated emails


fatfingers23 (commented):
Looks good and works on the PDS. Thank you!

danielroe (approved):

ghostdevv:
@Kai-ros @fatfingers23 on each template there is a display none `div` on the html (not in the body), any idea why? I noticed because renovate warned us about the weird unicode characters in it. Unless I'm missing something should we not just remove the div?

fatfingers23:
@ghostdevv I'd imagine it was probably from just copying the upstream emails from the PDS and what they had. But it has actually changed a bit and we can add some branding to the PDSs emails without doing this hack https://bsky.app/profile/pds.dad/post/3mn4ikx4gxk24

Will still need to provide our own template for 2FA since it comes from Gatekeeper, I need to switch the others over to the new way to do branding and i will look at that display none div on it as well

Kai-ros:
Yep, Bailey had it right, at the time I copied the versions he gave me and tried not to tweak too much.

PR #742, opened 2026-02-02
refactor: extract tag component

essenmitsosse:
This removes the Uno shortcut `tag` and replaces it was a reusable tag component. It adds two version: a static version and a clickable one. The static one is used in two places and I think both of them would be clickable eventually, but I didn't want to make too many decisions in one PR. For now this surfaces all the places where we use a non-clickable tag and changes the styling accordingly (previously those also inherited a hover effect).

It also syncs some minor design differences between tags, as well:
- adds a `disabled` styling (inherited from the only disabled tag we used so far)
- adds an `active` styling (inherited from the only tag that had such so far)

This should also make sure all tags have proper focus styles.

### DIFF
--- a/app/components/AppHeader.vue
+++ b/app/components/AppHeader.vue
@@ -19,7 +19,7 @@ const showMobileMenu = shallowRef(false)
 const route = useRoute()
 const isMobile = useIsMobile()
 const isSearchExpandedManually = shallowRef(false)
-const searchBoxRef = shallowRef<{ focus: () => void } | null>(null)
+const searchBoxRef = useTemplateRef('searchBoxRef')
 
 // On search page, always show search expanded on mobile
 const isOnHomePage = computed(() => route.name === 'index')
--- a/app/components/CallToAction.vue
+++ b/app/components/CallToAction.vue
@@ -12,12 +12,12 @@ const socialLinks = {
       {{ $t('about.get_involved.title') }}
     </h2>
 
-    <div class="grid gap-4 sm:grid-cols-3">
+    <div class="grid gap-4 sm:grid-cols-3 sm:items-stretch sm:grid-rows-[auto,1fr,auto]">
       <a
         :href="socialLinks.github"
         target="_blank"
         rel="noopener noreferrer"
-        class="group flex flex-col gap-3 p-4 rounded-lg bg-bg-subtle hover:bg-bg-elevated border border-border hover:border-border-hover transition-all duration-200"
+        class="group grid gap-3 p-4 rounded-lg bg-bg-subtle hover:bg-bg-elevated border border-border hover:border-border-hover transition-all duration-200 sm:grid-rows-subgrid sm:row-span-3"
       >
         <div class="flex gap-2">
           <span class="i-carbon:logo-github shrink-0 mt-1 w-5 h-5 text-fg" aria-hidden="true" />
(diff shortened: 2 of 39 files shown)


app/components/Tag/Static.vue
```diff
@@ -0,0 +1,12 @@
+<script setup lang="ts">
+const props = withDefaults(defineProps<{ as?: string | Component }>(), { as: 'span' })
+</script>
+
+<template>
+  <component
+    :is="props.as"
```
danielroe: although it works, typically better not to use props in the template:

```suggestion
    :is="as"
```

app/components/Tag/Clickable.vue
```diff
@@ -0,0 +1,25 @@
...
+  <component
+    :is="props.as"
+    class="inline-flex items-center px-2 py-0.5 text-xs font-mono border rounded transition-colors duration-200 focus-visible:ring-2 focus-visible:ring-fg focus-visible:ring-offset-1"
+    :class="{
+      'bg-bg-muted text-fg-muted border-border hover:(text-fg border-border-hover)':
+        props.status === 'default',
+      'bg-fg text-bg border-fg hover:(text-text-bg/50)': props.status === 'active',
+      'opacity-50 cursor-not-allowed': props.disabled,
+    }"
+    :disabled="props.disabled"
```
danielroe:
```suggestion
    :class="{
      'bg-bg-muted text-fg-muted border-border hover:(text-fg border-border-hover)':
        status === 'default',
      'bg-fg text-bg border-fg hover:(text-text-bg/50)': status === 'active',
      'opacity-50 cursor-not-allowed': disabled,
    }"
    :disabled="disabled"
```

app/components/Tag/Clickable.vue
```diff
@@ -0,0 +1,25 @@
...
+  <component
+    :is="props.as"
+    class="inline-flex items-center px-2 py-0.5 text-xs font-mono border rounded transition-colors duration-200 focus-visible:ring-2 focus-visible:ring-fg focus-visible:ring-offset-1"
+    :class="{
+      'bg-bg-muted text-fg-muted border-border hover:(text-fg border-border-hover)':
+        status === 'default',
+      'bg-fg text-bg border-fg hover:(text-text-bg/50)': status === 'active',
+      'opacity-50 cursor-not-allowed': disabled,
+    }"
+    :disabled="disabled"
```
knowler, line 21: @essenmitsosse In HTML the `disabled` attribute is a <i>[boolean attribute](https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML)</i> and it applies when it is present. I’m seeing it appear in the keyword lists as `<a href=wherever disabled=false>`. It’s not a huge issue in this case since it has no effect on `<a>` elements (i.e. `disabled` isn’t mapped to anything ARIA-related for anchor elements, see: https://www.w3.org/TR/html-aam-1.0/#att-disabled), but if the element did support the mapping then it would always be disabled. Also, worth noting that for the `role=radio` case, we’d need to manually map `aria-disabled` ourselves (as an [ARIA boolean attribute](https://developer.mozilla.org/en-US/docs/Glossary/Boolean/ARIA) which does use the `"true"` and `"false"` values).

essenmitsosse: Thanks for the comments. Will take then into account for the next PR. But also feel confident that it is good not to rebuild UI Elements from scratch every time.

knowler: > But also feel confident that it is good not to rebuild UI Elements from scratch every time.

I’m also totally for not needing to rebuild things. What I find though is that it’s really easy to identify and create presentational abstractions (i.e. in this case the class names that produce a tag visually), but how presentational abstractions get applied to elements with various semantics is much harder.

I think we’ll likely need some kind of protocol for what components can be used with the `as` prop. For example, if we had a primitive `<Link>` component which could build on `<NuxtLink>`, but knows that if the `disabled` prop is passed to it to use `aria-disabled`, then that could solve the issue without turning stuff like `TagClickable` into a monster component that needs to know how to handle every case. It can just focus on how to apply presentation to the more primitive UI components that we’ve defined.

essenmitsosse: That's more or less what I wanted to sketch out. I think having specific components for their style + semantic combination is the most reasonable way to go.

PR #745, opened 2026-02-02
fix: throw early error for package 404s

iiio2:
I think we can hide internal HTTP detail. In the wrong package we get http detail like "[GET] "https://registry.npmjs.org/danielroelkmdasdpadoapsd": 404 Not Found" (https://npmx.dev/package/danielroelkmdasdpadoapsd).
But for wrong org or username we do not get such internal http detail message. (https://npmx.dev/~danielroelkmdasdpadoapsd).

### DIFF
--- a/app/error.vue
+++ b/app/error.vue
@@ -20,22 +20,6 @@ const statusText = computed(() => {
   }
 })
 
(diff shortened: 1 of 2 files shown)


iiio2:
Thanks.

PR #747, opened 2026-02-02
test: enable type checking of app tests

43081j:


nuxt.config.ts
```diff
@@ -246,7 +246,13 @@ export default defineNuxtConfig({
     tsConfig: {
       compilerOptions: {
         noUnusedLocals: true,
+        // TODO (jg): this isn't actually right since the app shouldn't be
+        // importing `.ts` files directly. But there is an app test which
+        // imports the CLI, which _does_ import `.ts` files directly.
+        // Figure out a better solution later.
```
danielroe: this is right, I think

danielroe: we plan to enable this by default in the next version of nuxt fyi

danielroe:
```suggestion
```

43081j: its a very big grey area

im going to explain on discord instead of here though 😅

PR #756, opened 2026-02-02
fix(ui): prevent the tooltip from clipping in overflow containers

hyoban:
before

CleanShot 2026-02-02 at 23 26 53@2x

after

CleanShot 2026-02-02 at 23 27 16@2x

### DIFF
--- a/app/components/Tooltip/Base.vue
+++ b/app/components/Tooltip/Base.vue
@@ -14,8 +14,8 @@ const props = defineProps<{
   tooltipAttr?: HTMLAttributes
 }>()
 
-const triggerRef = shallowRef<HTMLElement | null>(null)
-const tooltipRef = shallowRef<HTMLElement | null>(null)
+const triggerRef = useTemplateRef('triggerRef')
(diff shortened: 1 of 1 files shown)


app/components/Tooltip/Base.vue
```diff
@@ -12,34 +14,39 @@ const props = defineProps<{
...
 }>()
 
-const positionClasses: Record<string, string> = {
-  top: 'bottom-full inset-is-1/2 -translate-x-1/2 mb-1',
-  bottom: 'top-full inset-is-0 mt-1',
-  left: 'inset-ie-full top-1/2 -translate-y-1/2 me-2',
-  right: 'inset-is-full top-1/2 -translate-y-1/2 ms-2',
-}
+const triggerRef = shallowRef<HTMLElement | null>(null)
+const tooltipRef = shallowRef<HTMLElement | null>(null)
```
danielroe:
```suggestion
const triggerRef = useTemplateRef('triggerRef')
const tooltipRef = useTemplateRef('tooltipRef')
```

PR #759, opened 2026-02-02
chore: include nitro in frameworks

iiio2:
Actually in the framework list, everything is related to frontend or frontend focused. I think we can include one backend thing which may look good.


iiio2:
Thanks @danielroe . :blush:

PR #763, opened 2026-02-02
chore: make search button functional

iiio2:
In our current implementation, search button is not even functional. (https://npmx.dev/). If we click on search button, it does nothing. I think we can keep the button if it does any work. I think in previous implementation I've seen when we click on search button, it navigates to "/search" page. So I make that thing.

### DIFF
--- a/app/pages/index.vue
+++ b/app/pages/index.vue
@@ -19,7 +19,6 @@ const frameworks = ref([
 
 async function search() {
   const query = searchQuery.value.trim()
-  if (!query) return
   await navigateTo({
     path: '/search',
     query: query ? { q: query } : undefined,


danielroe:
I think it doesn't make sense to search for nothing... I think the previous implementation is correct...

iiio2:
@danielroe , that means when we click on search button, it should navigate to "/search" page ? That is what I've done now.

danielroe:
i'm saying it shouldn't navigate without a query

the current behaviour is deliberate and the result of a pr merged earlier today

iiio2:
i think our search is based on typing. But clicking a button is really tough after typing here. when we start typing, it automatically navigates to "/search" page. When the input field is empty, then the search button navigates to "/search" page after clicking the button.

iiio2:
@danielroe , Oh I see. Then I close the pr.

PR #768, opened 2026-02-02
feat: make sparkline graph match accent color

QuietImCoding:
The chart looked a little sad and gray so I used lightening and darkening the accent color to make it pop out more, only issue is the spark itself can be a bit less visible now, I think it's broadly a good tradeoff but wonder what you all think.

### DIFF
--- a/i18n/locales/zh-CN.json
+++ b/i18n/locales/zh-CN.json
@@ -122,14 +122,14 @@
     "verified_provenance": "已验证的来源",
     "view_permalink": "查看这个版本的链接",
     "navigation": "包导航",
-    "copy_name": "复制包名",
+    "copy_name": "拷贝包名",
     "deprecation": {
       "package": "这个包已经被弃用。",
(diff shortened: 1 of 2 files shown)


danielroe:
initially I asked for the chart not to be in the accent as it's not the most important piece of the page and if it's accented it draws the eye too much

but happy to get other opinions

graphieros:
> initially I asked for the chart not to be in the accent as it's not the most important piece of the page and if it's accented it draws the eye too much
> 
> but happy to get other opinions

I prefer the initial vision. The trend is what users are looking for, and even grey, it conveys the information, and having only the pulse with the accent gives a lot of personality

QuietImCoding:
Alright fair enough, i was curious what people thought about it. To me the page looked a bit too gray and so i felt like the accent color could be applied somewhat more judiciously throughout

danielroe:
thank you for the PR @QuietImCoding ❤️

PR #771, opened 2026-02-02
chore: update knip config & prep for prod mode

webpro:
Not using production mode may hide issues, as this PR shows. At the same time, some consider it good to be strict in fast-pacing projects, and others only annoying. Let me know if you want me to add it to CI anyway.

- Bumped Knip
  - A `package.json#scripts.postinstall` no longer requires production `dependencies` (simple-git-hooks, lint-staged), so it won't flag these in production mode
- Did not actually add `pnpm knip --production` to CI, but prepared for that:
  - Re-added a bunch of exclamation marks
  - ~Removed unused exports~
  - Added `@internal` for exports only imported by tests
  - One issue remains: isn't unocss a dev dependency?

```
❯ pnpm knip --production
Unused dependencies (2)
@unocss/preset-wind4  package.json:63:6
unocss                package.json:87:6
```

No need to decide/fix now, as the changes don't affect how Knip is currently used in default/dev mode (but I'd be happy to if you agree!).

### DIFF
--- a/cli/src/schemas.ts
+++ b/cli/src/schemas.ts
@@ -25,6 +25,7 @@ export const PackageNameSchema = v.pipe(
 /**
  * Validates an npm package name for new packages only
  * Stricter than PackageNameSchema - rejects legacy formats (uppercase, etc.)
+ * @internal
  */
 export const NewPackageNameSchema = v.pipe(
   v.string(),
@@ -76,6 +77,7 @@ export const ScopeTeamSchema = v.pipe(
 
 /**
  * Validates org roles
+ * @internal
  */
 export const OrgRoleSchema = v.picklist(
   ['developer', 'admin', 'owner'],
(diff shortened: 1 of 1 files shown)


danielroe:
none of the dependencies in a nuxt project are actually used in prod so I tend to separate dev vs prod deps between lint/tooling stuff  and the stuff required to actually build nuxt (which includes uno)

webpro:
Production mode isn't about dependencies actually being installed or imported at runtime. It's about what's imported from production source code (regardless of getting bundled or not).

webpro:
But yeah, some "faux production entries" are a result of that strategy, and that's fine from a Knip perspective, provided the `!` suffixes everywhere are acceptable.

webpro:
Rebased. Added a bunch of `@internal` tags for exports only imported in tests. A11y audit gives

> TimeoutError: Navigation timeout of 60000 ms exceeded

danielroe:
more than happy to enable `--production` if you think it gives us some extra protection 🙏

webpro:
yeah it's just that we'd need _both_ and run it twice

I'll open a separate PR for that

PR #772, opened 2026-02-02
fix: improve exact match gradient in search results

cassidoo:
For exact matches in the search results, this PR:

- Makes the gradient a bit more subtle (to-accent/10 instead of to-accent/20)
- Makes it prominent on the right side of the search result instead of the center

Screenshots:

<details><summary>Before</summary>
<p>

image

image

</p>
</details> 

and

<details><summary>After</summary>
<p>

image

image

</p>
</details>

### DIFF
--- a/.github/workflows/provenance.yml
+++ b/.github/workflows/provenance.yml
@@ -14,7 +14,7 @@ permissions:
   contents: read
 jobs:
   check-provenance:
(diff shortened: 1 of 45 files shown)


whitep4nth3r (approved):

whitep4nth3r:
Beautiful!

PR #730, opened 2026-02-02
feat: add angular in the framework list

santoshyadavdev:
@danielroe asked me to do it, lol

PR #775, opened 2026-02-02
fix: angular should be lowercase in the footer

cassidoo:
Angular is the only tech starting with a capital letter right now:

image


wojtekmaj:
[thank you](https://github.com/npmx-dev/npmx.dev/pull/730#discussion_r2754551433) :D

PR #777, opened 2026-02-02
chore: build project before playwright test

mihkeleidast:
Effectively does the same thing, but builds for playwright browser tests as a pre-step, so playwright won't have to wait for the build to succeed, rather just the webserver starting, which should be fast.

Lowered the timeout to the default value, but left it in the config, as it might still be useful to see/change later on.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "test:browser": "pnpm build:playwright && pnpm test:browser:prebuilt",
     "test:browser:prebuilt": "playwright test",
     "test:browser:ui": "pnpm build:playwright && pnpm test:browser:prebuilt --ui",
-    "test:browser:update": "playwright test --update-snapshots",
+    "test:browser:update": "pnpm build:playwright && pnpm test:browser:prebuilt --update-snapshots",
     "test:nuxt": "vite test --project nuxt",
     "test:types": "nuxt prepare && vue-tsc -b --noEmit && pnpm --filter npmx-connector test:types",
     "test:unit": "vite test --project unit",


package.json
```diff
@@ -28,12 +29,12 @@
     "generate:lexicons": "lex build --lexicons lexicons --out shared/types/lexicons --clear",
     "test": "vite test",
     "test:browser": "playwright test",
-    "test:browser:ui": "playwright test --ui",
+    "test:browser:ui": "pnpm build:playwright && playwright test --ui",
```
mihkeleidast: Can also move it to a separate step in the CI workflow - that could give us more visibility into if/how the build time changes over time. But kept it here now so the local browser tests run would be the same DX, just run the same command.

package.json
```diff
@@ -12,6 +12,7 @@
   "scripts": {
     "build": "nuxt build",
     "build:lunaria": "node ./lunaria/lunaria.ts",
+    "build:playwright": "NODE_ENV=test pnpm build",
```
mihkeleidast, line 15: not sure where the NODE_ENV matters, kept it for both for now

mihkeleidast:
From what I understood is that the main issue currently is that playwright has to wait for the build to complete before running browser tests - this should affect CI positively in that regard, i.e. it doesn't matter if the build takes 1-2 minutes longer (depending on flakyness in CPU/memory availability for example), playwright only starts after the build completes, and only has to wait for the local webserver to start.

danielroe:
playwright has to wait for the build to finish one way or another - either before playwright starts, or as part of the playwright process - so this is the same as simply adding a very high timeout for that start command

package.json
```diff
@@ -27,13 +28,14 @@
     "postinstall": "nuxt prepare && simple-git-hooks && pnpm generate:lexicons",
     "generate:lexicons": "lex build --lexicons lexicons --out shared/types/lexicons --clear",
     "test": "vite test",
-    "test:browser": "playwright test",
-    "test:browser:ui": "playwright test --ui",
+    "test:browser": "pnpm build:playwright && pnpm test:browser:prebuilt",
+    "test:browser:prebuilt": "playwright test",
+    "test:browser:ui": "pnpm build:playwright && pnpm test:browser:prebuilt --ui",
     "test:browser:update": "playwright test --update-snapshots",
```
danielroe:
```suggestion
    "test:browser:update": "pnpm build:playwright && pnpm test:browser:prebuilt --update-snapshots",
```

PR #779, opened 2026-02-02
fix(a11y): extract explicit tag components and enforced a11y

essenmitsosse:
This splits the existing `TagClickable` into 3 specific components:
- `TagButton`
- `TagLink`
- `TagRadioButtom`
All have their own implementation and enforce proper usage especially with respect a11y.

The idea is that their implementation should only be touched when updating the design system, needing to extend them or wanting to prevent misusage. Their are explicit strict and not flexible, so they can only be used as intended (no `<a href="" type="button">`).

The tag component here is an example if this is a pattern we want to continue with. I am leaving the PR as a draft for now, so we can have some discussion about wether or not this is a good direction.

### DIFF
--- a/test/nuxt/a11y.spec.ts
+++ b/test/nuxt/a11y.spec.ts
@@ -286,7 +286,7 @@ describe('component accessibility audits', () => {
       expect(results.violations).toEqual([])
     })
 
-    it('should have no accessibility violations when it\'s the current link', async () => {
+    it("should have no accessibility violations when it's the current link", async () => {
       const component = await mountSuspended(TagLink, {
         props: { href: 'http://example.com', current: true },
         slots: { default: 'Tag content' },


app/components/Tag/RadioButton.vue
```diff
@@ -0,0 +1,72 @@
...
+const uid = useId()
+const internalId = `${model.value}-${uid}`
+
+const checked = computed(() => model.value === props.value)
+
+const onChange = () => {
+  model.value = props.value
+}
+
+const emit = defineEmits<{
```
danielroe: this is done for you by `defineModel`

app/components/Tag/RadioButton.vue
```diff
@@ -0,0 +1,72 @@
...
+    :htmlFor="internalId"
+  >
+    <input
+      type="radio"
+      :name="props.value"
+      :id="internalId"
+      :value="props.value"
+      :checked="checked"
+      :disabled="props.disabled ? true : undefined"
+      @change="onChange"
```
danielroe, line 35: I think you should just do `v-model="model"` and drop value/checked/change

app/components/Tag/Link.vue
```diff
@@ -0,0 +1,36 @@
...
+    current?: boolean
+  } &
+    /** This makes sure the link always has either `to` or `href` */
+    (Required<Pick<NuxtLinkProps, 'to'>> | Required<Pick<NuxtLinkProps, 'href'>>) &
+    NuxtLinkProps
+>()
+</script>
+
+<template>
+  <span v-if="disabled" class="opacity-50"><slot /></span>
```
danielroe, line 24: we should probably have a number of the other styles so it looks visually similar or you'd experience layout shift when disabling a button

essenmitsosse: Yes, this was a quick one, because this isn't used anywhere yet. Just wanted to sketch it out.

app/components/Tag/RadioButton.vue
essenmitsosse: Oh it was getting late. I was moving stuff around and forgot the clean after me.

danielroe:
conceptually this seems good to me 👍

app/components/Tag/RadioButton.vue
essenmitsosse: This seems buggy in Vue. It doesn't really set the `checked` attribute. Reverting for now.

app/components/Tag/Button.vue
```diff
@@ -0,0 +1,31 @@
...
+    :class="[
+      pressed
+        ? 'bg-fg text-bg border-fg hover:(text-text-bg/50)'
+        : 'bg-bg-muted text-fg-muted border-border hover:(text-fg border-border-hover)',
+      {
+        'opacity-50 cursor-not-allowed': disabled,
+      },
+    ]"
+    type="button"
+    :disabled="disabled ? true : undefined"
```
knowler, line 26: This bit handles the whole issue with `disabled=false` showing up as a bug related to hydration?

app/components/Tag/RadioButton.vue
```diff
@@ -0,0 +1,60 @@
...
+    <input
+      type="radio"
+      :id="internalId"
+      :value="props.value"
+      :checked="checked"
+      :disabled="props.disabled ? true : undefined"
+      @change="onChange"
+      class="peer"
+    />
+    <label
```
knowler, line 38: We can wrap with the label here:

```vue
<label :htmlFor="internalId">
  <input :id="internalId" type="radio" />
  <slot />
</label>
```

I’d still keep the explicit label association using the `for` attribute and not rely on the automatic form association as that’s known to have issues with speech recognition software.

app/components/Tag/RadioButton.vue
```diff
@@ -0,0 +1,60 @@
...
+      class="bg-bg-muted text-fg-muted border-border hover:(text-fg border-border-hover) inline-flex items-center px-2 py-0.5 text-xs font-mono border rounded transition-colors duration-200 peer-focus:ring-2 peer-focus:ring-fg border-none peer-checked:(bg-fg text-bg border-fg hover:(text-text-bg/50)) peer-disabled:(opacity-50 pointer-events-none)"
+      :htmlFor="internalId"
+    >
+      <slot />
+    </label>
+  </div>
+</template>
+
+<style scoped>
+input[type='radio'] {
```
knowler: Is this mostly the `sr-only` class?

app/components/Filter/Panel.vue
```diff
@@ -310,16 +307,14 @@ const hasActiveFilters = computed(() => !!filterSummary.value)
             {{ $t('filters.keywords') }}
           </legend>
           <div class="flex flex-wrap gap-1.5" role="group" :aria-label="$t('filters.keywords')">
-            <TagClickable
+            <TagButton
```
knowler, line 310: Bikeshed: `TagButton` vs. `TagToggleButton` — I know the latter is longer, but it’s more descriptive of the type of button it is though that might feel like it clashes with `TagRadioButton` (though maybe that should just be `TagRadio`)?

app/components/Tag/Button.vue
essenmitsosse: It should yes.

app/components/Tag/RadioButton.vue
essenmitsosse: I had this at first but couldn't get pure CSS styling with the Uno utility classes working. Would you say it's a big problem to leave it like that?

app/components/Filter/Panel.vue
essenmitsosse: I would say the toggleability of this button component ist optional. The component should just work fine as a regular button that never gets `pressed`.

app/components/Tag/RadioButton.vue
essenmitsosse: Spot on!

app/components/Tag/RadioButton.vue
knowler: I think this is probably fine. I usually like the do the `<label>` since it works as a nice hit area, but I guess in this scenario the actual input is visually hidden, so there’s no weird gap between it.

essenmitsosse: Not sure if we should actually display the radio button inside the tag to make the behavior more clear, but I think it's fine for now and that can be solved in a separate PR.

essenmitsosse:
Thanks @knowler for the input. @danielroe I think this would be ready to go if you are fine with it and I would continue with the other components then, now that we have a rough roadmap.

danielroe (approved):

PR #780, opened 2026-02-02
feat(org): embed search scope as operator in url

### LINKED ISSUE #761: fix: Scoped search in org search
Scoped search isn't working on the organizations' page. I can switch search modes (_name, description, etc._), but nothing changes.

We need to set up a proper search (_name, description, keywords - search only by these fields; all and none - search by all_).

We also need to correctly configure the query in the search param. 

I slightly improved their formation in the previous task, but more complex support is probably needed here, since searches of the form `name:something+desc:lorem` are expected. And without breaking keywords

Moshyfawn:
Search scope is now part of the `q=param` using operator syntax like `?q=description:core` instead of separate `scope= param` in preparation for multi-scope search; currently left in single-scope mode. To keep the scope of this PR small, I wired the filter params to be active 1 at a time as multi scope search requires a bigger UI overhaul. [`buildSearchQuery`](https://github.com/Moshyfawn/npmx.dev/blob/d67214482a9e2c6059935b8199e3b50737df5614/app/pages/%40%5Borg%5D.vue#L7C1-L19C2) exists solely for ease of enabling multi args search in the future.

I decided against short keywords in the URL for now, although could adopt `kw`, `desc` etc.

Closes #761


Moshyfawn:
https://npmx-my2gci1n9-poetry.vercel.app/@vueuse is a good test case

Issue #764, opened 2026-02-02
Initial seach improvements


OrbisK:
The initial search works-around the fact that we navigate from the index page to the search page. We need to prevent loss of focus between page renders. Teleports dont work because they recreate the element. 

I can think of two options:

- position/style the layout input to be centred on the index page (same input).
- combine the index and search pages and show the index on initial load.

Feel free to add more.

MatteoGabriele:
Perhaps this issue relates to a PR I closed because it had been stale for too long. Maybe this is the perfect place to discuss it before making any changes. See #408.

dominikg:
+1 to removing the redirect until there is a solution that actually works well in all browsers and with all typing speeds. The initial experience right now for a first time visitor is jarring and might deter them from exploring the rest of it

alexdln:
I'll add a bit: the problem is that when switching between pages, there's a moment where the old input is already removed, and the new one is already mounted but hasn't been hydrated yet.

That is - debounces, slowdowns and delayed changes will not help, but will only reduce the chance of such a thing.

Basically, we could always have a header search field in the header (transparent by default). But then there won't be smooth animation (although I like it without it as well)

<details>
  <summary>Screenshot</summary>
<img width="527" height="457" alt="Image" src="https://github.com/user-attachments/assets/8e11bee1-5b69-4f90-b48c-367ab02afac2" />
<details>

alexdln:
This also happens on other pages - #1211

_The problem might be when we add queries as well, not only navigate to another page..._

MatteoGabriele:
I have a search version that uses a submit-only approach. (see #1107)
As @alexdln mentioned, no matter how much you debounce, there will always be users who type differently and at different speeds, which can cause this bug. It's too unreliable. 
Perhaps we should consider removing the query parameter.

PR #785, opened 2026-02-02
fix: initial search debounce

shamilkotta:
Resolves: #764 

Even though it is not the complete fix or improvement, hope it is good starting point
I've removed the immediate redirection for initial search

> leading: true option. It will cause function to be immediately called


danielroe:
thank you, but I think this isn't the right solution - this will just make it feel a bit slower.

(we previously merged a similar PR but ended up reverting it)

PR #791, opened 2026-02-02
test: add property based tests via fast-check

dubzzz:
This PR proposes to add some tests using property based testing to the project.

**Why?**

The idea behind property based testing is that uncovering all bugs and edge-cases is often impossible as one cannot think of all of them. With this technique in place, developers can rely on the framework to find the issue without to think about all possible dangerous combinations that could exist.

Contrary to using random values in tests, property based tests are deterministic and reproducible as everything relies on a seed.

**It may have found a bug...**

The tests on parseJsDocLinks that I added are failing. Before trying to fix anything (the code or the test) I wanted to get a few insights from people knowing this piece of code.

Current failure is for a text being: `{@link http://a.aa/&}`.

**More**

Last point, adding a native support to Vitest is (by far) the most upvoted idea at the moment: https://github.com/vitest-dev/vitest/discussions/2212. While waiting for that to be true, we can use `@fast-check/vitest`.


test/unit/shared/utils/async.spec.ts
```diff
@@ -92,4 +93,49 @@ describe('mapWithConcurrency', () => {
...
+          expect(results).toEqual(items)
+        },
+      ),
+    )
+  })
+
+  it('not run more than concurrency tasks in parallel', async () => {
+    await fc.assert(
+      fc.asyncProperty(
+        fc.array(fc.anything()), // TODO, support failing tasks too
```
dubzzz: I have not implemented the failing task part yet, but if the idea of adding property based testing to this repository feels sane I can work on adding this failure case (probably into a distinct test case mixing successes and failures but having a different expectation - highly depends on the behavior expected in case of failure)

dubzzz:
@43081j @danielroe Given we discussed of the idea together on Bluesky and Discord

danielroe:
I think this is correct:
> `http://a.aa/&amp;`

this is how vue-router renders links with `&` in them...

dubzzz:
@danielroe Indeed you're right it's correct, I did the test on my browser and `<a href="http://a.aa/&amp;" target="_blank" rel="noreferrer" class="docs-link">http://a.aa/&amp;</a>` has the right display and lead to the right url. I'm gonna change the expectation to capture that.

dubzzz:
Tests adapted, sorry for the false warning

dubzzz:
Thanks ! I'm gonna propose other tests in the coming days.

PR #796, opened 2026-02-03
feat(i18n): update Chinese translations

penjj:


i18n/locales/zh-CN.json
```diff
@@ -202,7 +203,8 @@
     "readme": {
       "title": "Readme",
       "no_readme": "没有可用的 README。",
-      "view_on_github": "在 GitHub 上查看"
+      "view_on_github": "在 GitHub 上查看",
+      "toc_title": "目录"
```
9romise: I think the English term `Outline` is closer to `大纲` in Chinese.

penjj: 🤔 I think so. (Corrected)

PR #798, opened 2026-02-03
chore: fix navigation to home

iiio2:
This is for the mobile version. 

In the Navbar we get logo + search icon in any page like "/about", "/compare" etc. When we click on the search icon it expands a searchbar. Currently when we click on the logo, it also expands a searchbar. I mean I think when we click on logo we should navigate to "/" page. In the desktop version, when we click on the logo, it navigates to "/" page.

### DIFF
--- a/app/components/AppHeader.vue
+++ b/app/components/AppHeader.vue
@@ -194,15 +194,14 @@ onKeyStroke(
     <div class="absolute inset-0 bg-bg/80 backdrop-blur-md" />
     <nav
       :aria-label="$t('nav.main_navigation')"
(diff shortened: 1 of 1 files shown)


danielroe:
it's intentional, to reduce clicks when searching.

what we could do is add an additional search button in the header, that expands into the search bar when clicked. it could be located next to the mobile menu button on the right

iiio2:
@danielroe , I think I've done it now. 


image-search

iiio2:
Thanks @danielroe , @ghostdevv .

PR #799, opened 2026-02-03
refactor(a11y): add headings to call to action cards

knowler:
This refactors the cards in the CTA section (used on the home and about pages) to have headings (i.e. `<h3>` elements) and improves the focus rings. The links are now within the cards rather than wrapping the cards. I’m using [the tried and true trick](https://inclusive-components.design/cards/) of making an sub-element within the link to cover the card, but then using `z-index` to keep heading and paragraph text on top so it can be directly selected.

For accessibility, this makes these call to action cards navigable (i.e. since they have headings AT can navigate to them) and the accessible names of the links are more focused (i.e. they don’t include the entire card’s contents).

Glad to receive feedback on my utility class name ordering.

Aside: eventually a `<Card>` component (with a clickable variant) could be a good abstraction.

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -31,8 +31,8 @@ jobs:
       - name: 📦 Install dependencies
         run: pnpm install
 
-      - name: 📦 Install browsers
-        run: pnpm playwright install
+      - name: 🌐 Install browser
+        run: pnpm playwright install chromium-headless-shell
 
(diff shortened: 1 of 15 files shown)


danielroe:
this will probably need to be updated after https://github.com/npmx-dev/npmx.dev/pull/748 merges.

danielroe (approved):

Issue #802, opened 2026-02-03
[BUG] Mobile Menu: Scroll-to-Top Jump and Search Input Obstruction


NandkishorJadoun:
## Summary
On mobile and tablet devices, closing the hamburger menu triggers an unintended shift upward behaviour. Additionally, the search input becomes inaccessible when the menu is active, causing layout displacement when clicked.

## Problem Description
There are two primary issues occurring on small screen sizes:

1. **Closing the Menu:** When the hamburger menu is open and the user clicks the 'Close' button, the page slides back to the top instead of staying at the user's current scroll position.
2. **Interacting with Search:** When the hamburger menu is active, clicking the search input is impossible. Instead of focusing the input, the page shifts upward by some margin.

https://github.com/user-attachments/assets/019bc16a-f930-4d52-9f03-79f4e6521e11
https://github.com/user-attachments/assets/99dc0ba1-1c59-4f6e-8256-5db05733587f

## Steps to Reproduce
1. View the site on a **mobile or tablet** viewport.
2. Scroll down to the middle or bottom of any page.
3. Open the **hamburger menu**.
4. Perform either of the following:
   - Click the **close button** -> *Observed: Page shifts upward.*
   - Attempt to click the **search input** -> *Observed: Input is blocked and page shifts upward.*

## Expected Behavior
- **Scroll Retention:** The page should maintain its current scroll position after the menu is closed.
- **Input Accessibility:** The search bar should be interactive (or the menu should handle focus/z-index correctly so it doesn't block other UI elements).

PR #804, opened 2026-02-03
fix: move mobile nav menu higher in z-order

iiio2:
resolves https://github.com/npmx-dev/npmx.dev/issues/802


NandkishorJadoun:
@iiio2 thanks for the pr and for looking into this! I just tested the deploy preview on my mobile device, but unfortunately, the issue still persists.

when i scroll down and close the menu, the page still slides to the top. also the search input remains inaccessible while the menu is active. Happy to provide a screen recording if that helps

iiio2:
@NandkishorJadoun , yes please. you can share a screen recording.

NandkishorJadoun:
https://github.com/user-attachments/assets/c477caa8-f13d-4bf5-ab1e-ab83f707f876

iiio2:
@NandkishorJadoun , can you see now with new code ? I think we've corrected it.

iiio2:
I think the problem was in z-index. When the sidebar comes, it should be on top, the main content remains under it. 
For example, In youtube,  I've seen such behavior. so when sidebar comes, we cannot click on main content elements like navbar + others.

npmx:

image

danielroe (approved):

iiio2:
Thanks. @danielroe , @NandkishorJadoun .

PR #812, opened 2026-02-03
fix(ui): optimize table style

zyyv:
Previously, the table header's centering style was inconsistent with the alignment of the `td` content.
- The columns before `description` are now `left-aligned` to display more content, which looks more visually appealing when the width is long.
- The columns after `description` are now `right-aligned` because their values ​​are short and occupy less width, leaving more space for the left side.
- Fixed line breaks in table headers.
- Fixed the issue where the background color of `TagClickable` was lost under `tr:hover`.
- `hover:text-accent-muted` replace to `hover:text-accent-fallback`

| Before | After |
| ------ | ------ |
|   image     |    image    |

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -31,8 +31,8 @@ jobs:
       - name: 📦 Install dependencies
         run: pnpm install
 
-      - name: 📦 Install browsers
-        run: pnpm playwright install
+      - name: 🌐 Install browser
+        run: pnpm playwright install chromium-headless-shell
 
       - name: 🌐 Compare translations
         run: pnpm i18n:check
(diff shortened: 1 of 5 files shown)


uno.config.ts
```diff
@@ -152,6 +152,13 @@ export default defineConfig({
     ['badge-purple', 'bg-badge-purple/10 text-badge-purple'],
     ['badge-pink', 'bg-badge-pink/10 text-badge-pink'],
     ['badge-subtle', 'bg-bg-subtle text-fg-subtle'],
+
+    // table
```
danielroe: we've had some discussions about moving away from uno shortcuts - would you inline this in the components instead? 🙏

zyyv: Done 👌

PR #813, opened 2026-02-03
chore: run knip without installing the repo

wojtekmaj:


wojtekmaj:
Not feasible at the moment: knip loads the config files which are importing other dependencies. Not worth it at the moment.

PR #814, opened 2026-02-03
chore: rename `compare` directory to `Compare`

iiio2:
In the components/ directory, every folder name is capitalized except compare/ folder. I think for consistency we can rename `compare` to `Compare`


iiio2:
Thanks.

PR #819, opened 2026-02-03
ci: split a11y tests using matrix

wojtekmaj:
This PR splits a11y tests into two (light and dark). It's currently the longest running CI job (5m 40s on average), and I expect splitting jobs to provide roughly 1 minute of CI run time savings on average, making it roughly as long as the 2nd longest running job, effectively removing the bottleneck.

Before:

image

https://github.com/npmx-dev/npmx.dev/actions/runs/21621466460

After:

image

https://github.com/npmx-dev/npmx.dev/actions/runs/21625057371

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,6 +96,9 @@ jobs:
 
   a11y:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        mode: [dark, light]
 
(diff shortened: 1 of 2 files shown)


wojtekmaj:
@danielroe Note this will require changes to which jobs are required to pass in order to merge PRs. This is quite disruptive unfortunately.

PR #823, opened 2026-02-03
fix(ui): disable smooth scrolling

danielroe:
missed this in https://github.com/npmx-dev/npmx.dev/pull/254 - it makes it feel a bit jerky when navigating to another page after already having been scrolled down


wojtekmaj:
Nit: Typo in PR summary

danielroe:
I don't know what it is, but I definitely don't want smoth scrolling either

PR #826, opened 2026-02-03
ci: don't update test snapshots in autofix workflow

danielroe:


danielroe:
@wojtekmaj in general I find it helpful - a reviewer can then _assess_ whether the snapshot change was correct (e.g. it can provide easy before/after visualisation).

but in this project there are no snapshots yet, so it seems pointless

PR #828, opened 2026-02-03
test: use fixtures for server side api fetches

danielroe:
this is a first step to using api mocking _on the server_ to make CI tests (lighthouse + playwright) faster and less flaky

the majority of the time is still building the app, but this should help

**before**
SCR-20260203-lbhn

**after**
SCR-20260203-lbkr

### DIFF
--- a/modules/runtime/server/cache.ts
+++ b/modules/runtime/server/cache.ts
@@ -714,7 +714,7 @@ export default defineNitroPlugin(nitroApp => {
     const urlStr =
       typeof input === 'string' ? input : input instanceof URL ? input.toString() : input.url
 
-    if (urlStr.startsWith('/') || urlStr.includes('woff')) {
+    if (urlStr.startsWith('/') || urlStr.includes('woff') || urlStr.includes('fonts')) {
       return await originalFetch(input, init)
     }
 


modules/runtime/server/cache.ts
```diff
@@ -0,0 +1,380 @@
+import type { CachedFetchResult } from '#shared/utils/fetch-cache-config'
+import { join } from 'node:path'
+
+/**
+ * Test fixtures plugin for CI environments.
```
jhroemer, line 5:
```suggestion
 * Test fixtures plugin
```
Can/will be used locally too?

modules/runtime/server/cache.ts
```diff
@@ -0,0 +1,380 @@
...
+ *
+ * This plugin intercepts all cachedFetch calls
+ * and serves pre-recorded fixture data instead of hitting the real npm API.
+ *
+ * This ensures:
+ * - Tests are deterministic and don't depend on external API availability
```
jhroemer, line 11: This section seems more general around the testing philosophy and like something that could be added [here](https://github.com/npmx-dev/npmx.dev?tab=contributing-ov-file#end-to-end-tests) instead?

scripts/generate-fixtures.ts
```diff
@@ -0,0 +1,332 @@
...
+    console.error(`  Failed to fetch user ${username}:`, error)
+    throw error
+  }
+}
+
+// ============================================================================
+// Main
+// ============================================================================
+
+async function main(): Promise<void> {
```
jhroemer, line 387: If I understand correctly this either (re)-generates everything, or strictly the ones you ask for (w. args)? Re-generating could lead to annoyance since you'd have to update test expectations, and re-generate screenshots too (I guess there's none of that currently, but still). Thinking out aloud, a small CLI might be neat, letting you select from the pre-defined options in `REQUIRED_PACKAGES` and what you want to generate and what not.

test/e2e/create-command.spec.ts
```diff
@@ -1,9 +1,6 @@
 import { expect, test } from '@nuxt/test-utils/playwright'
 
 test.describe('Create Command', () => {
-  // TODO: these tests depend on external npm registry API - we should add data fixtures
```
jhroemer, line 4: 🚀

scripts/generate-fixtures.ts
danielroe: I imagine we will only be adding new snapshots rather than updating existing ones.... but happy for a PR to improve this 🙏

scripts/generate-fixtures.ts
```diff
@@ -0,0 +1,332 @@
...
+  'nuxt',
+  'vue',
+  'lodash',
+  'vite',
+  'next',
+  'color',
+  'ufo',
+  'is-odd',
+  'date-fns',
+  'lodash.merge',
```
wojtekmaj: That list is a bit long, is there something unique about each and every item on this list?

danielroe: no, we could probably cut it down. but does it hurt?

test/e2e/url-compatibility.spec.ts
wojtekmaj: Changing these to literally any other person on the planet will halve the size of our fixtures needed 😆

scripts/generate-fixtures.ts
wojtekmaj: Not really, but why have more code if we could have less. In particular, date-fns fixture is enormous. Just parsing JSON that is nearing 100k lines may become a performance bottleneck.

danielroe: will see how far I can cut it down

test/e2e/url-compatibility.spec.ts
danielroe: good point!!

scripts/generate-fixtures.ts
danielroe: updated - see what you think! ✅

PR #837, opened 2026-02-03
perf: handle session cleanup in parallel

iiio2:


iiio2:
Thanks @danielroe . :blush:

PR #841, opened 2026-02-03
feat: add server side atproto blog apis

jonathanyeong:
This pulls apart the feature work done on this PR https://github.com/npmx-dev/npmx.dev/pull/257.

In this commit, we're adding the atproto server side blog apis that will fetch bluesky comments and author profiles. This 
will be used in a later PR that will implement the frontend blog functionality.

### DIFF
--- a/shared/utils/constants.ts
+++ b/shared/utils/constants.ts
@@ -1,4 +1,4 @@
-import * as dev from '#shared/types/lexicons/dev'
+import * as dev from '../types/lexicons/dev'
 
 // Duration
 export const CACHE_MAX_AGE_ONE_MINUTE = 60


shared/utils/parse-basic-frontmatter.ts
```diff
@@ -0,0 +1,34 @@
...
+      if (value === 'true') acc[key] = true
+      else if (value === 'false') acc[key] = false
+      else if (/^-?\d+$/.test(value)) acc[key] = parseInt(value, 10)
+      else if (/^-?\d+\.\d+$/.test(value)) acc[key] = parseFloat(value)
+      else if (value.startsWith('[') && value.endsWith(']')) {
+        acc[key] = value
+          .slice(1, -1)
+          .split(',')
+          .map(s => s.trim().replace(/^["']|["']$/g, ''))
+      } else acc[key] = value
```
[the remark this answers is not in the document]
jonathanyeong: @Kai-ros is this comment valid?

modules/standard-site-sync.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  const document = site.standard.document.$build({
+    site: siteUrl as `${string}:${string}`,
+    path: data.path,
+    title: data.title,
+    description: data.description ?? data.excerpt,
+    tags: data.tags,
+    // This can be extended to update the site.standard.document .updatedAt if it is changed and use the posts date here
+    publishedAt: new Date(data.date).toISOString(),
+  })
```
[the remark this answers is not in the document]
jonathanyeong: @Kai-ros is this comment valid?

server/api/atproto/author-profiles.get.ts
```diff
@@ -81,7 +81,7 @@ export default defineCachedEventHandler(
     maxAge: CACHE_MAX_AGE_ONE_DAY,
     getKey: event => {
       const { authors } = getQuery(event)
-      return `author-profiles:${authors}`
+      return `author-profiles:${authors ?? ''}`
```
Kai-ros: We should default to `npmx.dev` here if there are no listed authors imo, that way we can make it easier for Patak to post official announcements.

server/api/atproto/author-profiles.get.ts
```diff
@@ -0,0 +1,87 @@
...
+      return {
+        authors: authors.map(author => ({
+          ...author,
+          avatar: null,
+          profileUrl: null,
+        })),
+      }
+    }
+
+    const response = await $fetch<ProfilesResponse>(`${BLUESKY_API}app.bsky.actor.getProfiles`, {
```
fatfingers23, line 60: May have already seen and handling client side but `app.bsky.actor.getProfiles` can only handle 25 profiles at a time 

https://docs.bsky.app/docs/api/app-bsky-actor-get-profiles

jonathanyeong: Yeah good point! Thinking about this some more, do we expect 25+ authors on a single blog post? I was going to fetch authors for individual blog posts. Rather than all the authors from every blog post. 

Happy to make changes though, I think slicing an Author array up into 25 element blocks would be the most straightforward. Let me know!

server/api/atproto/author-profiles.get.ts
jonathanyeong: Yah good point! Updated

server/api/atproto/author-profiles.get.ts
fatfingers23: OH I think we are good to go as is then lol. I was thinking it was loading the "Author of the bluesky post reply". Nah we're good just something to note if we do put all 100+ people on a blog post.

modules/standard-site-sync.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  const dateInMicroSeconds = new Date(result.output.date).getTime() * DATE_TO_MICROSECONDS
+
+  // Clock id(3) needs to be the same everytime to get the same TID from a timestamp
+  const tid = TID.fromTime(dateInMicroSeconds, CLOCK_ID_THREE)
+
+  // client.put is async and needs to be awaited
```
fatfingers23, line 128: Will need to make sure the client is logged into as well. Also want to only login when we know we're going do writes so we don't hit session rate limits

https://npmx.dev/package/@atproto/lex#user-content-authenticated-client-with-password

server/api/atproto/bluesky-comments.get.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    const { uri } = parsed.output
+
+    try {
+      // Fetch thread, likes, and post metadata in parallel
+      const [threadResponse, likesResponse, postsResponse] = await Promise.all([
+        $fetch<ThreadResponse>(`${BLUESKY_API}app.bsky.feed.getPostThread`, {
+          query: { uri, depth: 10 },
+        }).catch((err: Error) => {
```
OrbisK:
```suggestion
        $fetch<ThreadResponse>('/app.bsky.feed.getPostThread', {
          baseUrl: BLUESKY_API,
          query: { uri, depth: 10 },
        }).catch((err: Error) => {
```

This way it will always be merged correctly (even if someone removes the trailing slashes).

You could also think about:
```ts
const $bluesky = $fetch.create({baseUrl: BLUESKY_API})
// etc
```

package.json
```diff
@@ -74,6 +75,7 @@
     "@vueuse/router": "^14.2.0",
     "defu": "6.1.4",
     "focus-trap": "^7.8.0",
+    "globby": "16.1.0",
```
OrbisK: @43081j maybe you can help here. I am not that deep into tinyglobby migrations 😅

jonathanyeong: @Kai-ros I noticed we had both globby and tiny globby in our blog post fe branch - https://github.com/npmx-dev/npmx.dev/pull/257. Do you have some more context around tinyglobby migrations? Should I bring in tiny globby over globby?

43081j: ooh yeah definitely use `tinyglobby` instead of `globby`!

it provides the same functionality, at least for what we'll use it for

there is a migration guide from globby here:
https://superchupu.dev/tinyglobby/migration

server/api/atproto/bluesky-comments.get.ts
jonathanyeong: Ooh that's much cleaner, thank you! I went with `$bluesky` since we're using `BLUESKY_API` in multiple places.

package.json
jonathanyeong: Awesome! Thanks for the resource. Looks like we aren't doing anything too fancy (no gitignore) so should be a drop-in replacement.

shared/utils/constants.ts
```diff
@@ -1,4 +1,4 @@
-import * as dev from '#shared/types/lexicons/dev'
+import * as dev from '../types/lexicons/dev'
```
jonathanyeong, line 1: @fatfingers23 I had to update this because it was failing in CI for me https://github.com/npmx-dev/npmx.dev/actions/runs/21649231813/job/62409192743?pr=841

```
Error: shared/utils/constants.ts(1,22): error TS2307: Cannot find module '#shared/types/lexicons/dev' or its corresponding type declarations.
```

Let me know if this change is okay!

server/api/atproto/bluesky-comments.get.ts
jonathanyeong: Updated, thanks for the suggestion!

package.json
jonathanyeong: Updated!

shared/utils/constants.ts
fatfingers23: LGTM! I think `#shared` is the correct way and I had it wrong

danielroe (approved):

shared/utils/constants.ts
jonathanyeong: Weird, I wonder why `#shared` for me is failing 🤔

PR #842, opened 2026-02-03
fix: add missing import

graphieros:
Which broke the charts ;)

### DIFF
--- a/app/composables/useCharts.ts
+++ b/app/composables/useCharts.ts
@@ -45,12 +45,6 @@ export type PackageDownloadEvolutionOptions =
 
 type DailyDownloadsResponse = { downloads: Array<{ day: string; downloads: number }> }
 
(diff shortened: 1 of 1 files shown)


app/composables/useCharts.ts
```diff
@@ -45,12 +46,6 @@ export type PackageDownloadEvolutionOptions =
 
 type DailyDownloadsResponse = { downloads: Array<{ day: string; downloads: number }> }
 
-declare function fetchNpmDownloadsRange(
```
whitep4nth3r, line 48: just checking you meant to delete this

graphieros: And add the import too:)

whitep4nth3r: oh I get it, they're the same name :D

PR #843, opened 2026-02-03
docs: add clarity around naming and target audience

whitep4nth3r:


CONTRIBUTING.md
```diff
@@ -8,11 +8,16 @@ Thank you for your interest in contributing! ❤️ This document provides guide
 
 ## Goals
 
-We want to create 'a fast, modern browser for the npm registry.' This means, among other things:
```
whitep4nth3r, line 11: this change now makes it match the README

CONTRIBUTING.md
```diff
@@ -8,11 +8,16 @@ Thank you for your interest in contributing! ❤️ This document provides guide
...
-- We don't aim to replace the [npmjs.com](https://www.npmjs.com/) registry, just provide a better UI and DX.
-- Layout shift, flakiness, slowness is The Worst. We need to continually iterate to create the most performant, best DX possible.
-- We want to provide information in the best way. We don't want noise, cluttered display, or confusing UI. If in doubt: choose simplicity.
+We're not replacing the [npmjs.com](https://www.npmjs.com/) registry, but instead providing an elevated developer experience through a fast, modern UI.
+
+What npmx offers:
+
+- **Speed** &ndash; Fast searching, filtering, and navigation.
+- **Simplicity** &ndash; Get the information you need when you need it in an intuitive UI.
+- **Compatibility** &ndash; Replace `npmjs.com` with `xnpmjs.com` or `npmx.dev` in any URL and it just works.
```
serhalp: I wonder if this should be more precise, like "URL compatibility"? We don't intend to have compatibility with the app as a whole (I think).

whitep4nth3r: You're probably right, I was trying to go for single word buzzwords :D

CONTRIBUTING.md
```diff
@@ -8,11 +8,16 @@ Thank you for your interest in contributing! ❤️ This document provides guide
 
 ## Goals
 
-We want to create 'a fast, modern browser for the npm registry.' This means, among other things:
+The goal of [npmx.dev](https://npmx.dev) is to build a fast, modern browser for the npm registry.
 
-- We don't aim to replace the [npmjs.com](https://www.npmjs.com/) registry, just provide a better UI and DX.
-- Layout shift, flakiness, slowness is The Worst. We need to continually iterate to create the most performant, best DX possible.
-- We want to provide information in the best way. We don't want noise, cluttered display, or confusing UI. If in doubt: choose simplicity.
+We're not replacing the [npmjs.com](https://www.npmjs.com/) registry, but instead providing an elevated developer experience through a fast, modern UI.
```
serhalp: 💭 This was already here, but I wonder if `We're not replacing the [npmjs.com](https://www.npmjs.com/) registry` could be clearer? In my brain, the registry is called "the NPM registry" and npmjs.com is the site, so "the npmjs.com registry" doesn't help make it abundantly clear that we don't aim to replace the registry.

CONTRIBUTING.md
```diff
@@ -8,11 +8,16 @@ Thank you for your interest in contributing! ❤️ This document provides guide
...
+
+What npmx offers:
+
+- **Speed** &ndash; Fast searching, filtering, and navigation.
+- **Simplicity** &ndash; Get the information you need when you need it in an intuitive UI.
+- **URL Compatibility** &ndash; Replace `npmjs.com` with `xnpmjs.com` or `npmx.dev` in any URL and it just works.
+- **Enhanced admin experience** &ndash; Manage your packages, teams, and organizations from the browser, powered by your local npm CLI.
```
danielroe: I wanted this section to be about the goals/feel rather than the features.

serhalp (approved):
LGTM

CONTRIBUTING.md
whitep4nth3r: done

CONTRIBUTING.md
whitep4nth3r: ah, well, initially this part was just copied from the old readme anyway, let me work on that

whitep4nth3r: Just updated but could be better but I think I need more concrete info.

CONTRIBUTING.md
```diff
@@ -8,11 +8,19 @@ Thank you for your interest in contributing! ❤️ This document provides guide
...
-- We don't aim to replace the [npmjs.com](https://www.npmjs.com/) registry, just provide a better UI and DX.
-- Layout shift, flakiness, slowness is The Worst. We need to continually iterate to create the most performant, best DX possible.
-- We want to provide information in the best way. We don't want noise, cluttered display, or confusing UI. If in doubt: choose simplicity.
+### Core values
+
+- Speed
+- Simplicity
+- Community-first
+
+### Target audience
```
whitep4nth3r, line 19: @danielroe added target audience

serhalp (approved):
💯

Issue #832, opened 2026-02-03
fix: faild to render og on twitter (current X)


ubugeeei:
> https://discord.com/channels/1464542801676206113/1464544758020968448/1468206356707278869

I found about https://npmx.dev/ has invalid twitter og and they use nuxt-og-image or external deps issues.

https://cards-dev.x.com/validator

<img width="1060" height="633" alt="Image" src="https://github.com/user-attachments/assets/1c79c556-639f-4ad9-b267-649918e35d58" />

https://x.com/ubugeeei/status/2018643238475940039?s=20

<img width="536" height="161" alt="Image" src="https://github.com/user-attachments/assets/36b5c0f1-2e96-450d-ab32-a35691b97d21" />

PR #845, opened 2026-02-03
fix: explicit seo meta tags for social cards

RYGRIT:
fixes: #832

https://cards-dev.x.com/validator prompt

WARN:  The image URL https://npmx.dev/__og-image__/static/og.png specified by the 'twitter:image:src' metatag may be restricted by the site's robots.txt file, which will prevent Twitter from fetching it.

I will add `Allow: /__og-image__/*` to the robots.txt file. This method didn't fix the problem either; I need help.


danielroe (approved):

app/pages/index.vue
```diff
@@ -36,7 +36,12 @@ const handleInput = isTouchDevice()
 
 useSeoMeta({
   title: () => $t('seo.home.title'),
+  ogTitle: () => $t('seo.home.title'),
+  twitterTitle: () => $t('seo.home.title'),
+  twitterCard: 'summary_large_image',
   description: () => $t('seo.home.description'),
+  ogDescription: () => $t('seo.home.description'),
+  twitterDescription: () => $t('seo.home.description'),
```
danielroe, line 44: this should probably be done on all pages, but we can merge this to test.

RYGRIT: Yes, that's what I was thinking too, and I'll continue to improve it. Since `summary_large_image` has remained unchanged, should we consider putting it into the `nuxt.config.ts app` configuration?

RYGRIT: It works very well!

image

PR #847, opened 2026-02-03
feat: allow toggle relative dates on click `DateTime`

9romise:
This PR also unifies the `NuxtTime` in `‎app/components/BuildEnvironment.vue‎` to the global `DateTime` component.


danielroe (commented):
is there anywhere the datetime is used within a button or clickable card? I think this might be the case on package search pages.... in this case, embedding a button inside a link will be problematic

9romise:
We have datetime in the home page (build time), package page (last published, version publish date), compare page (last published), and yes, the package search page.

9romise:
Perhaps we could add a prop to control whether it renders as a `span` or a clickable `button`?

danielroe:
I think the problem is that semantically you don't expect `<DateTime>` to render a button. maybe instead add a new `<DateTimeButton>` that just wraps it, and then we use that judiciously where needed?

wojtekmaj (changes requested):
Accessibility issue: button isn't labeled in a way that describes what will happen when you activate it. It's a button... With a date. That's all. 

Unfortunately I don't immediately see a solution that would both keep the text using DateTime readable _and_ properly label the button.

PR #848, opened 2026-02-03
chore: enable sourcemaps in production

danielroe:


userquin:
Uhmmm, maybe we can use build-env.ts module, right now we don't have production env. once there just switch the flag using env (include sources when `env !== production`).

danielroe:
but we want it even in production - the only time we don't want it is in GH Actions as it takes longer to build.

PR #855, opened 2026-02-03
refactor(ui): distributions of elements in package header

MatteoGabriele:
These changes resolve several issues related to the current layout and element placement. They guarantee that the entire header experiences minimal element movement and overlaps.

The header, with its variable package name and numerous elements (version, provenance, badges, code, docs, etc.), can be very hard to maintain.
This becomes especially tricky as the screen size shrinks, causing misaligned elements.

The idea is to limit the number of variables at play:
- the package name sits alone, with only the repository star count, which is a very important metric for how users judge whether to use a library or another
- versions and badges sit right under the package name
- the description stays the same
- Various links and tabs are next to each other, which also makes more sense since the mobile view has them all together.

### Notes
- The org/name has been renamed simply using the provider name to avoid long names that eat space and repetition with the header title
- The forks link has been removed. I can put it back, if everyone thinks it's very useful, but tbh, I feel it's the least important info I can get from the view, and it just takes space and visual attention. 

Let me know what you think and if it needs adjustments

I've attached two videos: one with a small name and one with a longer name, to demonstrate that it works well in both cases.

https://github.com/user-attachments/assets/c071a069-16e8-4a54-8be8-06d315276172

https://github.com/user-attachments/assets/bcc6b394-f3ef-405c-9957-8e9d0756addd

### DIFF
--- a/app/components/Header/MobileMenu.client.vue
+++ b/app/components/Header/MobileMenu.client.vue
@@ -290,13 +290,6 @@ onUnmounted(deactivate)
                 <span class="flex-1">{{ $t('account_menu.connect_atmosphere') }}</span>
               </button>
             </div>
-
-            <!-- Footer -->
-            <div class="p-4 border-t border-border mt-auto">
-              <p class="font-mono text-xs text-fg-subtle text-center">
-                {{ $t('non_affiliation_disclaimer') }}
-              </p>
-            </div>
           </nav>
         </Transition>
       </div>


graphieros:
Looks great !
We discussed in chat about the removal of the fork-count, which does not reach consensus.
I personally think this metric has some merit, but would not mind not having if it improves the layout.

wojtekmaj (approved):
Works nicely for me

jhroemer:
Awesome stuff, thanks for this @MatteoGabriele

danielroe:
I think I ended up somewhat implementing this in https://github.com/npmx-dev/npmx.dev/pull/1188

would you see if there's anything I missed or you still think needs to be done there, and if so, resolve conflicts? 🙏 

I'll mark as draft in the interim.

MatteoGabriele:
Many improvements have been implemented, so I believe the PR is no longer needed. However, we might still want to consider relocating the tabs elsewhere, rather than next to the name, to prevent UI misalignment caused by longer names. Overall, it looks good, and I will close this one. 🙏

PR #860, opened 2026-02-03
fix(i18n): fallback on missing plural translation

BobbieGoede:
~~This is incomplete.~~

~~Only handles Ukrainian at the moment, and for some reason the warning inside the dev if statement is not being logged.~~

Falls back to the highest plural form if given plural amount is not translated, also logs a warning in dev mode if this happens. 

This results in inaccurate plurals if translations are incomplete, but the alternative is an error page.


config/i18n.ts
```diff
@@ -78,6 +78,25 @@ export const countryLocaleVariants: Record<string, (LocaleObjectData & { country
...
 
+function createPluralRule(locale: string, mapping: Record<string, number>) {
+  return (choice: number, choicesLength: number) => {
+    const name = new Intl.PluralRules(locale).select(choice)
+    const plural = mapping[name] || 0
+
+    // In case translation doesn't have all plural forms, use the last available form
+    if (plural > choicesLength - 1) {
+      if (import.meta.dev) {
+        console.warn(
```
userquin, line 90: add `// oxlint-disable-next-line no-console -- warn logging`

userquin:
Looks like we need to add a test...

PR #866, opened 2026-02-03
fix: support tree files in image path resolver

### LINKED ISSUE #865: Fix: Images are not displayed in older versions
Often, older versions (like [Nuxt](https://npmx.dev/package/nuxt/v/4.0.0-alpha.4)) don't display all images, even though they exist and the path is the same as the latest version.

Image

The path processing logic is slightly underdeveloped, in progress

alexdln:
In older versions, we get paths in .md files in the format `/owner/repo/tree/path`

Our current resolver is only able to modify the `/owner/repo/blob/path` format.

A small tweak is needed to support the first format as well. It's only used for img files, so it should work correct. For now, I've only added it for GitHub. I hope someone will help for other systems as well (_I'm asking on discord_)

The problem is both in the markdown files on the package pages and in the markdown files in code view

Closes #865

P.S. I also noticed here that the chevron-down is missing in our package-code, but it’s too minor change, so I added this fix here as well

Source with such urls - https://cdn.jsdelivr.net/npm/nuxt@4.0.0-alpha.4/README.md


alexdln:
I think we can just merge this part for now, since that's the majority of packages and cases, and then we'll add others. I'm afraid to change them without confirmation because there is a risk of breaking more

If we do this, I can create an additional issue (if necessary and this is a potential case)

PR #871, opened 2026-02-03
ci: split ci tests

essenmitsosse:
This splits the unit, component and type tests into three separate checks to speed up getting their results. Additionally this makes it easier to see what failed.

- Result of type check: ~-50s (-50%)
- Result of unit tests: ~-70s (-70%)
- Result of componen tests: ~-20s (-20%)

All other durations are unchanged

Also added some emojis, which will require some changes to GitHub branch protection settings.

### Before/After

Screenshot 2026-02-03 at 20 19 41
Screenshot 2026-02-03 at 20 32 20

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,11 +8,21 @@ Thank you for your interest in contributing! ❤️ This document provides guide
 
 ## Goals
 
-We want to create 'a fast, modern browser for the npm registry.' This means, among other things:
+The goal of [npmx.dev](https://npmx.dev) is to build a fast, modern and open-source browser for the npm registry, prioritizing speed, simplicity and a community-driven developer experience.
 
-- We don't aim to replace the [npmjs.com](https://www.npmjs.com/) registry, just provide a better UI and DX.
-- Layout shift, flakiness, slowness is The Worst. We need to continually iterate to create the most performant, best DX possible.
-- We want to provide information in the best way. We don't want noise, cluttered display, or confusing UI. If in doubt: choose simplicity.
+### Core values
+
+- Speed
+- Simplicity
+- Community-first
+
+### Target audience
+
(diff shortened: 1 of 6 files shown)


essenmitsosse:
I know, this is always annoying on GitHub. Not sure if @danielroe can power through with admin overwrite though?

danielroe:
I love the emojis and granular feedback!

but let me play devil's advocate for a moment. the overall duration of CI doesn't decrease, and the total runner time goes up, right?

I think this is because we end up duplicate initialising the container, installing deps, and also the extra cost of starting vitest with vite-plus (= a second postinstall run pre each test).

1 idea (separate):
I wonder if we could build once and reuse the build between the two a11y tests and the playwright tests?

essenmitsosse:
Yes we don't save much in total run time like that. I wasn't sure if we care about that? Building once and reusing the build is possible. I think I have the setup for that flying around somewhere. But it has a slight overhead for starting the workers, with some potential delay in-between, as well as downloading the build file. As well as it makes the setup a bit more complicated. So I'd only do that if total build minutes is something we worry about.

essenmitsosse:
The advantage of that would be: If a commit fails for the build, you get one failed run, and not 3 and you don't have to wonder if they all failed for the same reason or not. But that's pretty minor I'd say.

serhalp:
>  I wonder if we could build once and reuse the build between the two a11y tests and the playwright tests?

Strangely enough, across every CI platform and all sorts of scenarios, I've pretty much never seen this be worthwhile. The overhead of compressing (some sort of zip/tarball archive under the hood), uploading, downloading, and decompressing (this is often the biggest culprit) really adds up! You'd need the step that builds whatever artifact to be quite slow relative to the size of the artifact [and number of total files—this matters a lot, like with `node_modules/` and such] for it to be worthwhile.

PR #873, opened 2026-02-03
feat: add suggested packages to comparison package selector

thasmo:
Just a [demonstration](https://npmxdev-git-fork-thasmo-suggested-packages-poetry.vercel.app/compare?packages=vue) for "suggested" packages.

### DIFF
--- a/app/components/Compare/PackageSelector.vue
+++ b/app/components/Compare/PackageSelector.vue
@@ -17,6 +17,16 @@ const { data: searchData, status } = useNpmSearch(inputValue, { size: 15 })
 
 const isSearching = computed(() => status.value === 'pending')
 
+const suggestedPackagesCandidates = ['vue', 'svelte', 'solid-js', 'react']
+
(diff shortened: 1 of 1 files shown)


thasmo:
>  I think we could gather some analytics data from /compare to compile a list of "often compared with..."

Yep, that would work. I hoped there's some API, but I can't find any. Also using keyword search is not an option as keyword quality is quite low I'd say. But gathering "often compared with" data would work.

serhalp:
FYI I have this over here: https://github.com/npmx-dev/npmx.dev/pull/801

thasmo:
@serhalp, that's awesome, and I think it's perfect for this type of information. But I also think the "suggested" or "often compared with" packages are different, as in you would compare e.g. next with nuxt or alpine with jquery but you wouldn't want to replace one with the other necessarily.

ghostdevv:
I think we can close this as it was added in #801 - but please reopen if I'm wrong!

PR #877, opened 2026-02-03
fix(a11y): make span metrics badges focusable to render tooltip content

julien-deramond:
On https://npmx.dev/package/bootstrap, three metric badges are displayed:
- **Types** (link)
- **ESM** (span)
- **CJS** (span)

All three badges show informative tooltips on hover. However, only "Types" is keyboard-focusable, preventing keyboard users from accessing the ESM and CJS tooltip information. This creates an accessibility barrier for keyboard navigation:

https://github.com/user-attachments/assets/46234bd5-8477-4865-9ace-432986d86107

Adding `tabindex="0"` to the span elements makes all badges keyboard-focusable so that tooltip content is accessible to all users:

https://github.com/user-attachments/assets/6a5facfe-760d-48dc-b3b9-907f07c6c609

Please note that "Types" is not always a link, and can be a span too when the package has no type. You can check the rendering with https://npmx.dev/package/@ng-bootstrap/ng-bootstrap.

The focus ring rendering is different, but I know there are other issues reported to tackle this specific issue globally.


app/components/Package/MetricsBadges.vue
```diff
@@ -59,6 +59,7 @@ const typesHref = computed(() => {
         <component
           :is="typesHref ? NuxtLink : 'span'"
           :to="typesHref"
+          :tabindex="!typesHref ? 0 : undefined"
```
julien-deramond, line 62: This line is highlighted by the Codecov report as an issue.

I've kept it for now as it's similar to `:is="typesHref ? NuxtLink : 'span'"` already in the file.

I could possibly fix it with something like the following, but I preferred some guidance from the maintainers before changing anything to my initial fix :)

```vue
<script>
const typesTabIndex = computed(() => {
  return !typesHref.value ? 0 : undefined
})
</script>

<component
  :tabindex="typesTabIndex"
>
</component>
```

Note: feel free to modify directly my branch if it's quicker while you're reviewing.

PR #896, opened 2026-02-03
feat: show bluesky avatar in header

danielroe:
SCR-20260203-ueee

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -15,7 +15,7 @@ permissions:
 jobs:
   code:
     name: 🤖 Autofix code
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
 
     steps:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
@@ -24,7 +24,7 @@ jobs:
         with:
           node-version: lts/*
 
-      - uses: pnpm/action-setup@1e1c8eafbd745f64b1ef30a7d7ed7965034c486c
+      - uses: pnpm/action-setup@1e1c8eafbd745f64b1ef30a7d7ed7965034c486c # 1e1c8eafbd745f64b1ef30a7d7ed7965034c486c
         name: 🟧 Install pnpm
         with:
           cache: true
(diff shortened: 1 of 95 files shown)


server/api/auth/atproto.get.ts
```diff
@@ -58,8 +67,33 @@ export default defineEventHandler(async event => {
...
+      )
+      if (profileResponse.ok) {
+        const record = (await profileResponse.json()) as { value: ProfileRecord }
+        const avatarBlob = record.value.avatar
+        if (avatarBlob?.ref?.$link) {
+          // Construct the blob URL from the user's PDS
+          avatar = `${miniDoc.pds}/xrpc/com.atproto.sync.getBlob?did=${encodeURIComponent(agent.did!)}&cid=${encodeURIComponent(avatarBlob.ref.$link)}`
+        }
+      }
+    } catch {
```
fatfingers23, line 91: This catch will happen a good bit for our npmx.social users. They do not get a `app.bsky.actor.profile` on signup. Nothing we have to change now since the cloud will be good back up still

server/api/auth/atproto.get.ts
```diff
@@ -58,8 +67,33 @@ export default defineEventHandler(async event => {
...
+      const profileResponse = await fetch(
+        `${miniDoc.pds}/xrpc/com.atproto.repo.getRecord?repo=${encodeURIComponent(agent.did!)}&collection=app.bsky.actor.profile&rkey=self`,
+        { headers: { 'User-Agent': 'npmx' } },
+      )
+      if (profileResponse.ok) {
+        const record = (await profileResponse.json()) as { value: ProfileRecord }
+        const avatarBlob = record.value.avatar
+        if (avatarBlob?.ref?.$link) {
+          // Construct the blob URL from the user's PDS
+          avatar = `${miniDoc.pds}/xrpc/com.atproto.sync.getBlob?did=${encodeURIComponent(agent.did!)}&cid=${encodeURIComponent(avatarBlob.ref.$link)}`
```
fatfingers23: Since it's a bluesky profile we can leverage their CDN to show these images. Usually loads a bit faster as well

`https://cdn.bsky.app/img/feed_thumbnail/plain/{agent.did!}/{avatarBlob.ref.$link)}@jpeg`

danielroe: Is there any equivalent (of what I'm doing in the PR) that isn't bluesky-specific?

fatfingers23: Whoops sorry I missed this. Nah. I think using the Bluesky profile for now and loading from their CDN is good since that's what most people have @zeucapua is working on a profile lexicon with an avatar blob ref we can probably map over to as well. Default to bsky, if the npmx profile has one use it, etc

PR #902, opened 2026-02-04
chore: use commitlint to standardize the commit messages

kalu5:
Use commitlint to standardize the commit messages .
Avoid CI failures and repeated modifications of commit messages and PR titles

### DIFF
--- a/commitlint.config.ts
+++ b/commitlint.config.ts
@@ -1,0 +1,18 @@
+import type { UserConfig } from '@commitlint/types'
+import { RuleConfigSeverity } from '@commitlint/types'
+
+const Configuration: UserConfig = {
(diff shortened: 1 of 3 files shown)


danielroe:
because we squash prs, really it's only the pr title that needs to match, and we already have a test for that in ci

I think this would add too much friction 

thank you though 🙏

kalu5:
> because we squash prs, really it's only the pr title that needs to match, and we already have a test for that in ci
> 
> I think this would add too much friction
> 
> thank you though 🙏

Ok, I just think it's a bit troublesome to modify the title after the CI fails when there is non - standard submission information.

It's also good to maintain a higher degree of flexibility.

Issue #870, opened 2026-02-03
feat: atproto oauth session refresh, testing, and edgecase handling


fatfingers23:
We need to do some more testing and error handling of atproto OAuth Sessions to make sure they are solid.

- [ ] Test to see how we handle oauth session expired/invalidation. Can revoke oauth via your pds at a url like https://bsky.social/account and wait a bit to test
- [x] make sure oauth sessions are refreshing access tokens as expected. Access tokens are 30mins.  (Maybe #905 should hopefully resolve this)
- [ ] On login make sure we are handling and showing errors for the user if they type in an invalid URL or OAuth returns an error (#905 has basic errors on a error page, but in the auth modal would be better)
- [ ] Decide what to do if the OAuth session is no longer valid. Do we just redirect them back to the PDS to relogin?

Can access the oauth client and session from this event handler 
https://github.com/npmx-dev/npmx.dev/blob/07ce0e403514557fe7e8365ffbb7f4fe99c444a8/server/utils/atproto/oauth.ts#L63

shi-gg:
access tokens are 60 minutes
https://github.com/bluesky-social/atproto/blob/7b9a98a763636c5f66a06da11fe6013f29dd9157/packages/oauth/oauth-provider/src/constants.ts#L36

fatfingers23:
ahhh was going by the docs

https://atproto.com/specs/oauth#tokens-and-session-lifetime

PR #905, opened 2026-02-04
fix: small oauth fixes to extend sessions (hopefully)

fatfingers23:
Working on #870. Takes an hour for each access token to expire. So it's slow going 

- Add fallbacks for if the call to the minidoc fails so we can depend on the did being there fully if an oauth session was successful
- Moved the NodeClient to use handleResolver instead of depending on Bluesky's apis (default if not set)
- Re worked some of the useSession to hopefully not leave any stale values while the OAuthClient updates the store
- Return the error page with an error on OAuth redirect failures. I'd like to move this up to the AuthModal, but did not have luck with anything I tried. May be returning json for either a redirect client side or to show an error
- moved `session.get.ts` to use a regular h3 event handler to see if there may be a race condition and an old refresh token is getting set and sent along. I did see on my last run that it had passed in a brand new session to the store (had a console log). But not sure if it's maybe throwing an expection there on that update or something else

The big error seems to be `Refresh token replayed` every time the token refreshes (access token has 1 hour lifetime. From reading the atproto repo looks like it is sending a refresh token that does not match and my guess is it's sending the previous one and it is not being set or unset somehwere

### DIFF
--- a/server/utils/atproto/oauth-session-store.ts
+++ b/server/utils/atproto/oauth-session-store.ts
@@ -27,6 +27,7 @@ export class OAuthSessionStore implements NodeSavedSessionStore {
         '[oauth session store] Failed to set session:',
         error instanceof Error ? error.message : 'Unknown error',
       )
+      throw error
     }
   }
 


fatfingers23:
I think this PR is good to go. I don't think it will close every thing out for #870, but hoping some changes help with the session refreshes. I've had better luck with it locally testing and I think it's worth submitting for that.

danielroe (approved):

PR #906, opened 2026-02-04
fix: consistent error display in development and production

iiio2:
development and production environment are displaying different error message for non-existent package. I think the local environment is correct here. In production it is likely due to how Nuxt serializes/hydrates errors differently - in production maybe there are ISR/SSR stuffs. 

Production:
prod

Development:
dev

### DIFF
--- a/app/error.vue
+++ b/app/error.vue
@@ -5,9 +5,9 @@ const props = defineProps<{
   error: NuxtError
 }>()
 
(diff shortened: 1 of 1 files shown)


danielroe (approved):

iiio2:
Thanks @danielroe . :blush:

Issue #875, opened 2026-02-03
Display npm scores


graphieros:
Package scores could be displayed somewhere on the package page ?
Perhaps using small gauges (could be css only, or using VueUiGizmo from vue-data-ui), or any other cool solution.

npm API:
[https://api-docs.npms.io/#api-Package-GetPackageInfo](https://api-docs.npms.io/#api-Package-GetPackageInfo)

Inspirations:

<img width="126" height="114" alt="Image" src="https://github.com/user-attachments/assets/17c2068d-c12e-4d07-9970-08a8a0f58a5e" />
<br>

<img width="430" height="233" alt="Image" src="https://github.com/user-attachments/assets/cd986f97-5c7c-4010-8a21-e07566138e64" />

graphieros:
This issue requires further discussions regarding the scoring algorithm.
See #1099

PR #914, opened 2026-02-04
feat: add package scores display 

afterfiveguy:
Closes https://github.com/npmx-dev/npmx.dev/issues/875 

Adds:

- If score exist score bars with smoothed percentage based coloring ( % bewteen red - orange - green )
- If not "score does not exist"
- Icon to clarify the source of the information

image


server/api/registry/score/[...pkg].get.ts
```diff
@@ -0,0 +1,49 @@
+import * as v from 'valibot'
+import { PackageRouteParamsSchema } from '#shared/schemas/package'
+import { CACHE_MAX_AGE_ONE_HOUR } from '#shared/utils/constants'
+
+const NPMS_API = 'https://api.npms.io/v2/package'
```
Comment (43081j) [inline server/api/registry/score/[...pkg].get.ts]: the badge logic also uses the `npms.io`, so we should move this to the constants file and reuse it

Comment (afterfiveguy) [reply server/api/registry/score/[...pkg].get.ts]: Indeed, this is now implemented in the last commit.

serhalp:
we should probably discuss this one a bit on Discord, as it prominently displays some third-party, opinionated metrics that we don't control, changing the overall tone of npmx a bit? 👀

43081j:
@MhemedAbderrahmen we have discussed this and think its best we create our own transparent score.

if you'd like to help, or just see the reasoning, i've opened #1099 

Closing this as we will tackle it another way in future once we have our own in-house score.

Your effort here isn't lost as we will want to render our own score the roughly the same way you have done here eventually. So it'd be nice if you could join in the discussion and help us get there

PR #920, opened 2026-02-04
style: improve contributors layout

iiio2:
our contributors grid looks different in different mobile version specially on right side there is space. i think we can slightly improve it.

Example 1:
two

Example 2: 
one

New: 
new


danielroe (approved):

iiio2:
Thanks.

PR #929, opened 2026-02-04
fix(ui): improve MetricsBadges skeleton alignment

wojtekmaj:
Skeleton was misaligned due to:
- Skeleton being positioned differently (quirky because of baseline alignment of parent flex element)
- Skeleton being too small
- MetricsBadged introducing extra margins unexpectedly
- Missing "white skeleton" for Like button

Before:

https://github.com/user-attachments/assets/c320b568-78ca-4ef4-a2c1-f62f18ad016c

It's actually even worse than that because also the right toolbar with docs, code and compare may jump to the next line because of it.

After:

https://github.com/user-attachments/assets/68ffaeb1-ecfb-4e55-929c-bbc604da57b3

### DIFF
--- a/app/components/Package/MetricsBadges.vue
+++ b/app/components/Package/MetricsBadges.vue
@@ -12,7 +12,7 @@ const { data: analysis, status } = usePackageAnalysis(
   () => props.version,
 )
 
-const isLoading = computed(() => status.value === 'pending')
+const isLoading = computed(() => status.value !== 'error' && !analysis.value)
 
 // ESM support
 const hasEsm = computed(() => {


app/pages/package/[...package].vue
```diff
@@ -569,16 +569,17 @@ defineOgImageComponent('Package', {
...
-              <div class="flex items-center gap-1.5 self-baseline ms-1 sm:ms-2">
-                <SkeletonBlock class="w-8 h-5 rounded" />
-                <SkeletonBlock class="w-12 h-5 rounded" />
-                <SkeletonBlock class="w-5 h-5 rounded" />
-              </div>
-            </template>
-          </ClientOnly>
+              <template #fallback>
+                <div
+                  class="flex items-center gap-1.5 list-none m-0 p-0 relative top-[5px] self-baseline ms-1 sm:ms-2"
```
Comment (wojtekmaj) [inline app/pages/package/[...package].vue:618]: The `relative top-[5px]` sucks a bit, but gets the job done. Honestly, I have no better idea how to fix it while keeping baseline alignment.

Comment (danielroe) [reply app/pages/package/[...package].vue:618]: what about having the text visible so it's just the icon that loads? we could even replace it with a spinner to give better feedback

Comment (wojtekmaj) [reply app/pages/package/[...package].vue:618]: That's probably a good idea, but if I may have a say, I'd do it as a next iterative improvement. It's likely not going to be such an easy fix as this was.

Comment (danielroe) [reply app/pages/package/[...package].vue:618]: pushed a first stab - what do you think?

Comment (wojtekmaj) [reply app/pages/package/[...package].vue:618]: Sweet!

app/pages/package/[...package].vue
```diff
@@ -576,7 +620,7 @@ defineOgImageComponent('Package', {
                   <SkeletonBlock class="w-16 h-5.5 rounded" />
                   <SkeletonBlock class="w-13 h-5.5 rounded" />
                   <SkeletonBlock class="w-13 h-5.5 rounded" />
-                  <div class="w-13 h5.5" />
+                  <SkeletonBlock class="w-13 h-5.5 rounded bg-bg-subtle" />
```
Comment (wojtekmaj) [inline app/pages/package/[...package].vue:623]: I intentionally used a white div as the like button is itself white, but I'm okay with this.

Issue #918, opened 2026-02-04
Inconsitent ways to return to homepage


HarryLack:
In smaller viewports (~640px), the logo icon is collapsed into the search bar leaving no way to return to the homepage. 
Additionally, the settings page has a back button in the section title, while the about page has a back to home link in the footer so could also do with a consistency pass.

Settings
<img width="828" height="255" alt="Image" src="https://github.com/user-attachments/assets/a6e05623-ecee-4db2-ac4c-9a3b0c521d90" />
About
<img width="828" height="255" alt="Image" src="https://github.com/user-attachments/assets/de3d552c-0ce0-4415-9252-e4522edddf3c" />

Suggested solution as per @danielroe
> basically I'm totally up for splitting out a separate search icon probably putting it next to the mobile menu and then leaving the logo to return to the home page

RYGRIT:
I've also thought about this issue. Generally, clicking the logo should take you back to the homepage, but I think the option to go back to the previous page should also be added to the settings, about, and compare pages to maintain consistency.

PR #936, opened 2026-02-04
chore: consistent back button across pages

iiio2:
resolves https://github.com/npmx-dev/npmx.dev/issues/918

page-one
page-two
page-three


iiio2:
Thanks.

PR #944, opened 2026-02-04
fix: normalise package licence field

wojtekmaj:
Before:

image

https://npmx.dev/package/g

After:

image

image

### DIFF
--- a/shared/types/npm-registry.ts
+++ b/shared/types/npm-registry.ts
@@ -12,6 +12,7 @@ import type { ReadmeResponse } from './readme'
 // Re-export official npm types for packument/manifest
 export type { PackumentVersion, Manifest, ManifestVersion, PackageJSON } from '@npm/types'
 
+// TODO: Remove this type override when @npm/types fixes the license field typing
 export type Packument = Omit<PackumentWithoutLicenseObjects, 'license'> & {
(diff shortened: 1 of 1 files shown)


shared/types/npm-registry.ts
```diff
@@ -6,17 +6,16 @@
  * @see https://github.com/npm/registry/blob/main/docs/REGISTRY-API.md
  */
 
-import type { PackumentVersion } from '@npm/types'
+import type { Packument as PackumentWithLicenceBug, PackumentVersion } from '@npm/types'
```
43081j:
```suggestion
import type { Packument as BasePackument, PackumentVersion } from '@npm/types'
```

it feels less workaround-ish then. i know it is a workaround but its going to be here a long time

wojtekmaj: That stops it from being self documenting. :(

43081j: pinged you on discord but imo we should achieve this via a comment on the `type` rather than making people infer it from the symbol name

wojtekmaj:
See also: https://github.com/npm/types/issues/27

wojtekmaj:
Upstream fix: https://github.com/npm/types/pull/185

Issue #937, opened 2026-02-04
fix: package sidebar glitching


afterfiveguy:
I believe this bug was caused due to this [PR ](https://github.com/npmx-dev/npmx.dev/pull/805/changes)

I couldn't get the correct scenario on how it happens, but it has to do with the scrollbar and collapsible items.

https://github.com/user-attachments/assets/d31818ac-c7c4-49f1-b263-0e1063e1406c

PR #947, opened 2026-02-04
fix: package sidebar glitching

iiio2:
resolves https://github.com/npmx-dev/npmx.dev/issues/937

### DIFF
--- a/app/pages/package/[...package].vue
+++ b/app/pages/package/[...package].vue
@@ -1186,7 +1186,7 @@ onKeyStroke(
       <div class="area-sidebar">
         <!-- Sidebar -->
         <div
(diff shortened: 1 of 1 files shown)


graphieros:
It appears to still be glitching in preview

iiio2:
I think the cause of glitching is hover:pe-0.5 class from the sidebar div. This was causing the sidebar to change its padding-end on hover (from pe-2.5 to pe-0.5), which likely caused the "glitching" effect. I have tested in desktop + mobile and now it works well.

danielroe (approved):

iiio2:
Thanks. :blush:

PR #948, opened 2026-02-04
fix(i18n): add translation for 'likes' key

Pascalmh:
When hovering over the "Likes"-Icon on the Package-Page the `package.links.like` key is not translated.

image

--- 

After the fix:

image


wojtekmaj:
I'd suggest going with #957 instead as an even more complete solution

PR #958, opened 2026-02-04
draft: check og logic without global rule

alexdln:
Sorry, just test, related - https://discord.com/channels/1464542801676206113/1464544758020968448/1468646668957126912


alexdln:
Recreated - #962 for clean history

Issue #959, opened 2026-02-04
About page spacing issue


shuuji3:
On mobile viewport, https://npmx.dev/about page has extra space on the right side.

https://github.com/user-attachments/assets/984640e4-bbb6-4e4f-8870-da5403692b09

PR #965, opened 2026-02-04
fix: horizontal overflow

Flo0806:
fixes: #959 

---

Fixes wrong horizontal scrolling


43081j (approved):
it does fix it but i was curious why it really overflows at all. it seems to be the contributors grid that somehow causes the gap.

i.e. if you `overflow-x: hidden` only the contributors list, it works as expected.

but this will act as a catch-all so lets go with it

Issue #954, opened 2026-02-04
feat: show fixed versions for vulnerabilities


nitodeco:
Currently its shown that there *are* vulnerabilities, but not *which* version fixes them.

[Example](https://npmx.dev/package/next/v/15.5.6):

<img width="777" height="180" alt="Image" src="https://github.com/user-attachments/assets/8980c204-0c25-456c-badc-de38f9f6b6b5" />

<img width="773" height="125" alt="Image" src="https://github.com/user-attachments/assets/84c8ad1d-9a28-4860-a2c9-a12d52b62e94" />

A great value addition would be to show what earliest version after the affected one resolves the vulnerability.

In the case of Next.js 15.5.6, this would be 15.5.9 for the deprecation, and 16.1.5 for 0 vulns.

Flo0806:
@nitodeco @danielroe like this? 

<img width="856" height="202" alt="Image" src="https://github.com/user-attachments/assets/586f8a42-58ba-4e7c-b158-85a96b398977" />

nitodeco:
@Flo0806 Exactly, neat!

PR #967, opened 2026-02-04
feat: show fixed version for vulnerabilities

Flo0806:
Fixes: #954 

---

Get OSV-fixed version(s) for vulnerabilities if available and shoe it inside vulnerabilities banner.

### DIFF
--- a/i18n/schema.json
+++ b/i18n/schema.json
@@ -1128,6 +1128,9 @@
                 }
               },
               "additionalProperties": false
+            },
+            "fixed_in_title": {
+              "type": "string"
             }
           },
           "additionalProperties": false


server/utils/dependency-analysis.ts
```diff
@@ -144,6 +146,34 @@ function getVulnerabilityUrl(vuln: OsvVulnerability): string {
...
+  for (const range of packageAffected.ranges) {
+    for (const event of range.events) {
+      if (event.fixed) {
+        return event.fixed
+      }
+    }
+  }
+
+  return undefined
+}
```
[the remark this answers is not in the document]
danielroe: @Flo0806 did you see this comment?

danielroe (commented):
in general I think we should avoid creating ai translations as we currently don't have a way of asking for them to be reviewed by native speakers

server/utils/dependency-analysis.ts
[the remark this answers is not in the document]
Flo0806: Yes, this evening I'll fix it. I just  wanted to keep it simple. But I try to make it with all available informations

Flo0806:
> in general I think we should avoid creating ai translations as we currently don't have a way of asking for them to be reviewed by native speakers

So I should let translations **empty**? Or is there a "team" to help? My last request got lost because there was too much information in Discord i18n.

danielroe:
yes, just leave the translations empty 🙏

Flo0806:
@danielroe now it picks the current next fixed version of a vulnerability. Additional I added some tests.

trueberryless:
Native speaker here, for this one we could keep the German translation as is, it's correct 👍

server/utils/dependency-analysis.ts
```diff
@@ -144,6 +148,64 @@ function getVulnerabilityUrl(vuln: OsvVulnerability): string {
   return `https://osv.dev/vulnerability/${vuln.id}`
 }
 
+/**
+ * Check if a version falls within an OSV range (between introduced and fixed).
+ * OSV ranges use events: introduced starts vulnerability, fixed ends it.
+ */
+function isVersionInRange(version: string, range: OsvRange): boolean {
+  const introduced = range.events.find(e => e.introduced)?.introduced
+  const fixed = range.events.find(e => e.fixed)?.fixed
```
ghostdevv: could there be multiple of the same kind of event for a given package?

ghostdevv (changes requested):

server/utils/dependency-analysis.ts
Flo0806: Normally not from what I've read

app/components/Package/VulnerabilityTree.vue
```diff
@@ -150,6 +150,13 @@ function getDepthStyle(depth: string | undefined) {
                   {{ vuln.id }}
                 </a>
                 <span class="truncate w-0 flex-1">{{ vuln.summary }}</span>
+                <span
+                  v-if="vuln.fixedIn"
+                  class="shrink-0 font-mono text-emerald-600 dark:text-emerald-400"
+                  :title="$t('package.vulnerabilities.fixed_in_title', { version: vuln.fixedIn })"
+                >
+                  → {{ vuln.fixedIn }}
+                </span>
```
danielroe: should this be a link so you can click through to see the fixed version?

and should we show the version constraint so that it's clear it's not _just_ fixed in that one version?

-> ^3.0.0 or -> >= 3.0.0

Flo0806: sure!

server/utils/dependency-analysis.ts
serhalp: 😁

I gotchu:

Screenshot 2026-02-07 at 20 01 40

https://npmx-j14rd1m0y-poetry.vercel.app/package/next/v/15.5.8

Check out https://github.com/advisories/GHSA-5f7q-jpqc-wp7h

danielroe (approved):

serhalp:
👀 https://github.com/npmx-dev/npmx.dev/pull/967#discussion_r2778316328

^ this still seems to not quite behave as one would expect. I think the minimal repro is something like:

- vulnerable: [`>= 3.4.5-foo.2, < 3.4.8`, `>= 3.5.9-foo.7, < 3.5.9-foo.15`]
- patched: [`3.4.8`, `3.5.9-foo.15`]

In this case if you're looking at e.g. vulnerable `3.4.6` we currently suggest (I think) bumping to patched `3.5.9-foo.15`, which is _technically_ not inaccurate, but we should ideally suggest `3.4.8` — or keep it simple and show all options.

danielroe:
in practice it was the same - it returned the 'first' fix which would have been 3.4.8 in your example (and in the sample OSV data I looked at). but I've pushed a fix to make that explicit

PR #976, opened 2026-02-05
fix: deduplicate when dependencies === transitive dependencies

gameroman:
## Before

image

{D7F66734-DECA-457E-86D0-DE5DEC93432F}

## After

image

{3A1D5502-7EDF-49F3-87CE-E38CA7F542B9}


gameroman:
Yeah, makes sense I guess, will do so when I get home

PR #977, opened 2026-02-05
Exclude btn and tooltip from selection of install commands 

benjavicente:
Makes the buttons and the tooltip non selectable for the install terminal commands.

https://github.com/user-attachments/assets/cdc83cca-efcb-477a-b1c9-a4210c54db82


benjavicente:
I can't edit the title, will open another PR

PR #984, opened 2026-02-05
fix: hide likes when data is being fetched

sacrosanctic:
the likes flashes a 0 briefly before showing the actual value.

I opted to just hide it for now as a quick fix.

### DIFF
--- a/app/pages/package/[...package].vue
+++ b/app/pages/package/[...package].vue
@@ -399,10 +399,13 @@ const { user } = useAtproto()
 
 const authModal = useModal('auth-modal')
 
-const { data: likesData } = useFetch(() => `/api/social/likes/${packageName.value}`, {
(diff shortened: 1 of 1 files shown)


btea:
Perhaps adding a loading effect would be better. #982

sacrosanctic:
I figured this is just a quick fix for a bigger redesign down the line. Since there's a lot of UI elements being added at the moment.

jhroemer:
> Perhaps adding a loading effect would be better. #982

I agree, a simple `<SkeletonInline>` with the right size would make the effect less jarring. I stumbled on the same thing @sacrosanctic btw so thanks for opening the PR.

ghostdevv:
Looks like this was resolved in #982 :sweat_smile: - let me know if I'm wrong :pray:

PR #986, opened 2026-02-05
chore: remove duplicated footer content

iiio2:
"not affiliated with npm inc." this is a part of footer content. But this content is showing also in our sidebar. I think we may remove the duplicate content here. 

Before: 
before

After:
after


iiio2:
Thanks @43081j .

PR #988, opened 2026-02-05
style: improve background color for overflowed in Select

rzzf:
| Before | After |
| - | - |
| image | image |


danielroe:
I think this bg might have been addressing an issue rendering on Linux

danielroe (approved):

rzzf:
Thanks for the review. Since I don’t have a Linux environment, I wasn’t able to test it, so I switched to an alternative implementation.

PR #991, opened 2026-02-05
feat: add trail effect on sparkline pulse

graphieros:
- bump vue-data-ui to 3.14.6
- enable the trail effect in the sparkline chart config

| Before | After |
|--------|------|
|BEFORE (1)|AFTER (1)|

The effect can be tempered or disabled in `WeeklyDownloadStats.vue` at `config.style.line.pulse.trail`:

```js
trail: {
  show: true, // set to false to revert to previous behavior
  length: 20,
  opacity: 0.75, // can be reduced to temper the effect (ej. 0.5)
}

```
I think it looks cool ^^

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "vite-plugin-pwa": "1.2.0",
     "vite-plus": "0.0.0-833c515fa25cef20905a7f9affb156dfa6f151ab",
     "vue": "3.5.27",
-    "vue-data-ui": "3.14.6"
+    "vue-data-ui": "3.14.7"
   },
   "devDependencies": {
     "@intlify/core-base": "11.2.8",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -198,8 +198,8 @@ importers:
         specifier: 3.5.27
         version: 3.5.27(typescript@5.9.3)
       vue-data-ui:
-        specifier: 3.14.6
-        version: 3.14.6(vue@3.5.27(typescript@5.9.3))
+        specifier: 3.14.7
+        version: 3.14.7(vue@3.5.27(typescript@5.9.3))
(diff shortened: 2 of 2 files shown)


garthdw (approved):

shuuji3:
@graphieros Although this is not directly related to this PR, I think we also should provide an option to disable animation from accessibility perspective (some may found it difficult to focus on the main content when there's moving object on the page).

Does vue-data-ui already have a functionality to do this such as detecting `prefers-reduced-motion`? But probably it's better have npmx setting item since `prefers-reduced-motion` requires to configure browser or OS setting.

ref. Understanding Success Criterion 2.2.2: Pause, Stop, Hide | WAI | W3C - https://www.w3.org/WAI/WCAG21/Understanding/pause-stop-hide.html

graphieros:
> @graphieros Although this is not directly related to this PR, I think we also should provide an option to disable animation from accessibility perspective (some may found it difficult to focus on the main content when there's moving object on the page).
> 
> Does vue-data-ui already have a functionality to do this such as detecting `prefers-reduced-motion`? But probably it's better have npmx setting item since `prefers-reduced-motion` requires to configure browser or OS setting.
> 
> ref. Understanding Success Criterion 2.2.2: Pause, Stop, Hide | WAI | W3C - https://www.w3.org/WAI/WCAG21/Understanding/pause-stop-hide.html

Yes, the pulse does not show when prefers-reduced-motion, it is imposed by the library

graphieros:
I added a comment in the config to make that information available :)

shuuji3:
Great, thank you! 🙂

graphieros:
keeping it open for now, I noticed I left a dreaded log in my last vue-data-ui build (shaaame) 😭
More coffee please!

Issue #916, opened 2026-02-04
feat: Enter key should open the first package when searching


martinszeltins:
Currently when you search for a package like "vue", you need to either use tab or "down" key to select the package then press enter to open it.

It would be nice to be able to search for a package and then press "Enter" to open the first one found. It would make navigation even faster!

Related: https://github.com/npmx-dev/npmx.dev/issues/575

<img width="1738" height="865" alt="Image" src="https://github.com/user-attachments/assets/47d394cf-e7c9-413e-9f32-b310c2a2ec20" />

serhalp:
Closed by #993.

PR #993, opened 2026-02-05
feat: enter key open the first package when searching

kalu5:
feat #916

### DIFF
--- a/app/pages/search.vue
+++ b/app/pages/search.vue
@@ -541,46 +541,57 @@ function focusElement(el: HTMLElement) {
   el.scrollIntoView({ block: 'nearest', behavior: 'smooth' })
 }
 
-// Find latest package name
-function findLatestPackageName() {
-  const packageName = displayResults.value?.[0]?.package.name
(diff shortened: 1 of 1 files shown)


app/pages/search.vue
```diff
@@ -541,7 +541,48 @@ function focusElement(el: HTMLElement) {
...
+    if (latestPackageName) return navigateToPackage(latestPackageName)
+    // Waiting for the latest search results (maximum 1.5 seconds)
+    let waitSearchResultInterval: ReturnType<typeof setInterval> | null
+    function clearSearchResultInterval() {
+      if (waitSearchResultInterval) clearInterval(waitSearchResultInterval)
+      waitSearchResultInterval = null
+    }
+    waitSearchResultInterval = setInterval(() => {
+      const latestPackageName = findLatestPackageName()
+      if (latestPackageName) {
```
danielroe: should we check the search input is still focused and user hasn't started trying to navigate 

in this case, rather than using a timeout, we could watch the search results

PR #1002, opened 2026-02-05
chore: add missing connect to npm cli option in mobile version

iiio2:
currently add to npm cli is missing in mobile version. I think maybe we can add this.

clin


danielroe:
this is deliberate as you are unlikely to have the npm cli installed on your phone

wojtekmaj:
> this is deliberate as you are unlikely to have the npm cli installed on your phone

iiio2:
Oh I see, Thanks.

PR #904, opened 2026-02-04
feat: accessible rings, accent colors & visual changes

jellydeck:
closes #611 
closes #633

- use accent colors for focus rings in navbar
- AppLogo component which changes color as per accent color
- fixes in spacing for visual alignment
- compare page searchbar icon aligned
- copy package name button updates itself rather than calling popover for better UX

PR #1011, opened 2026-02-05
fix: remove border opacity on toggle

jellydeck:
accessible pretty switches

https://github.com/user-attachments/assets/c846184c-567d-4613-ad78-6b0fb4b90ff8

as per discussion to be more accessible 
 https://github.com/npmx-dev/npmx.dev/pull/904#discussion_r2765347606


lukewarlow:
Yeah this is better, in high contrast it's pretty easy to spot now.

image

PR #1012, opened 2026-02-05
chore(docs): rebuild `better-sqlite3` on install

danielroe:


ghostdevv:
@danielroe do you remember why this was needed?

danielroe:
I think it might have been a deploy issue on vercel - the binding got lost somehow with the build cache

PR #1019, opened 2026-02-05
chore: add e18e lint plugin

43081j:
This also disables `no-console` since pretty much everything in this
repo can run server-side, where console logs are valuable when done
right.

### DIFF
--- a/server/api/atproto/author-profiles.get.ts
+++ b/server/api/atproto/author-profiles.get.ts
@@ -49,11 +49,7 @@ export default defineCachedEventHandler(
 
     if (handles.length === 0) {
       return {
-        authors: authors.map(author => ({
-          ...author,
-          avatar: null,
-          profileUrl: null,
-        })),
+        authors: authors.map(author => Object.assign(author, { avatar: null, profileUrl: null })),
       }
     }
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -33,10 +33,12 @@ jobs:
 
       - uses: pnpm/action-setup@1e1c8eafbd745f64b1ef30a7d7ed7965034c486c # 1e1c8eafbd745f64b1ef30a7d7ed7965034c486c
         name: 🟧 Install pnpm
-        # pnpm cache skipped deliberately as the project is not actually installed here
```
serhalp, line 36: since this is no longer true, shouldn't we add the cache option here?

.oxlintrc.json
```diff
@@ -1,18 +1,25 @@
...
   "plugins": ["unicorn", "typescript", "oxc", "vue", "vitest"],
+  "jsPlugins": ["@e18e/eslint-plugin"],
   "categories": {
     "correctness": "error",
     "suspicious": "warn",
     "perf": "warn"
   },
   "rules": {
-    "no-console": "warn",
+    "no-console": "off",
```
serhalp: I never understood the purpose of `warn`s, so I'm good with this 😁. Longer term should we be using a li'l logger module and turning this to `"error"`? (maybe keep off for `scripts/`)

43081j: we really need to review logging overall.

only truly error-level things should be logged at error level, for example. this isn't currently the case iirc

i'll start a discussion around this 👍

43081j:
FYI this does mean the lint job can often take longer now (if it is the first to populate the cache)

but having the workaround `lint.ts` was causing a lot of friction and means we can't use lint plugins. we should revisit how we can speed it back up

.oxlintrc.json
danielroe: would it be possible to disable this just on a per-directory case? so we disable it for server/ and cli/ but keep it for app/ ?

danielroe:
@43081j are you happy with those small tweaks?

43081j:
looks good to me 👍

PR #1023, opened 2026-02-05
chore: cache node_modules in CI jobs

43081j:
Possibly not dominik-friendly but we shall see.


43081j:
closing this as it conflicts with pnpm cache. ill find another way to speed things up 👍

PR #1024, opened 2026-02-05
chore: fix auth error message

iiio2:
Suppose, if we click the Connect to Atmosphere modal and then provide invalid username in the input field and click on connect button, it throws error as 410 - Server Error.

I think 401 error message is not related to server error. 401 should not unauthorized. Server error starts from 500. 

Wrong:
wrong

Correct: 
right


danielroe:
would you take a look at the failing tests? 🙏

iiio2:
@danielroe , I think I've done it.

ghostdevv (approved):

iiio2:
Thanks.

PR #1027, opened 2026-02-05
refactor: use built-in `$fetch` helper

iiio2:
in some parts, we have used nuxt's `$fetch` helper, in some parts we have used native `fetch` in server directory. I think for consistency we can use `$fetch` helper everywhere.


ghostdevv:
Will put as draft for now since Vercel is failing, mark is as ready when you are! :smile: /nbr

ghostdevv (changes requested):
@iiio2 are you interested in updating this PR?

PR #1031, opened 2026-02-05
chore: include new deps to vite optimized deps

userquin:
From time to time we add some dependencies that should be optimized, this PR adds the following dependencies:
- `@vueuse/integrations/useFocusTrap`
- `fast-npm-meta`
- `@floating-ui/vue`

image


trueberryless (approved):
LGTM 🎉

EDM115:
@userquin (sorry for the ping) I always wondered, is there any guidelines out there that rules which deps should/shouldn't be optimized by Vite ?

PR #158, opened 2026-01-27
perf: use vue router to parse package page paths

danielroe:

PR #1041, opened 2026-02-05
refactor: use object-syntax/named routes

danielroe:
closes https://github.com/npmx-dev/npmx.dev/pull/158

### DIFF
--- a/server/middleware/canonical-redirects.global.ts
+++ b/server/middleware/canonical-redirects.global.ts
@@ -13,6 +13,7 @@
  * - /docs/*         → /package-docs/*
  */
 const pages = [
+  '/200.html',
   '/about',
   '/compare',
   '/org',


app/components/Terminal/Install.vue
```diff
@@ -149,7 +149,7 @@ const copyCreateCommand = () => copyCreate(getFullCreateCommand())
               ></code
             >
             <NuxtLink
-              :to="`/package/${typesPackageName}`"
+              :to="packageRoute(typesPackageName!)"
```
alexdln, line 152: Looks risky. I think we should either specify that the type is required in the props, or make it optional inside the method and `if (!name) return undefined`.

server/middleware/canonical-redirects.global.ts
```diff
@@ -0,0 +1,77 @@
...
+
+const cacheControl = 's-maxage=3600, stale-while-revalidate=36000'
+
+export default defineEventHandler(async event => {
+  const routeRules = getRouteRules(event)
+  if (Object.keys(routeRules).length > 1) {
+    return
+  }
+
+  const [path = '/', query] = event.path.split('?')
```
alexdln, line 36: Could be sth like https://npmx.dev/nuxt#user-content-getting-started

It seems to be working now, apparently the path doesn't include query and hash and perhaps this logic is unnecessary

server/middleware/canonical-redirects.global.ts
```diff
@@ -0,0 +1,77 @@
...
+    const args = [pkgMatch.groups.org, pkgMatch.groups.name].filter(Boolean).join('/')
+    setHeader(event, 'cache-control', cacheControl)
+    return sendRedirect(event, `/package/${args}` + (query ? '?' + query : ''), 301)
+  }
+
+  // /@org/pkg/v/version or /@org/pkg@version → /package/org/pkg/v/version
+  // /pkg/v/version or /pkg@version → /package/pkg/v/version
+  const pkgVersionMatch =
+    path.match(/^\/(?:(?<org>@[^/]+)\/)?(?<name>[^/@]+)\/v\/(?<version>[^/]+)$/) ||
+    path.match(/^\/(?:(?<org>@[^/]+)\/)?(?<name>[^/@]+)@(?<version>[^/]+)$/)
```
alexdln, line 59: same as 

`path.match(/^\/(?:(?<org>@[^/]+)\/)?(?<name>[^/@]+)(?:\/v\/|@)(?<version>[^/]+)$/)`

(not-captured group with `\/v\/` or `@`)

alexdln:
@danielroe I was just checking how it will work now and caught these little moments. They're all minor, so it just to not to miss them 🙃

server/middleware/canonical-redirects.global.ts
danielroe: hash is typically not sent to the server at all

PR #1046, opened 2026-02-05
feat(i18n): detect missing and dynamic translation keys

sybers:
# Add i18n validation to CI

## What's this about?

Added a script that checks if all translation keys used in the code actually exist in our locale files. It also flags dynamic keys like $t(someVariable) which can't be statically validated.

## Changes

- New script scripts/find-invalid-translations.ts powered by vue-i18n-extract
- New CI job that runs on every PR
- Fixed a few components that had hardcoded strings or weird key patterns

## Why bother?

Right now, if someone uses $t('typo.in.key'), nothing breaks during build, it just shows up broken for users.

This catches those issues before they get merged. The CI will fail on:
- Missing keys (typos, forgotten translations)
- Dynamic keys (impossible to validate statically)
- Unused keys are just warnings, no need to block PRs for cleanup tasks.

## Long term

As we add more locales, manually checking translations becomes impossible. Better to automate it now than debug weird blank strings in production later 😄


danielroe:
I also think we might consider auto-pruning unused keys 🙏

sybers:
> I also think we might consider auto-pruning unused keys 🙏

Good idea, will open a PR for this as well :)

Issue #1028, opened 2026-02-05
feat: use input-checkbox for toggle instead of button


alexdln:
We currently use a `button` as the primary element for `toggle`. This is generally correct, but it's not universal and convenient.

We need to use `input type="checkbox"` along with a `label` as the base element.

It has a cleaner separation of concerns, better component layout, is more scalable, works without JS, is native to forms (_get toggle data from formData_), and allows us to do more with CSS (incl. `:has(input:checked)`)

_It is important to keep focus behavior and update a11y tests_

abbeyperini:
Claiming this to work on in the next few days.

PR #1049, opened 2026-02-05
fix: settings toggle should use input

abbeyperini:
Update settings page toggle to use an `<input type="checkbox" />` instead of a button. Wrap the whole thing in a label and hide the checkbox behind styled toggle, so the click/touch target is pretty huge. Use translate animation and style it to look about the same as before. Use theme color so all color modes are supported. Use locale dir to support right to left languages.

Closes: #1028


app/components/Settings/Toggle.client.vue
```diff
@@ -1,54 +1,134 @@
...
-    </span>
-  </button>
+  <label :for="id">
+    <span class="toggle--label-text">{{ label }}</span>
+    <input
+      role="switch"
+      type="checkbox"
+      :id
+      class="toggle--checkbox"
+      :aria-checked="checked"
```
knowler: We can get rid of the redundant `aria-checked` as well.
```suggestion
```

app/components/Settings/Toggle.client.vue
```diff
@@ -1,54 +1,134 @@
...
+      :checked="checked"
+      @click="checked = !checked"
```
knowler: I don’t know how controlled/uncontrolled components work in Vue, but can we use `checked` how it works in HTML (i.e. as the initial state) and then let it be uncontrolled after that, updating state with the `input` event? I don’t know if that’s idiomatic for Vue.

app/components/Settings/Toggle.client.vue
```diff
@@ -1,54 +1,134 @@
...
+  }
+  60% {
+    left: 3px;
+    width: 40px;
+  }
+  100% {
+    left: 20px;
+    width: 20px;
+  }
+}
```
knowler: Can we use `translate` + `scale` instead here? They have better performance characteristics (i.e. they animate on the compositor rather than needing to do layout + paint).

knowler (commented):
This is a good change. I noticed these had some affordances I don’t typically associate with switches (i.e. activation on <kbd>Enter</kbd>).

app/components/Settings/Toggle.client.vue
abbeyperini: Yep, I should be using v-model. I'm not done translating the old React code I copied to Vue

app/components/Settings/Toggle.client.vue
abbeyperini: Didn't even need width/scale or keyframes 👍

app/components/Settings/Toggle.client.vue
```diff
@@ -1,81 +1,121 @@
...
-      border-color: Highlight;
-
-      & > span {
-        background: HighlightText;
-      }
-    }
-  }
+.toggle--checkbox:checked.rtl + .toggle--background:before {
+  transform: translate(-17px);
 }
```
knowler, line 151: I imagine this might have gotten dropped when merging as I believe [it was added after](#1150) you started your PR. Can we incorporate it somehow? If you don’t have Windows, you can [emulate forced colours in Chrome devtools](https://devtoolstips.org/tips/en/emulate-forced-colors/).

app/components/Settings/Toggle.client.vue
```diff
@@ -1,81 +1,121 @@
...
+
+label:has(input:hover) .toggle--background {
+  background: var(--fg-muted);
 }
-html[dir='rtl'] button[aria-checked='true'] > span:last-of-type > span {
-  translate: calc(-100%);
+
+/* Circle that moves */
+.toggle--checkbox:checked + .toggle--background:before {
+  animation-fill-mode: forwards;
```
knowler: We can drop this

```suggestion
```

app/components/Settings/Toggle.client.vue
```diff
@@ -1,81 +1,121 @@
...
+  <label :for="id">
+    <span class="toggle--label-text">{{ label }}</span>
+    <input
+      role="switch"
+      type="checkbox"
+      :id
+      class="toggle--checkbox"
+      :class="dir"
+      v-model="checked"
+    />
```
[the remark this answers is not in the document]
abbeyperini: This is called [same name shorthand](https://vuejs.org/guide/essentials/template-syntax.html#same-name-shorthand) and has been available since Vue 3.4. we're using Vue 3.5.27

app/components/Settings/Toggle.client.vue
```diff
@@ -1,81 +1,121 @@
...
+}
+
+/* Support rtl locales */
+.toggle--checkbox.ltr + .toggle--background:before {
+  left: 3px;
+}
+
+.toggle--checkbox.rtl + .toggle--background:before {
+  right: 3px;
+}
```
knowler: We should be able to `inset-inline-start` for this.


```suggestion
.toggle--checkbox + .toggle--background::before {
  inset-inline-start: 3px;
}
```

app/components/Settings/Toggle.client.vue
```diff
@@ -1,81 +1,121 @@
...
-html[dir='rtl'] button[aria-checked='true'] > span:last-of-type > span {
-  translate: calc(-100%);
+
+/* Circle that moves */
+.toggle--checkbox:checked + .toggle--background:before {
+  animation-fill-mode: forwards;
+  transition: transform 200ms ease-in-out;
+}
+
+.toggle--background:before {
```
knowler: nit: pseudo-elements should use `::`

```suggestion
.toggle--background::before {
```

danielroe:
this is a great change!

just one comment: the radii/position of the dot isn't quite right. here's a comparison

<details>
<summary>before</summary>


SCR-20260211-baky

</details>


<details>
<summary>after</summary>

SCR-20260211-baty


</details>

app/components/Settings/Toggle.client.vue
```diff
@@ -1,81 +1,121 @@
...
-  translate: calc(100%);
+
+label:has(input:hover) .toggle--background {
+  background: var(--fg-muted);
 }
-html[dir='rtl'] button[aria-checked='true'] > span:last-of-type > span {
-  translate: calc(-100%);
+
+/* Circle that moves */
+.toggle--checkbox:checked + .toggle--background:before {
```
knowler: nit: pseudo-elements should use `::`

```suggestion
.toggle--checkbox:checked + .toggle--background::before {
```

app/components/Settings/Toggle.client.vue
```diff
@@ -1,81 +1,121 @@
...
+  animation-fill-mode: forwards;
+  transition: transform 200ms ease-in-out;
+}
+
+.toggle--background:before {
+  animation-fill-mode: forwards;
```
knowler: Can drop
```suggestion
```

app/components/Settings/Toggle.client.vue
```diff
@@ -1,81 +1,121 @@
...
+/* Support rtl locales */
+.toggle--checkbox.ltr + .toggle--background:before {
+  left: 3px;
+}
+
+.toggle--checkbox.rtl + .toggle--background:before {
+  right: 3px;
+}
+
+.toggle--checkbox:checked.ltr + .toggle--background:before {
```
knowler: nit: pseudo-elements should use `::`

```suggestion
.toggle--checkbox:checked.ltr + .toggle--background::before {
```

app/components/Settings/Toggle.client.vue
```diff
@@ -1,81 +1,121 @@
...
-    &[aria-checked='true'] > span:last-of-type {
-      background: Highlight;
-      border-color: Highlight;
-
-      & > span {
-        background: HighlightText;
-      }
-    }
-  }
+.toggle--checkbox:checked.rtl + .toggle--background:before {
```
knowler: nit: pseudo-elements should use `::`

```suggestion
.toggle--checkbox:checked.rtl + .toggle--background::before {
```

app/components/Settings/Toggle.client.vue
```diff
@@ -1,81 +1,121 @@
...
+defineEmits(['update:modelValue'])
 const checked = defineModel<boolean>({
-  default: false,
+  required: true,
+})
+const id = useId()
+const { locale, locales } = useI18n()
+const dir = computed(() => {
+  const localeObj = locales.value.find(item => item.code === locale.value)
+  return localeObj?.dir ?? 'ltr'
```
knowler: I think we could use [the `:dir()` pseudo-class function](https://developer.mozilla.org/docs/Web/CSS/Reference/Selectors/:dir) for this? [Its browser support](https://caniuse.com/css-dir-pseudo) is the same as `:has()`.

app/components/Settings/Toggle.client.vue
abbeyperini: Has it not been incorporated into the theme in main.css? We should also add it to the tests since we're already doing dark and light mode

knowler: > Has it not been incorporated into the theme in main.css?

It probably should be, but it isn’t right now (there are some forced colours styles there, but they don’t apply to this).

knowler: I see `@media (prefers-contrast: more)` for some styles, but that doesn’t apply for forced colours unfortunately. Why don’t we create a separate PR for that and we can ignore here.

abbeyperini: Oh yeah, I was suggesting issues, not volunteering lol

app/components/Settings/Toggle.client.vue
```diff
@@ -1,81 +1,100 @@
...
+
+label:has(input:hover) .toggle--background {
+  background: var(--fg-muted);
+}
+
+/* Circle that moves */
+.toggle--checkbox:checked + .toggle--background::before {
+  transition: transform 200ms ease-in-out;
+}
+
```
knowler: This transition is already being applied below.

```suggestion
```

app/components/Settings/Toggle.client.vue
```diff
@@ -1,81 +1,100 @@
...
+  width: 44px;
+  height: 24px;
+  background: var(--fg-subtle);
+  border-radius: 9999px;
+  border: 1px solid var(--fg);
+  display: flex;
+  position: relative;
+}
+
+label:has(input:focus) .toggle--background {
```
knowler:
```suggestion
label:has(input:focus-visible) .toggle--background {
```

app/components/Settings/Toggle.client.vue
```diff
@@ -1,81 +1,100 @@
...
+/* background */
+.toggle--background {
+  width: 44px;
+  height: 24px;
+  background: var(--fg-subtle);
+  border-radius: 9999px;
+  border: 1px solid var(--fg);
+  display: flex;
+  position: relative;
+}
```
knowler, line 132: For the hover/focus transitions for the track:
```suggestion
  transition: background-color 100ms ease-in, border-color 100ms ease-in;
}
```

My code formatting might be off (maybe lint will catch it).

knowler (commented):
Sorry, one last set of changes all related to the transitions.

app/components/Settings/Toggle.client.vue
abbeyperini: Found a bug - in high contrast mode, the label isn't showing 1/2 the time. I'll fix that when I finish these styles in the morning.

https://github.com/user-attachments/assets/643b2d53-65cd-4ccf-978c-3e7392b041a4

app/components/Settings/Toggle.client.vue
```diff
@@ -1,81 +1,96 @@
...
-      background: Highlight;
-      border-color: Highlight;
+label:has(input:checked) .toggle--background {
+  background: var(--fg);
+  border-color: var(--fg);
+}
+
+label:has(input:hover) .toggle--background {
+  background: var(--fg-muted);
+}
```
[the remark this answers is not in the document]
abbeyperini: That's on purpose.

> This may explain the styling discrepancy noted in the review comments.

That discussion is about high contrast, not hover.

knowler (approved):
Looks good to me! As usually, I’ll defer to others with regard to Vue related stuff.

I will follow up with the forced colors mode stuff at the theme-level.

app/components/Settings/Toggle.client.vue
```diff
@@ -1,81 +1,96 @@
 <script setup lang="ts">
-defineProps<{
-  label?: string
+const props = defineProps<{
+  label: string
   description?: string
   class?: string
 }>()
-
+defineEmits(['update:modelValue'])
```
danielroe: is this used?

abbeyperini: I meant to look up if defineModel defined the emits as well as the prop and forgot 😮‍💨

app/components/Settings/Toggle.client.vue
```diff
@@ -77,76 +75,96 @@ const checked = defineModel<boolean>({
       <span v-else-if="label" class="text-sm text-fg font-medium text-start">
         {{ label }}
       </span>
+      <input
+        role="switch"
+        type="checkbox"
+        :id
+        class="toggle-checkbox opacity-0"
```
knowler: @danielroe It is possible to completely style this using utility classes if we wanted to. We don’t need the additional element or to use `label:has()`.

Just need to leverage `before:` where the `<input>` becomes the track and `::before` is the thumb.

Example (not styled the same): https://play.tailwindcss.com/TRertzjTg2

~I presume Uno allows for equivalent class generation.~ [Uno version](https://unocss.dev/play/#html=DwWwhglgdgBAxgGzAZ2QXgEQAcC0A2GAdxwDMBXBBGcADx2IQHMMA%2BAKBhmCQCMBTKohToMJBHxowwCCIyg4eKAdD4wAVmWQAXCCQCeCvlsJ8%2BUVh04wAIgHsYyWyCMALaI0udg0LGS2erLT0sPjQ4Fz44AGseWxoAzgAnW3E0ZEIILXCE%2BCRUTBzOMQkcKFsoPkKpLBCwRLAoOD5S8sqrdpgXfCIcAEYABhhcXqroGQqcRkSIABMYKdmcOBTkHAAmKuSyKBm%2BGdIKKh5GSfqDAHZ%2BwZm6qIAuY9OwAwH%2Bqq16qGRMiHKllNsiWQMBmZHqOj%2Bryq4UiUT2DxO0Bmsls%2BCuINudxh0XhjyRKJwAFYrlUSLY4JocAA3CDfHjiO62PzjZq2EgkZBGPowMkU1Y0ukMplaFnrHnkykCiD0viM5kqBQIMjNYlvKr8MmJWXLKBaMxaHAAbQA5MaALowDWA2VdAAslr4mtlxHtVq1CKeBgAzOibol7m7ZY8ps8cJdBhj-ViIjiZg9HdaPSHvSSOlZA3ctjs9gdKA6nViUjhkFgGn11Qn3R8Gt8IfJlghAcgqiBbHXi2ASEHK7Lq18fn8%2B8hNSBqG3fvJkJ3uwXQeCJ%2BtU2n4DG4XGM0OkHqcHRyJQchgAuwvAB6XgCdjAE-gaAsIA&config=JYWwDg9gTgLgBAbzgEwKYDNgDtUGEJaYDmANHGFKgM6owDq2yALHAL5zpQQhwDkArlggBjKlV4AoCagAekWCgwBDfgBt4aTDnyFgRABQIJccpRowqALjgBtCtVoMszfQEoAuiQmtXEoA&css=PQKgBA6gTglgLgUzAYwK4Gc4HsC2YDCAyoWABYJQIA0YAhgHYAmYcUD6AZllDhWOqgAOg7nAB0YAGLcwCAB60cggDYIAXGBDAAUKDBi0mXGADe2sGC704AWgDuCGAHNScDQFYADJ4Dc5sAACtMLKAJ5gggCMLPK2ABR2pPBIcsoAlH4WAEa0yADWTlBYqEw2yFjK3Bpw5LxxAOTllVDoYpSMYgAs3vUZ2gC%2BmsBAA&options=N4IgLgTghgdgzgMwPYQLYAkyoDYgFwJTZwCmAvkA) (just needed to add the `before:content-['']` class to get it working like the TW one).

danielroe: pushed something - hoping I haven't totally messed things up merging in main 🙏

I trust @abbeyperini will come back and clean up my code...

app/components/Settings/Toggle.client.vue
abbeyperini: Daniel added this one

app/components/Settings/Toggle.client.vue
abbeyperini: Ok the latest commit should fix forced contrast styles. I will accept style tag to inline utility class translations if you've got them.

https://github.com/user-attachments/assets/ec23cb18-c471-4e77-9b0b-97ca95e82067

abbeyperini: Also, my friend made this excellent [Color Blind & Low Vision Simulator Chrome Extension](https://chromewebstore.google.com/detail/color-blind-low-vision-si/lgflijofdempkdlaknbfmacfidjjhjbb)

app/components/Settings/Toggle.client.vue
abbeyperini: Just trying to make the label optional and forgetting the style tag in the server component. 

Thanks for adding the tooltip and prefers-reduced-motion check. It's always fun to see how other people would right the same thing.

app/components/Settings/Toggle.client.vue
knowler: Did a quick test with Windows High Contrast Mode (Night sky theme) and Edge on Windows 11 and this is what I found:

1. Looks like the label disappears when its enabled.
2. There’s also a slight flicker for the “thumb” when it changes colour.
3. When enabled the background and text colour is probably failing the minimum contrast ratio.

https://github.com/user-attachments/assets/69c1ec2f-a330-4137-8430-ef7e844d7e2d

knowler: Worth noting that when using a light contrast theme (e.g. “Desert”) these issues aren’t present, but I might still avoid the label colour inversion:

The “Display” settings for NPMX shown using the Desert theme for Windows High Contrast Mode. The labels are inverted when the control is enabled, but the contrast of everything is fine.

abbeyperini: Yeah, the label text color + background color/highlight is the crux of the issue. If `--fg` works in light mode, it doesn't work in dark mode.

abbeyperini: Ok I think I found the right Highlight/Canvas combo and reduced the flash. Just gotta test on a windows machine

knowler (approved):
Thanks for working through those forced colours styles, @abbeyperini! LGTM!

PR #1052, opened 2026-02-05
feat: suggest installing dependency as devDependency when appropriate

### LINKED ISSUE #530: Suggest installing dependency as devDependency when deemed appropriate
The command to install the package is always identical regardless of the package purpose.

Some packages are indisputably better to be installed as devDependency. This helps developers avoid adding a normal dependency by mistake which may result in said dependency pulled by all of its users. Suboptimal to say the least. 

Moreover, call me stupid but I never remember the command to install devDependencies on npm. I end up installing it normally and manually fixing package.json 🤦🏻‍♂️

To avoid this, we could:
- Maintain a list of popular packages which are extremely likely to be installed as a devDependency, like eslint, vite, esbuild, oxfmt, typescript, and more.
- Scan package's README for hints; in many cases install command is there, and if the maintainers themselves are suggesting installation as devDependency, then we could do the same as well.

What do you think?

wojtekmaj:
Closes #530

Examples:

image

https://npmxdev-git-fork-wojtekmaj-dev-dependency-suggestion-npmx.vercel.app/package/@types/react

image

https://npmxdev-git-fork-wojtekmaj-dev-dependency-suggestion-npmx.vercel.app/package/vite-tsconfig-paths

### DIFF
--- a/app/components/Terminal/Install.vue
+++ b/app/components/Terminal/Install.vue
@@ -183,16 +183,17 @@ const copyDevInstallCommand = () =>
                 >{{ i > 0 ? ' ' : '' }}{{ part }}</span
               ></code
             >
-            <button
+            <ButtonBase
               type="button"
-              class="px-2 py-0.5 font-mono text-xs text-fg-muted bg-bg-subtle/80 border border-border rounded transition-colors duration-200 opacity-0 group-hover/devinstallcmd:opacity-100 hover:(text-fg border-border-hover) active:scale-95 focus-visible:opacity-100 focus-visible:outline-accent/70 select-none"
+              size="small"
+              class="text-fg-muted bg-bg-subtle/80 border-border opacity-0 group-hover/devinstallcmd:opacity-100 active:scale-95 focus-visible:opacity-100 select-none"
               :aria-label="$t('package.get_started.copy_dev_command')"
               @click.stop="copyDevInstallCommand"
             >
               <span aria-live="polite">{{
                 devInstallCopied ? $t('common.copied') : $t('common.copy')
               }}</span>
-            </button>
+            </ButtonBase>
           </div>
         </template>
 
--- a/shared/utils/package-analysis.ts
+++ b/shared/utils/package-analysis.ts
@@ -34,8 +34,6 @@ export interface ExtendedPackageJson {
   dependencies?: Record<string, string>
   devDependencies?: Record<string, string>
   peerDependencies?: Record<string, string>
-  readme?: string
-  readmeFilename?: string
   /** npm maintainers (returned by registry API) */
   maintainers?: Array<{ name: string; email?: string }>
   /** Repository info (returned by registry API) */
--- a/i18n/schema.json
+++ b/i18n/schema.json
@@ -667,6 +667,12 @@
             "copy_command": {
               "type": "string"
             },
+            "copy_dev_command": {
+              "type": "string"
+            },
+            "dev_dependency_hint": {
+              "type": "string"
+            },
             "view_types": {
               "type": "string"
             }
--- a/server/api/registry/analysis/[...pkg].get.ts
+++ b/server/api/registry/analysis/[...pkg].get.ts
@@ -25,6 +25,10 @@ import { parseRepoUrl } from '#shared/utils/git-providers'
 import { encodePackageName } from '#shared/utils/npm'
 import { getLatestVersion, getLatestVersionBatch } from 'fast-npm-meta'
 
+interface AnalysisPackageJson extends ExtendedPackageJson {
+  readme?: string
+}
+
 export default defineCachedEventHandler(
   async event => {
     // Parse package name and optional version from path
@@ -42,7 +46,7 @@ export default defineCachedEventHandler(
       // Fetch package data
       const encodedName = encodePackageName(packageName)
       const versionSuffix = version ? `/${version}` : '/latest'
-      const pkg = await $fetch<ExtendedPackageJson>(
+      const pkg = await $fetch<AnalysisPackageJson>(
         `${NPM_REGISTRY}/${encodedName}${versionSuffix}`,
       )
 


app/components/Terminal/Install.vue
```diff
@@ -131,6 +158,41 @@ const copyCreateCommand = () => copyCreate(getFullCreateCommand())
...
+              class="px-2 py-0.5 font-mono text-xs text-fg-muted bg-bg-subtle/80 border border-border rounded transition-colors duration-200 opacity-0 group-hover/devinstallcmd:opacity-100 hover:(text-fg border-border-hover) active:scale-95 focus-visible:opacity-100 focus-visible:outline-accent/70 select-none"
+              :aria-label="$t('package.get_started.copy_dev_command')"
+              @click.stop="copyDevInstallCommand"
+            >
+              <span aria-live="polite">{{
+                devInstallCopied ? $t('common.copied') : $t('common.copy')
+              }}</span>
+            </button>
+          </div>
+        </template>
```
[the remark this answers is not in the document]
wojtekmaj: This is managed via CSS

app/utils/install-command.ts
```diff
@@ -68,6 +68,117 @@ export interface InstallCommandOptions {
...
+function escapeRegExp(text: string): string {
+  return text.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
+}
+
+function hasReadmeDevInstallHint(packageName: string, readmeHtml?: string | null): boolean {
+  if (!readmeHtml) return false
+
+  const escapedName = escapeRegExp(packageName)
+  const escapedNpmName = escapeRegExp(`npm:${packageName}`)
+  const packageSpec = `(?:${escapedName}|${escapedNpmName})(?:@[\\w.-]+)?`
```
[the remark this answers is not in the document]
wojtekmaj: Eeeh I don't think it's a realistic issue

app/utils/install-command.ts
```diff
@@ -68,6 +68,117 @@ export interface InstallCommandOptions {
...
+    new RegExp(
+      String.raw`(?:npm|pnpm|yarn|bun|vlt)\s+(?:install|add|i)\s+${packageSpec}\s+(?:--save-dev|-d)`,
+      'i',
+    ),
+    // deno add -D npm:pkg
+    new RegExp(String.raw`deno\s+add\s+(?:--dev|-d)\s+${packageSpec}`, 'i'),
+  ]
+
+  return patterns.some(pattern => pattern.test(readmeHtml))
+}
```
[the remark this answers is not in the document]
wojtekmaj: Fair point, fixed

i18n/locales/en.json
```diff
@@ -199,6 +199,8 @@
       "title": "Get started",
       "pm_label": "Package manager",
       "copy_command": "Copy install command",
+      "copy_dev_command": "Copy dev install command",
+      "dev_dependency_hint": "Usually installed as a dev dependency",
```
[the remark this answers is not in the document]
wojtekmaj: Meh

app/utils/install-command.ts
```diff
@@ -68,6 +68,117 @@ export interface InstallCommandOptions {
...
+      reason: 'readme-hint',
+    }
+  }
+
+  return { recommended: false }
+}
+
+export function getDevDependencyFlag(packageManager: PackageManagerId): '-D' | '-d' {
+  return packageManager === 'bun' ? '-d' : '-D'
 }
```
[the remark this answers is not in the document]
wojtekmaj: Lol, no:
https://docs.deno.com/runtime/reference/cli/add/

> --dev
> Short flag: -D
> 
> Add the package as a dev dependency. Note: This only applies when adding to a package.json file.

app/utils/install-command.ts
```diff
@@ -68,6 +68,116 @@ export interface InstallCommandOptions {
...
+    namePart.startsWith('webpack-') ||
+    namePart.startsWith('babel-')
+  )
+}
+
+function escapeRegExp(text: string): string {
+  return text.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
+}
+
+function hasReadmeDevInstallHint(packageName: string, readmeHtml?: string | null): boolean {
```
danielroe: this would be nice to move into the api endpoint to reduce the code we run on the client

wojtekmaj: Done!

app/components/Terminal/Install.vue
```diff
@@ -131,6 +159,41 @@ const copyCreateCommand = () => copyCreate(getFullCreateCommand())
...
+            <span class="text-fg-subtle font-mono text-sm select-none shrink-0">$</span>
+            <code class="font-mono text-sm min-w-0"
+              ><span
+                v-for="(part, i) in getDevInstallPartsForPM(pm.id)"
+                :key="i"
+                :class="i === 0 ? 'text-fg' : 'text-fg-muted'"
+                >{{ i > 0 ? ' ' : '' }}{{ part }}</span
+              ></code
+            >
+            <button
```
ghostdevv: can this use the button component?

shared/utils/dev-dependency.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+function escapeRegExp(text: string): string {
+  return text.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
+}
+
+function hasReadmeDevInstallHint(packageName: string, readmeContent?: string | null): boolean {
+  if (!readmeContent) return false
+
+  const escapedName = escapeRegExp(packageName)
```
ghostdevv: I think our node is new enough for [RegExp.escape()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape)?

shared/utils/package-analysis.ts
```diff
@@ -34,6 +34,8 @@ export interface ExtendedPackageJson {
   dependencies?: Record<string, string>
   devDependencies?: Record<string, string>
   peerDependencies?: Record<string, string>
+  readme?: string
```
ghostdevv: we shouldn't rely on this field imo

ghostdevv (changes requested):

shared/utils/dev-dependency.ts
wojtekmaj: Would require changing target in tsconfig

ghostdevv: Opened https://github.com/npmx-dev/npmx.dev/pull/1439 so maybe we can use it after all :thinking:

wojtekmaj: Let's park it for now. I think that PR might spark a bit longer discussion :)

PR #1054, opened 2026-02-05
fix(ui): hide awkward empty state for weekly downloads for new packages

### LINKED ISSUE #1044: Awkward empty state for weekly downloads section
In case of new packages that do not have any weekly downloads just yet, the following chart is shown:

Image

An empty chart with no data on neither X or Y axis.

In such cases, npmjs.com just hides the chart.

wojtekmaj:
Closes #1044

Too new packages now have Weekly downloads hidden:

image

Well established packages of course still have the section:

image

### DIFF
--- a/app/components/Package/WeeklyDownloadStats.vue
+++ b/app/components/Package/WeeklyDownloadStats.vue
@@ -215,7 +215,7 @@ const config = computed(() => {
 </script>
 
 <template>
-  <div v-if="isLoadingWeeklyDownloads || hasWeeklyDownloads" class="space-y-8">
+  <div class="space-y-8">
     <CollapsibleSection id="downloads" :title="$t('package.downloads.title')">
       <template #actions>
         <ButtonBase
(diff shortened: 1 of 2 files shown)


app/components/Package/WeeklyDownloadStats.vue
```diff
@@ -85,10 +87,13 @@ const pulseColor = computed(() => {
 })
 
 const weeklyDownloads = shallowRef<WeeklyDownloadPoint[]>([])
+const isLoadingWeeklyDownloads = shallowRef(false)
```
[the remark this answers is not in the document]
wojtekmaj: Sure

serhalp:
My hunch is it's more confusing to hide this entirely. We might get future bug reports like "the downloads chart is missing". Could we keep just the heading and show "No downloads yet"? 🤔

wojtekmaj:
> My hunch is it's more confusing to hide this entirely. We might get future bug reports like "the downloads chart is missing". Could we keep just the heading and show "No downloads yet"? 🤔

Good idea, I'll do this today

app/components/Package/WeeklyDownloadStats.vue
```diff
@@ -212,44 +221,52 @@ const config = computed(() => {
...
-              <!-- Title row: date range (24px height) -->
-              <div class="h-6 flex items-center ps-3">
-                <SkeletonInline class="h-3 w-36" />
-              </div>
-              <!-- Chart area: data label left, sparkline right -->
-              <div class="aspect-[500/80] flex items-center">
-                <!-- Data label (covers ~42% width) -->
-                <div class="w-[42%] flex items-center ps-0.5">
-                  <SkeletonInline class="h-7 w-24" />
+        <template v-if="isLoadingWeeklyDownloads || hasWeeklyDownloads">
```
43081j, line 234: does this mean if we're loading the downloads but ultimately have none in the end, we'll flash the chart up then remove it?

wojtekmaj: I guess we're optimistically assuming it WILL load, as it does for 99,9% of cases

43081j:
also agree we should show a "no data" message or something instead, like a subtle overlay of some sort

i see you're already planning on changing it so will re-review once you push 👍

ghostdevv:
(marked as draft while you're making changes)

serhalp (approved):
🎉

PR #1056, opened 2026-02-05
test: add tests for error-handler

43081j:


package.json
```diff
@@ -114,6 +114,7 @@
     "@vue/test-utils": "2.4.6",
     "axe-core": "4.11.1",
     "fast-check": "4.5.3",
+    "h3": "1.15.5",
```
43081j, line 117: its a transitive dependency but we need it as a direct devDependency for sake of resolving types in tests

test/unit/server/utils/error-handler.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  it('throws a 404 with the first issue message for valibot errors', () => {
+    const schema = v.object({ name: v.pipe(v.string(), v.minLength(1, 'Name is required')) })
+
+    let valibotError: unknown
+    try {
+      v.parse(schema, { name: '' })
+    } catch (e) {
+      valibotError = e
+    }
```
ghostdevv, line 23: I really wanted to use safe parse and construct the v.ValiError yourself but it somehow feels like more work 🫠

PR #1057, opened 2026-02-05
docs: remove gitpod as a playground platform

sarah11918:
Gitpod has rebranded as [Ona](https://ona.com) and is more enterprise-oriented than individual online IDE friendly now.

There are a few options you could take here, but as an avid Gitpod user (and community hero, even!) for years who has had to get over it, my personal suggestion is just to remove it.

#### Alternatives to this PR

You could instead choose to update the entry to Ona (link and text), but it has no nice icon in the Simple Icons set. And it's something that is less likely for an individual to use as a playground anymore.

You could instead choose to just leave the link and let the redirect to happen, but that's kicking the can down the road and advertising a service that has been sunset. (i.e. close this PR! That's fine!)

An alternative platform that *is* more or less a drop-in replacement for Gitpod that you don't currently have listed is [Firebase Studio](https://firebase.studio), but I don't think it has its own icon either. (There is a distinct Firebase Studio logo separate from just Firebase.)


ghostdevv (approved):
If you wanna add Firebase studio it should be easy if you're able to find the icon - under the hood it's using unplugin icons but I'm not sure if it's using that directly or [@nuxt/icon](https://nuxt.com/modules/icon) - regardless you could add it

Regardless, let's get rid of this xD

PR #1064, opened 2026-02-06
style: improve border visibility in select package section

iiio2:
I think the border in select package section is less visible in both dark and light mode. We can slightly improve it. 

Before: The border used --border which is very faint (oklch 0.269 in dark mode, 0.8514 in light mode)

After: The border now uses --border-hover which is more visible (oklch 0.371 in dark mode, 0.715 in light mode)

before: 
border-two

after:
border-one


iiio2:
Thanks.

PR #1065, opened 2026-02-06
chore: resolve broken alert rendering in nested list

RYGRIT:
Fixed a small rendering issue here. GitHub alerts (like [[!IMPORTANT](https://github.com/orgs/community/discussions/16925)]) don't render correctly when indented inside lists. I switched it to standard bold text (**Important:**) .

image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -334,7 +334,7 @@ To add a new locale:
    cp i18n/locales/uk-UA.json lunaria/files/uk-UA.json
    ```
 
-   > **Important:**
+   > ⚠**Important:**
    > This file must be committed. Lunaria uses git history to track translation progress, so the build will fail if this file is missing.
 
 5. If the language is `right-to-left`, add `dir: 'rtl'` (see `ar-EG` in config for example)


CONTRIBUTING.md
```diff
@@ -334,7 +334,7 @@ To add a new locale:
    cp i18n/locales/uk-UA.json lunaria/files/uk-UA.json
    ```
 
-   > [!IMPORTANT]
+   > **Important:**
```
Sybren-crwk: Would it help to be even more visible to add an icon in front of it like ⚠️ ?

RYGRIT: Yes, you're right.

CONTRIBUTING.md
```diff
@@ -334,7 +334,7 @@ To add a new locale:
    cp i18n/locales/uk-UA.json lunaria/files/uk-UA.json
    ```
 
-   > [!IMPORTANT]
+   > ⚠**Important:**
    > This file must be committed. Lunaria uses git history to track translation progress, so the build will fail if this file is missing.
```
OrbisK, line 338: It might be an option to just move it to the top level here.

danielroe (approved):

PR #1067, opened 2026-02-06
fix: prevent modal from taking full width on mobile screen

iiio2:
Currently in mobile screen, modal is taking full width. I think we can improve this. 

Before:
modal-previous

After:
modal-now


ghostdevv (approved):

iiio2:
Thanks.

PR #1071, opened 2026-02-06
feat: extract button and link component, unify and improve design

essenmitsosse:
This one is quite aggressive. It does some additional work on the button and link components and replaces _all_ buttons and links on the frontpage and package-page.

The basic idea behind this is to have a fixed set of component that can be used and will automatically have the desired designs. This way, if we want to redesign buttons/links later, we don't have to hunt them down in the whole code base. This should improve DX and also ensure Consistent design and a11y across the site.

This currently supports the following:
- link, primary button, secondary button and tag (which is a small button)
- automatically marking pressed buttons/current links 
- adding keyboard shortcuts
- adding an icon
- for external links automatically adding an icon and opening in new window
- for anchor links automatically adding a link icon on hover

Additionally this does the following design changes:
- make buttons look more clickable
- add an underline to links
- use accent color for buttons and links on hover/focus*
- primary buttons are always in accent color*
- add a button group (currently only used for the internal links on the top of the package page) 
- some buttons and links might now be larger, but before their click targets where pretty small on mobile

* this removes white buttons in dark mode and black buttons in light mode 

The reason behind this is, that before a lot of clickable things where really hard to spot. This _does_ add some visual noise, so I think it might be controversial. My line of thinking here is: Make the UI easy to understand and improve visual noise as a second step, instead of the other way around.

Additionally this aligns the static tag design with the button/link tag design, but ensures it is clear what is clickable and what is not.

Open Todos after this is merged:
- other pages still have the old buttons/links and should also be migrated
- some other inputs like drop down triggers should get the same treatment
- adding Storybook to make testing these components easier

I know this is tricky to review since it does changes on multiple levels (code structure, a11y, design etc.), so let me know if it would make sense to split this.

Screenshot 2026-02-06 at 06 33 39
Screenshot 2026-02-06 at 13 25 15
Screenshot 2026-02-06 at 13 25 34

### DIFF
--- a/app/pages/package/[[org]]/[name].vue
+++ b/app/pages/package/[[org]]/[name].vue
@@ -713,7 +713,9 @@ onKeyStroke(
           </div>
 
           <!-- External links -->
-          <ul class="flex flex-wrap items-center gap-x-3 gap-y-1.5 sm:gap-4 list-none m-0 p-0 mt-3">
+          <ul
+            class="flex flex-wrap items-center gap-x-3 gap-y-1.5 sm:gap-4 list-none m-0 p-0 mt-3 text-sm"
+          >
             <li v-if="repositoryUrl">
               <LinkBase :to="repositoryUrl" :classicon="repoProviderIcon">
                 <span v-if="repoRef">
@@ -1070,7 +1072,7 @@ onKeyStroke(
             </LinkBase>
           </h2>
           <ClientOnly>
-            <div class="flex items-center gap-2">
+            <div class="flex gap-2">
               <!-- Copy readme as Markdown button -->
               <TooltipApp
                 v-if="readmeData?.md"
@@ -1145,7 +1147,7 @@ onKeyStroke(
       <div class="area-sidebar">
         <!-- Sidebar -->
         <div
-          class="sidebar-scroll sticky top-34 space-y-6 sm:space-y-8 min-w-0 overflow-y-auto pe-2.5 lg:(max-h-[calc(100dvh-8.5rem)] overscroll-contain) xl:(top-22 pt-2 max-h-[calc(100dvh-6rem)])"
+          class="sidebar-scroll sticky top-34 space-y-6 sm:space-y-8 min-w-0 overflow-y-auto pe-2.5 lg:(max-h-[calc(100dvh-8.5rem)] overscroll-contain) xl:(top-22 pt-2 max-h-[calc(100dvh-6rem)]) pt-1"
         >
           <!-- Team access controls (for scoped packages when connected) -->
           <ClientOnly>
--- a/app/pages/index.vue
+++ b/app/pages/index.vue
@@ -121,7 +121,7 @@ defineOgImageComponent('Default', {
       >
         <ul class="flex flex-wrap items-center justify-center gap-x-6 gap-y-3 list-none m-0 p-0">
           <li v-for="framework in SHOWCASED_FRAMEWORKS" :key="framework.name">
-            <LinkBase :to="packageRoute(framework.package)" class="gap-2">
+            <LinkBase :to="packageRoute(framework.package)" class="gap-2 text-sm">
               <span
                 class="w-1 h-1 rounded-full bg-accent group-hover:bg-fg transition-colors duration-200"
               />


jellydeck:
can you convert pr to draft, since it's wip

essenmitsosse:
@jellydeck from my perspective it would be ready. Unless we decide that all links and buttons should be converted first. But I'd rather merge it sooner then later.

The WIP part was meant for after this PR.

jellydeck:
few nitpicks
- currently, links look like buttons, which might cause more confusion later on. Links should be underlined, and on hover or click, the underline should move down using offset. I understand visual change but we should not mess with user expectations. tho debatable, i'd love to hear more thoughts on this.
- It’d be better to keep the button padding as it was to avoid layout and outline issues across pages.
- input `:focus-visible` styles are changed as using `*:focus-visible`. I think the previous layered approach looked better than using only an outline/ring
- also on hover we can use accent color, 
`to-accent/20`

essenmitsosse:
@jellydeck thanks for the comments.

> - currently, links look like buttons, which might cause more confusion later on. Links should be underlined, and on hover or click, the underline should move down using offset. I understand visual change but we should not mess with user expectations. tho debatable, i'd love to hear more thoughts on this.

I didn't change which links look like buttons and which are just text. I think that is another discussion to be had, if any links should look like buttons. We could also add underlines to the button-like links, but I found that overkill. But would leave that open for debate.

I think the distinction between "clicking this will do a navigation" and "clicking this will perform an action" is a bit blurry anyway. 

Text-only links do now have an underline though.

> - It’d be better to keep the button padding as it was to avoid layout and outline issues across pages.

Which change to you refer to? I tried to keep the old sizes, did I miss anything?

> - input `:focus-visible` styles are changed as using `*:focus-visible`. I think the previous layered approach looked better than using only an outline/ring

I didn't change how they looked, I just changed the query. Which layered approach did you mean? 

Some things had the style from the `.css` file _AND_ utility classes which led to inconsistencies. We can still change how `:focus` looks, this PR doesn't really have an opinion about it, is just tries to unify how it looks. Would love to see your suggestion.

> also on hover we can use accent color, `to-accent/20`

The accent color is used on hover, just in a different way. Again this is just a draft, I am not hung up on this exact design. Just tried to do something that looks more clickable and also uses the accent color more consistently. Once all components are replaced it will be easier to change these designs anyway, because there will only be two files to consider.

jellydeck:
image

By "links" I meant this (first row from top), which looks like buttons and takes the same space since it includes padding, whereas before they were text without padding. Also, showing icons for every external link can look quite repetitive but they have their purpose.

The main reason for having .css & utility classes was to override base styles applied via main.css, which is now much better after refactoring.

---

it's out of scope for this pr, but here's what i see:
visually scan the image and see the number of buttons. quite a lot, which can even overwhelm users with so many things. hence for better ux it'd be better to have simple underlined text.
image

app/components/Button/Base.vue
```diff
@@ -0,0 +1,59 @@
...
+})
+</script>
+
+<template>
+  <button
+    ref="el"
+    class="group cursor-pointer inline-flex gap-x-1 items-center justify-center font-mono border border-border rounded-md transition-all duration-200 disabled:(opacity-40 cursor-not-allowed border-transparent) aria-pressed:(bg-fg text-bg border-fg hover:enabled:(text-bg/50)) bg-gradient-to-t dark:bg-gradient-to-b"
+    :class="{
+      'text-sm px-4 py-2': variant !== 'tag',
+      'text-xs px-2 py-0.5': variant === 'tag',
```
alexdln: How about "size" prop ("none" | "sm" | "md" for now)?

app/components/Button/Base.vue
```diff
@@ -0,0 +1,59 @@
...
+      'from-fg/10 via-transparent to-transparent text-fg hover:enabled:(bg-accent/20 border-accent) focus-visible:enabled:(bg-accent/20 border-accent)':
+        variant === 'tag' || variant === 'secondary',
+      'text-black from-accent via-accent to-accent/30 hover:enabled:(bg-accent/50) focus-visible:enabled:(bg-accent/50)':
```
alexdln: "tag" take too much responsibility - and sizes, and styles, while secondary and primary only about color.

I will prefer to use size for sizes and variant for colors.

Also you now have default colors and additional here - they can probably overwrite each other in some situations. Better - 

```
'default': '...default styles',
'secondary': '...secondary styles',
etc.
```
Also add none please

app/components/Button/Base.vue
```diff
@@ -0,0 +1,59 @@
...
+        variant === 'primary',
```
alexdln, line 41: Artifact?

app/components/Button/Base.vue
```diff
@@ -0,0 +1,59 @@
...
+    }"
+    :type="props.type"
+    :disabled="disabled ? true : undefined"
```
alexdln: Attrs should work out of the box, without specifying here?

app/components/Button/Group.vue
```diff
@@ -0,0 +1,14 @@
+<script setup lang="ts">
+const props = defineProps<{
+  as?: string | Component
+}>()
+</script>
+
+<template>
+  <component
+    :is="props.as || 'nav'"
+    class="flex items-center shrink-0 ms-auto [&>*:not(:first-child)]:rounded-s-none [&>*:not(:first-child)]:border-s-0 [&>*:not(:last-child)]:rounded-e-none"
```
alexdln, line 10: How about "gap" prop?

app/components/Link/Base.vue
```diff
@@ -0,0 +1,117 @@
...
+      'text-fg hover:(no-underline text-accent) focus-visible:(no-underline text-accent) transition-colors duration-200':
+        variant === 'link',
+      'font-mono border border-border rounded-md transition-all duration-200 aria-current:(bg-fg text-bg border-fg hover:(text-bg/50)) bg-gradient-to-t dark:bg-gradient-to-b':
+        variant !== 'link',
+      'text-sm px-4 py-2': variant !== 'tag' && variant !== 'link',
+      'text-xs px-2 py-0.5': variant === 'tag',
+      'from-fg/10 via-transparent to-transparent text-fg hover:(bg-accent/20 border-accent) focus-visible:(bg-accent/20 border-accent)':
+        variant === 'tag' || variant === 'button-secondary',
+      'text-black from-accent via-accent to-accent/30 hover:(bg-accent/50) focus-visible:(bg-accent/50)':
+        variant === 'button-primary',
```
alexdln: Same as for buttons

app/components/Link/Base.vue
```diff
@@ -0,0 +1,117 @@
...
+  /></span>
+  <NuxtLink
+    v-else
+    class="group inline-flex gap-x-1 items-center justify-center"
+    :class="{
+      'underline-offset-[0.2rem] underline decoration-1 decoration-fg/50':
+        !isLinkAnchor && variant === 'link',
+      'text-fg hover:(no-underline text-accent) focus-visible:(no-underline text-accent) transition-colors duration-200':
+        variant === 'link',
+      'font-mono border border-border rounded-md transition-all duration-200 aria-current:(bg-fg text-bg border-fg hover:(text-bg/50)) bg-gradient-to-t dark:bg-gradient-to-b':
```
alexdln: currentLink with background, are you sure? Feels it would look bad, for example, in the footer

app/components/Package/MetricsBadges.vue
```diff
@@ -58,35 +58,23 @@ const typesHref = computed(() => {
...
-          :tabindex="!typesHref ? 0 : undefined"
-          class="flex items-center gap-1 px-1.5 py-0.5 font-mono text-xs rounded transition-colors duration-200 focus-visible:(outline-2 outline-accent)"
-          :class="[
+        <LinkBase v-if="typesHref" variant="tag" :to="typesHref" classicon="i-carbon-checkmark">
+          {{ $t('package.metrics.types_label') }}
+        </LinkBase>
+        <TagStatic
+          v-else
+          :variant="hasTypes && !isLoading ? 'default' : 'disabled'"
+          :tabindex="0"
```
alexdln, line 73: Should it be enabled if there is no types? (same in other TagStatic here)

app/components/Tag/Static.vue
```diff
@@ -1,12 +1,22 @@
...
+  { as: 'span', variant: 'default' },
+)
 </script>
 
 <template>
   <component
     :is="as"
-    class="inline-flex items-center px-2 py-0.5 text-xs font-mono text-fg-muted bg-bg-muted border border-border rounded"
+    class="bg-bg-muted text-fg-muted inline-flex gap-x-1 items-center justify-center font-mono border border-transparent rounded-md text-xs px-2 py-0.5"
+    :class="{ 'opacity-40': variant === 'disabled' }"
```
alexdln: I think it's worth adding styles for disabled:(...), since if `as=button` -  not to specify an additional prop

app/components/Link/Base.vue
```diff
@@ -0,0 +1,117 @@
...
+  <span
+    v-if="disabled"
+    :class="{
+      'opacity-50 inline-flex gap-x-1 items-center justify-center font-mono border border-transparent rounded-md':
+        variant !== 'link',
+      'text-sm px-4 py-2': variant !== 'tag' && variant !== 'link',
+      'text-xs px-2 py-0.5': variant === 'tag',
+      'bg-bg-muted text-fg-muted': variant === 'tag',
+      'text-bg bg-fg': variant === 'button-primary',
+      'bg-transparent text-fg': variant === 'button-secondary',
```
alexdln, line 70: Should we reuse button styles?

alexdln:
Great work, thanks

Regarding the styles here was already written about, so and I've left comments regarding the implementation.

I'm also a little confused by the hard-coded logic for determining whether it's a "new tab with a label and only these rel elements" and this set of options (perhaps a habit from being able to do <Link as={Button} /> to merge styles).

But perhaps the current implementation is for the better.

P.S. just remembered - link variants can also be useful for Button

app/components/Button/Base.vue
```diff
@@ -0,0 +1,59 @@
...
+
+    'classicon'?: string
+  }>(),
+  {
+    type: 'button',
+    variant: 'secondary',
+  },
+)
+
+const el = ref<HTMLButtonElement | null>(null)
```
OrbisK:
```suggestion
const el = useTemplateRef('el')
```

app/components/Button/Base.vue
essenmitsosse: Yeah, the `tag` name is a bit historic, because that's what the class where the original "design" came form was called. This would also separate it from the variant and allow for a small primary button.

What would `none` do?

app/components/Button/Base.vue
essenmitsosse: Splitting the classes to avoid repetition. But separating `variant` and `size` should solve this as well.

app/components/Button/Base.vue
essenmitsosse: This fixes a problem where Vue would render `disabled="false"` which is invalid. Will add a comment.

app/components/Button/Group.vue
essenmitsosse: The button group is supposed to have no visual gap. Of course we can still debate that design decision.

app/components/Link/Base.vue
essenmitsosse: Ah that was by accident. Good catch. Also it is not really used anywhere currently, which we should also change.

app/components/Package/MetricsBadges.vue
essenmitsosse: That's how it was styled before, did not question that, just replicated the logic when it is grayed out and when not.

essenmitsosse: Maybe `disabled` is a bad word here, because it has semantic meaning that wouldn't apply to the static tag anyway.

alexdln: I mean tabIndex. For disabled elements it may not be needed

app/components/Tag/Static.vue
essenmitsosse: This isn't supposed to be used as a button. Unfortunately I can't exclude `button` form the type of the `as` attribute. The `ButtonBase` is what should be used as a button, otherwise we are back to defining all buttons things manually.

app/components/Button/Group.vue
alexdln: Oh, got it, thanks

app/components/Link/Base.vue
essenmitsosse: Are you asking about code re-use or wether the links should look like buttons?

app/components/Button/Base.vue
alexdln: Sometimes we'll need to control sizes from external for some specific cases. To avoid attempts to merge and overwrite classes, it's more convenient to specify none and write the necessary ones

app/components/Button/Base.vue
alexdln: Oh, sorry, missed out what happened

app/components/Tag/Static.vue
alexdln: Got it. How about to use aria-disabled?

app/components/Button/Base.vue
essenmitsosse: I see. I think what I am trying to do is avoid opening doors for too many custom buttons. I would rather prefer to wait for a situation where this is necessary instead and then maybe add another `size`. I think what is powerful about this, it will make offering things like adjust spacing in the design more easy. If this turns out to be to taxing, we can still reverse and add `none` later. You think that could work?

app/components/Button/Base.vue
essenmitsosse: Somehow my comment here got lost. Yeah, I agree. Will split variant and size. Add somepoint we might also want other variants, in which case this will get cumbersome.

app/components/Package/MetricsBadges.vue
essenmitsosse: Ah. Yes the tabindex is necessary for the hover on the tag to work. Otherwise you can't access that info via keyboard. Not sure if it should be accessible tough. Thought I was fixing a minor bug. Should it be removed?

essenmitsosse:
@alexdln @jellydeck: My apologies. I could have sworn those links where buttons before. Of course that change was to drastic, because I never intended it. Changed it back to the link design.

I'm also not sure if we should ditch the "external" link marker here.

Screenshot 2026-02-06 at 13 15 43

app/components/Tag/Static.vue
essenmitsosse: I'd rather rename it back to `ghost` as it was before. Think that is more fitting and less confusing.

app/components/Button/Base.vue
essenmitsosse: Done. There are now two props `variant` and `size`

app/components/Link/Base.vue
essenmitsosse: `aria-current` styles don't seem to work with Uno. Will open an issue after this is closed about it. For now it's not super critical. In a footer you don't necessarily expect current links to be highlighted and the menu currently doesn't have it either — so it's not a regression.

essenmitsosse:
@alexdln I think I addresse all your comments, except for the two I left unresovled.

danielroe:
unifying button + link components is good, but I'm not a fan of the intensity of the gradient design on all buttons.

is it possible to explore a 'flatter' feel that's more consistent with what we have now?

though it might just be the white -> black gradient here that I mind - the dark themed buttons seem much nicer: 

SCR-20260206-lprt

note this, if no theme is selected:
SCR-20260206-lqkf


another point: I think we need consistent borders in the 'button bars'

https://github.com/user-attachments/assets/a75ff3ba-5c28-40f6-adef-c362c577713c

app/components/Link/Base.vue
alexdln: Both :)

jellydeck:
> @alexdln @jellydeck: My apologies. I could have sworn those links where buttons before. Of course that change was to drastic, because I never intended it. Changed it back to the link design.


no worries, since this pr was intended to be a base component for link and button keep the old styles as it is.  we can discuss & design them later to keep things managable

app/components/Package/MetricsBadges.vue
alexdln: Do we need hover for disabled tags?

essenmitsosse:
@danielroe forgot to mention the border bug in the PR message. Just wanted to make sure this is the direction we want to go to, before fixing it.

essenmitsosse:
@danielroe should I remove the gradient for now, so it doesn't block this PR? Then I can do another one where we discuss "more clickable looking buttons". Just wanted to showcase the general direction here, but that's not bound to the other changes.

I can also take out the button group for now and redo that one.

app/components/Package/MetricsBadges.vue
alexdln: It just seems to me that for enabled ones we need tabIndex and hover, but for disabled ones we don’t

essenmitsosse: They are not really disabled. They should just show that a feature is missing. Changed the variant to `ghost`, as it was before.

alexdln:
@essenmitsosse 
I'm for removing them. They're hard to make look good, especially in white shade - we'll probably need new classes with different styles for light and dark themes. Overall, it's too complex and imo should be removed for now

danielroe:
sure, let's take out the gradient for now. happy if you want to iterate on the button group - I think it's a nice UI touch.

app/components/Link/Base.vue
essenmitsosse: I am still thinking about the code reuse part. Doesn't make me super happy either.

As for links shouldn't look like buttons: I am not sure either. But I would say for the scope of this PR I would leave them as is. But I also think we should have a discussion about it.

essenmitsosse:
Removed the gradients and also the accent colors (for now). So it's closer to where we were before. Will move further redesign work to a later PR.

This also solves the issue with the button group borders.

Screenshot 2026-02-06 at 14 22 36

alexdln:
Hmm, I am not sure about the header menu... Let's revert it for now too? And how a "like" looks like (same for deps).

And the fact that npm/github/etc will open in a new tab is pretty obvious...

P.S. Will it be possible to separate decomposition and design changes or it is critical for this PR already?

danielroe:
there seems to be duplicate focus states here (double ring)

https://github.com/user-attachments/assets/84a36adb-86e9-453b-b5e6-9893a481ad19

essenmitsosse:
@danielroe what Browser is this? Is that on the preview or locally?

danielroe:
this is firefox (well, Zen) and it's the preview

essenmitsosse:
@danielroe should be solved now

app/assets/main.css
```diff
@@ -180,49 +181,20 @@ body {
...
-a:hover {
-  color: var(--accent);
-  text-decoration-color: var(--accent);
-}
-
-a:focus-visible {
-  outline: 2px solid var(--accent);
-  outline-offset: 2px;
-  border-radius: 4px;
+*:focus-visible,
```
knowler:
```suggestion
:focus-visible,
```

The universal selector would be implied here.

app/components/Button/Group.vue
```diff
@@ -0,0 +1,14 @@
+<script setup lang="ts">
+const props = defineProps<{
+  as?: string | Component
+}>()
+</script>
+
+<template>
+  <component
+    :is="props.as || 'nav'"
```
knowler: I’m a bit unsure about `<nav>` as a default. I think if we had that, then we should enforce a label.

I think it might be better to just use `<div>` as the default (is that already the default of `<component>`?) and then use `<nav>` more specifically.

app/components/Package/MetricsBadges.vue
knowler: Anything that’s focusable should also have an ARIA role, so in this case they should not have `tabindex=0`. This goes for other instances of `tabindex=0` on `<TagStatic>`.

essenmitsosse:
@knowler: Can't reply to your last comment. The `tabindex` wasn't added in the PR, so I would say it's unrelated. We should fix it then, but I'd do so in a separate PR because this one is already blown out of proportion.

app/components/Link/Base.vue
essenmitsosse: I would say both things can still be addressed at a later point. Overall this should increase code reuse and I don't think it needs to be perfect just now.

app/pages/package/[[org]]/[name].vue
```diff
@@ -651,55 +647,38 @@ onKeyStroke(
           </div>
 
           <!-- Internal navigation: Docs + Code + Compare (hidden on mobile, shown in external links instead) -->
-          <nav
+          <ButtonGroup
             v-if="resolvedVersion"
+            as="div"
```
Comment (knowler) [inline app/pages/package/[[org]]/[name].vue]: This one can be a `<nav>` still:
```suggestion
            as="nav"
```

Comment (essenmitsosse) [reply app/pages/package/[[org]]/[name].vue]: Oh my bad, that was supped to be `nav`. Good catch! It's getting late ...

essenmitsosse:
Thanks everyone for your input. I think I addressed all the issues. There still some work to be done, but I think this PR is ready to go from my point @danielroe.

danielroe:
in my mind:

- let's keep the current font-mono links as monospaced
- could we make the underline slightly muted?
- let's not fade out links when highlighted. this also seems inconsistent as it's done in some places and not in others.
- can we keep buttons in the metrics bar inline-block? (see deps box)
- I agree we need an 'external link' indication even for pure icon buttons, but it's a bit intrusive right now when it's two icons next to each other - can we temporarily revert this and then plan a good design for this?
   
   this also goes for the links in the metrics bar, which are tight anyway, and this makes them take up more space than needed

alexdln:
Sorry for taking so long to respond, or not responding at all, I'm hyper-focused again somehow...

<details>
<summary>Readme header</summary>
image

1. The trust icon has become a bit strange to me - elongated and low. It used to have a very pleasant effect, and it's not the kind of element that should be clearly highlighted
2. A like is a very familiar element to everyone and everyone is very accustomed to it as a simple heart; it does not need any additional view markers
3. The shortcut became too tight. Overall, it's a little better, but the problem of it not getting enough air has been highlighted now
4. Two icons in a button are useless. I think if we want to highlight a critical moment as a new tab, we need to create new icons or add a tooltip, but this clutters the area a bit
</details>

<details>
<summary>Home page</summary>
image

1. In the light theme, the hover overlay has become too light. Also, hiding the underline on hover is a rather unconventional practice. If you do implement an underline—I'm in favor of the format in the readme—it's muted by default, and bright on hover
2. Besides the brightness being gone, accent color is gone completely as well. In fact, this setting is practically unused now, even though it's the user's preferred experience
3. A button-style navigation bar is also a rare solution. Users are used to seeing links, but here they get the feel of buttons. I don't feel such an element fits in the header
</details>
<details>
<summary>Dialogs</summary>
image

Missed close button (same for charts and connect to cli)
</details>
<details>
<summary>Settings page</summary>
image

back button, label and menu options missed cursor-pointer hover effect
</details>
<details>
<summary>Readme sidebar and content</summary>
image

cursor-pointer on chevrons and dropdowns (also for copy buttons in readme and connect-to-cli dialog) and 2 icons in button without air

</details>
<details>
<summary>Docs and Code</summary>
image

hover on version-select - same for code folders
</details>

essenmitsosse:
@danielroe @alexdln pushed the following changes:
- links use mono font
- link underline slightly muted
- link underline not hidden on hover
- font size in footer reduced to what it was before
- links use accent color on hover (this is still problematic when no accent color is selected, but that was an issue before as well and should be its own PR)
- links that have an icon now don't show an external link indicator — this is probably not the perfect logic but works pretty reasonable for now
- reverted the deps info layout
- modal close button is fixed
- more spacing around keyboard shortcut
- reverted the main menu back to the old design while using the new components — this is not perfect but at least not a design regression

Screenshot 2026-02-07 at 00 57 16


> - The trust icon has become a bit strange to me - elongated and low. It used to have a very pleasant effect, and it's not the kind of element that should be clearly highlighted
> - A like is a very familiar element to everyone and everyone is very accustomed to it as a simple heart; it does not need any additional view markers

The old version didn't make it clear at all you could interact with it. Before non of the buttons looked clickable, now all do. I agree these buttons needs a special design, but I'd also rather make that a separate discussion. This PR aims to make all buttons look like buttons and unify their design. The PR fixed one part of the problem and I think some small regressions are fine, since we are still pre-release.

Just trying to keep the scope from creeping here.

@danielroe I think I am done … 🫩

PR #1072, opened 2026-02-06
fix(a11y): move readme after sidebar in DOM order

### LINKED ISSUE #1069: Focus order does not follow visual order in mobile layout of the package view
I’m noticing a failure of [WCAG 2.2 SC 2.4.3 Focus Order (Level A)](https://www.w3.org/WAI/WCAG22/quickref/#focus-order) for the mobile layout of [the package view](https://github.com/npmx-dev/npmx.dev/blob/6acba09af095e74e3366e3e9ff47dad5a6c8461d/app/pages/package/%5B...package%5D.vue):

> If a [web page](https://www.w3.org/TR/WCAG22/#dfn-web-page-s) can be [navigated sequentially](https://www.w3.org/TR/WCAG22/#dfn-navigated-sequentially) and the navigation sequences affect meaning or operation, focusable components receive focus in an order that preserves meaning and operability.

I encountered this issue while testing with VoiceOver on iOS 26.2.1.

The associated [“Understanding Focus Order”](https://www.w3.org/WAI/WCAG22/Understanding/focus-order.html) document states:

> For example, a screen reader user interacts with the programmatically determined reading order, while a sighted keyboard user interacts with the visual presentation of the web page. Care should be taken so that the focus order makes sense to both of these sets of users and does not appear to either of them to jump around randomly. 

“[J]umping around randomly” is an apt description of what’s taking place. While it might be easy to assume that this would not affect a person using a screen reader, the truth is [not everyone who uses a screen reader is totally blind](https://tetralogical.com/blog/2025/10/14/common-misconceptions-about-screen-readers/#misconception-1%3A-only-blind-people-use-screen-readers).

The cause of this issue is how CSS grid is being used.

https://github.com/npmx-dev/npmx.dev/blob/6acba09af095e74e3366e3e9ff47dad5a6c8461d/app/pages/package/%5B...package%5D.vue#L1279C1-L1288C2

This visual grid order does not match the DOM order. Unfortunately, there are very few cases where we can alter the visual order with grid or flexbox. With that said there are some solutions…

## Possible solutions

First off, a non-solution to this problem is to use `tabindex` with positive values. [This is not a good path to go down](https://adrianroselli.com/2014/11/dont-use-tabindex-greater-than-0.html) as it’s bound to break.

### Let the visual grid order follow the DOM order

The easiest solution is to just let the visual order match the DOM order. I think this is worth considering given the fact that the DOM order is what’s presented to those using the desktop view. One improvement to the mobile experience to go along with this change could be to make the README section collapsible, so users could easily scroll past it.

### Leverage shadow DOM composition

Shadow DOM composition using `<slot>` elements can alter the visual and focus order of the shadow host’s light DOM content (i.e. a shadow host composes two distinct trees—[unlike Vue/React, the elements remain in place in the DOM](https://knowler.dev/blog/on-slot-elements#slots-on-the-web-are-different)). To do this we would need to watch the break point with JavaScript and then change which slots each part of the layout was assigned to either using declarative `named` slot assignment (i.e. the default) or [imperative `manual` slot assignment](https://knowler.dev/blog/an-intro-to-manual-slot-assignment). I imagine this would be better for rendering performance than moving around the light DOM with Vue as the light DOM would remain in place (minimizing DOM mutations).

I will admit that this solution is more complex and I’d argue for the first solution before we consider it.

### Move the DOM around with Vue

I don’t think this is a good idea, but I don’t know Vue very well. Maybe it’s got some tricks up its sleeve.

### Future solution: `reading-flow: grid-order`

Chromium-based browsers ([version 137+](https://caniuse.com/wf-reading-flow)) now have the [`reading-flow`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/reading-flow) and [`reading-order`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/reading-order) CSS properties. These allow users to augment the reading/focus order. The `grid-order` value for the `reading-flow` property would make the visual and focus order match for this layout.

It is possible to leverage this today as a progressive enhancement using `@supports (reading-flow: grid-order)` then setting the grid template how it currently is and `reading-flow: grid-order`.

Keep in mind that [the majority of mobile screen reader users are likely using VoiceOver on iOS](https://webaim.org/projects/screenreadersurvey10/#mobilescreenreaders) if the stats are correct.

knowler:
Resolves #1069 which reports a [WCAG 2.2 SC 2.4.3 Focus Order (Level A)](https://www.w3.org/WAI/WCAG22/quickref/#focus-order) failure.

This PR moves the README after the sidebar in the DOM order, so that it matches on mobile. On tablet and desktop, this changes the tab order, so the focus will jump to the sidebar which we determined was okay, since going from one column to the next and then back is not uncalled for (e.g. it’s often the case with table of contents). The grid layout was slightly tweaked for the tablet view which just moves the sidebar up a bit so that the focus jump makes more sense.

### DIFF
--- a/app/components/Package/Skeleton.vue
+++ b/app/components/Package/Skeleton.vue
@@ -173,33 +173,6 @@
     <!-- Vulns area (empty placeholder to hold grid space) -->
     <div class="area-vulns" />
 
-    <!-- README — matches area-readme in [...name].vue -->
-    <section class="area-readme min-w-0 scroll-mt-20">
-      <div class="flex flex-wrap items-center justify-between mb-3 px-1">
-        <h2 class="text-xs font-mono text-fg-subtle uppercase tracking-wider">
-          {{ $t('package.readme.title') }}
-        </h2>
-      </div>
-      <!-- Simulated README content -->
-      <div class="space-y-4">
-        <!-- Heading -->
-        <SkeletonBlock class="h-7 w-2/3" />
-        <!-- Paragraphs -->
-        <SkeletonBlock class="h-4 w-full" />
-        <SkeletonBlock class="h-4 w-full" />
-        <SkeletonBlock class="h-4 w-4/5" />
-        <!-- Gap for section break -->
-        <SkeletonBlock class="h-6 w-1/2 mt-6" />
-        <SkeletonBlock class="h-4 w-full" />
-        <SkeletonBlock class="h-4 w-full" />
-        <SkeletonBlock class="h-4 w-3/4" />
-        <!-- Code block placeholder -->
-        <SkeletonBlock class="h-24 w-full rounded-lg mt-4" />
-        <SkeletonBlock class="h-4 w-full" />
-        <SkeletonBlock class="h-4 w-5/6" />
-      </div>
-    </section>
-
     <!-- Sidebar — matches area-sidebar in [...name].vue -->
     <div class="area-sidebar">
       <div
@@ -376,6 +349,33 @@
         </div>
       </div>
     </div>
+
+    <!-- README — matches area-readme in [...name].vue -->
+    <section class="area-readme min-w-0 scroll-mt-20">
+      <div class="flex flex-wrap items-center justify-between mb-3 px-1">
+        <h2 class="text-xs font-mono text-fg-subtle uppercase tracking-wider">
+          {{ $t('package.readme.title') }}
+        </h2>
+      </div>
+      <!-- Simulated README content -->
+      <div class="space-y-4">
+        <!-- Heading -->
+        <SkeletonBlock class="h-7 w-2/3" />
+        <!-- Paragraphs -->
+        <SkeletonBlock class="h-4 w-full" />
+        <SkeletonBlock class="h-4 w-full" />
+        <SkeletonBlock class="h-4 w-4/5" />
+        <!-- Gap for section break -->
+        <SkeletonBlock class="h-6 w-1/2 mt-6" />
+        <SkeletonBlock class="h-4 w-full" />
+        <SkeletonBlock class="h-4 w-full" />
+        <SkeletonBlock class="h-4 w-3/4" />
+        <!-- Code block placeholder -->
+        <SkeletonBlock class="h-24 w-full rounded-lg mt-4" />
+        <SkeletonBlock class="h-4 w-full" />
+        <SkeletonBlock class="h-4 w-5/6" />
+      </div>
+    </section>
   </article>
 </template>
 
@@ -390,8 +390,8 @@
     'details'
     'install'
     'vulns'
-    'readme'
-    'sidebar';
+    'sidebar'
+    'readme';
 }
 
 /* Tablet/medium: install/vulns full width, readme+sidebar side by side */
@@ -400,8 +400,8 @@
     grid-template-columns: 2fr 1fr;
     grid-template-areas:
       'details details'
-      'install install'
-      'vulns   vulns'
+      'install sidebar'
+      'vulns   sidebar'
       'readme  sidebar';
     grid-template-rows: auto auto auto auto 1fr;
   }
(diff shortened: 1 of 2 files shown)


alexdln:
I agree that the order needs to be changed ([discord](https://discord.com/channels/1464542801676206113/1464542802565140506/1469043389965013198)), but I disagree about the reason. If the problem is only with A11Y, we need to change the order of the elements in the DOM (make \<Sidebar> higher than \<Readme>).

As for the visual reordering, I think it's better to decide what experience we want to achieve (maybe for example collapsing sidebar sections on mobile devices is better)

knowler:
In https://github.com/npmx-dev/npmx.dev/pull/1072#issuecomment-3859839570, @alexdln said:

> If the problem is only with A11Y, we need to change the order of the elements in the DOM (make <Sidebar> higher than <Readme>).

If we did that it would cause the same WCAG failure for the “tablet” and desktop layouts because they would no longer match their grid order. Moving DOM specifically when the mobile breakpoint matches (i.e. using `MediaQueryList.addListener()` or whatever Vue abstraction exists) could solve that, however, that will inherently be less performant than using CSS itself and it would leave the no-JS experience inaccessible (because it wouldn’t be able to change the order).

It’s that complexity and “whack-a-mole” problem that I’m trying to address with this PR. To make matters simple: **DOM order must match the visual order for every breakpoint** (without the help of JavaScript). I’m doing so at the cost of design, but it was a design that should never have been tenable because it did not consider accessibility.

With all of that said, if we want to revisit the visual order that’s totally fine. The purpose of this PR less so to make a design change and more so to remediate the accessibility issue in a way that is most performant and universal (i.e. it works with whatever no-JS experience there is).

alexdln:
> If we did that it would cause the same WCAG failure for the “tablet” and desktop

It shouldn't. Navigation, toc, and priority elements to the right and left of main content are fairly standard practice.

The current complaint was that focus jumping abruptly to the bottom of the page, and that shouldn't happen. With the option I'm talking about, focus should move correctly, since our right panel is sticky

knowler:
In https://github.com/npmx-dev/npmx.dev/pull/1072#issuecomment-3860904778, @alexdln said:

> It shouldn't. Navigation, toc, and priority elements to the right and left of main content are fairly standard practice.

That’s a really fair point.

> The current complaint was that focus jumping abruptly to the bottom of the page, and that shouldn't happen. With the option I'm talking about, focus should move correctly, since our right panel is sticky

With your option applied, I think the desktop view is fine. The focus jump remains in view.

In the “tablet” view (I use a 14" laptop so I get this view if I open the vertical-tab-sidebar in my browser, which is pretty much always), it is a bit strange when the focus initially moves to the right column instead of the left column when traversing through the page.

At this breakpoint, the “Getting started” + “vulnerabilities” sections are both comically wide when they don’t need to be. If we make both of them only take up one column and bump the sidebar up, that improves the experience to be more like desktop (and makes it look nicer IMO).

<details><summary>Proposed grid layout change for tablet view</summary>
<p>

```diff
 @media (min-width: 1024px) {
   .package-page {
     grid-template-columns: 2fr 1fr;
     grid-template-areas:
       'header  header'
       'details details'
-      'install install'
+      'install sidebar'
-      'vulns   vulns'
+      'vulns   sidebar'
       'readme  sidebar';
     grid-template-rows: auto auto auto auto 1fr;
   }
 }
```

</p>
</details>

What do you think?

> As for the visual reordering, I think it's better to decide what experience we want to achieve (maybe for example collapsing sidebar sections on mobile devices is better)

I agree and, yes, I think collapsing the sidebar sections on mobile would be a good fix.

alexdln:
lgtm about tablet ([source](https://npmx.dev/package/robindoc/v/1.2.5))

image

ghostdevv:
(marking as draft while you work on it)

PR #1075, opened 2026-02-06
fix(i18n): update german translations

chronoB:
I ran `pnpm i18n:check:fix de-DE` and added the translations.


chronoB:
Oh I missed that they were already done overnight in https://github.com/npmx-dev/npmx.dev/pull/1033. amazing!

PR #1076, opened 2026-02-06
fix: open graph image on reddit

NandkishorJadoun:
will add more details later

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -201,6 +201,8 @@ export default defineNuxtConfig({
   ogImage: {
     defaults: {
       component: 'Default',
+      width: 1200,
+      height: 630,
     },
   },
 


ghostdevv:
will take a look at this after the vacation/recharging period :eyes:

nuxt.config.ts
```diff
@@ -201,6 +201,8 @@ export default defineNuxtConfig({
   ogImage: {
     defaults: {
       component: 'Default',
+      width: 1200,
+      height: 630,
```
ghostdevv, line 205:
```suggestion
      height: 627,
```

There is some research that says this is the most correct here: https://github.com/cloudinary-community/svelte-cloudinary/issues/159#issuecomment-3359399957

ghostdevv: @harlan-zw you're the author of the og image lib right? wdyt?

harlan-zw:
Nearly every platform renders perfectly fine with 2:1 these days and is the default for Nuxt OG Image currently.

image

I'd recommend sticking to that unless you see something dramatically bad.

alexdln:
We're also intentionally leaving plenty of space on the sides, so we don't have any critical reasons to minor changes to it for a specific platform. I'll close the PR, please reopen it if you want to continue working on it

Issue #1073, opened 2026-02-06
Expand clickable area of file list in code viewer


shuuji3:
### Issue
Hovering file row highlights the entire square area but the clickable area is only a part of them.

### Expected behavior
Users can click the same area highlighted to open a file.

### Screenshot
<img width="1488" height="426" alt="Image" src="https://github.com/user-attachments/assets/6ffc8e8f-af5e-487b-9b80-82796e54de1b" />

(https://npmx.dev/package-code/bun-types/v/1.3.8)

PR #1080, opened 2026-02-06
fix: clickable rows & mobile scrollbar overflow & dynamic line width

RYGRIT:
this pr:
+ fixes: #1073
+ The left and right scrollbars on mobile devices extend beyond the code container.
+ The previously hard-coded line number width has been changed to be dynamically calculated to prevent insufficient width.

| before  | after |
| --- | --- |
| image | image |

### DIFF
--- a/app/components/Code/Viewer.vue
+++ b/app/components/Code/Viewer.vue
@@ -16,11 +16,9 @@ const lineNumbers = computed(() => {
   return Array.from({ length: props.lines }, (_, i) => i + 1)
 })
 
-// Dynamically compute line number column width based on digit count
-const lineNumberWidth = computed(() => {
-  const digits = String(props.lines).length
-  // ch unit matches monospace digit width + 1.5rem for px-3 padding
-  return `calc(${digits}ch + 1.5rem)`
+// Used for CSS calculation of line number column width
+const lineDigits = computed(() => {
+  return String(props.lines).length
 })
 
 // Check if a line is selected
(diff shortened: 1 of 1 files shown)


RYGRIT:
I found a minor issue; I'll fix it later.

app/components/Code/Viewer.vue
```diff
@@ -55,10 +62,11 @@ watch(
 </script>
 
 <template>
-  <div class="code-viewer flex min-h-full">
+  <div class="code-viewer flex min-h-full max-w-full">
     <!-- Line numbers column -->
     <div
-      class="line-numbers shrink-0 bg-bg-subtle border-ie border-border text-end select-none"
+      class="line-numbers shrink-0 bg-bg-subtle border-r border-solid border-border text-end select-none relative"
```
danielroe:
```suggestion
      class="line-numbers shrink-0 bg-bg-subtle border-ie border-solid border-border text-end select-none relative"
```

RYGRIT: I'm sorry, I overlooked that.

app/components/Code/Viewer.vue
```diff
@@ -16,6 +16,13 @@ const lineNumbers = computed(() => {
   return Array.from({ length: props.lines }, (_, i) => i + 1)
 })
 
+// Dynamically compute line number column width based on digit count
+const lineNumberWidth = computed(() => {
```
danielroe: I think this could be more performant if we set a css variable on the entire column, and reference it in each line....

RYGRIT: Ah, I see your point. Moving the calc() logic to CSS keeps the styling concerns (like the 1.5rem padding offset) out of the JavaScript, which improves maintainability.

I'll update it to pass the digit count as a CSS variable (e.g. --line-digits) to the container, and let CSS handle the width calculation. Thanks!

PR #1083, opened 2026-02-06
test: avoid redundant html validation

essenmitsosse:
Currently the HTML validation via `@nuxtjs/html-validator` runs 4 times:
- in Vercel
- for ci: "🖥️ Browser tests"
- for ci: "♿ Accessibility audit (light)"
- for ci: "♿ Accessibility audit (dark)"

If there is any html validation error, all 4 fail, making things look worse then they are.

This changes it so, that the validation is skipped for all 3 ci tests and only runs on Vercel.

~This also removes the script `build:playwright`, because I don't think anyone will use that locally and it wasn't even used consequently in all tests.~

I think ideally the html validation shouldn't run on Vercel (so you still get a deploy preview), and run somewhere else (ci: "🖥️ Browser tests" or a separate test), but that would require to changing the Vercel build script.


nuxt.config.ts
```diff
@@ -18,7 +21,7 @@ export default defineNuxtConfig({
       })
     },
     '@unocss/nuxt',
-    '@nuxtjs/html-validator',
+    process.env.NODE_ENV === 'test' ? null : '@nuxtjs/html-validator',
```
OrbisK: I wonder if this would work:

```ts
export default defineNuxtConfig({
  modules: [
        // ...
        '@unocss/nuxt',
        '@nuxt/scripts',
        '@nuxt/a11y',
        '@nuxt/fonts',
        'nuxt-og-image',
        '@nuxt/test-utils',
        '@vite-pwa/nuxt',
        '@vueuse/nuxt',
        '@nuxtjs/i18n',
        '@nuxtjs/color-mode',
  ],
  $test: {
    modules: [
        '@nuxtjs/html-validator'
    ]
  },

OrbisK: https://nuxt.com/docs/4.x/getting-started/configuration#environment-overrides

danielroe: yes, this is the equivalent as what the PR does. however, see my comment below

danielroe:
@OrbisK's suggestion is the native way to do this

however, enabling this only in test mode will prevent it being added on prod deployments, but will also prevent it being loaded in development, where it's useful.

and it will still run for both browser + a11y tests....

let me push a change.

danielroe (approved):

essenmitsosse:
Yeah great, opt-in is even better.

PR #1085, opened 2026-02-06
fix(i18n): improve german translation

essenmitsosse:
- Removes a few anglicisms
- Fixes wrong casing for german footer

Before ("datenschutzerklärung" needs to be capitalized):
Screenshot 2026-02-06 at 11 34 23

@userquin: could you check if there was a specific reason for lower casing this word via CSS? Don't want to break any other language

edit: Also updated the tagline to the new version and made sure we don't get unnecessary line wrapping.

### DIFF
--- a/app/pages/index.vue
+++ b/app/pages/index.vue
@@ -56,7 +56,7 @@ defineOgImageComponent('Default', {
         </h1>
 
         <p
-          class="text-fg-muted text-lg sm:text-xl max-w-md mb-12 lg:mb-14 motion-safe:animate-slide-up motion-safe:animate-fill-both"
+          class="text-fg-muted text-lg sm:text-xl max-w-xl mb-12 lg:mb-14 motion-safe:animate-slide-up motion-safe:animate-fill-both"
           style="animation-delay: 0.1s"
         >
           {{ $t('tagline') }}


app/components/AppFooter.vue
```diff
@@ -20,7 +20,7 @@ const isHome = computed(() => route.name === 'index')
           </NuxtLink>
           <NuxtLink
             to="/privacy"
-            class="link-subtle font-mono text-xs min-h-11 flex items-center gap-1 lowercase"
+            class="link-subtle font-mono text-xs min-h-11 flex items-center gap-1"
```
userquin, line 23: revert this change, we need to disccuss this, we also have some uppercase usages in headings, or just await core team response

essenmitsosse: Thanks for chiming in. I would be fine with uppercase in headings for now, at least regarding German. Lowercase would also be fine if we reeeeally want it, but then the whole footer should be lowercased.

trueberryless:
Awesome enhancements.

I know it's closed already, but why did you replace all "Operationen" with "Aktionen". IMHO "Operationen" fits much better. Idk, for me all the phrases read somehow weird with "Aktionen". Would you (@essenmitsosse ) be so kind and give a little bit more reasoning for your idea regarding this? I'd like to just understand it better 👍

essenmitsosse:
@trueberryless i didn't even notice this was closed already. Saw it more as a suggestion and wanted to hear some feedback. Honestly we can just roll it back, I'm not really stuck on it.

trueberryless:
Thanks for the quick reply and feedback. I think I might revert it then - or you, whoever does the next translation ☺️

PR #1092, opened 2026-02-06
fix(ui): correct column dropdown positioning and mobile overflow

NandkishorJadoun:
### Description

Fixed a layout issue with the Column Toggle dropdown where it was causing horizontal overflow on mobile devices and incorrectly overlapping the trigger button

| Before | After |
| :--- | :--- |
| <video src="https://github.com/user-attachments/assets/b6dbcfb7-e81b-4997-9bd3-6bc8918b433f" height="400" controls></video> |  |

### DIFF
--- a/app/components/ColumnPicker.vue
+++ b/app/components/ColumnPicker.vue
@@ -87,7 +87,7 @@ function handleReset() {
         v-if="isOpen"
         ref="menuRef"
         :id="menuId"
-        class="absolute top-full right-0 sm:inset-is-auto sm:inset-ie-0 mt-2 w-60 bg-bg-subtle border border-border rounded-lg shadow-lg z-20"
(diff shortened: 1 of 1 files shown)


app/components/ColumnPicker.vue
```diff
@@ -87,7 +87,7 @@ function handleReset() {
         v-if="isOpen"
         ref="menuRef"
         :id="menuId"
-        class="absolute inset-is-0 sm:inset-is-auto sm:inset-ie-0 mt-2 w-60 bg-bg-subtle border border-border rounded-lg shadow-lg z-20"
+        class="absolute top-full right-0 sm:inset-is-auto sm:inset-ie-0 mt-2 w-60 bg-bg-subtle border border-border rounded-lg shadow-lg z-20"
```
danielroe:
```suggestion
        class="absolute top-full inset-ie-0 sm:inset-is-auto sm:inset-ie-0 mt-2 w-60 bg-bg-subtle border border-border rounded-lg shadow-lg z-20"
```

danielroe (approved):

PR #1107, opened 2026-02-06
refactor(ui): fetch results on submit instead of on type

MatteoGabriele:
- Merges the two search boxes on the homepage and header.
- Changes the update to occur on submit rather than on typing. This reduces requests to the registry API, minimizes router updates to the input value, which can cause issues like missing letters, and prevents empty responses when results are available.
- Fixes a minor accessibility issue where focus could get stuck on the hidden search form button in the header.

### DIFF
--- a/app/components/AppHeader.vue
+++ b/app/components/AppHeader.vue
@@ -25,14 +25,14 @@ const searchBoxRef = useTemplateRef('searchBoxRef')
 const { searchQuery, updateSearchQuery } = usePackageSearchQuery()
 
 const router = useRouter()
-function handleSubmitSearch() {
-  if (searchQuery.value === '') {
+function handleSubmitSearch(queryValue: string) {
+  if (queryValue === '') {
     return
   }
 
   router.replace({
     name: 'search',
-    query: { ...route.query, q: searchQuery.value },
+    query: { ...route.query, q: queryValue },
   })
 }
 


danielroe:
would you check the e2e tests? they test a lot of functionality that has changed, so I expect they need to be updated....

app/components/AppHeader.vue
```diff
@@ -21,6 +21,26 @@ const isMobile = useIsMobile()
...
+
+async function handleSearchSubmit() {
+  if (!searchQuery.value) {
+    return
+  }
+
+  await navigateTo({
+    name: 'search',
+    query: { q: searchQuery.value },
+  })
```
ghostdevv: what's the difference in nuxt between the `router.push` that was happening before, and this?

app/components/SearchBox.vue
```diff
@@ -0,0 +1,95 @@
...
+defineExpose({
+  focus,
+})
+</script>
+
+<template>
+  <search class="w-full @container">
+    <form method="GET" action="/search" class="relative" @submit.prevent="handleSubmit">
+      <label for="search-box" class="sr-only">
+        {{ $t('search.label') }}
```
ghostdevv, line 42: `search.label` is currently set to "Search npm packages" so perhaps this text should be passed in as a prop for if/when this component is reused

app/components/SearchBox.vue
```diff
@@ -0,0 +1,95 @@
...
+            /
+          </span>
+
+          <input
+            id="search-box"
+            ref="inputRef"
+            v-model.trim="searchQuery"
+            type="search"
+            name="q"
+            :placeholder="$t('search.placeholder')"
```
ghostdevv, line 64: Same here

app/components/SearchBox.vue
```diff
@@ -0,0 +1,95 @@
...
+                ? 'px-1.5 py-0.5 @md:ps-4 @md:pe-4'
+                : 'rounded-md px-2.5 @md:ps-4 @md:pe-4 py-2 text-bg bg-fg/90 hover:bg-fg! group-focus-within:bg-fg/80'
+            "
+          >
+            <span
+              class="inline-block i-carbon:search align-middle w-4 h-4 @md:me-2"
+              aria-hidden="true"
+            ></span>
+            <span class="sr-only @md:not-sr-only">
+              {{ $t('search.button') }}
```
ghostdevv, line 88: this is currently set to "search" so should be ok, assuming that the other languages are also translated plainly

app/components/AppHeader.vue
MatteoGabriele: In this case, I think either option would work. 
Previously, this component had no route handling, so I copy-pasted the function from the homepage, as they serve the same purpose. I didn't check whether it was navigateTo or router.push, since they essentially perform the same action. I believe navigateTo is a helper function that can also be used on the server side.

app/components/SearchBox.vue
MatteoGabriele: I don't think this will be used in a different context: I just moved it because it's now not only a header component but also shares the same look and functionality as the one on the homepage.
If we ever need more abstraction, there's always time to add a prop for the label and placeholder, in my opinion.

PS: The same goes for the other comments regarding the copy.

MatteoGabriele:
@ghostdevv I see that "exact" behaviour as well now.
I'm going to review that part of the logic because, now that we always press enter, that functionality might no longer be working as expected.

MatteoGabriele:
@ghostdevv I've removed the "Enter to exact" functionality since we now use Enter to actually search. What do you think?
Not sure what you meant by the throttling part, since we now work on submit instead of on-type.
thanks for your help

app/components/SearchBox.vue
ghostdevv: ahh I misread your original message

ghostdevv:
I just wonder if we can't solve the search issue without sacrificing on enter to exact or the results updating as you type... like the registry requests is annoying but can be fine tuned/cached I think and is worth it for the user experience, and the other issues should be solvable :thinking: wdyt?

> Not sure what you meant by the throttling part, since we now work on submit instead of on-type.

just as an alternative to debounce which has a nicer UX - but it only matters if we can fix the underlying problems

MatteoGabriele:
@ghostdevv I have re-implemented an update-on-type feature, but instead of using a watcher, I am now debouncing the search input update so it is not linked to the route update. It seems to be working better, avoiding the weird update loop we had before. I also re-enabled the enter key functionality.

Let me know what you think of this version.

PS: The index still uses the Enter key to prevent accidental page change while typing, which could cause loss of focus or characters, or a disruptive experience.

MatteoGabriele:
I'm actually finding issues again with the update-on-type. I will revert my last commit.

MatteoGabriele:
@ghostdevv I made a change that might resolve the issue. The code might needs some cleaning, but I wanted to hear your thoughts first.

I'm using a shared state variable, updating the URL only on submit or when the promise resolves. The input field is no longer connected to the URL preventing update loops by syncing only once on landing. 

wdyt?

MatteoGabriele:
I'm going to close this PR because the merge conflicts are getting larger and larger, and I might have found a way to fix the update-on-type issue. I also don't want to tie this decision to the input component refactor.
I'm going to open a new PR with only smaller changes to the search logic.

PR #1113, opened 2026-02-06
feat: profile page

### LINKED ISSUE #616: feat: user profiles
Allow users to set profile details and showcase them on a profile page. This is where social features related to the specific users will live, e.g likes and collections, while also showing the username, handle, any links, etc.

Will live under `/profile/<handle>` URL, e.g. `/profile/zeu.dev`

### LINKED ISSUE #923: See your own likes
Add an option to see your likes somewhere. I guess it would require building a "logged in profile page" first?

zeucapua:
Implements `dev.npmx.actor.profile` lexicon and `/profile/[handle]` page that displays the record.

When the user logs in via oauth, it will automatically create a profile record in their PDS with their domain as the display name. A user can edit their profile, which also includes the optional description and website fields. Users can also like and unlike packages from profile pages.

Fixes #616 
Fixes #923

https://github.com/user-attachments/assets/90efe87b-542f-4ea7-9465-c14472ba6381

### DIFF
--- a/lunaria/files/en-GB.json
+++ b/lunaria/files/en-GB.json
@@ -163,7 +163,7 @@
     "invite": {
       "message": "It doesn't look like they're using npmx yet. Want to tell them about it?",
       "share_button": "Share on Bluesky",
-      "compose_text": "Hey {'@'}{handle}! Have you checked out npmx.dev yet? It's a fast, modern browser for the npm registry.\nhttps://npmx.dev"
+      "compose_text": "Hey {'@'}{handle}! Have you checked out npmx.dev yet? It's a browser for the npm registry that's fast, modern, and open-source.\nhttps://npmx.dev"
     }
   },
   "package": {
--- a/lunaria/files/en-US.json
+++ b/lunaria/files/en-US.json
@@ -161,9 +161,9 @@
     "not_found": "Profile Not Found",
     "not_found_message": "The profile for {handle} could not be found.",
     "invite": {
-      "message": "It doesn't look like they're using npmx... yet. Want to tell them about it?",
+      "message": "It doesn't look like they're using npmx yet. Want to tell them about it?",
       "share_button": "Share on Bluesky",
-      "compose_text": "Hey {'@'}{handle}! Have you checked out npmx.dev yet? It's a fast, modern browser for the npm registry.\nhttps://npmx.dev"
+      "compose_text": "Hey {'@'}{handle}! Have you checked out npmx.dev yet? It's a browser for the npm registry that's fast, modern, and open-source.\nhttps://npmx.dev"
     }
   },
   "package": {


shared/types/userSession.ts
```diff
@@ -7,6 +7,11 @@ export interface UserServerSession {
     pds: string
     avatar?: string
   }
+  profile: {
```
fatfingers23, line 14: Can probably add this in to the `public` part. My thoughts there was anything in public we return to the client side and the rest we don't

server/api/social/profile/[identifier]/index.put.ts
```diff
@@ -0,0 +1,46 @@
+import { parse } from 'valibot'
+import { Client } from '@atproto/lex'
+import * as dev from '#shared/types/lexicons/dev'
+import type { NPMXProfile } from '~~/shared/types/social'
+import { ProfileEditBodySchema } from '~~/shared/schemas/social'
```
Comment (fatfingers23) [inline server/api/social/profile/[identifier]/index.put.ts]: Should  be `#shared` instead of `~~/shared` (it gets me everytime too)

server/api/auth/atproto.get.ts
```diff
@@ -279,3 +283,35 @@ async function getAvatar(did: DidString, pds: string) {
   }
   return avatar
 }
+
+async function getNpmxProfile(handle: string, authSession: OAuthSession) {
+  const client = new Client(authSession)
+
+  // get existing npmx profile OR create a new one
+  const profileUri = `at://${client.did}/dev.npmx.actor.profile/self`
+  const profileResponse = await fetch(
```
fatfingers23: nitpick (more of an fyi really): not worried about it since I think this is the only place using it. But we can also bring in the `blue.microcosm.repo.getRecordByUri` lexicon like we did for the minidoc one if we want to use atproto/lex clients with this call

server/utils/atproto/utils/profile.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  private async slingshotMiniDoc(handle: string) {
+    const miniDocKey = CACHE_PROFILE_MINI_DOC(handle)
+    const cachedMiniDoc = await this.cache.get<MiniDoc>(miniDocKey)
+
+    let miniDoc
+    if (cachedMiniDoc) {
+      miniDoc = cachedMiniDoc
+    } else {
+      const resolveUrl = `https://${SLINGSHOT_HOST}/xrpc/blue.microcosm.identity.resolveMiniDoc?identifier=${encodeURIComponent(handle)}`
```
fatfingers23: Can use the slingshot client like in `server/utils/atproto/utils/identity.ts` here as well if you'd like

server/api/auth/atproto.get.ts
zeucapua: Getting an error trying to get rpc scope for `getRecordByUri` into the oauth scopes, will leave for now and put a note to update in the future

server/utils/atproto/utils/profile.ts
zeucapua: Will add a TODO, since it's only in this one class for now.

PR #1116, opened 2026-02-06
feat: rework fontsize classes with pixels to rem

Sybren-crwk:
This PR is a proposal so please feel free to give your feedback.

I noticed that there were a bunch of text fontsize classes used as `text-[8px]`, `text-[9px]`, `text-[10px]` and `text-[11px]`.
First I wanted to change font sizes from pixels to rem for better scalability and a11y. 
Then I choose the rem size of `0.625rem`, which is `10px` taken with base is `16px`, since this was the most common used `text-[*px]` classes in the project. 
To make it easier to use and more consistent I extended the Uno-CSS theme with a new text class `text-xxs`. And applied this overall in the project.

Screenshot 2026-02-06 at 23 46 16

### DIFF
--- a/app/components/Package/DownloadAnalytics.vue
+++ b/app/components/Package/DownloadAnalytics.vue
@@ -30,8 +30,15 @@ const { accentColors, selectedAccentColor } = useAccentColor()
 const colorMode = useColorMode()
 const resolvedMode = shallowRef<'light' | 'dark'>('light')
 const rootEl = shallowRef<HTMLElement | null>(null)
+const isZoomed = shallowRef(false)
+
+function setIsZoom({ isZoom }: { isZoom: boolean }) {
+  isZoomed.value = isZoom
+}
 
 const { width } = useElementSize(rootEl)
+
+const compactNumberFormatter = useCompactNumberFormatter()
 
 onMounted(async () => {
   rootEl.value = document.documentElement
@@ -307,6 +314,33 @@ const xAxisLabel = computed(() => {
 
 const selectedGranularity = shallowRef<ChartTimeGranularity>('weekly')
 const displayedGranularity = shallowRef<ChartTimeGranularity>('weekly')
+
+const isEndDateOnPeriodEnd = computed(() => {
+  const g = selectedGranularity.value
+  if (g !== 'monthly' && g !== 'yearly') return false
+
+  const iso = String(endDate.value ?? '').slice(0, 10)
+  if (!/^\d{4}-\d{2}-\d{2}$/.test(iso)) return false
+
+  const [year, month, day] = iso.split('-').map(Number)
+  if (!year || !month || !day) return false
+
+  // Monthly: endDate is the last day of its month (UTC)
+  if (g === 'monthly') {
+    const lastDayOfMonth = new Date(Date.UTC(year, month, 0)).getUTCDate()
+    return day === lastDayOfMonth
+  }
+
+  // Yearly: endDate is the last day of the year (UTC)
+  return month === 12 && day === 31
+})
+
+const isEstimationGranularity = computed(
+  () => displayedGranularity.value === 'monthly' || displayedGranularity.value === 'yearly',
+)
+const shouldRenderEstimationOverlay = computed(
+  () => !pending.value && isEstimationGranularity.value,
+)
 
 const startDate = shallowRef<string>('') // YYYY-MM-DD
 const endDate = shallowRef<string>('') // YYYY-MM-DD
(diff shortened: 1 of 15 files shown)


uno.config.ts
```diff
@@ -40,6 +40,10 @@ export default defineConfig({
       mono: "'Geist Mono', monospace",
       sans: "'Geist', system-ui, -apple-system, sans-serif",
     },
+    text: {
+      // This should add to the existing Wind4 preset text sizes
```
alexdln:
```suggestion
    text: {
```

It's better not to add comments that don't provide new information, especially without a clear explanation of how and why

uno.config.ts
```diff
@@ -40,6 +40,10 @@ export default defineConfig({
...
+      xxs: { fontSize: '0.625rem' },
```
alexdln: Please use the 2xs format, it follows the 2xl, 3xl format - which works out of the box

app/components/Package/Versions.vue
```diff
@@ -771,7 +771,7 @@ function getTagVersions(tag: string): VersionDisplay[] {
                     <span
                       v-for="tag in v.tags"
                       :key="tag"
-                      class="text-[8px] font-semibold text-fg-subtle uppercase tracking-wide"
+                      class="text-xxs font-semibold text-fg-subtle uppercase tracking-wide"
```
alexdln: I agree that 8px should be changed to 10px, but when I looked, there were many places where it seriously disrupted the layout and the design. Let's add 3xs/4xs first to not regress in unrelated issue and then trying to increase all sizes. 

There are a lot of small elements on the package page and in the filters. If you're confident that it can definitely avoid certain dimensions and everything still looks clear, obvious, and without shifting the layout - that's great

danielroe:
it was a [deliberate decision](https://github.com/npmx-dev/npmx.dev/pull/980) to use px for font size, to allow browsers to scale font size separately from ui

but we probably shouldn't hard code px in the classes

knowler:
@danielroe that one changed the spacing (padding/margin/gap/sizes) to be pixel based. Font size is still rem based, so this is good.

alexdln:
@danielroe there seems to be a bit of confusion here - we currently use `px` for spacings (_paddings, margins, widths, etc._) and rem for font-sizes. In this issue, the places where the font-size was set to px (like `text-[8px]`) are changed to rem (_an old story, I changed this in size-picker too_)

Then it could have been changed to `text-[0.5rem]`, but the author created a new class in which the size is specified in rem (_what we expect for now from font-sizes_)

`text-[X_rem]`

`margin-[X_px]`

danielroe:
ah - my apologies!

app/components/Package/Versions.vue
Sybren-crwk: This does makes a lot of sense to add multiple xs classes. I ended up with 5 sadly enough since the pixles ranged from 8 to 11px.
```
'2xs': { fontSize: '0.6875rem' }, // 11px
'3xs': { fontSize: '0.625rem' }, // 10px
'4xs': { fontSize: '0.5625rem' }, // 9px
'5xs': { fontSize: '0.5rem' }, // 8px
```

But this will be safer and should indeed not break any small elements.

uno.config.ts
Sybren-crwk: Great format idea! I applied this.

app/components/Package/Versions.vue
alexdln: Oh goddess, I was hoping there would be 2-3 of them 😅

alexdln:
Looks great, thanks for working on it ❤️

Would you like to continue this part and as a result get rid of the smallest dimensions?

Seems like minor changes to the layout or related elements may be needed in some places, but I hope it would be a very small part

alexdln (approved):

Sybren-crwk:
> Looks great, thanks for working on it ❤️
> 
> Would you like to continue this part and as a result get rid of the smallest dimensions?
> 
> Seems like minor changes to the layout or related elements may be needed in some places, but I hope it would be a very small part

@alexdln You are welcome!
I will put it on my todo list for the coming days to look into this. I assume the goal would be to move towards just`2xs` and maybe `3xs` class and fix the layout where needed, isn't it?

alexdln:
In some places, the logic of the content structure may be lost. For example if both the text and the additional text will have the same size after update. In some places, an icon can take away attention and look visually cumbersome next to small text.

And yes, remove as much classes as possible (_honestly even 12px seems crazy to me, and here it's 8px_)

Would you create an issue or would you prefer that I create one?

PR #1125, opened 2026-02-07
fix: align deprecated version icon and text

iiio2:
I think we can improve the alignment. 

Before:
before

After:
after


danielroe (approved):

iiio2:
Thanks.

PR #1137, opened 2026-02-07
feat: remove space and `@` from user input atproto handle

shuuji3:
This change prevents common mistakes for atproto handle input. Many apps includes `@`-prefix so users can easily copy their handle with a extra character `@` and sometimes spaces in their UI.

Currenctly, I have to strictly input my handle `shuuji3.xyz`, now I can input ` @shuuji3.xyz `.

Screenshot of auth modal input is ' @shuuji3.xyz '

For reference, this is Bluesky app for `@npmx.dev` profile:

Screenshot of bluesky profile @npmx.dev including '@' char

### DIFF
--- a/app/components/Header/AuthModal.client.vue
+++ b/app/components/Header/AuthModal.client.vue
@@ -22,9 +22,22 @@ async function handleCreateAccount() {
 
 async function handleLogin() {
   if (handleInput.value) {
-    await authRedirect(handleInput.value.trim().replace('@', ''))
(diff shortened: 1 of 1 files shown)


danielroe (commented):
what about doing the validation at the input level? ie stripping invalid characters directly on paste/type?

shuuji3:
Yeah, that's better approach 👍🏻  I'll update later today.

PR #1143, opened 2026-02-07
fix: proxy external images in package readmes

### LINKED ISSUE #1138: Privacy issue: third-party images in README expose visitor data
@danielroe suggested me to open a public issue instead of a private one, so here it is.

### Description

Images referenced in a package’s README are loaded directly from third-party servers.
This allows anyone who controls such an image to collect data about visitors who view the package page on https://npmx.dev

This is not an exploit in the traditional sense, but it creates an unintended privacy leak for users browsing packages.

### What happens

When a README contains an image like:

```md
![example](https://attacker-controlled-domain.com/image.png)
```

the browser requests that image **directly from the external server** when someone opens the package page.
As a result, the image host can log detailed request metadata.

### Proof of concept

Steps to reproduce:

1. Host an image on a server you control (for example, a simple server or Vercel Serverless Function).
2. Publish an npm package whose README includes that image.
3. Open the package page on https://npmx.dev.
4. Observe incoming requests on the image server.

Example test package:

> [!WARNING]
> The PoC below uses my own server with a real request logger.
> No data is shared publicly, but requests are genuinely received and logged to demonstrate the issue

👉 https://npmx.dev/package/npmx-image-issue

When someone visits that page, the image server receives a request containing data such as:

* IP address
* User-Agent
* Accept / Accept-Language headers
* Referrer (`https://npmx.dev/`)
* Approximate location (depending on hosting provider)
* Browser and OS information

Example (redacted) request headers:

```text
accept-language: en-US,en;q=0.9
user-agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) Chrome/145
referer: https://npmx.dev/
x-forwarded-for: XXX.XXX.XXX.XX
x-vercel-ip-country: XX
x-vercel-ip-city: XXXXX
...
```

### Why this is a problem

Anyone browsing packages may unknowingly leak personal and device-related information to arbitrary third parties.

This affects **all visitors**, not just package authors.

### Expected behavior

Viewing a package page should not directly expose visitor data to untrusted third-party servers.

### Suggested solution

Proxy README images through a trusted image proxy, except for a small allowlist of well-known domains (for example, `img.shields.io`).

This is the same approach GitHub uses to prevent IP and metadata leakage:

* GitHub image proxy ("Camo"): [https://github.com/atmos/camo](https://github.com/atmos/camo)
* GitHub docs:
  [https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-anonymized-urls](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-anonymized-urls)
* Blog post:
  [https://github.blog/news-insights/product-news/proxying-user-images/](https://github.blog/news-insights/product-news/proxying-user-images/)

Let's implement a fix for this issue!

liuxiaopai-ai:
## Summary

Resolves #1138

README images from third-party servers are currently loaded directly by the visitor's browser, which exposes their IP address, User-Agent, and other request metadata to potentially malicious image hosts.

This PR introduces a server-side image proxy (similar to [GitHub's camo proxy](https://github.blog/2014-01-28-proxying-user-images/)) that routes untrusted external images through `/api/registry/image-proxy`, so visitors never make direct requests to third-party servers.

### New files
- **`server/api/registry/image-proxy/index.get.ts`** — Cached Nitro API endpoint that fetches images server-side and forwards them to the client. Includes:
  - URL validation (HTTP/HTTPS only)
  - SSRF protection (blocks localhost, private IPs, .local/.internal domains)
  - Content-Type verification (only `image/*` responses)
  - 10 MB size limit to prevent abuse
  - Security headers (`X-Content-Type-Options: nosniff`, restrictive CSP)
  - 24-hour server-side caching via `defineCachedEventHandler`

- **`shared/utils/image-proxy.ts`** — Shared utilities:
  - `isTrustedImageDomain()` — Allowlist of trusted domains (GitHub, GitLab, jsdelivr, shields.io, etc.) that don't need proxying
  - `isAllowedImageUrl()` — Validates URLs are safe to proxy (blocks SSRF vectors)
  - `toProxiedImageUrl()` — Converts external image URLs to proxied URLs; trusted domains pass through unchanged

### Modified files
- **`server/utils/readme.ts`** — Updated `resolveImageUrl()` to route resolved image URLs through `toProxiedImageUrl()`, which proxies untrusted external images while leaving trusted sources (GitHub raw, jsdelivr, shields.io) unchanged.

### Tests
- **`test/unit/shared/utils/image-proxy.spec.ts`** — 19 unit tests covering trusted domain detection, URL validation, SSRF blocking, and proxy URL generation.
- **`test/unit/server/utils/readme.spec.ts`** — 6 new integration tests verifying that README rendering correctly proxies untrusted images and passes through trusted ones.

## How it works

1. When a README is rendered, all image URLs go through `resolveImageUrl()`
2. After resolving relative paths (to GitHub raw, jsdelivr, etc.), `toProxiedImageUrl()` checks if the domain is trusted
3. Trusted domains (GitHub, shields.io, jsdelivr, etc.) → returned as-is
4. Untrusted domains → rewritten to `/api/registry/image-proxy?url=<encoded_url>`
5. The proxy endpoint fetches the image server-side and streams it to the client
6. The visitor's browser never contacts the third-party server directly


server/api/registry/image-proxy/index.get.ts
```diff
@@ -0,0 +1,119 @@
...
+      const contentLength = response.headers.get('content-length')
+      const MAX_SIZE = 10 * 1024 * 1024 // 10 MB
+      if (contentLength && Number.parseInt(contentLength, 10) > MAX_SIZE) {
+        throw createError({
+          statusCode: 413,
+          message: 'Image too large.',
+        })
+      }
+
+      const imageBuffer = await response.arrayBuffer()
```
danielroe: I wonder if we can just stream the buffer rather than loading it fully?

shared/utils/image-proxy.ts
```diff
@@ -0,0 +1,118 @@
...
+export function isTrustedImageDomain(url: string): boolean {
+  try {
+    const parsed = new URL(url)
+    const hostname = parsed.hostname.toLowerCase()
+    return TRUSTED_IMAGE_DOMAINS.some(
+      domain => hostname === domain || hostname.endsWith(`.${domain}`),
+    )
+  } catch {
+    return false
+  }
```
ghostdevv:
```suggestion
  const parsed = URL.parse(url)
  if (!parsed) return false

  return TRUSTED_IMAGE_DOMAINS.some(
    domain => domain === parsed.hostname || parsed.hostname.endsWith(`.${domain}`),
  )
```

`URL.parse` won't throw if it fails to parse

shared/utils/image-proxy.ts
```diff
@@ -0,0 +1,118 @@
...
+      hostname.startsWith('172.') ||
+      hostname.endsWith('.local') ||
+      hostname.endsWith('.internal')
+    ) {
+      return false
+    }
+    return true
+  } catch {
+    return false
+  }
```
ghostdevv:
```suggestion
  const parsed = URL.parse(url)
  if (!parsed) return false

  // Only allow HTTP and HTTPS protocols
  if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
    return false
  }

  // Block localhost / private IPs to prevent SSRF
  const hostname = parsed.hostname.toLowerCase()
  if (
    hostname === 'localhost' ||
    hostname === '127.0.0.1' ||
    hostname === '::1' ||
    hostname === '0.0.0.0' ||
    hostname.startsWith('10.') ||
    hostname.startsWith('192.168.') ||
    hostname.startsWith('172.') ||
    hostname.endsWith('.local') ||
    hostname.endsWith('.internal')
  ) {
    return false
  }

  return true
```

shared/utils/image-proxy.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  try {
+    const parsed = new URL(url)
+    if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
+      return url
+    }
+  } catch {
+    // Not an absolute URL, return as-is (relative URLs are fine)
+    return url
+  }
```
ghostdevv:
```suggestion
  const parsed = URL.parse(url)

  if (!parsed || (parsed.protocol !== 'http:' && parsed.protocol !== 'https:')) {
    return url
  }
```

ghostdevv (dismissed):

server/api/registry/image-proxy/index.get.ts
43081j: the content length can be a lie afaik, so this could be loading a very large response into memory before it even checks the size.

i think we need to investigate a couple of changes here:

- can we use `sendStream` to pipe this through instead of reading it? (what Daniel said)
- in order to do that, does that mean we can't use `defineCachedEventHandler` anymore since that presumably needs to cache the entire response?
- should we even be using `defineCachedEventHandler`? it doesn't seem sane to me to cache all images we proxy through, we should just rely on cache headers and clients for that IMO

i would pipe the stream back through, not cache it, and add an intermediate pipe that cuts it off when it reaches the `MAX_SIZE`

okineadev:
You probably need to add "Resolves https://github.com/npmx-dev/npmx.dev/security/advisories/GHSA-fhmf-6j9f-wmmq" to the PR description to close the private Security Issue at the same time.

okineadev:
We need to make sure this PR works exactly like Camo - https://github.com/atmos/camo/blob/master/server.coffee

okineadev:
Please wait 🙏

danielroe:
@okineadev can you explain more?

okineadev:
> @okineadev can you explain more?

First, we need to use the same request and response headers as Camo - https://github.com/atmos/camo/blob/master/server.coffee#L39-L57
Second, we probably need to have a list of allowed MIME content types - https://github.com/atmos/camo/blob/master/mime-types.json
Thirdly - we need to somehow encrypt all this and take care of protection against abuse

I don't know much about encryption, I asked Claude to explain how Camo does it (see https://github.com/atmos/camo#url-formats)

Like we need to make it so that only npmx can proxy only approved URLs (i.e. only from the readme of any package on npmx) and not any that the user specifies directly, as can be done now.

We have two options - either update the current code or ... just port https://github.com/atmos/camo/blob/master/server.js to this project

okineadev:
There is another similar well-commented project - https://github.com/rehypejs/camomile

It might be useful

PR #1146, opened 2026-02-07
chore: use `getEnv` environment abstraction in lunaria module

serhalp:
We'd like to avoid hardcoding platform-specific logic and assumptions in this codebase. We'll eventually add documentation for zero-config deployment to any platform.

### DIFF
--- a/modules/lunaria.ts
+++ b/modules/lunaria.ts
@@ -1,9 +1,9 @@
-import process from 'node:process'
 import { defineNuxtModule, useNuxt } from 'nuxt/kit'
 import { execSync } from 'node:child_process'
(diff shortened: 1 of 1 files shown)


modules/lunaria.ts
```diff
@@ -33,9 +33,10 @@ export default defineNuxtModule({
             cwd: nuxt.options.rootDir,
           })
         } catch (e) {
-          // do not throw when building for staging
```
serhalp, line 36: AFAICT "staging" here was a misnomer for preview environments. We don't yet have a staging env.

danielroe (approved):

Issue #1130, opened 2026-02-07
Icons disappear in Windows High Contrast Mode


knowler:
Currently when viewing with [Windows High Contrast Mode](https://support.microsoft.com/windows/change-color-contrast-in-windows-fedc744c-90ac-69df-aed5-c8a90125e696) (WHCM) enabled almost none of the icons show up. This is because WHCM overrides the colours.

<p><figure>
<img width="1425" height="693" alt="npmx.dev running in Edge with Windows High Contrast Mode enabled. The app is in the package view for nuxt and all over none of the icons for buttons and other UI elements are displayed." src="https://github.com/user-attachments/assets/86015976-2176-4232-8a4a-2a0836f66658" />
<figcaption>npmx.dev running in Microsoft Edge with Windows High Contrast Mode enabled.</figcaption>
</figure></p>

The icon classes all seem to use `color: inherit`, so we can use [`forced-color-adjust: preserve-parent-color`](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/forced-color-adjust#preserve-parent-color). We can provide this to [the `extraProperties` option](https://unocss.dev/presets/icons#extra-properties) for the icons preset in the UnoCSS config:

https://github.com/npmx-dev/npmx.dev/blob/3cf7fef97cef6be97e8f9ace628c88757a1b0c84/uno.config.ts#L24-L26

If you don’t have Windows, you can [emulate forced colors in Chromium-browsers](https://devtoolstips.org/tips/en/emulate-forced-colors/).

userquin:
I use this rules:
```css
@media (forced-colors: active) {
    .icon[class*="i-"], .icon[class^="i-"] {

        color: CanvasText !important;

        /* background-color: ButtonText; */
    }
    .button-active [class*="i-"],
    .button-active [class^="i-"] {
        background-color: Highlight !important;
    }
}
```

RYGRIT:
> I use this rules:
> 
> ```css
>  @media (forced-colors: active) {
>     .icon[class*="i-"], .icon[class^="i-"] {
> 
>         color: CanvasText !important;
> 
>         /* background-color: ButtonText; */
>     }
>     .button-active [class*="i-"],
>     .button-active [class^="i-"] {
>         background-color: Highlight !important;
>     }
> }```

Yeah, that's the rule I'm using, but I found a problem with the switch and color selection in the settings page, so I fixed it.

PR #1150, opened 2026-02-07
fix: icons disappear in Windows High Contrast Mode

RYGRIT:
fixes: #1130 

https://github.com/user-attachments/assets/1c08e3a4-9b42-4343-a743-3e8477588459


userquin:
nice ❤️

app/assets/main.css
```diff
@@ -155,6 +155,62 @@
...
+  }
+
+  button[role='switch'][aria-checked='true'] > span:last-of-type {
+    background: Highlight;
+    border-color: Highlight;
+  }
+
+  button[role='switch'][aria-checked='true'] > span:last-of-type > span {
+    background: HighlightText;
+  }
```
Sybren-crwk: Nit pick: css does support nesting now, I do think this could be nested down.

app/assets/main.css
```diff
@@ -155,6 +155,62 @@
   }
 }
 
+/* Keep icon masks visible when Windows forced colors overrides theme tokens. */
+@media (forced-colors: active) {
+  :where([class^='i-'], [class*=' i-']) {
+    forced-color-adjust: none;
+    color: CanvasText;
+  }
```
knowler: @userquin This overrides the effect of `forced-color-adjust: preserve-parent-color` that is being applied with the UnoConfig. Which do you think is better?

RYGRIT:
> Can we colocate any page specific changes with the corresponding page?

@knowler You are absolutely right! I re-tested the pages used in most scenarios and found that the default `preserve-parent-color` property handles the icons correctly in most cases.

To keep the code clean, I've temporarily commented out the global override styles in `main.css`, but kept them as a reference in case we encounter any special situations where masking/inheritance fails (e.g., in certain button states or with transparent backgrounds).

Additionally, as you suggested, I have moved all other specific styles (homepage dots, settings switches) to their respective components.

app/assets/main.css
RYGRIT: Updated to use CSS nesting in the components. Thanks for the suggestion!'❤️

danielroe (approved):

app/assets/main.css
userquin: oh, sorry, I didn't see this comment: that's why I use icon class in the selector (check my comment in linked issue), if not present I add the corresponding colors

PR #1153, opened 2026-02-07
feat: use default cursor for non-link items

### LINKED ISSUE #1074: feat: Use default cursor for non-link items
Modern frameworks like shadcn and NuxtUI and others are moving towards using regular, normal default cursor for non-link items and leave the "hand" cursor for actual links that have a href. This is also consistent with how the OS works.

This would bring it up to speed with modern practices.

For example:
Image

kalu5:
resolves #1074

### DIFF
--- a/app/assets/main.css
+++ b/app/assets/main.css
@@ -236,21 +236,6 @@ dd {
   margin: 0;
 }
 
-/* Reset button styles */
-button {
-  background: transparent;
-  border: none;
-  font: inherit;
-  padding: 0;
-}
-
-button:focus-visible,
-select:focus-visible {
-  outline: 2px solid var(--accent);
-  outline-offset: 2px;
-  border-radius: 4px;
-}
-
 /* Selection */
 ::selection {
   background-color: var(--fg-muted);


app/assets/main.css
```diff
@@ -238,7 +238,17 @@ dd {
...
+  border: none;
+  font: inherit;
+  padding: 0;
+}
+
+button:focus-visible,
+select:focus-visible {
+  outline: 2px solid var(--accent);
+  outline-offset: 2px;
+  border-radius: 4px;
```
alexdln: It might break places where we only have something like `rounded-l-md` classes

app/assets/main.css
```diff
@@ -238,7 +238,17 @@ dd {
...
+  background: transparent;
+  border: none;
+  font: inherit;
+  padding: 0;
+}
+
+button:focus-visible,
+select:focus-visible {
+  outline: 2px solid var(--accent);
+  outline-offset: 2px;
```
alexdln: We have a lot of places where the focus is done through the ring (_box shadow_). Those together might look bad, so let's remove it for now to avoid this case

And then solve a separate issue to remove the rings and focus through the outline only

app/assets/main.css
```diff
@@ -238,7 +238,17 @@ dd {
 
 /* Reset button styles */
 button {
-  cursor: pointer;
+  background: transparent;
+  border: none;
+  font: inherit;
+  padding: 0;
```
alexdln: Same as in previous comments - if you wanna keep it - please check that we don't have any cases where we relied on default styles (_like border being solid or padding always being there and only x/y being modifiable_)

app/pages/settings.vue
```diff
@@ -162,7 +162,7 @@ const setLocale: typeof setNuxti18nLocale = locale => {
                 <select
                   id="language-select"
                   :value="locale"
-                  class="w-full sm:w-auto min-w-48 bg-bg border border-border rounded-md px-3 py-2 text-sm text-fg focus-visible:outline-accent/70 cursor-pointer duration-200 transition-colors hover:border-fg-subtle"
+                  class="w-full sm:w-auto min-w-48 bg-bg border border-border rounded-md px-3 py-2 text-sm text-fg focus-visible:outline-accent/70 duration-200 transition-colors hover:border-fg-subtle"
```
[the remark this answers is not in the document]
alexdln: Not related to the current changes, we will work on this separately

alexdln:
Hi @kalu5, you've closed the issue. Is there anything I can do to help?

_The code is great btw, thanks for working on it_

kalu5:
> The code is great btw, thanks for working on it

Well, I noticed that I re-based some of the previous commits and there are still some conflicts with the main branch. I will resubmit a new PR #1216 . Thank you for your attention.

PR #1155, opened 2026-02-07
feat(i18n): cleanup unused translation keys

sybers:
This PR adds a script which removes all unused translation keys found by the `i18n:report` command from all translation files.

It also runs it, and adds a check which fails the CI if there are some unused translations (it used to be a warning only).

### DIFF
--- a/i18n/locales/mr-IN.json
+++ b/i18n/locales/mr-IN.json
@@ -172,6 +172,7 @@
       "fund": "निधी",
       "compare": "तुलना करा"
     },
+    "likes": {},
     "docs": {
       "not_available": "दस्तऐवज उपलब्ध नाहीत",
       "not_available_detail": "आम्ही या आवृत्तीसाठी दस्तऐवज तयार करू शकलो नाही."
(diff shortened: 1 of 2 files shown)


i18n/locales/ar.json
```diff
@@ -650,14 +607,6 @@
     "search_placeholder_description": "فلترة حسب الوصف…",
     "search_placeholder_keywords": "فلترة حسب الكلمات المفتاحية…",
     "search_placeholder_all": "فلتر حسب الكل أو استخدم name: desc: kw:",
-    "scope_name": "الإسم",
```
BobbieGoede, line 653: A quick search reveals this key is actually used (but not in a statically analyzable way) in `Panel.vue`

BobbieGoede (changes requested):

danielroe (dismissed):

i18n/locales/ar.json
danielroe: hmm. can we fix that?

PR #1158, opened 2026-02-07
fix(i18n): update Japanese translation

shuuji3:
Sorry, this is the second update today:
- Add three new strings
- Update the tagline that I forgot in the previous PR


shuuji3:
Thank you! 😄

PR #1163, opened 2026-02-07
feat: tweak button & link styles

jellydeck:
update the general style of buttons, group buttons and links to be consistent with our design language.

also, removed global css as Input component seems to be done.
Thanks @alexdln @essenmitsosse for feedback, enjoyed working together :)

update :
current version of npmx looks good and all the changes i wanted are addressed in one way or another.
hence closing it
thanks @alexdln @essenmitsosse for help buddies


app/assets/main.css
```diff
@@ -186,13 +186,28 @@ body {
...
+}
+
+input:focus {
+  border: 1px solid var(--accent);
+}
+
+input[type='text']:focus-visible,
+input[type='search']:focus-visible {
+  outline: 2px solid color-mix(in oklch, var(--accent) 70%, transparent);
 }
```
alexdln: Honestly, I'm not a big fan of global styles in modern apps. They always end up with overrides, !important , and other hacks. I'd prefer a basic Button and Link component that all others can build on

app/components/Button/Base.vue
```diff
@@ -32,14 +32,13 @@ defineExpose({
 <template>
   <button
     ref="el"
-    class="group cursor-pointer inline-flex gap-x-1 items-center justify-center font-mono border border-border rounded-md transition-all duration-200 disabled:(opacity-40 cursor-not-allowed border-transparent)"
+    class="group cursor-pointer inline-flex gap-x-1.5 items-center justify-center rounded-md active:rounded-4xl font-mono border border-solid [transition:all_200ms,border-radius_100ms] disabled:(opacity-40 cursor-not-allowed border-transparent) after:(content-[''] absolute inset-0 rounded-inherit pointer-events-none)"
```
alexdln: I think since we've already started enumerating for transition, it's better to list all the options individually. This reduces unexpected effects and improves stability

alexdln: Does after have any effect in this case? It's pointer-events-none, so it's essentially outside the hover analyzing zone

app/components/Button/Group.vue
```diff
@@ -7,7 +7,7 @@ const props = defineProps<{
 <template>
   <component
     :is="props.as || 'div'"
-    class="flex items-center shrink-0 ms-auto [&>*:not(:first-child)]:rounded-s-none [&>*:not(:first-child)]:border-s-0 [&>*:not(:last-child)]:rounded-e-none"
+    class="flex ms-auto rounded-2xl [&>*:not(:first-child)]:rounded-s-none [&>*:not(:first-child)]:border-s-0 [&>*:not(:last-child)]:rounded-e-none"
```
alexdln: Can you share where it is later? It feels like there's too much rounding, and it goes from slight rounding to full rounding in terms of UI basics

app/assets/main.css
jellydeck: all styles moved to buttons & links, only kept inputs for now

app/components/Button/Base.vue
jellydeck: and it's done.

jellydeck: oops, lemme fix that

app/components/Button/Group.vue
jellydeck: i was messing around, not intended effect.

app/components/Link/Base.vue
```diff
@@ -62,35 +62,41 @@ const isButtonMedium = computed(() => props.size === 'medium' && props.variant !
...
+        'font-mono text-fg hover:(decoration-accent) focus-visible:(decoration-accent text-accent) transition-colors duration-200':
+          isLink,
+        'font-mono border border-solid border-transparent rounded-md transition-all duration-200':
+          isButton,
+        'text-sm px-4 py-2': isButtonMedium,
+        'text-xs px-2 py-0.5': isButtonSmall,
+        'text-fg bg-transparent hover:(bg-fg/10 border-fg/10)': variant === 'button-secondary',
+        'text-bg bg-fg border-fg hover:(bg-fg/80)': variant === 'button-primary',
+      },
+      $attrs.class,
```
alexdln: configure as prop please

app/components/Link/Base.vue
```diff
@@ -62,35 +62,41 @@ const isButtonMedium = computed(() => props.size === 'medium' && props.variant !
   <span
     v-if="disabled"
     :class="{
-      'opacity-50 inline-flex gap-x-1 items-center justify-center font-mono border border-transparent rounded-md':
+      'opacity-50 inline-flex gap-x-1 items-center justify-center font-mono border  border-red rounded-md':
```
alexdln:
```suggestion
      'opacity-50 inline-flex gap-x-1 items-center justify-center font-mono border border-red rounded-md':
```

app/components/Link/Base.vue
jellydeck: done

thasmo:
1. focus ring is not in foreground
image

2. accent color is not used on some elements
image
image
image

3. cut off focus-ring (same on prod)
image

app/components/Button/Base.vue
```diff
@@ -32,15 +33,8 @@ defineExpose({
...
-    :class="{
-      'text-sm px-4 py-2': size === 'medium',
-      'text-xs px-2 py-0.5': size === 'small',
-      'bg-transparent text-fg hover:enabled:(bg-fg/10) focus-visible:enabled:(bg-fg/10) aria-pressed:(bg-fg text-bg border-fg hover:enabled:(bg-fg text-bg/50))':
-        variant === 'secondary',
-      'text-bg bg-fg hover:enabled:(bg-fg/50) focus-visible:enabled:(bg-fg/50) aria-pressed:(bg-fg text-bg border-fg hover:enabled:(text-bg/50))':
-        variant === 'primary',
-    }"
+    class="rounded-md outline-none group"
+    :class="props.class"
```
essenmitsosse: This is redundant. Vue does this on its own.

app/components/Button/Base.vue
```diff
@@ -54,17 +48,29 @@ defineExpose({
...
+      class="group cursor-pointer inline-flex gap-x-1.5 relative items-center justify-center rounded-md active:scale-[0.98] font-mono border border-solid border-border transition-[background-color,color,border,outline] duration-200 transition-[border-radius_100ms] after:(content-[''] absolute inset--0.5 rounded-md) outline-transparent group-focus-visible:(outline-2 outline-accent outline-offset-2)"
+      :class="{
+        'text-sm px-4 py-2': size === 'medium',
+        'text-xs px-2 py-0.5': size === 'small',
+        'text-fg bg-bg hover:(bg-fg/10 border-fg/10)': variant === 'secondary',
+        'text-bg bg-fg border-fg hover:(bg-fg/80 rounded-xl) active:rounded-xl':
+          variant === 'primary',
+        'opacity-40 cursor-not-allowed border-transparent': disabled,
+      }"
     >
```
essenmitsosse: What's the reason for the additional wrapper?

app/components/Link/Base.vue
```diff
@@ -62,35 +63,40 @@ const isButtonMedium = computed(() => props.size === 'medium' && props.variant !
   <span
     v-if="disabled"
     :class="{
-      'opacity-50 inline-flex gap-x-1 items-center justify-center font-mono border border-transparent rounded-md':
+      'opacity-50 inline-flex gap-x-1 items-center justify-center font-mono border  rounded-md':
```
essenmitsosse: Why was `border-transparent` removed? The idea was for an inactive link button not to have a border to make it visually more obvious it can not be clicked. I think now it will go to some default border color.

app/components/Link/Base.vue
```diff
@@ -62,35 +63,40 @@ const isButtonMedium = computed(() => props.size === 'medium' && props.variant !
...
-      'text-bg bg-fg hover:(bg-fg/50) focus-visible:(bg-fg/50)': variant === 'button-primary',
-    }"
+    class="group inline-flex gap-x-1 items-center justify-center rounded-sm outline-transparent active:scale-[0.98] focus-visible:(outline-2 outline-offset-2 outline-accent)"
+    :class="[
+      {
+        'underline-offset-[0.2rem] underline decoration-1 decoration-fg/30':
+          !isLinkAnchor && isLink,
+        'font-mono text-fg hover:(decoration-accent) focus-visible:(decoration-accent text-accent) transition-colors duration-200':
+          isLink,
+        'border border-solid border-border rounded-md transition-all duration-200': isButton,
```
essenmitsosse: I also initially removed `font-mono`, but @danielroe asked for it to be but back (for now) — I think changing this should be part of a broader design concept.

essenmitsosse: `border-solid` should be redundant.

app/components/Link/Base.vue
```diff
@@ -108,7 +114,7 @@ const isButtonMedium = computed(() => props.size === 'medium' && props.variant !
     />
     <kbd
       v-if="keyshortcut"
-      class="ms-2 inline-flex items-center justify-center w-4 h-4 text-xs text-fg bg-bg-muted border border-border rounded no-underline"
+      class="ms-2 inline-flex items-center justify-center w-4 h-4 text-xs group-hover:text-accent bg-bg-muted border border-border rounded no-underline"
```
essenmitsosse: I see what you want to do here. I would say though, that the keyboard short cut is not part of what you click on, therefor shouldn't be highlighted. But that is philosophical.

app/components/Button/Base.vue
jellydeck: for searchbox on home page, we pass `absolute` class directly to button which messed with internal classes. to remove that clash i took wrapper approach. so all classes passed directly to button would only handle layout cases and inside with wrapper we manage text & icons

app/components/Link/Base.vue
jellydeck: good catch, that was meant for else case. i'll fix it

app/components/Link/Base.vue
jellydeck: i added it while working on class prop, will be fixed :)

app/components/Link/Base.vue
jellydeck: i see it from this angle, if we're in text field i should make it highlight so that user don't confuse `/` as placeholder or single character added to text field. so this makes clear that it is indeed for decoration only.

app/components/Button/Base.vue
jellydeck: can you elaborate? using ` class="rounded-md outline-none group"` or props.class?
edit: so just got to know that vue merges class props automatically, still i'd consider the explicit approach coming from react/svelte world. if this is not a convention for vue developer lemme know i'll remove it from props.

app/components/Link/Base.vue
essenmitsosse: But that would only work if you hover it to check. If we are worried about it not being clear, we should more generally change the design.

app/components/Button/Base.vue
essenmitsosse: Would it work to wrap the button in the searchbox instead of adding a wrapper to every button? I wouldn't want to leak the problem of one location into the implementation of an abstract button.

app/components/Button/Base.vue
essenmitsosse: sorry, I wasn't clear. Adding `class` to props and then doing `:class="props.class"` is not necessary. Vue will automatically pass through all attributes. This is unnecessarily verbose.

jellydeck: gotcha! damn vue has good dx

app/components/Link/Base.vue
jellydeck: so currently in our design we have buttons with single character buttons, i think it's better to stick to `text-fg` as accent doesn't look contrasty enough

app/components/Button/Base.vue
jellydeck: pushed commits, lemme know what you think

app/components/Button/Base.vue
```diff
@@ -28,16 +27,15 @@ defineExpose({
 <template>
   <button
     ref="el"
-    class="group cursor-pointer gap-x-1 items-center justify-center font-mono border border-border rounded-md transition-all duration-200 disabled:(opacity-40 cursor-not-allowed border-transparent)"
+    class="group cursor-pointer inline-flex gap-x-1.5 relative items-center justify-center rounded-md active:scale-[0.98] font-mono border border-solid border-border transition-[background-color,color,border,outline] duration-200 outline-transparent focus-visible:(outline-2 outline-accent outline-offset-2)"
```
essenmitsosse: `inline-flex` needs to go, this is no handled further down.

essenmitsosse: What does `active:scale-[0.98]` do?

essenmitsosse: `border-solid` should be redundant?

essenmitsosse: Is `transition-[background-color,color,border,outline]` necessary? This will add a lot of classes, is there anything we don't want to animate?

app/components/Button/Base.vue
```diff
@@ -28,16 +27,15 @@ defineExpose({
...
       'flex': block,
       'text-sm px-4 py-2': size === 'medium',
       'text-xs px-2 py-0.5': size === 'small',
-      'bg-transparent text-fg hover:enabled:(bg-fg/10) focus-visible:enabled:(bg-fg/10) aria-pressed:(bg-fg/10 border-fg/20 hover:enabled:(bg-fg/20 text-fg/50))':
-        variant === 'secondary',
-      'text-bg bg-fg hover:enabled:(bg-fg/50) focus-visible:enabled:(bg-fg/50) aria-pressed:(bg-fg text-bg border-fg hover:enabled:(text-bg/50))':
-        variant === 'primary',
+      'text-fg bg-bg hover:(bg-fg/10 border-fg/10)': variant === 'secondary',
+      'text-bg bg-fg border-fg hover:(bg-fg/80)': variant === 'primary',
+      'opacity-40 cursor-not-allowed border-transparent': disabled,
```
essenmitsosse, line 40: Why did you remove the `disabled:(opacity-40 cursor-not-allowed border-transparent)"` implementation? This handles it in CSS only without the need to change the HTML.

app/components/Compare/FacetSelector.vue
```diff
@@ -69,7 +69,7 @@ function isCategoryNoneSelected(category: string): boolean {
           :disabled="facet.comingSoon"
           :aria-pressed="isFacetSelected(facet.id)"
           :aria-label="facet.label"
-          class="inline-flex items-center gap-1 px-1.5 py-0.5 font-mono text-xs rounded border transition-colors duration-200 focus-visible:outline-accent/70"
+          class="inline-flex items-center gap-1 px-1.5 py-0.5 font-mono text-xs rounded border transition-colors duration-200 focus-visible:(outline-2 outline-accent outline-offset-2)"
```
essenmitsosse: `focus-visible:(outline-2 outline-accent outline-offset-2)` should be redundant

app/components/Link/Base.vue
```diff
@@ -70,26 +70,24 @@ const isButtonMedium = computed(() => props.size === 'medium' && !isLink.value)
...
       'flex': block,
       'inline-flex': !block,
-      'underline-offset-[0.2rem] underline decoration-1 decoration-fg/30':
-        !isLinkAnchor && isLink && !noUnderline,
-      'justify-start font-mono text-fg hover:(decoration-accent text-accent) focus-visible:(decoration-accent text-accent) transition-colors duration-200':
+      'underline-offset-[0.2rem] underline decoration-1 decoration-fg/30': !isLinkAnchor && isLink,
+      'font-mono text-fg hover:(decoration-accent) focus-visible:(decoration-accent outline-offset-2 text-accent) transition-colors duration-200':
         isLink,
-      'justify-center font-mono border border-border rounded-md transition-all duration-200':
+      'border border-solid border-border rounded-md transition-all duration-200 focus-visible:outline-offset-2':
```
essenmitsosse: `justify-center` is necessary for full width buttons

app/components/Link/Base.vue
```diff
@@ -70,26 +70,24 @@ const isButtonMedium = computed(() => props.size === 'medium' && !isLink.value)
...
     }"
-    ><slot
-  /></span>
+  >
+    <slot />
+  </span>
   <NuxtLink
     v-else
-    class="group/link gap-x-1 items-center"
+    class="group inline-flex gap-x-1 items-center justify-center rounded-sm outline-transparent active:scale-[0.98] focus-visible:(outline-2 outline-accent)"
```
essenmitsosse: `justify-center` wouldn't work, since this is a `flex` or `inline-flex`

essenmitsosse: Why is the focus definition slightly different then the one for button?

essenmitsosse:
Screenshot 2026-02-09 at 13 28 25

There is a regression here, where the menu links/buttons have their outline back. This should be overwritten (see prod)

app/components/Button/Base.vue
jellydeck: active scale class if for press effect when we click on button, without border-solid class border weren't showing up.
`border` class just sets below
image
unlike tailwind i think we have to specify border-solid here

i was suggested on discord to only animate properties that we use
`transition-[background-color,color,border,outline]`

app/components/Button/Base.vue
jellydeck: will do

app/components/Compare/FacetSelector.vue
jellydeck: done

app/components/Tag/Static.vue
```diff
@@ -13,7 +13,7 @@ const props = withDefaults(
 <template>
   <component
     :is="as"
-    class="bg-bg-muted text-fg-muted inline-flex gap-x-1 items-center justify-center font-mono border border-transparent rounded-md text-xs px-2 py-0.5"
+    class="bg-bg-muted text-fg-muted inline-flex gap-x-1 items-center justify-center font-mono border border-transparent rounded-md text-xs px-2 py-0.5 focus-visible:(outline-2 outline-accent)"
```
essenmitsosse: This is a static element and should be able to have focus. If someone uses it as a button or link that would be wrong and it's a case that should be prevented.

app/components/Link/Base.vue
```diff
@@ -70,26 +70,24 @@ const isButtonMedium = computed(() => props.size === 'medium' && !isLink.value)
...
     :class="{
       'flex': block,
       'inline-flex': !block,
-      'underline-offset-[0.2rem] underline decoration-1 decoration-fg/30':
-        !isLinkAnchor && isLink && !noUnderline,
-      'justify-start font-mono text-fg hover:(decoration-accent text-accent) focus-visible:(decoration-accent text-accent) transition-colors duration-200':
+      'underline-offset-[0.2rem] underline decoration-1 decoration-fg/30': !isLinkAnchor && isLink,
+      'font-mono text-fg hover:(decoration-accent) focus-visible:(decoration-accent outline-offset-2 text-accent) transition-colors duration-200':
```
essenmitsosse: Is there a reason `hover:text-accent` was removed? Now the hover effect is much more subtle.

app/components/Tag/Static.vue
jellydeck: so no focus-visible i suppose, but without that it shows default outlines of browser

essenmitsosse:
Screenshot 2026-02-09 at 13 33 24

This looks very tight now

jellydeck:
> Screenshot 2026-02-09 at 13 33 24
> 
> This looks very tight now

not intended, will do fix once i'll be back on desktop

app/components/Link/Base.vue
jellydeck: underline text should use their underline as effect hence `hover:text-accent`

app/components/Button/Base.vue
essenmitsosse: But there was a border before? 🤔

essenmitsosse: I would leave the `transition-` change to a separate PR, if it does not change the design, to not bloat this one to much.

essenmitsosse: > active scale class if for press effect when we click on button,

Ahhh! got it!

app/components/Tag/Static.vue
essenmitsosse: Is there a case where this element can be focused?

jellydeck: removed focus-visible outlines since most of them shows a tooltip

jellydeck:
> 1. focus ring is not in foreground
> 
> 
> image
> 
>     2. accent color is not used on some elements
> 
> 
> image image image
> 
>     3. cut off focus-ring (same on prod)
> 
> 
> image

done 👍

app/components/Link/Base.vue
jellydeck: i used `outline-offset-2` in buttons to get a correct outline around button that's good looking and accessible

jellydeck:
> Screenshot 2026-02-09 at 13 28 25
> 
> There is a regression here, where the menu links/buttons have their outline back. This should be overwritten (see prod)

done in latest commit 👍

app/components/Header/SearchBox.vue
```diff
@@ -127,7 +128,7 @@ defineExpose({ focus })
             @blur="isSearchFocused = false"
             size="small"
           />
-          <button type="submit" class="sr-only">{{ $t('search.button') }}</button>
+          <button type="submit" class="sr-only" tabindex="-1">{{ $t('search.button') }}</button>
```
essenmitsosse: Not sure what the consequences are of adding `tabindex=-1` to a `sr-only` button.

app/components/Link/Base.vue
```diff
@@ -70,26 +70,23 @@ const isButtonMedium = computed(() => props.size === 'medium' && !isLink.value)
...
-        isButton,
+      'justify-center border border-solid border-border rounded-md  outline-offset-2': isButton,
       'text-sm px-4 py-2': isButtonMedium,
       'text-xs px-2 py-0.5': isButtonSmall,
-      'bg-transparent text-fg hover:(bg-fg/10 text-accent) focus-visible:(bg-fg/10 text-accent) aria-[current=true]:(bg-fg/10 text-accent border-fg/20 hover:enabled:(bg-fg/20 text-fg/50))':
-        variant === 'button-secondary',
-      'text-bg bg-fg hover:(bg-fg/50 text-accent) focus-visible:(bg-fg/50) aria-current:(bg-fg text-bg border-fg hover:enabled:(text-bg/50))':
-        variant === 'button-primary',
+      'text-bg bg-fg border-fg hover:(bg-fg/80)': variant === 'button-primary',
+      'text-fg bg-bg hover:(bg-fg/10 border-fg/10)': variant === 'button-secondary',
```
essenmitsosse: This removes the `aria-current` styling, I guess this might be by accident?

Also this is it on purpose to remove the focus styling?

app/components/Link/Base.vue
```diff
@@ -70,26 +70,23 @@ const isButtonMedium = computed(() => props.size === 'medium' && !isLink.value)
...
-    class="group/link gap-x-1 items-center"
+    class="group cursor-pointer gap-x-1.5 items-center rounded-sm outline-transparent active:scale-[0.98] focus-visible:(outline-2 outline-accent) transition-colors duration-200"
     :class="{
       'flex': block,
       'inline-flex': !block,
-      'underline-offset-[0.2rem] underline decoration-1 decoration-fg/30':
-        !isLinkAnchor && isLink && !noUnderline,
-      'justify-start font-mono text-fg hover:(decoration-accent text-accent) focus-visible:(decoration-accent text-accent) transition-colors duration-200':
+      'underline-offset-[0.2rem] underline decoration-1 decoration-fg/30': !isLinkAnchor && isLink,
+      'font-mono text-fg hover:(decoration-accent) focus-visible:(text-accent outline-offset-2)':
```
essenmitsosse: - `justify-start` for `isLink` is important.

app/components/Link/Base.vue
```diff
@@ -70,26 +70,23 @@ const isButtonMedium = computed(() => props.size === 'medium' && !isLink.value)
...
     v-else
-    class="group/link gap-x-1 items-center"
+    class="group cursor-pointer gap-x-1.5 items-center rounded-sm outline-transparent active:scale-[0.98] focus-visible:(outline-2 outline-accent) transition-colors duration-200"
     :class="{
       'flex': block,
       'inline-flex': !block,
-      'underline-offset-[0.2rem] underline decoration-1 decoration-fg/30':
-        !isLinkAnchor && isLink && !noUnderline,
-      'justify-start font-mono text-fg hover:(decoration-accent text-accent) focus-visible:(decoration-accent text-accent) transition-colors duration-200':
+      'underline-offset-[0.2rem] underline decoration-1 decoration-fg/30': !isLinkAnchor && isLink,
```
essenmitsosse: - `!noUnderline` is missing

app/components/Link/Base.vue
```diff
@@ -70,26 +70,23 @@ const isButtonMedium = computed(() => props.size === 'medium' && !isLink.value)
...
+      'font-mono text-fg hover:(decoration-accent) focus-visible:(text-accent outline-offset-2)':
         isLink,
-      'justify-center font-mono border border-border rounded-md transition-all duration-200':
-        isButton,
+      'justify-center border border-solid border-border rounded-md  outline-offset-2': isButton,
```
essenmitsosse: `font-mono` is missing

app/components/Button/Base.vue
essenmitsosse: This is still not resolved.

essenmitsosse:
by the way your screenshots are not showing for me

app/components/Link/Base.vue
jellydeck: my bad, i thought it was something that was going to be implemented via prop. didn't knew it was used in package code section

jellydeck:
> by the way your screenshots are not showing for me

can you quote that reply?

essenmitsosse:
> > by the way your screenshots are not showing for me
> 
> can you quote that reply?

These: https://github.com/npmx-dev/npmx.dev/pull/1163#issuecomment-3875767520 + https://github.com/npmx-dev/npmx.dev/pull/1163#issuecomment-3875772495

app/components/Link/Base.vue
essenmitsosse: Yeah, we really need Storybook to pin down what is a feature.

jellydeck:
[#1163 (comment)](https://github.com/npmx-dev/npmx.dev/pull/1163#issuecomment-3875767520) is quote reply to this comment https://github.com/npmx-dev/npmx.dev/pull/1163#issuecomment-3867137491

[#1163 (comment)](https://github.com/npmx-dev/npmx.dev/pull/1163#issuecomment-3875772495) is 
547088109-bea84a9b-6590-482e-a4e7-6379a1fc2ebf


both are quote reply, i didn't upload any ss myself.

essenmitsosse:
Ah you're right. Still weird it doesn't show the images though. But never mind.

essenmitsosse:
@jellydeck FYI: there is now also a styled link in the css in `Readme.vue`. For now we need to keep this in sync with the style from `LinkBase`.

app/components/Readme.vue
```diff
@@ -135,26 +135,26 @@ function handleClick(event: MouseEvent) {
...
-.readme :deep(a:focus-visible),
-.readme :deep(button:focus-visible) {
-  outline: 2px solid var(--accent);
-  outline-offset: 2px;
-  border-radius: 4px;
+.readme :deep(a[target='_blank']:has(img)) {
+  /* Hide icon for links containing images */
+  &::after {
+    display: none;
+  }
```
essenmitsosse, line 157: A good catch!

app/components/Button/Base.vue
```diff
@@ -28,16 +27,15 @@ defineExpose({
...
       'flex': block,
       'text-sm px-4 py-2': size === 'medium',
       'text-xs px-2 py-0.5': size === 'small',
-      'bg-transparent text-fg hover:enabled:(bg-fg/10) focus-visible:enabled:(bg-fg/10) aria-pressed:(bg-fg/10 border-fg/20 hover:enabled:(bg-fg/20 text-fg/50))':
-        variant === 'secondary',
-      'text-bg bg-fg hover:enabled:(bg-fg/50) focus-visible:enabled:(bg-fg/50) aria-pressed:(bg-fg text-bg border-fg hover:enabled:(text-bg/50))':
-        variant === 'primary',
+      'text-bg bg-fg border-fg hover:(bg-fg/80)': variant === 'primary',
+      'text-fg bg-bg hover:(bg-fg/10 border-fg/10)': variant === 'secondary',
+      'opacity-40 cursor-not-allowed border-transparent': disabled,
```
alexdln, line 40: `border-transparent` will be at the same time as `border-border`

alexdln, line 40: Formally we have a chance that button will be disabled without this prop. It would be better to style disabled buttons using `:disabled`

app/components/Button/Group.vue
```diff
@@ -7,7 +7,7 @@ const props = defineProps<{
 <template>
   <component
     :is="props.as || 'div'"
-    class="flex items-center shrink-0 ms-auto [&>*:not(:first-child)]:rounded-s-none [&>*:not(:first-child)]:border-s-0 [&>*:not(:last-child)]:rounded-e-none"
+    class="inline-flex items-stretch [&>*:not(:first-child)]:(-ms-px rounded-is-none) [&>*:not(:last-child)]:rounded-ie-none [&>*:hover]:(relative z-40) [&>*:focus-visible]:(relative z-40)"
```
alexdln: I think we need to turn off scale here as well

app/components/Header/AccountMenu.server.vue
```diff
@@ -1,10 +1,8 @@
 <template>
-  <div class="relative flex min-w-28 justify-end">
```
alexdln, line 2: min-w-28 was to reduce CLS for authorized users (same in AccountMenu)

app/components/Link/Base.vue
```diff
@@ -70,26 +70,27 @@ const isButtonMedium = computed(() => props.size === 'medium' && !isLink.value)
...
     }"
-    ><slot
-  /></span>
+  >
+    <slot />
+  </span>
   <NuxtLink
     v-else
-    class="group/link gap-x-1 items-center"
+    class="group cursor-pointer gap-x-1.5 items-center rounded-sm outline-transparent active:scale-[0.98] focus-visible:(outline-2 outline-accent) transition-colors duration-200"
```
alexdln: restore group/link please. 

Users of the component from the outside shouldn't be bothered by the presence of `group` (_and they shouldn't have artifacts if they also use group internally_)

app/components/Button/Base.vue
jellydeck: `disabled:(opacity-40 cursor-not-allowed border-transparent)`
https://github.com/npmx-dev/npmx.dev/pull/1163/changes/BASE..be463823d307f24f07a17f37866da9912c961c00#diff-4274e1791d85b6346c5a195d790469b3792aa4a49ddd2874f043048e118ae598R30

do you mean this?

app/components/Button/Group.vue
jellydeck: disabling scale effect on active and hover

app/components/Header/AccountMenu.server.vue
jellydeck: i see, i'll use same `min-w-28` then

app/components/Link/Base.vue
jellydeck: alrighty, now all link styles are under `group/link` variant

app/components/Button/Base.vue
essenmitsosse: There is also a comment about this being a regression that is marked as resolved, but it isn't.

app/components/Link/Base.vue
```diff
@@ -70,26 +70,27 @@ const isButtonMedium = computed(() => props.size === 'medium' && !isLink.value)
...
     :class="{
       'flex': block,
       'inline-flex': !block,
       'underline-offset-[0.2rem] underline decoration-1 decoration-fg/30':
         !isLinkAnchor && isLink && !noUnderline,
-      'justify-start font-mono text-fg hover:(decoration-accent text-accent) focus-visible:(decoration-accent text-accent) transition-colors duration-200':
+      'justify-start font-mono text-fg hover:(decoration-accent) focus-visible:(text-accent outline-offset-2)':
         isLink,
-      'justify-center font-mono border border-border rounded-md transition-all duration-200':
+      'justify-center font-mono border border-solid border-border rounded-md  outline-offset-2':
```
essenmitsosse: These `outline-offsets-2` should be set on the first `class` because they should apply to all links regardless of type. Also the use is inconsistant, sometimes it's always set, sometimes only on `focus-visible`.

app/components/Link/Base.vue
```diff
@@ -70,26 +70,27 @@ const isButtonMedium = computed(() => props.size === 'medium' && !isLink.value)
...
         isButton,
       'text-sm px-4 py-2': isButtonMedium,
       'text-xs px-2 py-0.5': isButtonSmall,
-      'bg-transparent text-fg hover:(bg-fg/10 text-accent) focus-visible:(bg-fg/10 text-accent) aria-[current=true]:(bg-fg/10 text-accent border-fg/20 hover:enabled:(bg-fg/20 text-fg/50))':
-        variant === 'button-secondary',
-      'text-bg bg-fg hover:(bg-fg/50 text-accent) focus-visible:(bg-fg/50) aria-current:(bg-fg text-bg border-fg hover:enabled:(text-bg/50))':
+      'text-bg bg-fg border-fg hover:(bg-fg/80) aria-[current=true]:(bg-fg/80)':
         variant === 'button-primary',
+      'text-fg bg-bg hover:(bg-fg/10 border-fg/10) aria-[current=true]:(bg-fg/10 border-fg/10)':
+        variant === 'button-secondary',
```
essenmitsosse, line 93: This doesn't change the style anymore when the button is focused. Is that on purpose or by accident? The same goes for not having the text in accent colors, for button-styled links.

app/components/Link/Base.vue
```diff
@@ -70,26 +70,27 @@ const isButtonMedium = computed(() => props.size === 'medium' && !isLink.value)
...
     v-else
-    class="group/link gap-x-1 items-center"
+    class="group/link cursor-pointer gap-x-1.5 items-center rounded-sm outline-transparent active:scale-[0.98] focus-visible:(outline-2 outline-accent) transition-colors duration-200"
     :class="{
       'flex': block,
       'inline-flex': !block,
       'underline-offset-[0.2rem] underline decoration-1 decoration-fg/30':
         !isLinkAnchor && isLink && !noUnderline,
-      'justify-start font-mono text-fg hover:(decoration-accent text-accent) focus-visible:(decoration-accent text-accent) transition-colors duration-200':
+      'justify-start font-mono text-fg hover:(decoration-accent) focus-visible:(text-accent outline-offset-2)':
```
essenmitsosse: This changes doesn't change the text color to accent on hover, but on focus. This seems inconsistant with the button styles.

app/components/Link/Base.vue
```diff
@@ -105,7 +106,7 @@ const isButtonMedium = computed(() => props.size === 'medium' && !isLink.value)
     />
     <span
       v-else-if="isLinkAnchor && isLink"
-      class="i-carbon:link size-[1em] opacity-0 group-hover/link:opacity-100 transition-opacity duration-200"
+      class="i-carbon:link size-[1em] opacity-0 group-hover/link:opacity-100 group-focus-visible/link:opacity-100 transition-opacity duration-200"
```
essenmitsosse, line 109: Good catch!

app/components/Package/ManagerSelect.vue
```diff
@@ -88,7 +88,7 @@ function handleKeydown(event: KeyboardEvent) {
   <button
     ref="triggerRef"
     type="button"
-    class="cursor-pointer flex items-center gap-1.5 px-2 py-2 font-mono text-xs text-fg-muted bg-bg-subtle border border-border-subtle border-solid rounded-md transition-colors duration-150 hover:(text-fg border-border-hover) active:scale-95 focus:border-border-hover focus-visible:outline-accent/70"
+    class="flex items-center gap-1.5 px-2 py-2 font-mono text-xs text-fg-muted bg-bg-subtle border border-border-subtle border-solid rounded-md transition-colors duration-150 hover:(text-fg border-border-hover) active:scale-95 focus:border-border-hover outline-transparent focus-visible:(outline-2 outline-accent outline-offset-2) hover:text-fg cursor-pointer"
```
essenmitsosse, line 91: There are a lot of vanilla `button` in this PR which are updated. But most of them should be migrated to `ButtonBase`, instead of manually syncing the style. It is fine like that, but that change will happen soon anyway.

app/components/Link/Base.vue
jellydeck: for button i used `offset-2` otherwise it already has `focus-visible:(outline-2 outline-accent)` . this is to create a better seperation.

app/components/Link/Base.vue
jellydeck: since it's going to be a button and on first class we have `focus-visible:(outline-2 outline-accent)`  and `outline-offset-2` on button.
it will have focus style applied as intended.

app/components/Package/ManagerSelect.vue
jellydeck: yeah, i wanted to keep scope of pr limited hence just updated to sync with current buttons style for now

app/components/Link/Base.vue
jellydeck: text-accent was just too strong, will use shade of fg color if we still need a dim variant

essenmitsosse: image

I think especially on light mode it can be too subtle now.

essenmitsosse:
image

Those two have different variants, which is barely visible now.

jellydeck:
@essenmitsosse 
looks like class specificity issue on tag. fixed now.

image

jellydeck:
current version of npmx looks good and all the changes i wanted are addressed in one way or another.
hence closing it
thanks @alexdln @essenmitsosse for help buddies

PR #1181, opened 2026-02-07
feat: add PWA web manifest (without service worker)

### LINKED ISSUE #1124: PWA
Would be awesome of npmx was a simple PWA, no need for anything fancy!

userquin:
I need to resolve a permanent redirection when requesting the webmanifest on local dev server or build + preview, works in the preview.

@danielroe  wants to add the service worker, we'll need to:
- review manifest options (adding categories)
- review denylist and runtime caching
- review ui logic to handle offline
- prerender search page: this page on page refresh when offline will redirect to home page with the query param search
- change PR title (I cannot change it, missing edit button)

**NOTE**: open pr preview in private browsing

resolves #1124


nuxt.config.ts
```diff
@@ -114,6 +114,13 @@ export default defineNuxtConfig({
     '/_v/view': { proxy: 'https://npmx.dev/_vercel/insights/view' },
     '/_v/event': { proxy: 'https://npmx.dev/_vercel/insights/event' },
     '/_v/session': { proxy: 'https://npmx.dev/_vercel/insights/session' },
+    // PWA manifest
+    '/manifest.webmanifest': {
+      headers: {
+        'Content-Type': 'application/manifest+json',
+        'Cache-Control': 'public, max-age=0, must-revalidate',
```
wojtekmaj, line 197: Why not some minimal caching, e.g.:

```suggestion
        'Cache-Control': 'public, max-age=3600',
```

?

userquin: The webmanifest should be always checked like the service worker. Once web manifest stable we can change the cache control header.

nuxt.config.ts
okineadev, line 1: There is a problem with the icon, it is square, but on other shapes of icons (for example, round ones), the edges are white, this needs to be fixed.

It would also be great to add a monochrome version of the icon to have this icon themed as well.

okineadev (changes requested):

nuxt.config.ts
userquin: We need to change the svg icons again, will be fixed once the service worker fixed (search page)

alexdln: _If you need some specific designs - let me know_

userquin: I guess we only need to add some padding at pwa assets generator configuration...

userquin: Right now we cannot check PWA icons at devtools, we'll need to use external tools: https://issues.chromium.org/issues/492211941

alexdln: https://github.com/npmx-dev/npmx.dev/blob/77b9103a6b1ec172706a6438c92e7328dabfafce/public/pwa-64x64.png

How to add them and where? The icon itself looks fine. I can make it circle or just "./" with a transparent background, but I don't understand what the problem is...

userquin: The screenshot is the PWA icons when installed at android device, android "desktop" icon, we can use https://maskable.app/ to check the paddings.

The icon comes from the [preset](https://github.com/vite-pwa/assets-generator/blob/main/src/presets/minimal-2023.ts) (will be generated when running the pwa icons script  here https://github.com/npmx-dev/npmx.dev/blob/main/package.json#L31)

For example, the paddings can be added via custom preset (here the logic https://github.com/elk-zone/elk/blob/main/scripts/generate-pwa-icons.ts#L55-L80 , I added the manual preset al elk.zone then extracted to pwa assets generarator package) => we can add some padding to the svg or we can add the padding when generating the png files.

PR #1186, opened 2026-02-08
fix: provide payload fallback json

danielroe:


serhalp:
For my own nuxt edification, would you mind explaining what this does and what this PR is solving for? 🙏🏼

danielroe:
@serhalp this provides [fallback json files](https://vercel.com/docs/build-output-api/primitives#fallback-static-file) to speed up initial fetching of payloads when the cache is cold.

it's not a nuxt thing in particular, and we do need to solve by speeding up the rendering of payloads for package pages, instead.

PR #1189, opened 2026-02-08
feat: add server-synced user preferences infrastructure (#484)

### LINKED ISSUE #484: Persisted user preferences
Now that we're able to auth into the site, we should add the ability to persist user preferences.

Basically, anything in the settings page should be stored server-side (likely in the KV store) bound to the user's unique ID.

This doesn't need to touch atproto, it just needs to privately store the data and associated ID.

gusa4grr:
## Summary

Closes #484

Persist user preferences to the server for authenticated users so settings follow the user across devices and browsers. Anonymous users continue to use localStorage only.

Previously, all settings lived in a single `useSettings` composable backed by one `npmx-settings` localStorage key. This PR splits that into two concerns:

- **User preferences** (synced) — theme, accent color, locale, search provider, keyboard shortcuts, etc. Stored in `npmx-user-preferences` localStorage + server API.
- **Local settings** (device-only) — chart filters, sidebar collapse state, connector config. Remain in `npmx-settings` localStorage, never synced.

See [`app/composables/userPreferences/README.md`](https://github.com/npmx-dev/npmx.dev/pull/1189/changes#diff-460695ce6ea82114fd1b7b0c7ad054af168bca4db9888d4c1ade012dcf8a3d1d) for a developer guide on working with preferences going forward.

---

### localStorage key migration

User-facing preferences moved from `npmx-settings` to `npmx-user-preferences`. A one-time pre-hydration migration runs automatically for existing users — no action required from end users. The migration:

- Copies `accentColorId`, `preferredBackgroundTheme`, `selectedLocale`, `relativeDates` from legacy → new key
- Does not overwrite values already present in the new key
- Cleans migrated keys from the legacy store
- Sets a `npmx-prefs-migrated` flag so it never runs again

### Composable renames

| Before                                    | After                                            |
| ----------------------------------------- | ------------------------------------------------ |
| `useSettings()`                           | Removed — use individual composables (see below) |
| `useKeyboardShortcuts()`                  | `useKeyboardShortcutsPreference()`               |
| `useRelativeDates()`                      | `useRelativeDatesPreference()`                   |
| `useInstantSearch()`                      | `useInstantSearchPreference()`                   |
| `useAccentColor()` (from useSettings)     | `useAccentColor()` (from `userPreferences/`)     |
| `useBackgroundTheme()` (from useSettings) | `useBackgroundTheme()` (from `userPreferences/`) |
| `useSearchProvider()` (from useSettings)  | `useSearchProvider()` (from `userPreferences/`)  |

---

<details>
<summary><h1>Core implementation details</h1></summary>

### User preferences store factory (`useUserPreferencesProvider`)

- Module-level singleton (`useLocalStorage` from VueUse) keyed on `npmx-user-preferences`.
- Anonymous users: localStorage only. Authenticated users: localStorage as cache, server API as source of truth.
- Provides reactive `data` ref, sync status signals (`isSyncing`, `isSynced`, `hasError`), and an `initSync()` entry point.
- Includes `__resetPreferencesForTest()` to reset singleton state between test runs.

### Server sync with debounced writes (`useUserPreferencesSync.client.ts`)

- Debounces preference changes with a **2-second** window before pushing to the server (`PUT /api/user/preferences`).
- Route guard (`router.beforeEach`) flushes pending changes on navigation so no updates are lost during SPA transitions.
- `beforeunload` listener uses `navigator.sendBeacon` to fire-and-forget the latest preferences when the tab closes.
- Both the route guard and `beforeunload` listener are registered once via module-level flags (`routeGuardRegistered`, `beforeUnloadRegistered`) to prevent duplicate handlers.
- Network errors during load set status to `'error'` (not `'idle'`) so the UI can surface problems.
- Singleton sync state ensures all consumers share the same status/error refs.

### Preference merge strategy (`preferences-merge.ts`)

- **First login (new user):** local preferences are preserved and pushed to the server, so anonymous customisations survive authentication.
- **Returning user:** server preferences take precedence; local values fill in any keys the server doesn't have yet (handles schema evolution).
- `arePreferencesEqual` compares only declared preference keys (ignores `updatedAt`).

### Non-preference local settings (`useUserLocalSettings`)

- Extracted chart-filter params, sidebar collapse state, connector config, and sparkline animation toggle into a separate `useUserLocalSettings` composable.
- Keeps the `npmx-settings` localStorage key for backward compatibility but no longer stores user preferences there.
- These settings are purely local and are not synced to the server.

### Preferences sync plugin (`preferences-sync.client.ts`)

- Registered as a Nuxt client plugin (runs before `i18n-loader` via `dependsOn`).
- Applies the stored color mode preference early (before component mount) to avoid a flash of wrong theme.
- Calls `initSync()` to kick off server sync for authenticated users at boot time.

### Theme select hydration fix

- Wrapped the theme `<select>` on the settings page in `<ClientOnly>` with a disabled fallback to prevent SSR hydration mismatches caused by `@nuxtjs/color-mode` injecting a value the server can't predict.

### Sync status indicator on settings page

- Shows cloud-upload (syncing), checkmark (synced), warning (error), or cloud (idle) icons for authenticated users.
- Uses a `<Transition>` for smooth state changes; wrapped in `<ClientOnly>` since sync state is client-only.

### Color mode preference composable (`useColorModePreference`)

- Bidirectional bridge between the user-preferences store and `@nuxtjs/color-mode`.
- `setColorMode()` updates both the preference ref and `colorMode.preference` atomically.
- `applyStoredColorMode()` is called early by the sync plugin to restore the saved mode before components mount.

### Individual preference composables (split into `userPreferences/`)

- Each preference concern is a standalone file under `app/composables/userPreferences/`:
  - `useAccentColor` — accent swatch with SSR guard for DOM mutations.
  - `useBackgroundTheme` — background shade with SSR guard.
  - `useColorModePreference` — color mode sync (see above).
  - `useInstantSearchPreference` — togglable instant search (`createSharedComposable`).
  - `useKeyboardShortcutsPreference` — togglable keyboard shortcuts with `data-kbd-shortcuts` DOM attribute sync (`createSharedComposable`).
  - `useRelativeDatesPreference` — read-only computed for relative date display.
  - `useSearchProvider` — npm/algolia toggle.
  - `useUserPreferencesState` — read-only access to the reactive preferences ref.
  - `useUserPreferencesSyncStatus` — exposes sync status signals for UI consumption.
  - `useInitUserPreferencesSync` — imperative `initSync` wrapper for the plugin.

### Legacy settings migration (`prehydrate.ts`)

- Runs inside `onPrehydrate` (before Vue hydration) to avoid layout shifts.
- One-time migration from `npmx-settings` → `npmx-user-preferences` for keys: `accentColorId`, `preferredBackgroundTheme`, `selectedLocale`, `relativeDates`.
- Does not overwrite keys that already exist in `npmx-user-preferences`.
- Cleans migrated keys from `npmx-settings` and sets a `npmx-prefs-migrated` flag so the migration never runs again.

### Shared schema (`shared/schemas/userPreferences.ts`)

- Valibot schema (`UserPreferencesSchema`) used by both client and server for validation.
- Exports types (`UserPreferences`, `AccentColorId`, `BackgroundThemeId`, `ColorModePreference`, `SearchProvider`), defaults (`DEFAULT_USER_PREFERENCES`), and the storage base key.

### localStorage hash provider (`useLocalStorageHashProvider`)

- Generic composable for localStorage-backed reactive objects with `defu` defaults merging.
- Arrays are replaced (not deep-merged) to preserve user intent.
- Used by `usePackageListPreferences` for managing list view/column/pagination settings independently of user preferences.

### API endpoints

- `GET /api/user/preferences` — returns stored preferences or `null` for first-time users (allows client to distinguish "no prefs" from "default prefs").
- `PUT /api/user/preferences` — validates body against `UserPreferencesSchema`, stores with server-managed `updatedAt` timestamp.
- `POST /api/user/preferences` — re-exports PUT handler (supports `sendBeacon` which sends POST).
- All endpoints require OAuth session; return 401 if session is invalid.

### Storage (`UserPreferencesStore`)

- Nitro `useStorage` under `user-preferences` namespace (Upstash in production, `vercel-runtime-cache` in preview, `fsLite` in dev).
- `set()` returns the stored preferences (with `updatedAt`); `merge()` delegates to `set()` after resolving the base.

### Unit tests (vitest)

- `use-keyboard-shortcuts.spec.ts` — reactivity, DOM attribute sync.
- `use-package-list-preferences.spec.ts` — defaults, localStorage merge, array replacement.
- `use-install-command.spec.ts` — updated to use `useUserPreferencesState` and reset singleton between tests.
- `user-preferences-merge.spec.ts` — first-login preservation, server-wins for returning users, schema migration fill-in, `arePreferencesEqual` edge cases.
- Component specs (`DateTime`, `FacetRow`, `HeaderConnectorModal`, `PackageSelector`) updated for new mock paths and composable names.

### E2E tests (Playwright)

- `legacy-settings-migration.spec.ts` — 10 scenarios covering migration, flag gating, DOM application, edge cases (empty/missing storage, pre-existing user prefs).
- `hydration.spec.ts` — updated localStorage key from `npmx-settings` → `npmx-user-preferences`.
- `interactions.spec.ts` — updated localStorage key for keyboard shortcuts disabled state.

</details>

---

<details>
<summary><h2>How to test</h2></summary>

### Anonymous user flow

1. Open settings, change accent color / theme / toggle keyboard shortcuts
2. Refresh — settings persist via localStorage
3. Open DevTools → Application → localStorage → verify `npmx-user-preferences` key

### Authenticated user flow

1. Sign in via Bluesky OAuth
2. Change a preference on the settings page
3. Observe sync indicator (cloud icon → syncing → checkmark)
4. Open Network tab → confirm `PUT /api/user/preferences` fires after ~2s debounce
5. Clear localStorage, refresh → preferences should reload from server

### Legacy migration

1. In DevTools, clear all localStorage
2. Manually set legacy key:
   ```js
   localStorage.setItem(
     'npmx-settings',
     JSON.stringify({
       accentColorId: 'violet',
       relativeDates: true,
     }),
   )
   ```
3. Refresh the page
4. Verify `npmx-user-preferences` now has those values, `npmx-settings` no longer contains them, and `npmx-prefs-migrated` is `'1'`

### Navigation/unload flush

1. Sign in, change a preference
2. Immediately navigate to another page (before 2s debounce) → verify the PUT still fires (route guard)
3. Change a preference and close the tab → verify `sendBeacon` fires in Network tab

### Hydration (no flash)

1. Set a non-default accent color and background theme
2. Hard refresh — no flash of default colors before the custom ones appear

### Automated tests

```bash
pnpm test:unit                    # Unit tests (merge logic, composables)
pnpm test:nuxt                    # Nuxt component tests
pnpm test:browser                 # E2E tests (legacy migration, hydration, interactions)
```

</details>

<details>
<summary><h3>ToDo list [Done ✅]</h3></summary>

- [x] add support for `searchProvider`, which was added while this MR was open
- [x] add support for `connector`,  which was added while this MR was open
- [x] fully test locally all scenarios
- [x] properly add localization (if not yet proper 😅 )
- [x] preload and hydrate settings properly (advice needed. See questions below)
- [x] handle migration from old to new LS keys
- [x] handle properly saving data to KV storage on first login/signup  

</details>
<details>
<summary><h3>Outstanding questions (mostly due to the lack of experience in Nuxt ecosystem):</h3></summary>

* [Done ✅] - Do we want to cleanup "old" KV pairs from `npmx-settings` LS, as those will now live in separate place? 
* I used `client.ts` suffix for [useUserPreferencesSync.client.ts](https://github.com/npmx-dev/npmx.dev/compare/main...gusa4grr:feat-484-persist-user-preferences?expand=1#diff-c3b948492a406d31d4be0c4888950b429f3a412764f8d16778252310d5219504) to ensure it is client-side only. Is this the standard convention to prevent server-side execution?
* with this new logic, we need to fetch user preferences immediately upon/before the app loads. what is the recommended approach for this? I created the `preferences-sync.client.ts` plugin for now

I noticed initPreferencesOnPrehydrate, which retrieves some settings from LS on the client, but it doesn't appear to support data fetching. Few other places also using `onPrehydrate`.
I am curious as we can load preferences during SSR too and can hydrate client with the preferences right away (if logged in). What files/places should I look at, any suggestions?

</details>

<details>
<summary><h3>Needs to be discussed [Done ✅]</h3></summary>

During the implementation, I identified inconsistent local storage usage across the app:

image

* `npmx-color-mode` - used for color mode. It is adjustable via the settings page, but is also evaluated by `lighthouse` and referenced in the `nuxt.config` `colorMode` property
* `npmx-list-prefs` - used in search to modify the viewing experience
*  `npmx-settings` - contains settings found in `/settings` route as well as unrelated sidebar states on package page (see image below)
* 
image

Based on the feature requirements, I decided to create the `user-preferences` schema specifically for the `/settings` page configuration. However, the current `useSettings` hook combines both user-preferences and "sidebar states".

I want to align with the team on the execution strategy before finalizing these changes.

Proposed plan:
* color-mode - retain as-is (`npmx-color-mode`), but connect to user-preference service
* npmx-list-prefs - retain as-is for now (or potentially merge into npmx-settings). We can add this to the user-preferences service as a follow-up and remove this LS key then.
* npmx-settings - remove all user-preference related keys (leaving only the sidebar state). Merge with npmx-settings or leave as-is
* npmx-user-preferences - new LS key for user preferences. This will serve as the source of truth for anonymous users. For logged-in users, it will sync with the server, where the server state takes precedence on load

The solution I am drafting centralizes these options into a single user-preference service. However, if we include items outside of /settings, we need to consider:
* making it clear to the user what parts are persistent settings
* defining the correct "sync" lifecycle

</details>


43081j:
haven't reviewed the code yet, but on the proposed changes:

- color-mode - i agree leave it where it is but possibly sync it
- npmx-list-prefs - keep it clientside since it is likely to be changed often between views
- npmx-settings - agree
- npmx-user-preferences - agree

The usual pattern is this:

- Anything that persists across pages, sync it to the backend
- Anything that is one-time, temporary, or often toggled (like search filters), do not sync

we should then end up in a state where our prefs save to local storage and take effect immediately, but meanwhile get synced to the server in the background.

i think we probably also need a toggle somewhere to disable sync, in case you want different prefs per machine. same way chrome works today

gusa4grr:
@43081j 👋🏻 

Updated the MR description and pushed one more commit.
I'll leave it as a draft for now, as I haven't validated all the scenarios, but the core things are working fine!

if you have time - please look throught the code 🙂 Much appreciated!

I will try to find time during the work week to finalise this, as not much is left 🤞🏻

app/components/CollapsibleSection.vue
```diff
@@ -15,16 +14,17 @@ const props = withDefaults(defineProps<Props>(), {
...
 const buttonId = `${props.id}-collapsible-button`
 const contentId = `${props.id}-collapsible-content`
+const headingId = `${props.id}-heading`
 
 const isOpen = shallowRef(true)
 
 onPrehydrate(() => {
-  const settings = JSON.parse(localStorage.getItem('npmx-settings') || '{}')
-  const collapsed: string[] = settings?.sidebar?.collapsed || []
+  const sidebar = JSON.parse(localStorage.getItem('npmx-sidebar-preferences') || '{}')
```
43081j: can we extract this `npmx-sidebar-preferences` and the other keys to constants somewhere?

app/composables/useUserPreferences.ts
```diff
@@ -0,0 +1,133 @@
...
+    isSyncing: provider.isSyncing,
+    isSynced: provider.isSynced,
+    hasError: provider.hasError,
+    syncError: provider.syncError,
+    lastSyncedAt: provider.lastSyncedAt,
+    initSync: provider.initSync,
+  }
+}
+
+export function useRelativeDates() {
```
43081j: we should probably split these up into `app/composables/preferences/whatever.ts`

e.g. `app/composables/preferences/useRelativeDates.ts`

to follow convention of the rest of the repo

app/components/CollapsibleSection.vue
gusa4grr: it won't work, as it can't depend on things outside

https://nuxt.com/docs/4.x/api/composables/on-prehydrate#parameters

> A function (or stringified function) to run before Nuxt hydrates. It will be stringified and inlined in the HTML. Should not have external dependencies or reference variables outside the callback. Runs before Nuxt runtime initializes, so it should not rely on Nuxt or Vue context.

43081j: sad times. fair enough then 👍

app/components/Settings/BgThemePicker.vue
```diff
@@ -2,8 +2,8 @@
 const { backgroundThemes, selectedBackgroundTheme, setBackgroundTheme } = useBackgroundTheme()
 
 onPrehydrate(el => {
-  const settings = JSON.parse(localStorage.getItem('npmx-settings') || '{}')
-  const id = settings.preferredBackgroundTheme
+  const preferences = JSON.parse(localStorage.getItem('npmx-user-preferences') || '{}')
```
43081j: here too we should try/catch the `parse` call and probably just fall back to `{}` like we do if it isn't set

app/composables/userPreferences/useRelativeDates.ts
```diff
@@ -0,0 +1,4 @@
+export function useRelativeDates() {
```
43081j: maybe this should also be called `useRelativeDatesPreference`?

since nuxt magic will make it global, so the name might end up ambiguous in other contexts

app/composables/useUserPreferencesProvider.ts
```diff
@@ -0,0 +1,110 @@
...
+
+import type { RemovableRef } from '@vueuse/core'
+import { useLocalStorage } from '@vueuse/core'
+import { DEFAULT_USER_PREFERENCES, type UserPreferences } from '#shared/schemas/userPreferences'
+
+const STORAGE_KEY = 'npmx-user-preferences'
+
+function arePreferencesEqual(a: UserPreferences, b: UserPreferences): boolean {
+  const keys = Object.keys(DEFAULT_USER_PREFERENCES) as (keyof typeof DEFAULT_USER_PREFERENCES)[]
+  return keys.every(key => a[key] === b[key])
```
43081j: what if we ever add or remove a key? should we also check the `length` here?

gusa4grr:
here is the GD link with the recording on preview, how it works
https://drive.google.com/file/d/1C7Lq_QyvwzyN7qjZfnjFyVga6nUVvu_C/view?usp=sharing

app/composables/usePackageListPreferences.ts
43081j: maybe this isnt necessary anymore since `isHydrated` doesn't exist?

43081j:
overall looks good to me

can you check each of the coderabbit comments?

and we could do with a 2nd maintainer review on this - ill ping the other maintainers

app/composables/usePackageListPreferences.ts
gusa4grr: good catch! removed

gusa4grr:
> overall looks good to me
> 
> can you check each of the coderabbit comments?
> 
> and we could do with a 2nd maintainer review on this - ill ping the other maintainers

thanks for the review @43081j !

I just added the [LS migration from old to new key and E2E tests](https://github.com/npmx-dev/npmx.dev/pull/1189/changes/1c63c45132bdb64fc36b4997056c19c632b10455)

I am fixing the pipeline and coderabbit issues now. Was planning to ask people in Discord to help with review (which I will do after I make those fixes ⬆️ ).
still will appreciate you spreading the word ❤️

test/nuxt/components/compare/PackageSelector.spec.ts
```diff
@@ -1,22 +1,66 @@
...
+    status: ref('idle'),
+    isLoadingMore: ref(false),
+    hasMore: ref(false),
+    fetchMore: vi.fn(),
+    isRateLimited: ref(false),
+    suggestions: ref([]),
+    suggestionsLoading: ref(false),
+    packageAvailability: ref(null),
+  })
+})
```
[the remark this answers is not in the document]
gusa4grr: non-related to my MR, will skip this

shared/schemas/userPreferences.ts
```diff
@@ -0,0 +1,55 @@
...
+const ColorModePreferenceSchema = picklist(['light', 'dark', 'system'])
+
+const SearchProviderSchema = picklist(['npm', 'algolia'])
+
+export const UserPreferencesSchema = object({
+  /** Display dates as relative (e.g., "3 days ago") instead of absolute */
+  relativeDates: optional(boolean()),
+  /** Include @types/* package in install command for packages without built-in types */
+  includeTypesInInstall: optional(boolean()),
+  /** Accent color theme ID (e.g., "rose", "amber", "emerald") */
```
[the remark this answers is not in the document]
gusa4grr: 👍🏻

server/utils/preferences/user-preferences-store.ts
```diff
@@ -0,0 +1,49 @@
...
+    const base = existing ?? { ...DEFAULT_USER_PREFERENCES }
+
+    const merged: UserPreferences = {
+      ...base,
+      ...partial,
+      updatedAt: new Date().toISOString(),
+    }
+
+    await this.set(did, merged)
+    return merged
```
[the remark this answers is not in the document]
gusa4grr: 👍🏻

app/pages/settings.vue
```diff
@@ -35,7 +36,7 @@ defineOgImageComponent('Default', {
 })
 
 const setLocale: typeof setNuxti18nLocale = locale => {
-  settings.value.selectedLocale = locale
+  preferences.value.selectedLocale = locale
   return setNuxti18nLocale(locale)
 }
```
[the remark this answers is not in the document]
gusa4grr: the risk is theoretical, iiuc

app/composables/useUserPreferencesSync.client.ts
```diff
@@ -0,0 +1,169 @@
...
+    window.addEventListener('beforeunload', () => {
+      if (hasPendingChanges && isAuthenticated.value) {
+        const preferences = getPreferences()
+        navigator.sendBeacon(
+          '/api/user/preferences',
+          new Blob([JSON.stringify(preferences)], { type: 'application/json' }),
+        )
+      }
+    })
+  }
```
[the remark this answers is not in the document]
gusa4grr: while in larger context this is only called inside a plugin "preference-sync.client.ts", I'll add a small flag to omit creating extra listeners just in case :)

app/composables/useUserPreferencesSync.client.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+  }
+
+  function setupRouteGuard(getPreferences: () => UserPreferences): void {
+    router.beforeEach(async (_to, _from, next) => {
+      if (hasPendingChanges && isAuthenticated.value) {
+        void flushPendingSync(getPreferences())
+      }
+      next()
+    })
```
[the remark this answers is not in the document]
gusa4grr: same as [here](https://github.com/npmx-dev/npmx.dev/pull/1189#discussion_r2915042220)

app/composables/useUserPreferencesSync.client.ts
```diff
@@ -0,0 +1,169 @@
...
+    const serverPreferences = await fetchServerPreferences()
+
+    if (serverPreferences) {
+      showSyncedStatus()
+      return serverPreferences
+    }
+
+    state.status.value = 'idle'
+    return { ...DEFAULT_USER_PREFERENCES }
+  }
```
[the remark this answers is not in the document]
gusa4grr: 👍🏻

app/composables/useUserPreferencesProvider.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+export type HydratedUserPreferences = Required<Omit<UserPreferences, 'updatedAt'>> &
+  Pick<UserPreferences, 'updatedAt'>
+
+let dataRef: RemovableRef<HydratedUserPreferences> | null = null
+let syncInitialized = false
```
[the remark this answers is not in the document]
gusa4grr: while it is a valid concern, there is no test for `initSync`, and we can indeed use resetModules, when needed

app/composables/useUserPreferencesProvider.ts
gusa4grr: TypeScript will throw an error, effectively forcing the developer to add/remove a key from the `default` object. Adding a length check will actually make things less reliable and flaky

app/composables/userPreferences/useRelativeDates.ts
gusa4grr: good point, will rename this and others 👍🏻

test/nuxt/composables/use-package-list-preferences.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  it('handles array merging by replacement', () => {
+    mountWithSetup(() => {
+      const stored = { columns: [] }
+      setLocalStorage(stored)
+      const { preferences } = usePackageListPreferences()
+      onMounted(() => {
+        expect(preferences.value.columns).toEqual([])
+      })
+    })
```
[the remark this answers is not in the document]
gusa4grr: 👍🏻

test/nuxt/composables/use-install-command.spec.ts
```diff
@@ -217,9 +217,9 @@ describe('useInstallCommand', () => {
     })
 
     it('should only include main command when @types disabled via settings', () => {
-      // Get settings and disable includeTypesInInstall directly
-      const { settings } = useSettings()
-      settings.value.includeTypesInInstall = false
+      // Get preferences and disable includeTypesInInstall directly
+      const { preferences } = useUserPreferencesState()
+      preferences.value.includeTypesInInstall = false
```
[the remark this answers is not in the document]
gusa4grr: 👍🏻

app/plugins/i18n-loader.client.ts
```diff
@@ -1,20 +1,18 @@
...
-      //   Check if the value is a supported locale
-      locales.value.map(l => l.code).includes(settingsLocale) &&
-      //   Check if the value is not a current locale
-      settingsLocale !== locale.value
-    ) {
-      setLocale(settingsLocale)
+    const matchedLocale = locales.value.map(l => l.code).find(code => code === settingsLocale)
+
+    if (matchedLocale && matchedLocale !== locale.value) {
+      setLocale(matchedLocale)
```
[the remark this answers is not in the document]
gusa4grr: false positive, it can't be undefined

app/composables/userPreferences/useAccentColor.ts
```diff
@@ -0,0 +1,33 @@
...
+  })
+
+  function setAccentColor(id: AccentColorId | null) {
+    if (id) {
+      document.documentElement.style.setProperty('--accent-color', `var(--swatch-${id})`)
+    } else {
+      document.documentElement.style.removeProperty('--accent-color')
+    }
+    preferences.value.accentColorId = id
+  }
```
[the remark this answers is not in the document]
gusa4grr: 👍🏻

app/composables/userPreferences/useBackgroundTheme.ts
```diff
@@ -0,0 +1,27 @@
...
+  const { preferences } = useUserPreferencesState()
+
+  function setBackgroundTheme(id: BackgroundThemeId | null) {
+    if (id) {
+      document.documentElement.dataset.bgTheme = id
+    } else {
+      document.documentElement.removeAttribute('data-bg-theme')
+    }
+    preferences.value.preferredBackgroundTheme = id
+  }
```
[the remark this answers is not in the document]
gusa4grr: 👍🏻

app/composables/useUserPreferencesProvider.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    if (isAuthenticated.value) {
+      const serverPrefs = await loadFromServer()
+      if (serverPrefs) {
+        const merged = { ...preferences.value, ...serverPrefs }
+        if (!arePreferencesEqual(preferences.value, merged)) {
+          preferences.value = merged
+        }
+      }
+    }
```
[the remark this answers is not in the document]
gusa4grr: outdated/not applicable anymore

gusa4grr:
@43081j addressed comments. Updated [MR description](https://github.com/npmx-dev/npmx.dev/pull/1189#issue-3911743073) so you can read the changes summarised

gusa4grr:
there is a test failed, looked unrelated at start, but I have an idea what may went wrong. Will check hopefully tomorrow

shuuji3:
This is my quick investigation and I may be wrong, but I suspect rabbit's concern (https://github.com/npmx-dev/npmx.dev/pull/1189#discussion_r2838628106) might be related to the test failure. It seems that `syncInitialized` is not reset between multiple tests, and then `profile.handle` somehow got `undefined`, making always `true` here: https://github.com/fatfingers23/npmx.dev/blob/def75e8be1f72a58df5a5f2ce2fc52086f332f75/app/pages/profile/%5Bidentity%5D/index.vue#L87.

gusa4grr:
> This is my quick investigation and I may be wrong, but I suspect rabbit's concern ([#1189 (comment)](https://github.com/npmx-dev/npmx.dev/pull/1189#discussion_r2838628106)) might be related to the test failure. It seems that `syncInitialized` is not reset between multiple tests, and then `profile.handle` somehow got `undefined`, making always `true` here: https://github.com/fatfingers23/npmx.dev/blob/def75e8be1f72a58df5a5f2ce2fc52086f332f75/app/pages/profile/%5Bidentity%5D/index.vue#L87.

Thanks for the input!

It is not directly influencing it. The root cause of the failure is due to newly added calls to `useUserPreferenceState` and `useInitUserPreferencesSync` by the i18n plugin and the newly added `preferences-sync` plugin, respectively.

As the plugins run before the test file is executed (where the mocking happens), `useAtproto` (which is now expected to be called in user-preference composables) fires before the test, and the "attempt" to mock it in the test body is doomed. Spying is also impossible for `export const`

I found an open issue about that, but I did not find it helpful
https://github.com/nuxt/test-utils/issues/544

Not sure what the best workaround is in this case, as nothing declared in the test file can "intercept" this chain, as plugins will always be run during Nuxt initialization

The only thing I have in mind is:
-  Introduce some "mutable global variable" which will "intercept" the `useAtproto` response. This way, we'll mutate that in the test and reset after the test

but I don't want to make it too ugly and would rather have someone more experienced in the Nuxt ecosystem to advise here

cc @danielroe @shuuji3 @43081j

app/composables/atproto/useAtproto.ts
```diff
@@ -1,4 +1,24 @@
...
+
+export interface UseAtprotoReturn {
+  user: Ref<User | null | undefined>
+  pending: Ref<boolean>
+  logout: () => Promise<void>
+}
+
+declare global {
+  // eslint-disable-next-line no-var
+  var __useAtprotoMock: UseAtprotoReturn | undefined
```
43081j, line 14: is it not possible to mock nuxt composables? 😞 

it seems a shame to need this in prod sources to satisfy tests

CONTRIBUTING.md
```diff
@@ -166,6 +166,7 @@ The `.cache/` directory is a separate storage mount used for fetch-cache and atp
 app/                    # Nuxt 4 app directory
 ├── components/         # Vue components (PascalCase.vue)
 ├── composables/        # Vue composables (useFeature.ts)
+│   └── userPreferences/  # User preference composables (synced to server)
```
43081j, line 169: not sure this is necessary - we don't do it for any other composables

app/composables/atproto/useAtproto.ts
gusa4grr: I am not an expert in Nuxt, but I did not find a way how to mock/bypass "plugins" for specific test.

this is the reason, why I went this way

> As the plugins run before the test file is executed (where the mocking happens), useAtproto (which is now expected to be called in user-preference composables) fires before the test, and the "attempt" to mock it in the test body is doomed. Spying is also impossible for export const

CONTRIBUTING.md
gusa4grr: You mean creating a folder for composables?

I followed your suggestion from Feb 10 😅 
https://github.com/npmx-dev/npmx.dev/pull/1189#discussion_r2790578766

gusa4grr:
almost there 🙏🏻 
found some more things, I will describe them in more detail and most likely will create a separate MR for that 🙂

CONTRIBUTING.md
43081j: no i mean this literal change, the one here this comment is on.

it doesn't make sense to list only one composable sub-dir, so that means either list them all, or don't list any. but if we list them all, it'll very quickly go out of sync - so my preference would be to just not do this.

app/composables/atproto/useAtproto.ts
serhalp: Hey @gusa4grr check out `mockNuxtImport()`, used [like this](https://github.com/npmx-dev/npmx.dev/blob/38820a3a21e25cfdeff94dabce1c20d1e2372228/test/nuxt/components/Header/MobileMenu.spec.ts#L14-L16). You need to call it in module scope, not in a `describe` or `it`.

CONTRIBUTING.md
gusa4grr: ok, I will remove 👍🏻

app/composables/atproto/useAtproto.ts
gusa4grr: hey @serhalp !
It won't work because, as part of initial userPreferences fetch, `useAtproto` is now run inside the plugin. Plugins run before the test module executes, so `mockNuxtImport()` does nothing.

The other way, maybe, is to kill the [preference-sync.client.ts](https://github.com/npmx-dev/npmx.dev/pull/1189/changes#diff-70a4766332e3ae5323beabd2c5a2c548728162f0bacbd55413eaa3c491137ca1) plugin, and put the "initial" loading in some other entry point location. 

let me try this approach

43081j:
sorry this went very stale.

i caught up with the rest of the core team on this recently and we've decided the direction we want to go for now is:

- do not persist preferences serverside for now
- do unify/cleanup the existing preferences

this PR right now is both of these. if we can carve out the part which does server persistence, we can land the rest and decide on the server stuff another time in future.

if you don't have the time, let me know and possibly i can try get around to it

Issue #1174, opened 2026-02-07
Search results not appearing due to rate limiting / 429s


johnnyreilly:
Hello!

I was doing some testing on my phone (android pixel 8a) and found that typing `ts-loader` (one of my packages) into the search bar lead to no results:

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/111b0799-da78-4653-97af-235b18264c59" />

The URL at this point was https://npmx.dev/search?q=ts-loader

If you take the URL and open a brand new browser tab with that URL all behaves as expected:

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/d9c75d25-03f0-4dda-9b5b-7ccf4c7f0897" />

I'm not sure what the cause of this might be, but I thought I'd raise it. I'll try on a laptop when I get a moment and see if it is the same behaviour there as well.

Adebesin-Cell:
> Hello!
> 
> I was doing some testing on my phone (android pixel 8a) and found that typing `ts-loader` (one of my packages) into the search bar lead to no results:
> 
[...]

PR #1200, opened 2026-02-08
fix: surfacing 429s as not found to users

johnnyreilly:
Resolves: #1174 

This is a speculative PR which changes the approach to surface an explicit "Hit npm rate limit, try again in a moment" message when 429s happen. I wondered if this would be useful?

image

No worries if it's not - but I thought I'd share the idea! This doesn't contain all i18n locales. If the approach is helpful I think with AIs help I can probably provide that.

PS first time doing any Vue / Nuxt - please bear with my mistakes!


danielroe:
thank you ❤️

PR #174, opened 2026-01-27
feat: use/support algolia for package search and org/user package listing

Haroenv:
ref: #32

This is just a prototype to see what it takes to swap out the search for Algolia.

DONE:
- set up search via algolia JS client
- map the parameters to algolia parameters
- map the response to the same one as npm registry
- swap the query from debounced to non-debounced (it's faster)

TODO:
- wire up the ALGOLIA flag to some global flags system
- clean up types and mapping
- highlighting

What do you think?

PR #1204, opened 2026-02-08
feat: support algolia for package search + org/user package listing

danielroe:
remake of https://github.com/npmx-dev/npmx.dev/pull/174 as I didn't have permission to push

this builds on @haroenv's work to add algolia as a provider of data for org/user package listings, as well as the package search page

it also makes it the default, as it is _much_ faster

~~need to tweak the wording about algolia~~

SCR-20260208-klsn
SCR-20260208-klql

closes #174
resolves https://github.com/npmx-dev/npmx.dev/issues/32

### DIFF
--- a/app/composables/useSettings.ts
+++ b/app/composables/useSettings.ts
@@ -41,7 +41,7 @@ const DEFAULT_SETTINGS: AppSettings = {
   hidePlatformPackages: true,
   selectedLocale: null,
   preferredBackgroundTheme: null,
-  searchProvider: 'algolia',
+  searchProvider: import.meta.test ? 'npm' : 'algolia',
   sidebar: {
     collapsed: [],
   },
--- a/test/nuxt/components/compare/PackageSelector.spec.ts
+++ b/test/nuxt/components/compare/PackageSelector.spec.ts
@@ -1,5 +1,5 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest'
-import { nextTick, ref } from 'vue'
+import { ref } from 'vue'
 import { mountSuspended } from '@nuxt/test-utils/runtime'
 import PackageSelector from '~/components/Compare/PackageSelector.vue'
 
@@ -141,7 +141,6 @@ describe('PackageSelector', () => {
 
       const input = component.find('input')
       await input.setValue('lodash')
-      await nextTick()
       await input.trigger('keydown', { key: 'Enter' })
 
       const emitted = component.emitted('update:modelValue')
@@ -158,7 +157,6 @@ describe('PackageSelector', () => {
 
       const input = component.find('input')
       await input.setValue('no dep')
-      await nextTick()
       await input.trigger('keydown', { key: 'Enter' })
 
       const emitted = component.emitted('update:modelValue')
@@ -175,7 +173,6 @@ describe('PackageSelector', () => {
 
       const input = component.find('input')
       await input.setValue('lodash')
-      await nextTick()
       await input.trigger('keydown', { key: 'Enter' })
 
       // Input should be cleared
--- a/test/nuxt/setup.ts
+++ b/test/nuxt/setup.ts
@@ -23,14 +23,4 @@ if (typeof window !== 'undefined') {
       removeColorScheme: () => {},
     }
   }
-
-  // force npm search provider in tests for compatibility with our mocks
-  const settingsKey = 'npmx-settings'
-  try {
-    const existing = JSON.parse(localStorage.getItem(settingsKey) || '{}')
-    existing.searchProvider = 'npm'
-    localStorage.setItem(settingsKey, JSON.stringify(existing))
-  } catch {
-    localStorage.setItem(settingsKey, JSON.stringify({ searchProvider: 'npm' }))
-  }
 }


app/composables/npm/useAlgoliaSearch.ts
```diff
@@ -0,0 +1,213 @@
...
+/**
+ * Algolia search client for npm packages.
+ * Uses npm's public Algolia index (same as npmjs.com).
+ */
+let _searchClient: LiteClient | null = null
+
+function getAlgoliaClient(): LiteClient {
+  if (!_searchClient) {
+    // npm's public search-only Algolia credentials (same as npmjs.com uses)
+    _searchClient = algoliasearch('OFCNCOG2CU', 'f54e21fa3a2a0160595bb058179bfb1e')
```
43081j: these are unlikely to change, but is it possible we could have them in env vars instead? made available by whatever process publishes this stuff (i.e. template them in at deployment time or some such thing)

danielroe: yes, good point 👍

app/components/SearchProviderToggle.client.vue
```diff
@@ -0,0 +1,117 @@
...
+                {{ $t('settings.search_provider_npm_description') }}
+              </p>
+            </div>
+          </button>
+
+          <!-- Algolia option -->
+          <button
+            type="button"
+            role="menuitem"
+            class="w-full flex items-start gap-3 px-3 py-2.5 rounded-md text-start transition-colors hover:bg-bg-muted"
```
alexdln: Now they are glued and it is noticeable when hovering.

```suggestion
            class="w-full flex items-start gap-3 px-3 py-2.5 rounded-md text-start transition-colors hover:bg-bg-muted mt-1"
```

app/composables/npm/useAlgoliaSearch.ts
```diff
@@ -0,0 +1,235 @@
...
+          analyticsTags: ['npmx.dev'],
+          attributesToRetrieve: ATTRIBUTES_TO_RETRIEVE,
+          attributesToHighlight: [],
+        },
+      },
+    ])
+
+    const response = results[0] as SearchResponse<AlgoliaHit> | undefined
+    if (!response) {
+      return { isStale: false, objects: [], total: 0, time: new Date().toISOString() }
```
alexdln: Shouldn't we do some additional processing in this case?

It seems this is a case where we won't figure out whether nothing was actually found or a server error

app/composables/npm/useAlgoliaSearch.ts
```diff
@@ -0,0 +1,235 @@
...
+    const max = options.maxResults ?? 1000
+
+    const allHits: AlgoliaHit[] = []
+    let offset = 0
+    let serverTotal = 0
+    const batchSize = 200
+
+    // Algolia supports up to 1000 results per query with offset/length pagination
+    while (offset < max) {
+      const length = Math.min(batchSize, max - offset)
```
alexdln: It seems the logic is broken here a bit. It should probably be sth like `Math.max(max, totalPackages - downloadPackages)`

Because right now, the maximum downloaded is 200, and the total could end up in the minus

app/composables/npm/useUserPackages.ts
```diff
@@ -0,0 +1,226 @@
...
+    }
+    return asyncData.data.value
+  })
+
+  /** Whether there are more results available to load */
+  const hasMore = computed(() => {
+    // Algolia fetches everything in one go
+    if (searchProvider.value !== 'npm') return false
+    if (!cache.value) return true
+    return cache.value.objects.length < Math.min(cache.value.total, MAX_RESULTS)
```
alexdln: if `cache.value.objects.length` is 1700, `cache.value.total` is 1800, `MAX_RESULTS` - 1000 - this condition will return `false`. Is it planned?

app/pages/~[username]/index.vue
```diff
@@ -219,7 +177,7 @@ defineOgImageComponent('Default', {
 
     <!-- Loading state -->
     <LoadingSpinner
-      v-if="status === 'pending' && currentPage === 1"
+      v-if="status === 'pending' && packages.length === 0"
```
Comment (alexdln) [inline app/pages/~[username]/index.vue]: It seems there is a chance for an eternal spinner?

app/composables/npm/useUserPackages.ts
```diff
@@ -0,0 +1,229 @@
...
+  const cache = shallowRef<{
+    username: string
+    objects: NpmSearchResult[]
+    total: number
+  } | null>(null)
+
+  const isLoadingMore = shallowRef(false)
+
+  const asyncData = useLazyAsyncData(
+    () => `user-packages:${searchProvider.value}:${toValue(username)}`,
```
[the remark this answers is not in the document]
danielroe: no asyncData keys can be functions

Haroenv:
thanks for taking over @danielroe :D

Haroenv:
However, could we add the acknowledgment again?

danielroe:
the acknowledgement is in two places I believe - let's chat in discord and you can tell me where would be better 🙏


image

image

nuxt.config.ts
```diff
@@ -33,6 +33,14 @@ export default defineNuxtConfig({
...
       redisRestToken: process.env.UPSTASH_KV_REST_API_TOKEN || process.env.KV_REST_API_TOKEN || '',
     },
+    public: {
+      // Algolia npm-search index (maintained by Algolia & jsDelivr, used by yarnpkg.com et al.)
+      algolia: {
+        appId: 'OFCNCOG2CU',
+        apiKey: 'f54e21fa3a2a0160595bb058179bfb1e',
+        indexName: 'npm-search',
+      },
+    },
```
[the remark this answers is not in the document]
Haroenv: API key is correct to use, but I made one specifically for npmx.dev now: `e0359cd30aa2c30b80044a0a1fe47282`

PR #1217, opened 2026-02-08
chore: add proper types to OgImage `Package.d.vue.ts` workaround

serhalp:
It appears this file was here to break a circular type resolution. `nuxt-og-image` generates a type declaration that imports `Package.vue type`, which needs auto-import globals from `nuxt.d.ts`, which needs the `nuxt-og-image` type declaration, creating a cycle.

The old file used `any` which was a band-aid. I couldn't come up with a way to remove this file entirely, but this at least types the component props so consumers still get real type-checking.

### DIFF
--- a/app/components/OgImage/Package.d.vue.ts
+++ b/app/components/OgImage/Package.d.vue.ts
@@ -1,3 +1,16 @@
-declare const _default: any
+// This type declaration file is required to break a circular type resolution in vue-tsc.
+//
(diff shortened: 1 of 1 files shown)


43081j (approved):
too bad we have to do this, would be good to somehow solve it one day such that we can drop the workaround

however, looks good to me

danielroe:
this will be resolved in the next version of `nuxt-og-image`

PR #1218, opened 2026-02-08
chore: validate package existence before adding to compare

iiio2:
validate


serhalp:
🤔 @iiio2 could you explain what problem this is solving? I tried poking around in the app and I couldn't find a scenario this would change.

iiio2:
@serhalp , I thought I can provide validation when no package found.

danielroe:
I agree - we're providing pre-validated packages so we probably don't need to revalidate them 🙏

iiio2:
Ok, Thanks.

Issue #1210, opened 2026-02-08
Navigating to a package from the search doesn't update the page URL


nicolo-ribaudo:
Steps:
1. Go to https://npmx.dev
2. Type in the search bar `@babel/core`
3. Click on the first result

It opens the `@babel/core` page, but the URL is still https://npmx.dev/search?q=%40babel%2Fcore. This means that if I share it with somebody, they'll see the search results rather than the package.

Video:
https://github.com/user-attachments/assets/a5eaa264-49e6-4351-8ccd-120953d13477

ghostdevv:
I think this may have been resolved @nicolo-ribaudo - could you reopen if I'm wrong?

nicolo-ribaudo:
Still happening, just less frequently:

https://github.com/user-attachments/assets/28ae61c1-0572-4e9d-a105-8a203845dcac

(I do not have permission to reopen)

trueberryless:
> Still happening, just less frequently:
> Screencast.From.2026-02-10.13-05-37.webm
> 
> (I do not have permission to reopen)

Hmm, I personally can't reproduce this behavior. How frequent was this buggy behavior happening to you? Sorry to ask again, but since it has been 3 weeks it could be that this issue has already been resolved by now. So you could please maybe try it out again? That would be much appreciated 🙏

shuuji3:
I assume Nicolò's browser is Firefox. I also couldn't reproduce on  Firefox (version: 148.0 (64-bit) on Linux and 147.0.4 (aarch64) macOS).

Maybe this is related to debounce? 🤔 We could try increasing debounce wait time on dev env if the issue still persists.

alexdln:
Some improvements have been made to this behavior in #1431. Essentially, URL changes are now processed in the background and shouldn't be lost or affect results 🤔

serhalp:
Hey friends, is this issue still relevant?

nicolo-ribaudo:
Yes, I am still getting the behavior described in the original issue.

https://github.com/user-attachments/assets/72b2f876-158f-4713-9103-96f3e0616ad0

nicolo-ribaudo:
I am on Firefox 150.0a1 (2026-03-23), and this bug happens _almost always_ (I just tried 5 times, it happened 4 out of 5).

trueberryless:
Sorry for another ping, @nicolo-ribaudo, but does it still happen in your browser?
Can you see something in the browser console?

[...]

PR #1220, opened 2026-02-08
fix: preserve correct search query url

iiio2:
resolves https://github.com/npmx-dev/npmx.dev/issues/1210


ghostdevv (changes requested):
My understanding of the linked issue is that it wasn't updated the url, not that it was encoded incorrectly?

ghostdevv:
I will close this for now because regardless I think the underlying issue is resolved - please let me know if I'm wrong :pray:

Issue #1111, opened 2026-02-06
Surface keyboard shortcuts


whitep4nth3r:
Keyboard shortcuts are going to be a big win and a feature we keep talking about.

We need a way to surface all keyboard shortcuts.

I propose a keyboard shortcut to show shortcuts in a popover or something, and the shortcut for this is shown in the footer, perhaps?

RYGRIT:
I had a similar idea a few days ago. I strongly agree with adding a keyboard shortcut to trigger a pop-up window that displays all the shortcuts and explains what they do.

For example, the Vimium C browser plugin allows you to display keyboard shortcut help documentation using `?`.

<img width="1440" height="1472" alt="Image" src="https://github.com/user-attachments/assets/015daee4-839b-4857-90cc-ff38da3b8c03" />

whitep4nth3r:
I'm working on this.

PR #1221, opened 2026-02-08
feat: surface kb shortcuts

whitep4nth3r:
I'm not sure how good this is.

Initially thought about using native HTML popover etc, but thought I would try and make it more Vue like.

image
image

Closes https://github.com/npmx-dev/npmx.dev/issues/1111

### DIFF
--- a/app/components/AppHeader.vue
+++ b/app/components/AppHeader.vue
@@ -122,8 +122,12 @@ onKeyStroke(
 
       <!-- Center: Search bar + nav items -->
       <div
-        class="flex-1 flex items-center justify-center md:gap-6"
-        :class="{ 'hidden sm:flex': !isSearchExpanded }"
+        class="flex-1 flex items-center md:gap-6"
+        :class="{
+          'hidden sm:flex': !isSearchExpanded,
+          'justify-end': isOnHomePage,
+          'justify-center': !isOnHomePage,
+        }"
       >
         <!-- Search bar (hidden on mobile unless expanded) -->
         <HeaderSearchBox
--- a/app/components/Compare/PackageSelector.vue
+++ b/app/components/Compare/PackageSelector.vue
@@ -139,17 +139,16 @@ function handleBlur() {
 
     <!-- Add package input -->
     <div v-if="packages.length < maxPackages" class="relative">
-      <div class="relative group">
+      <div class="relative group flex items-center">
         <label for="package-search" class="sr-only">
           {{ $t('compare.selector.search_label') }}
         </label>
         <span
-          class="absolute inset-y-0 start-3 flex items-center text-fg-subtle pointer-events-none group-focus-within:text-accent"
-          aria-hidden="true"
+          class="absolute inset-is-3 text-fg-subtle font-mono text-md pointer-events-none transition-colors duration-200 motion-reduce:transition-none [.group:hover:not(:focus-within)_&]:text-fg/80 group-focus-within:text-accent z-1"
         >
-          <span class="i-carbon:search w-4 h-4" />
+          /
         </span>
-        <input
+        <InputBase
           id="package-search"
           v-model="inputValue"
           type="text"
@@ -158,7 +157,9 @@ function handleBlur() {
               ? $t('compare.selector.search_first')
               : $t('compare.selector.search_add')
           "
-          class="w-full bg-bg-subtle border border-border rounded-lg ps-10 pe-4 py-2.5 font-mono text-sm text-fg placeholder:text-fg-subtle motion-reduce:transition-none duration-200 focus:border-accent focus-visible:(outline-2 outline-accent/70)"
+          no-correct
+          size="medium"
+          class="w-full min-w-25 ps-7"
           aria-autocomplete="list"
           @focus="isInputFocused = true"
           @blur="handleBlur"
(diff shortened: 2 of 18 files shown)


app/components/AppFooter.vue
```diff
@@ -1,6 +1,47 @@
 <script setup lang="ts">
+import { ref, watch, nextTick } from 'vue'
+import { onClickOutside, onKeyDown } from '@vueuse/core'
+import { useFocusTrap } from '@vueuse/integrations/useFocusTrap'
+
 const route = useRoute()
 const isHome = computed(() => route.name === 'index')
+
+const triggerRef = ref<HTMLElement | null>(null)
+const popoverRef = ref<HTMLElement | null>(null)
```
danielroe: we should use useTemplateRef instead

app/components/AppFooter.vue
```diff
@@ -1,6 +1,47 @@
...
+  { ignore: [triggerRef] },
+)
+
+onKeyDown(
+  'Escape',
+  () => {
+    if (showPopover.value) showPopover.value = false
+  },
+  { dedupe: true },
+)
```
danielroe: Escape is also bound to go 'back' from the settings page so we probably need to prevent default as if you're on that page you get two actions for the price of one when pressing Escape

app/components/AppFooter.vue
```diff
@@ -1,6 +1,47 @@
...
+const triggerRef = ref<HTMLElement | null>(null)
+const popoverRef = ref<HTMLElement | null>(null)
+const showPopover = ref(false)
+
+const togglePopover = (e?: Event) => {
+  e?.stopPropagation()
+  showPopover.value = !showPopover.value
+}
+
+onClickOutside(
```
danielroe: I think we can use our existing Modal component which contains a lot of this logic already

app/components/AppFooter.vue
whitep4nth3r: done

app/components/AppFooter.vue
whitep4nth3r: good catch, done I think!

app/components/AppFooter.vue
whitep4nth3r: on it

whitep4nth3r:
@danielroe I have no idea what I am doing but take a look now.

app/components/AppFooter.vue
danielroe: this got reverted somehow when moving to modal

whitep4nth3r: 😭 fixed

app/components/AppFooter.vue
```diff
@@ -1,6 +1,42 @@
 <script setup lang="ts">
+import { ref, computed } from 'vue'
+import { onKeyDown } from '@vueuse/core'
+import Modal from './Modal.client.vue'
+
 const route = useRoute()
 const isHome = computed(() => route.name === 'index')
+
+const triggerRef = ref<HTMLElement | null>(null)
```
danielroe: this should also be useTemplateRef

app/components/AppFooter.vue
```diff
@@ -1,6 +1,42 @@
 <script setup lang="ts">
+import { ref, computed } from 'vue'
+import { onKeyDown } from '@vueuse/core'
+import Modal from './Modal.client.vue'
```
danielroe: don't import here - you'll lose the server side fallback

ghostdevv (dismissed):
could we add a keyboard shortcut for the keyboard shortcuts 😄 ctrl/cmd+? perhaps

whitep4nth3r:
@ghostdevv I did think about this but 

1. Didn't know the best option and 
2. Thought that given this is something people won't reference too often, I wanted to free up a shortcut for other things in the future and
3. How to surface this shortcut without putting a long shortcut inline with the link to the modal 😅

danielroe:
we already have `?` which highlights shortcuts in the UI, fwiw.

whitep4nth3r:
Thank you for cleaning up my nonsense @danielroe

whitep4nth3r:
> we already have `?` which highlights shortcuts in the UI, fwiw.

I guess I should put this in the shortcut modal?

danielroe:
not nonsense at all! and still wondering whether I should not have removed `aria-expanded` (but it didn't _seem_ semantic for a modal.... 🤔)

danielroe:
I'm updating the content atm - will push and you can edit/remove

whitep4nth3r:
@danielroe I approve your addition

danielroe:
🤝

whitep4nth3r:
@ghostdevv are you happy to proceed without a shortcut for the shortcuts for now?

whitep4nth3r:
@ghostdevv I'm gonna merge this and we can add something else later if desired.

ghostdevv:
> we already have `?` which highlights shortcuts in the UI, fwiw.

ah nice! It looks like github uses that to show the modal, so I wonder if we should too - Discord uses `ctrl + ?` but they're a chat app - `?` would work everywhere but the search box. IMO a modal is better, even though the currently behavior is very cool, because it more easily grabs your attention and alerts you to shortcuts that may not be on your page or viewport

ghostdevv:
> or the shortcuts for now?

ahhh sorry! only just seeing this - happy to createa a follow up pr if we reach a consensus

PR #1223, opened 2026-02-08
feat(ui): use icon-only when connected to both services

MatteoGabriele:
Adding avatars for each connected service has several problems:
- Visually, when a user has the same avatar everywhere, it creates multiple copies of the same image to be rendered
- The feature won't scale if we add more services

My proposal is to retain only the single cloud icon and add a green dot to indicate to the user that it is connected.

before
Screenshot 2026-02-08 at 17 00 12

after
Screenshot 2026-02-08 at 17 24 37

### DIFF
--- a/app/components/Header/AccountMenu.client.vue
+++ b/app/components/Header/AccountMenu.client.vue
@@ -18,9 +18,6 @@ const isOpen = shallowRef(false)
 /** Check if connected to at least one service */
 const hasAnyConnection = computed(() => isNpmConnected.value || !!atprotoUser.value)
 
-/** Check if connected to both services */
-const hasBothConnections = computed(() => isNpmConnected.value && !!atprotoUser.value)
-
 /** Only show count of active (pending/approved/running) operations */
 const operationCount = computed(() => activeOperations.value.length)
 
@@ -66,7 +63,7 @@ function openAuthModal() {
     >
       <span v-if="hasAnyConnection" class="flex items-center">
         <img
-          v-if="isNpmConnected && npmAvatar && !hasBothConnections"
+          v-if="npmAvatar && !atprotoUser?.avatar"
           :src="npmAvatar"
           :alt="npmUser || $t('account_menu.npm_cli')"
           width="24"
@@ -74,7 +71,7 @@ function openAuthModal() {
           class="w-6 h-6 rounded-full ring-2 ring-bg object-cover"
         />
         <img
-          v-else-if="atprotoUser?.avatar && !hasBothConnections"
+          v-else-if="atprotoUser?.avatar && !npmAvatar"
           :src="atprotoUser.avatar"
           :alt="atprotoUser.handle"
           width="24"


danielroe:
what about instead displaying the logo of the connected service? (npm, bluesky or atproto) and we revisit when we get lots more... (could display a `+4` or something)

MatteoGabriele:
That could indeed be a valid alternative. I will use the logos

Tobbe:
> what about instead displaying the logo of the connected service? (npm, bluesky or atproto) and we revisit when we get lots more... (could display a `+4` or something)

How do you pick what logo to display though, if the user is connected to more than one thing? (assuming they can be)

Could also have overlapping avatars. (But then again it gets tricky - which one do we show on top?)

MatteoGabriele:
There's also an "issue" with the ATProto icon: it kinda looks like I want to connect my email.


Screenshot 2026-02-08 at 18 31 04

MatteoGabriele:
@danielroe I've updated the PR with icons.
The npm icon is fine, but the one I picked for the ATProto might be misunderstood. Let me know what you think. I could use the Bluesky icon, but that would technically not be correct.

danielroe:
we need to update the mobile menu as well:

SCR-20260208-pqwx

I think the icon looks good.

app/components/Header/AccountMenu.client.vue
```diff
@@ -6,7 +6,6 @@ const {
   isConnected: isNpmConnected,
   isConnecting: isNpmConnecting,
   npmUser,
-  avatar: npmAvatar,
```
danielroe, line 9: still wondering if it makes sense to show avatars in the dropdown, even if not in the top level .... 🤔 

wdyt?

MatteoGabriele: I think it may look a bit _confusing_ and inconsistent.
If I have to go with avatars, which I do like because they give the impression that I've actually connected, I prefer the initial proposal.

whitep4nth3r:
Personal opinion, I want to see my face if I am just logged in via Bluesky.

So maybe for the first service, you show the avatar, and for subsequent, you do the other thing.

This is my preference:

1. Show the first logged in service avatar in the header
2. Show all other avatars next to the items in the dropdown list

e.g.

If you're logged into Bluesky only, show the Bluesky avatar in the header.
If you're logged into npm only, show the npm avatar in the header.

If you're logged into both, show the FIRST ONE in the list in the header, which at this point is Bluesky.

Additionally, the list items of available services should be in the same order all the time to avoid confusion.

e.g.

Bluesky first
npm second
etc

MatteoGabriele:
@whitep4nth3r @danielroe based on both your ideas, I have made the changes

when nothing is connected
Screenshot 2026-02-12 at 21 56 13

when atproto is connected
Screenshot 2026-02-12 at 21 56 33

when npm is connected 
Screenshot 2026-02-12 at 22 13 44

when both are connected
Screenshot 2026-02-12 at 22 14 05

Tobbe:
I think users are so used to seeing their own avatar in that spot, so I think we should have one there even if the user is connected to multiple services.  

The question then is - which one do we show? The first one the user connected? The most recent one? 
Have a setting for it on the settings page, where the user can pick the one they want to use? (maybe including a generic cloud icon?)

Always using the first one connected, as @whitep4nth3r suggested, does have the benefit of being less of a surprise, and potentially confusing when the avatar they're used to happens to change just because they connected some other service

MatteoGabriele:
@Tobbe I do agree with you. 

I proposed using the same connected icon (with a green dot) for every service to maintain consistency. People usually see their own avatar, but they rarely expect to see all avatars piled together when connected to multiple services.

Maybe we could do progressive enhancement.
Remove multiple avatars in the dropdown button, then add the possibility to set one of them as the main avatar.

Tobbe:
> but they rarely expect to see all avatars piled together when connected to multiple services.

I agree. That's why I suggested just showing one of them in the menu toggle button. 

I'm fine with going with Salma's suggestion of always just using the avatar from the service the user first connected.
But having it configurable (like you say, the possibility to set one of them as the main avatar) is also a valid option 👍

MatteoGabriele:
How do we know which one is picked first tho? We probably need to save it somewhere, right? 🤔

Tobbe:
> How do we know which one is picked first tho? We probably need to save it somewhere, right? 🤔

We don't already store the timestamp for when a service was connected, do we?

MatteoGabriele:
@Tobbe no, I believe we do not have such property available at the moment

ghostdevv:
Another idea :sweat_smile: 

image

Tobbe:
> Another idea 😅
> 
> image

And then someone adds Facebook login 😅 🟦

PR #1226, opened 2026-02-08
fix: only search on enter keypress

gameroman:
Searches on enter key press


gameroman:
I need to fix the tests now

ghostdevv:
I think this is a duplicate of #1107 unfortunately

gameroman:
@coderabbitai review

danielroe:
with recent improvements in this, I think we can keep this as search-as-you-type

PR #1229, opened 2026-02-08
chore: reduce uncaught errors in prod

43081j:
This adds a couple of minor fixes for reducing errors we're seeing in
production:

- Sessions can be `{}`, in which case, our types are incorrect
- When a module imports a built-in, the docs resolver tries to load it
  from `esm.sh`
- Use `URL.parse` instead of a try/catch

### DIFF
--- a/shared/types/userSession.ts
+++ b/shared/types/userSession.ts
@@ -1,15 +1,17 @@
 import type { NodeSavedSession, NodeSavedState } from '@atproto/oauth-client-node'
 
 export interface UserServerSession {
-  public: {
-    did: string
-    handle: string
-    pds: string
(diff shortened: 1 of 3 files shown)


shared/types/userSession.ts
```diff
@@ -1,15 +1,17 @@
...
-    avatar?: string
-  }
+  public?:
+    | {
+        did: string
+        handle: string
+        pds: string
+        avatar?: string
+      }
+    | undefined
```
ghostdevv, line 11: `?: ... | undefined` 😔

ghostdevv (approved):

danielroe (approved):

PR #1232, opened 2026-02-08
feat(i18n): update fr-FR translations

sybers:
Just adding some missing french translations :)


sybers:
@ghostdevv Yeah the Atmosphere wording has an accent, let me remove it

PR #1235, opened 2026-02-08
feat: use navigation builder for desktop and mobile links

bdbch:
This PR moves the navigation into it's own configuration object to make the configuration for both - desktop & mobile easier & less prone for errors.


bdbch:
Looking into the Type errors right now.

danielroe (approved):

PR #1237, opened 2026-02-08
perf: reduce initial css size

essenmitsosse:
This lazy loads components that rely on Vue data ui, which reduces ...


essenmitsosse:
This doesn't work. CSS doesn't get loaded anymore.

PR #1245, opened 2026-02-08
feat(ui): community version distribution

vinnymac:
Built on the backs of @graphieros's [vue-ui-xy](https://vue-data-ui.graphieros.com/docs#vue-ui-xy) and @NullVoxPopuli's [majors](https://majors.nullvoxpopuli.com/q?minors=false&old=false&packages=%40prisma%2Fclient) to add a npmx themed version distributions feature.

image

Also added hide old here because that's something I personally wanted for this. Idk if anyone else was working on a related feature here, so if this crosses paths with some ongoing work, just let me know.

### DIFF
--- a/app/components/Package/VersionDistribution.vue
+++ b/app/components/Package/VersionDistribution.vue
@@ -375,29 +375,35 @@ const endDate = computed(() => {
       </div>
 
       <div class="flex flex-col gap-4 w-full max-w-1/2">
-        <SettingsToggle
-          v-model="showRecentOnly"
-          :label="$t('package.versions.recent_versions_only')"
-          :tooltip="$t('package.versions.recent_versions_only_tooltip')"
-          tooltip-position="bottom"
-          :tooltip-to="inModal ? '#chart-modal' : undefined"
-          :tooltip-offset="8"
-          justify="between"
-          reverse-order
-          :class="pending ? 'opacity-50 pointer-events-none' : ''"
-        />
+        <TooltipApp
+          :text="$t('package.versions.recent_versions_only_tooltip')"
+          position="bottom"
+          :to="inModal ? '#chart-modal' : undefined"
+          :offset="8"
+        >
+          <SettingsToggle
+            v-model="showRecentOnly"
+            :label="$t('package.versions.recent_versions_only')"
+            justify="start"
+            reverse-order
+            :class="pending ? 'opacity-50 pointer-events-none' : ''"
+          />
+        </TooltipApp>
 
-        <SettingsToggle
-          v-model="showLowUsageVersions"
-          :label="$t('package.versions.show_low_usage')"
-          :tooltip="$t('package.versions.show_low_usage_tooltip')"
-          tooltip-position="bottom"
-          :tooltip-to="inModal ? '#chart-modal' : undefined"
-          :tooltip-offset="8"
-          justify="between"
-          reverse-order
-          :class="pending ? 'opacity-50 pointer-events-none' : ''"
-        />
+        <TooltipApp
+          :text="$t('package.versions.show_low_usage_tooltip')"
+          position="bottom"
+          :to="inModal ? '#chart-modal' : undefined"
+          :offset="8"
+        >
+          <SettingsToggle
+            v-model="showLowUsageVersions"
+            :label="$t('package.versions.show_low_usage')"
+            justify="start"
+            reverse-order
+            :class="pending ? 'opacity-50 pointer-events-none' : ''"
+          />
+        </TooltipApp>
       </div>
     </div>
 


NullVoxPopuli:
Oooo fancy !

vinnymac:
Need to change the title to feat(ui) oops

app/components/Settings/Toggle.client.vue
vinnymac, line 1: I'll fix the obvious issues here with modifying a global settings toggle 👍🏼

app/components/Package/VersionDistribution.vue
```diff
@@ -0,0 +1,326 @@
...
+      </ClientOnly>
+    </div>
+  </div>
+</template>
+
+<style scoped>
+/* Disable all transitions on SVG elements to prevent repositioning animation */
+:deep(.vue-ui-xy) svg rect {
+  transition: none !important;
+}
```
vinnymac, line 592: @graphieros I tried to disable the vue-ui-xy animations that I saw in the modal, as they looked a bit odd in this context (bars on the graph flying in from bottom left into position).

I couldn't figure out what the reasonable way to disable this was, so this is what I came up with for now. If there was something obvious I was missing, please let me know 🙏🏼

vinnymac:
@coderabbitai resume

vinnymac:
@coderabbitai resume

app/components/Package/VersionDistribution.vue
graphieros: What you did is fine!
It's part of the things to improve on the library side.

graphieros:
Looks great !
@vinnymac would you mind to check out what's done on `DownloadAnalytics.vue` (soon to be `TrendsChart.vue`, see #1310) regarding:

- svg slot
- custom legend placement
- watermark for SVG & PNG print
- dedicated slots for the context menu to customize icons
- xAxis labels could probably have a bigger font size 
- I wonder if the version toggles could be positioned before the labels
- you should not have to leave more bottom padding for rotated labels, the chart auto sizes for that (or tell me if you had issues with this ?)
- I think in mobile you can stretch the aspect ratio vertically
- checkout the css overrides in DownloadAnalytics.vue for the placement of the reset button of the minimap

app/components/Package/VersionDistribution.vue
```diff
@@ -0,0 +1,460 @@
...
+          xAxisLabels: {
+            show: xAxisLabels.value.length <= 25,
+            values: xAxisLabels.value,
+            fontSize: isMobile.value ? 14 : 12,
+            color: colors.value.fgSubtle,
+            rotation: xAxisLabels.value.length > 10 ? 45 : 0,
+          },
+        },
+      },
+      timeTag: {
```
graphieros: time tag can be removed (false by default)

app/components/Package/VersionDistribution.vue
```diff
@@ -0,0 +1,460 @@
...
+          yAxis: {
+            formatter: ({ value }: { value: number }) => {
+              return compactNumberFormatter.value.format(Number.isFinite(value) ? value : 0)
+            },
+            useNiceScale: true,
+          },
+          xAxisLabels: {
+            show: xAxisLabels.value.length <= 25,
+            values: xAxisLabels.value,
+            fontSize: isMobile.value ? 14 : 12,
```
graphieros: I think the font size could be bigger

app/components/Package/VersionDistribution.vue
```diff
@@ -0,0 +1,460 @@
...
+const chartConfig = computed(() => {
+  return {
+    theme: isDarkMode.value ? 'dark' : 'default',
+    chart: {
+      height: isMobile.value ? 500 : 400,
+      backgroundColor: colors.value.bg,
+      padding: {
+        top: 24,
+        right: 24,
+        bottom: xAxisLabels.value.length > 10 ? 100 : 72, // More space for rotated labels
```
graphieros: When labels rotate the chart resizes to avoid overflow automatically, so you don't need that much bottom padding

app/components/Package/VersionDistribution.vue
```diff
@@ -0,0 +1,460 @@
...
+        },
+        preview: {
+          fill: transparentizeOklch(accent.value, isDarkMode.value ? 0.95 : 0.92),
+          stroke: transparentizeOklch(accent.value, 0.5),
+          strokeWidth: 1,
+          strokeDasharray: 3,
+        },
+      },
+    },
+    userOptions: {
```
graphieros: userOptions is correctly declared on L.95, this can be removed

app/components/Package/VersionDistribution.vue
```diff
@@ -0,0 +1,460 @@
...
+        <span class="i-carbon:data-vis-4 w-8 h-8" />
+        <p>{{ $t('package.trends.no_data') }}</p>
+      </div>
+
+      <ClientOnly v-else-if="xyDataset.length > 0">
+        <div
+          class="chart-container w-full h-[400px] sm:h-[400px]"
+          :class="{ 'h-[500px]': isMobile }"
+        >
+          <VueUiXy :dataset="xyDataset" :config="chartConfig" class="[direction:ltr]" />
```
graphieros: To be added:

- svg scoped slot (see implementation in prod of DownloadAnalytics.vue, to inject a custom legend for SVG print, and npmx watermark)
- legend scoped slot for a single series
- all the scoped slots for the contextual menu:
  - menuIcon
  - optionCsv
  - optionImg
  - optionSvg
  - all the annotator-action-xxx scoped slots

app/components/Package/VersionDistribution.vue
```diff
@@ -0,0 +1,460 @@
...
+        </div>
```
graphieros, line 582: You can use the fallback slot provided by ClientOnly (see DownloadAnalytics.vue)

NullVoxPopuli:
It seems some logic is goofy with the 'old versions' toggle?

For example:
https://npmxdev-git-fork-vinnymac-vt-nullvoxpopuli-npmx.vercel.app/package/ember-source

Unchecked: i see 2.x and 6.x.
Checked: i see them all

Which seems backwards? Idk

Lookin' good so far!

vinnymac:
> Unchecked: i see 2.x and 6.x. Checked: i see them all
> 
> Which seems backwards? Idk
> 
> Lookin' good so far!

Just tried this and it worked for me, have you opened this link before by any chance? I have fixed caching related bugs on this branch that may require you to clear your browser  cache

NullVoxPopuli:
Hmm -- I experienced this in a private browser window 🤔

vinnymac:
@NullVoxPopuli oh! I read your message wrong. I actually flipped this toggle from "hide" to "show" recently, so now it defaults to hiding the old versions, so what you described is expected.

By default it will show versions younger than 1 year ago (older ones are hidden), and checked the box reveals them (shows you more versions)

NullVoxPopuli:
It's time based?

I don't think that is as effective a signal, because in the example above, we had an enterprise client who hadn't upgraded in over a decade pay for a security fix, so now 2.x shows up even though most users aren't on 2.x

And the whole major-release schedule is every year and a half for this particular project - and some other ecosystems have even longer between majors - perhaps a current version to current version minus 5 or some other range could be used?

Was time chosen because of frequent major releasors? (Angular, Next, etc?)

vinnymac:
> It's time based

Yea the old versions feature is time based. It defaults to hiding versions older than 1 years, and enabling the toggle displays those old versions. This was just something I wanted personally, but if you'd prefer it behave differently I am open to changing this / the defaults / etc.

NullVoxPopuli:
To me, it seems like 1 year only makes sense if the whole ecosystem is super on top of upgrades _and_ every library is doing multiple majors per year -- which would be quite disruptive if it were the default, i think

NullVoxPopuli:
image

should the labels be more left-aligned -- closer to the toggles?
and should the tooltip be more descriptive? 

<3

vinnymac:
Interesting, I didn't know they had tooltips on the toggles, I added my own to the labels instead when I first started this, not sure how new that is, but it definitely wasn't there before (maybe a bug fix?)

vinnymac:
> I noticed there's a bit of overlapping text when downloading a PNG so it would be worth looking at later, but wanted to get this in!

Do you have an example of this, screenshot?

Perhaps when the labels are rotated?

danielroe:
vue-ui-xy

graphieros:
> amazing ❤️
> 
> I noticed there's a bit of overlapping text when downloading a PNG so it would be worth looking at later, but wanted to get this in!

That's the watermark which is injected during the PNG generation. It takes a 'while' hence the flash. Some config adjustments are needed to avoid the overlap with time labels.

PR #1247, opened 2026-02-09
refactor: remove unnecessary prop from links

essenmitsosse:
In #1240 @alexdln fixed two regressions I had introduce in #1071:
1. a regression with the icon size in links (for some versions on the package page)
2. a regression with the truncation of long version names

Unfortunatly that did two things:
- by fixing 1.: introduce a new prop for the icon size, that would manually be set when changing the font size of a link (which would still make it easy to use the component and end up with a broken design)
- by fixing 2.: added an explicit margin to _all_ icons in links, which broke icon only links
    Screenshot 2026-02-09 at 00 53 42

This fixes this by:
- making the icon size relative to the font size of a link/button, no matter how the font size was set. This makes the components more robust and easier to use while ensuring a consistent design. This might still be insufficient in some cases, but not in any we currently have. The result should be identical to the fix @alexdln introduced
- refactor the links in `/Package/Versions` so that truncate works, but the links can keep their flex layout

To avoid `flex!` on links it additionally:
- adds a new `link-block` variant, to properly let a link be a block element, as well as changing all buttons and links with button design to be blocks by default (couldn't find any case where this would be a problem)

It also fixes a very rare case were the tag name would be truncated without need.

@alexdln thanks for spotting the regression! Would you be so kind and double check, I didn't make things worse again and missed anything?

### Preview:

Screenshot 2026-02-09 at 01 12 56

### DIFF
--- a/app/components/Code/FileTree.vue
+++ b/app/components/Code/FileTree.vue
@@ -50,7 +50,8 @@ watch(
       <!-- Directory -->
       <template v-if="node.type === 'directory'">
         <ButtonBase
-          class="w-full flex! justify-start! rounded-none! border-none!"
+          class="w-full justify-start! rounded-none! border-none!"
+          block
           :aria-pressed="isNodeActive(node)"
           :style="{ paddingLeft: `${depth * 12 + 12}px` }"
           @click="toggleDir(node.path)"
@@ -82,7 +83,8 @@ watch(
           variant="button-secondary"
           :to="getFileRoute(node.path)"
           :aria-current="currentPath === node.path"
-          class="w-full flex! justify-start! rounded-none! border-none!"
+          class="w-full justify-start! rounded-none! border-none!"
+          block
           :style="{ paddingLeft: `${depth * 12 + 32}px` }"
           :classicon="getFileIcon(node.name)"
         >
--- a/app/components/Compare/ComparisonGrid.vue
+++ b/app/components/Compare/ComparisonGrid.vue
@@ -42,7 +42,8 @@ function getReplacementTooltip(col: ComparisonGridColumn): string {
           <span class="inline-flex items-center gap-1.5 truncate">
             <LinkBase
               :to="packageRoute(col.name, col.version)"
-              class="text-sm flex! truncate"
+              class="text-sm truncate"
+              block
               :title="col.version ? `${col.name}@${col.version}` : col.name"
             >
               {{ col.name }}<template v-if="col.version">@{{ col.version }}</template>


app/components/Link/Base.vue
```diff
@@ -52,28 +51,21 @@ const isLinkAnchor = computed(
...
-const isButtonMedium = computed(() => props.size === 'medium' && props.variant !== 'link')
-
-const iconSizeClass = computed(
-  () => ICON_SIZE_MAP[props.iconSize || (isButtonSmall.value && 'sm') || 'md'],
-)
+const isLink = computed(() => props.variant === 'link' || props.variant === 'link-block')
+const isButton = computed(() => !isLink.value)
+const isButtonSmall = computed(() => props.size === 'small' && !isLink.value)
+const isButtonMedium = computed(() => props.size === 'medium' && !isLink.value)
+const isBlock = computed(() => isButton || props.variant === 'link-block')
```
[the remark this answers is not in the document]
danielroe: this is correct - it should be isButton.value here...

app/components/Link/Base.vue
```diff
@@ -84,12 +76,15 @@ const iconSizeClass = computed(
   /></span>
   <NuxtLink
     v-else
-    class="group/link inline-flex gap-x-1 items-center justify-center"
+    class="group/link gap-x-1 items-center"
     :class="{
+      'flex': isBlock,
+      'inline-flex': !isBlock,
```
alexdln: If I see a link-block prop being passed, I'm more likely to expect it to add display: block.

Remind me, why do we need flex here? Wouldn't inline-block be enough for the children so they just line up?

app/components/Link/Base.vue
```diff
@@ -52,28 +51,21 @@ const isLinkAnchor = computed(
...
-const isButtonMedium = computed(() => props.size === 'medium' && props.variant !== 'link')
-
-const iconSizeClass = computed(
-  () => ICON_SIZE_MAP[props.iconSize || (isButtonSmall.value && 'sm') || 'md'],
-)
+const isLink = computed(() => props.variant === 'link' || props.variant === 'link-block')
+const isButton = computed(() => !isLink.value)
+const isButtonSmall = computed(() => props.size === 'small' && !isLink.value)
+const isButtonMedium = computed(() => props.size === 'medium' && !isLink.value)
+const isBlock = computed(() => isButton.value || props.variant === 'link-block')
```
alexdln: Too much very specific conditions. If I provide size="small" for links I expect some size change

It should be limited, but comfortable to use and update.

Add prop inline: boolean, type: button | link, size: "sm" | "md" | "lg" (short for consistency with uno)

app/components/Link/Base.vue
alexdln: What if I need plain inline link?

For example, inline-flex won't be able to make a link start on one line and move to another

We definitely need LinkBaseBase...

app/components/Link/Base.vue
```diff
@@ -10,9 +10,8 @@ const props = withDefaults(
        * `type` should never be used, because this will always be a link.
        * */
       'type'?: never
-      'variant'?: 'button-primary' | 'button-secondary' | 'link'
+      'variant'?: 'button-primary' | 'button-secondary' | 'link' | 'link-block'
```
alexdln: Don't mix responsibilities. The prop should be responsible for a specific area

app/components/Button/Base.vue
```diff
@@ -32,7 +32,7 @@ defineExpose({
 <template>
   <button
     ref="el"
-    class="group cursor-pointer inline-flex gap-x-1 items-center justify-center font-mono border border-border rounded-md transition-all duration-200 disabled:(opacity-40 cursor-not-allowed border-transparent)"
+    class="group cursor-pointer flex gap-x-1 items-center justify-center font-mono border border-border rounded-md transition-all duration-200 disabled:(opacity-40 cursor-not-allowed border-transparent)"
     :class="{
       'text-sm px-4 py-2': size === 'medium',
```
alexdln, line 35: Let's use here sm | md | lg as well

app/components/Button/Base.vue
```diff
@@ -53,11 +53,7 @@ defineExpose({
     "
     :aria-keyshortcuts="keyshortcut"
   >
-    <span
-      v-if="classicon"
-      :class="[size === 'small' ? 'size-3' : 'size-4', classicon]"
-      aria-hidden="true"
-    />
+    <span v-if="classicon" class="size-[1em]" :class="classicon" aria-hidden="true" />
```
alexdln, line 54: On the side where the icon is, the padding should be a little smaller

app/components/Button/Base.vue
alexdln, line 62: Let's use start-icon and end-icon and allow there only icon class (sth like startIcon: \`i-carbon:${string}\`)

app/components/Link/Base.vue
essenmitsosse: Very good point, but that was an issue before as well.

app/components/Button/Base.vue
essenmitsosse: I think that would not work with Uno, because classes must be explicit strings

app/components/Button/Base.vue
essenmitsosse: Maybe I am misunderstand, but your comment seems like a design change? This PR doesn't try to change the design, but just fix the regressions.

app/components/Link/Base.vue
alexdln: You're right. Can you please move these comments into a separate issue to refine these elements?

app/components/Button/Base.vue
alexdln: This is about the type only. But it's also not related. I saw that the component was being polished and started to write it all down, sorry 😅

app/components/Link/Base.vue
essenmitsosse: What we need to take into consideration: Icon would only work in a limited way for such links, or at least their positioning would be limited to how you can position things inline.

app/components/Link/Base.vue
essenmitsosse: The idea is that the size of the link can be set by the text size of the parent (or directly on the link itself). Adding a prop to change the text size would be redundant. Links should be able to work in text of any size. I would love the `size` prop to only be applicable when you use a button variant, but unfortunately, Vue doesn't allow such "complex" types. Maybe it's best to split the link component to avoid this issue.

app/components/Link/Base.vue
essenmitsosse: I know what you mean. This goes back to my previous comment: Adding another prop, that is only relevant for a certain variant is also not the best design. Again: probably best to split the component. Would you be fine with this as an intermediate solution?

app/components/Button/Base.vue
alexdln: When there's an icon, to make the button look symmetrical, the indentation around the icon should be ~20% smaller. This makes it look more correct. But yeah, it's not related as well, sorry, I just started looking at all the changes as coderabbit

app/components/Button/Base.vue
essenmitsosse: See above comment, also renaming the variants seems outside the scope of this PR, since it doesn't touch the size variant at all.

alexdln: Not related, but copy to another issue please \:)

app/components/Link/Base.vue
alexdln: Not related, but copy to another issue please \:)

app/components/Button/Base.vue
essenmitsosse: I would love proper typing for the icon classes. Haven't found a way to get the types yet. Also being able to color the icon through the same prop currently is kind of nice.

app/components/Link/Base.vue
alexdln: Let's call this `inline` already and use it for both buttons and links

app/components/Button/Base.vue
essenmitsosse: Sure, let's improve this. But I also would love to hold out till we have Storybook before we dive into the deep end of this.

app/components/Link/Base.vue
alexdln: Sorry, missed comment

essenmitsosse: Yeah I see. Just one thing: Would it default to `true` for links but `false` for buttons?

essenmitsosse: Ah, just noticed buttons are currently inline as well. Ok I'll have a go on it.

alexdln: The point isn't that there are so many options, but that they're for completely different purposes. In your set of conditions below, you can see what you're using for what. Essentially, you're generating several other props from this one

And this list of conditions below is actually more logical for DX, and that's what you came up with. We're more comfortable writing and working with a component when we clearly understand what the result will be

`false` by default please. It's easier and more common to write just inline, not inline="false"

alexdln: This is also pretty standard practice when it's responsible for a narrow section: disabled, autocorrect, checked, hidden, open. It's probably even more standard in Vue, since it's positioned as closer to the HTML

essenmitsosse:
@alexdln Thanks for the input. You were right about the prop being overloaded. I went with having the components be inline (`inline-flex` by default with the prop `block` changing them to be block elements (`flex`).

I would handle the inline-inline links separately and opened an issue for the them (#1252).

Are you fine with this?

app/components/Link/Base.vue
alexdln: `block` is better, yes (_but now we've run into a terminology conflict. Would be better to work on that in the next iteration_)

alexdln:
How about putting the rest of the comments there too (or in another issue)?

alexdln:
Looks good to me, thanks for working on it and polishing it 🩵

app/components/Link/Base.vue
essenmitsosse: Yeah, never would have liked to set `true` as the default for the reasons you stated. What terminology conflict do you mean?

essenmitsosse:
@alexdln I think I got all your comments that are issue worthy. Thanks for the input!

alexdln:
@essenmitsosse Please update issues descriptions a bit so that newcomers to the project can more easily understand what happened - in case someone will want to take it right away

alexdln:
@essenmitsosse Also, to rework the props in general. Variant still has too much responsibility.

One for color, one for type (link / button)

PR #1249, opened 2026-02-09
chore(i18n): generate JSON Schema for locale files

serhalp:
## Summary

Generate a JSON Schema from the canonical `en.json` that provides some cool editor DX niceties (autocomplete, typo detection via `additionalProperties: false`) for all i18n locale files.

Screenshot 2026-02-08 at 15 00 16
Screenshot 2026-02-08 at 19 35 28

## Changes

- Add `scripts/generate-i18n-schema.ts` that generates a JSON Schema from the structure of `en.json` and writes it to `i18n/schema.json`.
- Add `"$schema"` reference to this schema to all 27 locale files
- Add `pnpm i18n:schema` and run it in the pre-commit hook (updates schema) and in CI (fails if schema is out of date)

## Notes

- In the pre-commit hooks I called the node script directly because in my testing on my machine pnpm adds a ~250ms overhead and it's important for good DX for these hooks to be as fast as possible.
- I initially thought we could generate the JSON Schema from the static analysis stuff we already do to detect unused keys. But it turned out to be much simpler and less wasteful to just use the `en.json` reference file as the source of truth.

### DIFF
--- a/lunaria/prepare-json-files.ts
+++ b/lunaria/prepare-json-files.ts
@@ -39,41 +39,37 @@ export async function prepareJsonFiles(): Promise<void> {
 
 type NestedObject = Record<string, unknown>
 
-export async function mergeLocaleObject(
-  locale: LocaleObject,
-  options: { copy?: boolean } = {},
-): Promise<NestedObject | undefined> {
-  const { copy = false } = options
+export async function mergeLocaleObject(locale: LocaleObject): Promise<NestedObject | undefined> {
   const files = locale.files ?? []
   if (locale.file || files.length === 1) {
     const json =
       (locale.file ? getFileName(locale.file) : undefined) ??
       (files[0] ? getFileName(files[0]) : undefined)
     if (!json) return undefined
-    if (copy) {
-      await fs.cp(path.resolve(`${localesFolder}/${json}`), path.resolve(`${destFolder}/${json}`))
-      return undefined
-    }
 
-    return await loadJsonFile<NestedObject>(json)
+    return await loadLocaleSourceJson<NestedObject>(json)
   }
 
   const firstFile = files[0]
   if (!firstFile) return undefined
-  const source = await loadJsonFile<NestedObject>(getFileName(firstFile))
+  const source = await loadLocaleSourceJson<NestedObject>(getFileName(firstFile))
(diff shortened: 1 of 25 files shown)


scripts/generate-i18n-schema.ts
```diff
@@ -0,0 +1,77 @@
...
+type LocaleJson = Json & { $schema: string }
+
+interface JsonSchema {
+  $schema?: string
+  title?: string
+  description?: string
+  type: string
+  properties?: Record<string, JsonSchema>
+  additionalProperties?: boolean
+}
```
serhalp, line 22: Technically I could pull in a dev dep to get a proper type for this... and we could pull in the JSON Schema meta-schema to validate our schema against the meta-schema... but none of that seemed worth it here.

package.json
```diff
@@ -145,9 +146,12 @@
   },
   "lint-staged": {
     "i18n/locales/*": [
-      "pnpm build:lunaria",
+      "node ./lunaria/lunaria.ts",
```
serhalp, line 149: Calling node scripts directly where possible because in my testing on my machine `pnpm ...` adds a ~250ms overhead and it's important for good DX for these hooks to be as fast as possible.

lunaria/prepare-json-files.ts
```diff
@@ -39,49 +39,45 @@ export async function prepareJsonFiles(): Promise<void> {
...
-): Promise<NestedObject | undefined> {
-  const { copy = false } = options
+export async function mergeLocaleObject(locale: LocaleObject): Promise<NestedObject | undefined> {
   const files = locale.files ?? []
   if (locale.file || files.length === 1) {
     const json =
       (locale.file ? getFileName(locale.file) : undefined) ??
       (files[0] ? getFileName(files[0]) : undefined)
     if (!json) return undefined
-    if (copy) {
```
serhalp, line 53: This was a teeny minor perf optimization (for a script that is run on demand). I had to remove it in order to exclude `$schema` from the generated files.

It's still quite fast, so this seems fine to me.

bluwy:
I'm not sure if the review request for `@npmx-dev/_` is automatic, but can we not ping it? Not all of us on that team are maintainers. Thanks 🙏

PR #1258, opened 2026-02-09
chore: rearrange footer items

iiio2:
There is one thing I want to say. We've added keyboard shortcuts in footer recently. Now in footer some items are related to within our app like about, privacy policy. Some footer items are external links which open in a new tab like docs.npmx.dev, chat.npmx.dev etc. And external links have also icon in the footer so that we can understand it is external. I think keyboard shortcuts is not external things. So I think we can group it together with the internal footer items like about and privacy policy. But currently keyboard shortcuts is with external links. So I think maybe it looks better if we rearrange this. 

 Before:
before

After:
now

### DIFF
--- a/app/components/AppFooter.vue
+++ b/app/components/AppFooter.vue
@@ -23,22 +23,9 @@ const showModal = () => modalRef.value?.showModal?.()
           <LinkBase :to="{ name: 'privacy' }">
             {{ $t('privacy_policy.title') }}
           </LinkBase>
-          <LinkBase to="https://docs.npmx.dev">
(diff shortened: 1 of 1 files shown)


whitep4nth3r:
This is a good change. How does the mobile menu match this thinking now? Are there any changes needed?

iiio2:
@whitep4nth3r , I think in mobile version, there is no keyboard shortcuts options anywhere currently. It is only for desktop.

iiio2:
Thanks @whitep4nth3r .

PR #1260, opened 2026-02-09
chore: remove shadow and border accent

iiio2:
Look, the search button has some shadow and border accent. I think we can improve it. 

Before:
wrong

After:
right


iiio2:
Thanks. :blush:

PR #1261, opened 2026-02-09
fix: restore default cursor and add commitment in docs

kalu5:
Restore the return cursor and delete the cursor of the basic button.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -395,19 +395,15 @@ For package links, use the auto-imported `packageRoute()` utility from `app/util
 
 ### Cursor and navigation
 
-We prefer consistency in user everyday experience
+**npmx** uses `cursor: pointer` only for links to match users’ everyday experience. For all other interactive elements, including buttons, use the default cursor (_or another appropriate cursor to indicate state_).
 
-npmx uses `cursor-pointer` only for links. For all other elements, including buttons, use the default cursor (_or other appropriate cursors to show state_).
-
-**What counts as a link**
-
-A link is any element that change or show meaningful content:
-
-- **Not meaningful content:** burger menu, "likes" button - the user does not get substantive information from them.
-- **Meaningful content:** a dialog with a chart, auth button, tabs with new data—they convey new or important information.
+> [!NOTE]
+> A link is any element that leads to another content (_go to another page, authorize_)
+> A button is any element that operates an action (_show tooltip, open menu, "like" package, open dropdown_)
+> If you're unsure which element to use - feel free to ask question in the issue or on discord
 
 > [!IMPORTANT]
-> Always prefer implementing such changes as real links so they can be opened in a new tab, shared or reloaded, and so the same content is available at a stable URL.
+> Always Prefer implementing navigation as real links whenever possible. This ensures they can be opened in a new tab, shared or reloaded, and so the same content is available at a stable URL
 
 ## RTL Support
 


kalu5:
I suggest adding the description of this problem to the contribution document to remind everyone to consider it when adding the cursor. 
Also pay attention to this issue during the review to avoid style overwriting.
What do you think? @alexdln

kalu5:
I found that the default style of Vue Data UI also uses "pointer". I'll submit an issue first.

essenmitsosse:
I think the consensus was that all clickable things should get cursor: pointer. You can still add `cursor-pointer` to the plain buttons that don't have it yet, but this will also be fixed once I replace them with the `ButtonBase` component.

kalu5:
> I think the consensus was that all clickable things should get cursor: pointer. You can still add `cursor-pointer` to the plain buttons that don't have it yet, but this will also be fixed once I replace them with the `ButtonBase` component.

Something like this: It is agreed that all buttons should use a unified component, such as ButtonBase. This component uses the default style. You can also use cursor: pointer to make it a Link Button.

essenmitsosse:
No I meant, also all buttons are supposed to just keep `cursor-pointer`. I think no one wanted an extra treatment for non-link buttons.

kalu5:
> No I meant, also all buttons are supposed to just keep `cursor-pointer`. I think no one wanted an extra treatment for non-link buttons.

Here is a discussion. #1074

whitep4nth3r:
To make this cursor-pointer discussion a final final, can this PR include some guidelines about it summing up the discussions and decisions please?

Rules about:

- what's defined in global CSS
- when should cursor pointer be used
- how to treat the ButtonBase component when it is a button and a link

Thank you!

danielroe:
yes! apologies for the slow review but I think we need to set some consensus here or I fear we'll have alternating PRs reverting/implementing cursor 😆

essenmitsosse:
> * how to treat the ButtonBase component when it is a button and a link

`ButtonBase` is always a button. `LinkBase` can look like a button.

CONTRIBUTING.md
```diff
@@ -356,6 +356,22 @@ Query parameters work as expected:
...
+
+**What counts as a link**
+
+A link is any element that change or show meaningful content:
+
+- **Not meaningful content:** burger menu, "likes" button - the user does not get substantive information from them.
+- **Meaningful content:** a dialog with a chart, auth button, tabs with new data—they convey new or important information.
+
+> [!IMPORTANT]
+> Always prefer implementing such changes as real links so they can be opened in a new tab, shared or reloaded, and so the same content is available at a stable URL.
```
knowler: Can we defer this “What counts as a link” bit for now?

CONTRIBUTING.md
```diff
@@ -356,6 +356,22 @@ Query parameters work as expected:
 <NuxtLink :to="{ name: 'compare', query: { packages: pkg.name } }">Compare</NuxtLink>
 ```
 
+### Cursor and navigation
+
+We prefer consistency in user everyday experience
+
+npmx uses `cursor-pointer` only for links. For all other elements, including buttons, use the default cursor (_or other appropriate cursors to show state_).
```
knowler: Can we both mention the CSS property-value combo: `cursor: pointer` and the utility class name? If we just had to go with one, I think just using the CSS property-value would be better.

alexdln: let's just CSS (`cursor: pointer`), good point

CONTRIBUTING.md
alexdln: Are you suggesting to remove this proposal and list entirely? (or even blockquote as well?)

alexdln: Then we won't close any questions. If there are any cases you'd like to skip, let's leave a note like "it's better to discuss such and such cases first"

knowler: Recording the decision that “we use `cursor: pointer` for links” is a sufficient start and I think the point of this PR, but defining _what a link is_ expands the scope which opens more discussion. I’m not sure I align with what you’ve got for what a link is, but I don’t want a disagreement/discussion to block what I think is the point of this PR from merging — we should add + workshop the link definition in a subsequent PR.

knowler: Resolved in [637ef09](https://github.com/npmx-dev/npmx.dev/pull/1261/commits/637ef091b027a6fc7cd57f60820922772f52bd38) after some discussion elsewhere.

alexdln: Updated it a bit to avoid including examples that haven't been agreed upon yet

\* We'll try to discuss and add examples more specifically later one by one

alexdln:
There are still some questions, but since we haven't made a final decision and the trigger-elements currently have different cursors, I'm merging it and leaving a new section in contributing guide that we use default cursor for buttons

We'll work through it further and, if necessary, we will update the contributing guide

Thanks for fixing ❤️

Issue #1192, opened 2026-02-08
The clickable area is too small in table search list


shuuji3:
In the table list, the current clickable area is only a part of the name cell dispite the entire table row is highlighted when hovering.

<img width="768" height="674" alt="Image" src="https://github.com/user-attachments/assets/02571359-8c5a-4b2a-b8ff-6dc94afdeb6f" />

example: https://npmx.dev/search?q=lunify.js

RYGRIT:
Hi @shuuji3, do you want the entire row to be clickable, or just a portion of it?

shuuji3:
We could make just the name column clickable, or the entire row, but I think it's more consistent to make the entire row clickable, given that the entire card is clickable in the other view mode (and the current highlight styling).

RYGRIT:
Yes, I think so too.

It seems l've discovered another problem 😂

https://github.com/user-attachments/assets/d26b933b-0d29-4c21-b655-41c079b2f47f

KirankumarAmbati:
I can work on this

RYGRIT:
@KirankumarAmbati Appreciate the offer, but I'm working on this. Might take a bit to push the PR, but I'm on it.

RYGRIT:
@shuuji3 Here's a quick demo of the fix in action.

https://github.com/user-attachments/assets/cc3153ed-ae0a-41a2-97ee-6cf3a62a5100

One thing I'd like your input on — for keyboard navigation, I opted to make the package name link the primary tab stop rather than adding tabindex to the entire row. This keeps tab navigation cleaner, especially in longer lists.

If you'd prefer the row itself to be focusable first (then tab into name/keywords), let me know and I can adjust.

shuuji3:
Yes, that makes sense and it looks good to me. I think if we make the entire row focusable first, we may remove the link on the name? But I'm not sure which is the best for accessibility perspective. Please create a PR once ready (or draft if it's work-in-progress), then we can correct the feedback then.

RYGRIT:
Thank you for your reply. I will submit my pull request (PR) later. Personally, I think the current behavior is better.

PR #1277, opened 2026-02-09
fix(ui): implement clickable package table rows

RYGRIT:
fixes: #1192

### DIFF
--- a/app/components/Package/TableRow.vue
+++ b/app/components/Package/TableRow.vue
@@ -44,7 +44,7 @@ const allMaintainersText = computed(() => {
 
 <template>
   <tr
-    class="group border-b border-border hover:bg-bg-muted transition-colors duration-200 focus-visible:ring-2 focus-visible:ring-fg focus-visible:ring-inset focus-visible:outline-none focus:bg-bg-muted"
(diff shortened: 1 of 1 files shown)


app/components/Package/TableRow.vue
```diff
@@ -198,3 +198,19 @@ const allMaintainersText = computed(() => {
     </td>
   </tr>
 </template>
+
+<style scoped>
```
Sybren-crwk, line 202: Why are we not using the Tailwind-ish utility classes in line in the template for this? 

I do believe this is not a too crazy thing since it only gets used once.

`after:content-[''] after:absolute after:inset-0 after:cursor-pointer focus-visible:after:outline-2 focus-visible:after:outline-offset-[-2px] focus-visible:after:outline-[var(--fg)]`

app/components/Package/TableRow.vue
```diff
@@ -198,3 +198,19 @@ const allMaintainersText = computed(() => {
...
+    position: absolute;
+    inset: 0;
+    cursor: pointer;
+  }
+
+  &:focus-visible::after {
+    outline: 2px solid var(--color-fg);
+    outline-offset: -2px;
+  }
+}
```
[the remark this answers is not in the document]
Sybren-crwk: I agree on this CSS variable usage. I'm also not finding `--color-fg` in the Uno CSS config.

app/components/Package/TableRow.vue
danielroe: this is also a good point

Issue #1244, opened 2026-02-08
Column selector go outside viewport on small screen


shuuji3:
On a small screen, the button wraps to the left side.

<details>
<summary>Screenshot</summary>
<img width="339" height="744" alt="Image" alt="Screenshot of search result opened column selector only shown part of right side" src="https://github.com/user-attachments/assets/47d15cbc-11df-4a01-a32a-83b0f0db5374">
</details>

RYGRIT:
I encountered this problem before. Due to my oversight, the same issue also appeared in other language versions. I saw some similar pull requests (PRs) that also addressed this problem (but only for certain languages).

I just tested it with different languages ​​and found that it works correctly in Japanese and Chinese environments. I suspect the same problem might exist in other language versions as well. Do you have any good solutions? If so, please let me know, and I will start working on fixing this issue later.

<details>
<summary>Chinese</summary>

<img width="339" height="744" alt="Image" src="https://github.com/user-attachments/assets/09dc01c2-6b56-40a5-b235-d6bc34341642" />
</details>

<details>
<summary>Japanese</summary>

<img width="344" height="744" alt="Image" src="https://github.com/user-attachments/assets/3e326cfb-cb33-4c72-b808-86d26839100d" />
</details>

<details>
<summary>Azerbaijani</summary>

<img width="346" height="741" alt="Image" src="https://github.com/user-attachments/assets/e42416d7-8cad-4026-a882-54143c2c9887" />
</details>

RYGRIT:
I'm considering hiding the label text on smaller screens (showing only the icon), but keeping the label accessible to screen readers. This would prevent text overflow issues in languages ​​with longer text strings. Do you think this approach is feasible?

shuuji3:
@RYGRIT sorry, I totally missed this comment by flood of notifications 😅 I think hiding the label is a good solution given each language has a different width for the left Relevance dropdown.

alexdln:
I think the main question here is how the dropdown works - there will always be a situation or language where this select appears at the beginning of a line (_for example in linked PR and 360px_)

Probably it should be implemented using `floating-ui`

PR #1284, opened 2026-02-09
fix(ui): hide text on mobile devices to prevent inconsistent display on small screens

RYGRIT:
related: #1244 

Currently, the "Columns" button text in the table view toolbar may cause layout overflow on small screens, especially in languages ​​with longer translated text.

My proposed solution is to display it as an icon button while ensuring that screen readers can still access the corresponding content.

What are your thoughts on this? If everyone agrees with my suggestion, I will proceed with refining this PR.

<details>
<summary>before</summary>
ac465657-ed1b-4f3b-bb0d-7cc0abab0b64
771c943b-9151-478b-ad24-08a32c364501
6782fff9-990e-4208-9650-a927ba578d8b
</details>

<details>
<summary>after</summary>
image
</details>

### DIFF
--- a/app/components/ColumnPicker.vue
+++ b/app/components/ColumnPicker.vue
@@ -77,7 +77,7 @@ function handleReset() {
       @click.stop="isOpen = !isOpen"
       classicon="i-carbon-column"
     >
-      {{ $t('filters.columns.title') }}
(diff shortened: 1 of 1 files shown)


ghostdevv (changes requested):
You interested in updating this? @RYGRIT

PR #1285, opened 2026-02-09
fix: redirect user to original path after user cancel oauth login

NandkishorJadoun:
## Description

This pr improves the authentication experience by handling cases where a user explicitly cancels the authorization request on the Bluesky/ATProto login page.

Previously, clicking "Cancel" would result in 401 "Handle not provided in query" error. This change introduces a check for the `access_denied` error and gracefully redirects the user back to their original location using the validated path stored in the session cookie.


server/api/auth/atproto.get.ts
```diff
@@ -70,6 +70,15 @@ export default defineEventHandler(async event => {
     handleResolver,
   })
 
+  const error = query.error
+
+  // user cancelled explicitly
+  if (error === 'access_denied') {
```
danielroe, line 76:
```suggestion
  // user cancelled explicitly
  if (query.error === 'access_denied') {
```

PR #1288, opened 2026-02-09
feat: add solid to framework list

danieljcafonso:
# What does it do?
Adds SolidJS to the framework list! Let me know if I went through the right steps or missed anything! 

Thank you for your time!

@joshuakgoldberg @eliancodes @atilafassina asked me to do it ❤️


carwack (approved):
Looks Solid to me!👌

PR #1297, opened 2026-02-09
feat: added embedded bluesky post for markdown files

Kai-ros:
This is imperfect work but I wanna make it available to others who might have more time and/or knowledge to fix the 2 outstanding issues:
- hydration mismatch with the client only component
- localization strings not working


lunaria/files/en-US.json
```diff
@@ -80,6 +63,10 @@
     "title": "Blog",
     "heading": "blog",
     "meta_description": "Insights and updates from the npmx community",
+    "atproto": {
+      "loading_bluesky_post": "Loading Bluesky post...",
+      "view_on_bluesky": "View this post on Bluesky."
```
whitep4nth3r:
```suggestion
      "view_on_bluesky": "View this post on Bluesky"
```

PR #1300, opened 2026-02-09
fix: locale aware build time in footer

shamilkotta:
Before
image

After:
image


danielroe (approved):

app/components/BuildEnvironment.vue
```diff
@@ -6,9 +6,9 @@ const { footer = false, buildInfo: buildInfoProp } = defineProps<{
   buildInfo?: BuildInfo
 }>()
 
-const { locale } = useI18n()
 const appConfig = useAppConfig()
 const buildInfo = computed(() => buildInfoProp || appConfig.buildInfo)
+const buildTime = computed(() => new Date(buildInfo.value.time))
```
danielroe, line 11: I don't think this is necessary - you should just be able to pass `buildInfo.time` into `<DateTime>`

PR #1303, opened 2026-02-10
feat: show translation progress in locale picker

serhalp:
image

Why:
- Increase up-front transparency of translation status
- Provide a passive call-to-action to come help us translate!

Inspired by https://opencollective.com/


danielroe:
what about instead having a button which opens a modal with all translation progress

whitep4nth3r:
> what about instead having a button which opens a modal with all translation progress

I like this idea.

Maybe next to the "help translate" we could have a progress bar of the total percentage of translated strings?

And then more info in the modal of each locale?

I've seen comments on bsky about how overwhelming sometimes the UI can be when we try to cram as much info into little areas as possible, just trying to avoid this. But it's a nice data point.

whitep4nth3r:
Just marking as draft as it's not ready yet and a couple of us are going down the list to see what we can review!

serhalp:
Thanks for the feedback, friends. I'll close this one since the proposed alternatives are different enough that there's nothing salvageable here. 🫡

PR #1310, opened 2026-02-10
feat(ui): generalize downloads chart and add social likes option

serhalp:
Generalize the download analytics chart into a trends chart that supports both downloads and likes, extensibly allowing for more facets in the future.

https://github.com/user-attachments/assets/cd998ead-98f6-44a8-b019-b6f12b222cab

- Add a facet selector (Downloads / Likes) shown on Compare page and in the single-package chart modal
- Add server endpoint and utility for fetching per-package likes evolution from ATProto

> [!NOTE]
> Full disclosure: I don't really know what I'm doing with the ATProto stuff. I tried to follow existing patterns. It seems to work.

### DIFF
--- a/lunaria/files/ar-EG.json
+++ b/lunaria/files/ar-EG.json
@@ -927,7 +927,8 @@
         "types_none": "الأنواع غير مضمّنة",
         "vulnerabilities_summary": "{count} ({critical} حرجة/{high} عالية)",
         "up_to_you": "الأمر متروك لك!"
-      }
+      },
+      "trends": {}
     }
   },
   "privacy_policy": {
--- a/lunaria/files/az-AZ.json
+++ b/lunaria/files/az-AZ.json
@@ -735,7 +735,8 @@
         "license": {},
         "vulnerabilities": {}
       },
-      "values": {}
+      "values": {},
+      "trends": {}
     }
   },
   "privacy_policy": {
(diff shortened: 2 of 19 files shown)


i18n/schema.json
serhalp, line 1: It's kinda nice to see this as a diff in PRs actually 😁

app/components/Package/TrendsChart.vue
```diff
@@ -1858,21 +1945,17 @@ const chartConfig = computed(() => {
           <div class="min-h-[260px]" />
```
graphieros, line 1945: I forgot about this placeholder.
We can probably have it fit the resolved size, to mitigate CLS when switching facets for the first time

app/components/Package/TrendsChart.vue
```diff
@@ -23,6 +23,9 @@ const props = defineProps<{
    */
   packageNames?: string[]
   createdIso?: string | null
+
+  /** When true, shows facet selector (e.g. Downloads / Likes). */
+  showFacetSelector?: boolean
```
Tobbe, line 28: Is this needed? Seems like it's never `false`

i18n/schema.json
danielroe: it is!

what genius built it?

danielroe (approved):

app/components/Package/TrendsChart.vue
serhalp: would you like to follow up on that? 😁

Issue #1254, opened 2026-02-09
Rename size prop for buttons and links


essenmitsosse:
We're using uno with the tailwind template in npmx, the most common format for specifying sizes is using letters is sm, md, lg, etc. A similar format is used for Nuxt-UI components. Therefore for better DX in npmx, it would be more convenient to use the same style (`<Button size="md">`) throughout our project.

For now we need to update the size prop in the button, link and input components.

_P.S. Don't forget to update the tests_

_Originally posted by @alexdln in https://github.com/npmx-dev/npmx.dev/pull/1247#discussion_r2780137461_

RYGRIT:
I working on this

RYGRIT:
@essenmitsosse @alexdln 

Hi everyone, I've submitted a draft PR for the LinkBase refactoring and also updated the test cases.

However, I'm not entirely sure if the current implementation fully meets the expectations of #1254. Could you please review it?

Any suggestions are welcome, thank you!

PR #1319, opened 2026-02-10
refactor(LinkBase/ButtonBase): add inline prop & rename size prop for buttons and links

RYGRIT:
resolves: #1254

### DIFF
--- a/app/components/Package/InstallScripts.vue
+++ b/app/components/Package/InstallScripts.vue
@@ -60,7 +60,7 @@ const isExpanded = shallowRef(false)
           <template v-if="installScripts.content?.[scriptName] && scriptParts[scriptName]">
             <template v-if="scriptParts[scriptName].prefix">
               {{ scriptParts[scriptName].prefix }}
-              <LinkBase :to="scriptParts[scriptName].link" text="sm">
+              <LinkBase :to="scriptParts[scriptName].link" size="sm">
                 {{ scriptParts[scriptName].filePath }}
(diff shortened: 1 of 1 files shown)


essenmitsosse:
Thanks for your work. Won't have time to fully review it until later. Just one thing:

I think `inline` was a more resonable default, with `block` being an optional prop. Inline is almost always fine, where as block is only needed in special cases, like when you want to apply `truncate`. I'd suggest to switch it back.

app/components/Link/Base.vue
```diff
@@ -50,47 +49,66 @@ const isLinkAnchor = computed(
...
+    case 'xs':
+      return 'text-xs'
+    case 'sm':
+      return 'text-sm'
+    case 'md':
+      return 'text-base'
+    case 'lg':
+      return 'text-lg'
+  }
+})
```
essenmitsosse, line 78: Setting a class like this doesn't work, because Uno won't pick these up. It only works "by accident", because these classes are all used somewhere else. Try setting `bg-indigo-300` here and it wouldn't be applied.

This needs to be inlined into the components class prop.

essenmitsosse:
The `primary`/`secondary` color variant for Links is now missing. Is that on purpose or part of the WIP?

ghostdevv:
I'm sorry this slipped through the cracks! I'm gonna merge #2069 instead as it's less conflicts to work with, but ty for the pr!

PR #1326, opened 2026-02-10
feat: add story viewer

### LINKED ISSUE #1298: Native stories viewer
I think it would be awesome if npmx.dev could show how the components of a UI library package look like. Luckily there's a sort-of standard for how to render UI components, and it's stories!

Imagine if a UI package like [React Aria](https://react-aria.adobe.com/react-aria-starter/?path=/docs/breadcrumbs--docs) or [Chakra UI use Storybook](https://storybook.chakra-ui.com) to document/develop/test their components, wouldn't it be great if users could inspect those components directly on npmx.dev? 

What I'm imagining, is a view very similar to the existing Code view, except a Stories/Storybook view. See these mock ups:

Image

Image

# But how?

I'm glad you asked! Luckily, there is already an API that would make this "fairly easy" to do.
Storybook has a little known feature (we should try to make it more known), that any npm package that has a `"storybook.url"` property in their `package.json` will be composed into projects that use that dependency. It is [used in the wild, but not much](https://github.com/search?q=/%22storybook%22%5Cs*:%5Cs*%5C%7B%5B%5E%7D%5D*%22url%22%5Cs*:%5Cs*%22%5B%5E%22%5D*%22%5B%5E%7D%5D*%5C%7D/+path:package.json+-is:fork&type=code) yet. An example is `@chakra-ui/react`:

https://npmx.dev/package-code/@chakra-ui/react/v/3.32.0/package.json#L36-L39

We could look that property up in the package, and if it's there, add the "Stories" button as shown above.

**Step 1 here would maybe just to do that, and make the button an external link to the Storybook.**

But we could also render all of this natively in npmx.dev. Storybooks are builtable, deployable SPAs (like the Chakra UI link above). All Storybooks have [an `/index.json`-file](https://storybook.chakra-ui.com/index.json) that is the source of truth for all the stories, their id's and human readable names. We would fetch that from the url to build the "story tree" on the left, similar to how we build the file tree in the code view. Then, navigating between each story would render the [story's isolated iframe](https://storybook.chakra-ui.com/iframe.html?globals=&args=&id=charts-bar-list--ascending&viewMode=story) in the center of the screen.

## Additional Considerations

This doesn't have to be Storybook-only, I'm sure this could work in a similar fashion with Histoire/Ladle as well, if they have similar constructs we could use. Both also use Stories as the terminology, so that would be compatible.
We could also use something more generic, like a new `"component-explorer"` field in `package.json`/`.npmxconfig` or similar. The reason I have proposed the `"storybook"`-field is because that is something that exists _today_ and already have use, so it would work from day-1 and it speaks in to npmx's general idea of being backwards compatible.

sacrosanctic:
closes #1298

I added both an integrated viewer and an external link just to see how it would feel.

The UI is really rough right now, and had a lot of assist from LLM, so I'll need to clean it up.

### DIFF
--- a/app/components/Storybook/FileTree.vue
+++ b/app/components/Storybook/FileTree.vue
@@ -26,7 +26,7 @@ function isNodeActive(node: StorybookFileTree): boolean {
 
 // Build route object for a story
 function getStoryRoute(node: StorybookFileTree): RouteLocationRaw {
-  if (node.type === 'story') {
+  if (node.type === 'story' && node.storyId) {
     return {
       name: 'stories',
       params: { path: props.basePath },
@@ -36,7 +36,7 @@ function getStoryRoute(node: StorybookFileTree): RouteLocationRaw {
   // For directories - navigate to first story in that directory
   if (node.type === 'directory') {
     const firstStory = getFirstStoryInDirectory(node)
-    if (firstStory) {
+    if (firstStory?.storyId) {
       return {
         name: 'stories',
         params: { path: props.basePath },
@@ -68,6 +68,48 @@ function getNodeIcon(node: StorybookFileTree): string {
 }
 
 const { toggleDir, isExpanded, autoExpandAncestors } = useStoryTreeState(props.baseUrl)
+
+const storyRefs = shallowReactive(new Map<string, HTMLElement>())
+
+function setStoryRef(storyId: string | null) {
+  return (el: Element | { $el?: Element } | null) => {
+    if (!storyId) return
+    if (!el) {
+      storyRefs.delete(storyId)
+      return
+    }
+    const target = '$el' in el ? el.$el : el
+    if (target instanceof HTMLElement) storyRefs.set(storyId, target)
+  }
+}
+
+function findStoryPath(tree: StorybookFileTree[], storyId: string): string | null {
+  for (const node of tree) {
+    if (node.type === 'story' && node.storyId === storyId) return node.path
+    if (node.type === 'directory' && node.children) {
+      const found = findStoryPath(node.children, storyId)
+      if (found) return found
+    }
+  }
+  return null
+}
+
+watch(
+  [() => props.currentStoryId, () => props.tree],
+  ([storyId, tree]) => {
+    if (!storyId || !tree.length || depth.value !== 0) return
+    const storyPath = findStoryPath(tree, storyId)
+    if (!storyPath) return
+    autoExpandAncestors(storyPath)
+    nextTick(() => {
+      const target = storyRefs.get(storyId)
+      if (target) {
+        target.scrollIntoView({ block: 'nearest', inline: 'nearest' })
+      }
+    })
+  },
+  { immediate: true },
+)
 
 // Handle directory click - toggle expansion and navigate to first story
 function handleDirectoryClick(node: StorybookFileTree) {
@@ -120,6 +162,7 @@ function handleDirectoryClick(node: StorybookFileTree) {
           block
           :style="{ paddingLeft: `${depth * 12 + 32}px` }"
           :classicon="getNodeIcon(node)"
+          :ref="setStoryRef(node.storyId)"
         >
           <span class="truncate">{{ node.name }}</span>
         </LinkBase>
--- a/app/pages/package-stories/[...path].vue
+++ b/app/pages/package-stories/[...path].vue
@@ -365,6 +365,7 @@ defineOgImageComponent('Default', {
         class="w-64 lg:w-72 border-ie border-border shrink-0 hidden md:block bg-bg-subtle sticky top-28 self-start h-[calc(100vh-7rem)] overflow-y-auto"
       >
         <StorybookFileTree
+          data-story-tree-scroll
           :tree="storybookTree"
           :current-story-id="currentStoryId"
           :base-url="`/package-stories/${packageName}/v/${version}`"
--- a/app/components/Storybook/MobileTreeDrawer.vue
+++ b/app/components/Storybook/MobileTreeDrawer.vue
@@ -59,6 +59,7 @@ watch(isOpen, open => (isLocked.value = open))
     <aside
       v-if="isOpen"
       class="md:hidden fixed inset-y-0 inset-is-0 z-50 w-72 bg-bg-subtle border-ie border-border overflow-y-auto"
+      data-story-tree-scroll
     >
       <div
         class="sticky top-0 bg-bg-subtle border-b border-border px-4 py-3 flex items-center justify-start"


app/components/Storybook/FileTree.vue
```diff
@@ -0,0 +1,129 @@
...
+// Build route object for a story
+function getStoryRoute(node: StorybookFileTree): RouteLocationRaw {
+  if (node.type === 'story') {
+    return {
+      name: 'stories',
+      params: { path: props.basePath },
+      query: { storyid: node.storyId },
+    }
+  }
+  // For directories - navigate to first story in that directory
```
JReinhold, line 36: I think instead we want to follow the existing navigation pattern in the Code view, where clicking a directory only expands it (and changes it icon) but doesn't select any stories - only navigating to a story actually selects it.
I know that Storybook itself does the latter, but here I think we want to be consistent with npmx instead.

app/components/Storybook/FileTree.vue
```diff
@@ -0,0 +1,129 @@
...
+      return getFileIcon(node.story.importPath)
+    }
+    // Default story icon
+    return 'i-vscode-icons-file-type-storybook'
+  }
+
+  return getFileIcon(node.name)
+}
+
+const { toggleDir, isExpanded, autoExpandAncestors } = useStoryTreeState(props.baseUrl)
```
JReinhold, line 70: I think `autoExpandAncestors` is the right idea, but it's not active at the moment.

app/pages/package-stories/[...path].vue
```diff
@@ -0,0 +1,428 @@
...
+        <div
+          v-if="currentStory"
+          class="sticky z-10 top-0 bg-bg border-b border-border px-4 py-2 flex items-center justify-between"
+        >
+          <div class="flex items-center gap-3 text-sm">
+            <span class="text-fg-muted">{{ $t('stories.story') }}</span>
+            <span class="text-fg font-medium">{{ currentStory.path }}</span>
+            <span v-if="currentStory.story?.tags" class="text-fg-subtle">
+              {{ currentStory.story.tags.join(', ') }}
+            </span>
```
Comment (JReinhold) [inline app/pages/package-stories/[...path].vue:387]: I don't think tags is valuable information here, in 99 % of the cases it's internal information.

app/pages/package/[[org]]/[name].vue
```diff
@@ -65,6 +70,20 @@ const { data: readmeData } = useLazyFetch<ReadmeResponse>(
   { default: () => ({ html: '', md: '', playgroundLinks: [], toc: [] }) },
 )
 
+const playgroundLinks = computed(() => [
+  ...readmeData.value.playgroundLinks,
+  ...(packageJson.value?.storybook
```
Comment (JReinhold) [inline app/pages/package/[[org]]/[name].vue:136]: packages can have a `storybook` property without having a `storybook.url` property, so we need to check for that here instead. Storybook addons have metadata in the same `storybook` property, but with no `url`, eg. https://npmxdev-git-fork-sacrosanctic-story-viewer-npmx.vercel.app/package-code/@storybook/addon-docs/v/10.2.8/package.json#L118-L124

Comment (sacrosanctic) [reply app/pages/package/[[org]]/[name].vue:136]: Do you know where I can get source the type definition for `package.json#storybook` and `/index.json`?

Comment (yannbf) [reply app/pages/package/[[org]]/[name].vue:136]: You can find the index type here:
`import { StoryIndex } from 'storybook/internal/types';`

As for the package.json definition, AFAIC there aren't any, we don't type it ourselves
https://github.com/storybookjs/storybook/blob/d29589dfa45bcb218479b013334747ed4655b2b9/code/core/src/common/utils/get-storybook-refs.ts#L28

yannbf:
This is incredible!! I was testing out in a project and noticed that the MDX icon works for some entries and not for others, e.g. both of these should have the mdx icon:

https://npmxdev-git-fork-sacrosanctic-story-viewer-npmx.vercel.app/package-stories/@storybook/design-system/v/7.15.19?storyid=typography-base--docs

https://npmxdev-git-fork-sacrosanctic-story-viewer-npmx.vercel.app/package-stories/@storybook/design-system/v/7.15.19?storyid=typography-styles--docs

They're both `type: "docs"` in the index.

app/composables/useStoryTreeState.ts
```diff
@@ -0,0 +1,36 @@
+import { computed } from 'vue'
+import { useState } from '#app'
+
+export function useStoryTreeState(baseUrl: string) {
+  const stateKey = computed(() => `npmx-story-tree${baseUrl}`)
```
huang-julien, line 5: you don't need `computed` without any reactive dependency. 
Did you mean `MaybeRefOrGetter<string>` instead of `baseUrl: string` as param ? Then you'd wrap with `toValue` -> `computed(() => `npmx-story-tree${toValue(baseUrl)}`)`

app/pages/package-stories/[...path].vue
```diff
@@ -0,0 +1,428 @@
...
+  return url
+})
+
+// Fetch Storybook index data
+const { data: storybookData, status: storybookStatus } = useLazyFetch<{
+  v: number
+  entries: Record<string, StorybookEntry>
+}>(
+  () => {
+    if (!packageJson.value?.storybook?.url) return ''
```
Comment (huang-julien) [inline app/pages/package-stories/[...path].vue]: returning `''` here will still trigger the fetch

app/pages/package-stories/[...path].vue
```diff
@@ -0,0 +1,428 @@
...
+
+// URL pattern for version selector - maintain current story if available
+const versionUrlPattern = computed(() => {
+  const base = `/package-stories/${packageName.value}/v/{version}`
+  // Directly include the current storyid if available
+  return currentStoryId.value ? `${base}?storyid=${currentStoryId.value}` : base
+})
+
+// Fetch package.json to get Storybook URL
+const { data: packageJson } = useFetch<{ storybook: { title: string; url: string } }>(() => {
```
Comment (huang-julien) [inline app/pages/package-stories/[...path].vue:60]: we need to `await` here.
I saw there's another `useFetch` at L66. We could have a single `useAsyncData()` that handles both requests. This way, error handling becomes a bit more easier.

app/pages/package-stories/[...path].vue
```diff
@@ -0,0 +1,428 @@
...
+
+// Storybook iframe URL
+const iframeUrl = computed(() => {
+  if (!packageJson.value?.storybook?.url || !currentStoryId.value) return ''
+  return `${packageJson.value.storybook.url}/iframe.html?id=${currentStoryId.value}`
+})
+
+// SEO Meta
+useHead({
+  title: () => {
```
Comment (huang-julien) [inline app/pages/package-stories/[...path].vue:224]: title is already handled at line 229. L219 to 226 can be removed

huang-julien:
Looks really cool ! :) Got a few comments to improve the PR

ghostdevv:
will mark as draft while you guys are working on it :pray:

yannbf:
After discussing with the npmx team, a simpler approach (just the playground link) was used. This was extracted into this PR which is merged: https://github.com/npmx-dev/npmx.dev/pull/1624 

This PR should remain as a potential exploration of a solution in the future, but should probably be closed. 

I'll leave two gifs here for record sake:
### The UX provided by this PR

A `stories` option on the package page, bringing you to a separate page that embeds the stories of the deployed storybook of that given package.




### An alternative proposal

Instead of having a dedicated page, there would be a list of stories in the sidebar of the package page, and upon hovering the story, you would see a little preview thumbnail.

Issue #1322, opened 2026-02-10
DPop key should never leave the server


matthieusieben:
DPoP keys are meant to protect the access-token in transit (from attacks like heartbleed, etc.). This is only efficient if the DPoP key is never sent over the wire.

The current implementation stores the OAuth session data in (encrypted, but still) cookies, which goes against this:

https://github.com/npmx-dev/npmx.dev/blob/b73edafe49f99a97487b91b586bd48253a18342c/server/api/auth/atproto.get.ts#L64-L65

The same is also true for "state data". The State store is used to keep track of the state of a particular oauth flow. In particular, it should allow to prevent replays. Being stored in cookies, replays can be made possible.

Both SessionStore and StateStore should be using backend databases to store their data.

matthieusieben:
Using cookies to store the oauth data has the advantage of ensuring that the right browser is using performing the oauth callback.

IF the storage of oauth data is moved to a backend database (as it should), the oauth callback logic should enforce that the right browser is performing the redirect.

This is what https://github.com/npmx-dev/npmx.dev/pull/1327 does.

matthieusieben:
Note that using cookies to store session data means that multiple parallel incoming HTTP requests might trigger multiple refreshes of the same oauth refresh token. Normally, this is protected using the `requestLock` (see [`getOAuthLock`](https://github.com/npmx-dev/npmx.dev/blob/24b9d92f96d557f5b48b39c8818a19802fea391e/server/utils/atproto/lock.ts#L54)). However, because all parallel request will have the same exact cookie value, the refresh will be performed using the same refresh token multiple times. This will cause the AS to invalidate the OAuth session. Reading oauth session data from a central place allows to ensure proper handling of parallel refresh attemts.

matthieusieben:
Note that an alternative solution would be to use browser side oauth sessions (instead of server side), using the `OAuthClientBrowser`.

fatfingers23:
Thanks for the PR earlier and this! After the discussion in #1327 do you still think it would be best to move it back to the backend again especially since we are going to add it to be a confidential client for the longer sessions? 


edit: Took some more time to read over it all and look at some previous OAuth implementation's I've done. Going get it all moved back server side again. Thanks again for checking over everything! Really want to make sure I nail the atproto stuff for npmx

matthieusieben:
Yes, the PR was complementary to fixing this issue!

matthieusieben:
Feel free to ping me as reviewer

PR #1327, opened 2026-02-10
fix: properly bind OAuth state data to a browser session

matthieusieben:
Currently, the OAuth state data (redirect path) is stored in a dedicated cookie. However, the `NodeOAuthClient` already provides a way to store this state data in the OAuth state store.

Relatedly, the OAuth state data should be stored server side (see https://github.com/npmx-dev/npmx.dev/issues/1322). Once it does, the server should verify that the browser for which it is performing the OAuth callback is indeed the browser that initiated that OAuth request.

This PR improves the handling of the OAuth state by:

1) Using the OAuth client state store to store OAuth state data (redirect uri), instead of a custom cookie.
2) Ensuring that the OAuth state data matches the right device (using an ephemeral session cookie)

### DIFF
--- a/server/api/auth/atproto.get.ts
+++ b/server/api/auth/atproto.get.ts
@@ -104,10 +104,14 @@ export default defineEventHandler(async event => {
         throw error
       }
     } catch (error) {
-      // user cancelled explicitly
-      if (query.error === 'access_denied' && error instanceof OAuthCallbackError && error.state) {
+      if (error instanceof OAuthCallbackError && error.state) {
+        // Always decode the state, to clean up the cookie
         const state = decodeOAuthState(event, error.state)
-        return sendRedirect(event, state.redirectPath)
+
+        // user cancelled explicitly
+        if (query.error === 'access_denied') {
+          return sendRedirect(event, state.redirectPath)
+        }
       }
 
       const message = error instanceof Error ? error.message : 'Authentication failed.'
@@ -124,8 +128,7 @@ type OAuthStateData = {
   redirectPath: string
 }
 
-const SID_COOKIE_PREFIX = 'atproto_oauth_sid'
-const SID_COOKIE_VALUE = '1'
+const OAUTH_REQUEST_COOKIE_PREFIX = 'atproto_oauth_req'
 
 /**
  * This function encodes the OAuth state by generating a random SID, storing it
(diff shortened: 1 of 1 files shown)


server/api/auth/atproto.get.ts
```diff
@@ -1,55 +1,20 @@
...
+import type { DidString } from '@atproto/lex'
 import { Client } from '@atproto/lex'
 import * as com from '#shared/types/lexicons/com'
 import * as app from '#shared/types/lexicons/app'
 import { isAtIdentifierString } from '@atproto/lex'
 // @ts-expect-error virtual file from oauth module
 import { clientUri } from '#oauth/config'
 
-//I did not have luck with other ones than these. I got this list from the PDS language picker
-const OAUTH_LOCALES = new Set(['en', 'fr-FR', 'ja-JP'])
```
matthieusieben, line 16: Other Atproto Authorization Server implementation exists out there. We should let the AS decide how to handle the locale passed in, based on what it supports.

server/api/auth/atproto.get.ts
```diff
@@ -95,79 +65,211 @@ export default defineEventHandler(async event => {
...
+    maxAge: 60 * 5,
+    httpOnly: true,
+    // secure only if NOT in dev mode
+    secure: !import.meta.dev,
+    sameSite: 'lax',
+    path: event.path.split('?', 1)[0],
+  })
+  return JSON.stringify({ data, sid })
+}
+
```
danielroe, line 174: `useSession` generates an encrypted, stateless cookie that validates it was issued by the server

doesn't that provide enough protection here?

matthieusieben: It does. This way of doing things allows to keep the "oauth sid" (the name might be poorly chosen) ephemeral without having to cleanup the `h3` cookie.

matthieusieben: For example, if the user initiates an oauth flow (causing the `useSession` cookie to contain that "sid" value), then the user performs a "back" navigation. In that case, the "sid" would probably not be removed from the `useSession` cookie, unless every time we read that cookie, we performed additional logic to clean things up. I wanted to avoid doing that.

matthieusieben: I tried to explain this in the JSdoc right above.

PR #1328, opened 2026-02-10
perf: try extracting css on a per-module basis

danielroe:


danielroe:
~> https://github.com/npmx-dev/npmx.dev/pull/1347

PR #1332, opened 2026-02-10
draft: check dist-chunk

alexdln:
Just testing


alexdln:
re-created - #1347 - with optimization for icons

PR #1338, opened 2026-02-10
fix: only show estimation segment in last datapoint

### LINKED ISSUE #1318: Only show estimation on the last datapoint in download charts
Currently for incomplete periods, both the estimation and the last datapoint are visible.

Only show the estimation in this case:

- swap the last datapoint with the estimation in the dataset passed to VueUiXy
- remove the nice scale adaptations that will not be necessary anymore as the estimation will be part of the chart's native data
- remove the opacifying line overlay currently positioned on the last datapoint
- add an overlay to cover the last segment, under the dashed segment already implemented

graphieros:
closes #1318

- Remove falling segment for partial month / year

image

### DIFF
--- a/app/components/Package/TrendsChart.vue
+++ b/app/components/Package/TrendsChart.vue
@@ -848,82 +848,18 @@ const chartData = computed<{
   return { dataset, dates }
 })
 
+const normalisedDataset = computed(() => {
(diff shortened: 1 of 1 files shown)


patak-cat:
Looks amazing! at least from a functionality POV

danielroe (approved):

PR #1339, opened 2026-02-10
feat: replace links in markdown instead of via javascript

essenmitsosse:
Currently all links in the readme are intercepted via JavaScript and if they go to npmjs.com redirected to npmx.dev. This has several problems, most of which stem from the fact, that the link in the URL is not the link we are going to.

This PR:
- replaces the link in the generated HTML ~and removes the JavaScript handler~ (JavaScript handler is still needed for code copy button)
- adds an indicator for external link (aligned with regular links)
- aligns the link styling with regular links)

Good example readme: [/package/is-odd](https://npmxdev-git-fork-essenmitsosse-replace-links-in-mar-d9d152-npmx.vercel.app/package/is-odd)

Screenshot 2026-02-10 at 15 25 23

### DIFF
--- a/app/components/Readme.vue
+++ b/app/components/Readme.vue
@@ -3,7 +3,6 @@ defineProps<{
   html: string
 }>()
 
-const router = useRouter()
 const { copy } = useClipboard()
 
 // Combined click handler for:


server/utils/readme.ts
```diff
@@ -368,6 +399,12 @@ ${html}
     const isExternal = resolvedHref.startsWith('http://') || resolvedHref.startsWith('https://')
     const relAttr = isExternal ? ' rel="nofollow noreferrer noopener"' : ''
     const targetAttr = isExternal ? ' target="_blank"' : ''
+    const isExternalIcon = isExternal
+      ? `<span
+      class="i-carbon:launch inline-block rtl-flip ms-1 size-[1em] opacity-50"
+      aria-hidden="true"
+    ></span>`
+      : ''
```
alexdln: What do you think about doing it with `::after` on the client side and just adding the attribute here (_or use_ `_target=blank`)?

There will be less HTML and classes will definitely be collected by uno

essenmitsosse: Good call!

alexdln (approved):

alexdln:
@essenmitsosse Just remembered one more weird behavior with this "new tab" markers

image

alexdln:
@essenmitsosse It seems that if you specify for `after`: `display: inline` and `content: "__"` it will work well, but it needs to be tested

essenmitsosse:
@alexdln can't make it inline because, then the icon does not work. I don't think there is a robust solution that doesn't make the problem worse then it currently is.

alexdln:
@essenmitsosse Have you checked by adding `content: "__"` (2x "_")?

alexdln:
image

danielroe:
the JS handler will still be needed for client-side navigation, which will be more performant, but you can update it to check if window.location is the same as the href origin.

essenmitsosse:
> the JS handler will still be needed for client-side navigation, which will be more performant, but you can update it to check if window.location is the same as the href origin.

@danielroe what do you mean by that? Why do we need a JS handler for client side navigation?

danielroe:
because nuxt doesn't listen to clicks on `<a>` elements, only on `<NuxtLink>`. so even though you are replacing the href, these links will result in a full refresh of the page, with a full server round trip

essenmitsosse:
> @essenmitsosse Have you checked by adding `content: "__"` (2x "_")?

1. now it works, I don't know why
2. WHAT KIND OF SORCERY IS THIS???

essenmitsosse:
> because nuxt doesn't listen to clicks on `<a>` elements, only on `<NuxtLink>`. so even though you are replacing the href, these links will result in a full refresh of the page, with a full server round trip

Aw shucks!

alexdln:
No sorcery, just uno adds a mask with icon under the current container. And since you had an inline with no content, its width was 0. So you just needed to increase the width.

You can't use width for an inline, but you can use content that won't be visible and preferably won't interfere with the mask and not wrapping

essenmitsosse:
> No sorcery, just uno adds a mask with icon under the current container. And since you had an inline with no content, its width was 0. So you just needed to increase the width.
> 
> You can't use width for an inline, but you can use content that won't be visible and preferably won't interfere with the mask and not wrapping

If that is not sorcery, what is then. 😂 But glad it works!

essenmitsosse:
> because nuxt doesn't listen to clicks on `<a>` elements, only on `<NuxtLink>`. so even though you are replacing the href, these links will result in a full refresh of the page, with a full server round trip

@danielroe ✅ 

I also added a check for modifier keys, so `CTRL` + click for new tab works as expected, which it didn't before.

app/components/Readme.vue
```diff
@@ -48,20 +52,11 @@ function handleClick(event: MouseEvent) {
   if (!href) return
 
   // Handle relative anchor links
-  if (href.startsWith('#')) {
+  if (href.startsWith('#') || href.startsWith('/')) {
     event.preventDefault()
     router.push(href)
```
danielroe: let's use `navigateTo(href)` - there's currently a little jump which happens that _might_ be resolved with it....

essenmitsosse: Done.

PR #1344, opened 2026-02-10
fix: keep package name visible

iiio2:
Look, our engines in vitest is really long. For that reason the package name is not visible in mobile version due to long values. I think we can improve it.

Before:
wrong

After:
correct


iiio2:
Thanks @43081j .

PR #1351, opened 2026-02-10
ci: only run mirror to tangled on npmx repo

jonathanyeong:
This will skip running this github action on forks which will always fail ([example](https://github.com/jonathanyeong/npmx.dev/actions/runs/21879526080/job/63157784276)).

I think it's okay to limit to github repository since the tangled remote is also hardcoded to `npmx`:

```
          git remote add tangled git@tangled.org:npmx.dev/npmx.dev
```

Github actions doc: https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-jobs-with-conditions

### DIFF
--- a/.github/workflows/mirror-tangled.yml
+++ b/.github/workflows/mirror-tangled.yml
@@ -13,6 +13,7 @@ permissions:
 jobs:
   mirror:
     name: 🕸️ Mirror to Tangled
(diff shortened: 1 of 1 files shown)


danielroe (approved):

jonathanyeong:
Confirmed job skipped on my fork https://github.com/jonathanyeong/npmx.dev/actions/runs/21881062641

Issue #1340, opened 2026-02-10
Scroll to bottom should show footer


vinnymac:
I noticed on mobile if I scroll to the bottom of the page, the floating buttons are in the way. It'd be nice if they got "stuck" above the footer area so I can be able to tap the sha link in the footer.

![image](https://github.com/user-attachments/assets/41f669e5-9840-484b-b018-5f9b1f6ee602)

Currently I can't tap or read the footer on mobile. I've only tested on one device, so maybe it works for others.

Note: I was using Mobile Safari, apologies in advance 😆

vinnymac:
Fix works for me on my device, great work 👍🏼

PR #1352, opened 2026-02-10
fix: scroll to bottom should show footer

ackzell:
[https://github.com/user-attachments/assets/2c99f960-f2b8-45ca-9a7e-c982ba969d00](https://github.com/user-attachments/assets/2c99f960-f2b8-45ca-9a7e-c982ba969d00)

My attempt at fixing #1340 

Using an intersection observer to determine the footer would be hidden behind the ButtonGroup, then re-positioning the buttons when on mobiled

### DIFF
--- a/app/components/Compare/FacetCard.vue
+++ b/app/components/Compare/FacetCard.vue
@@ -101,14 +101,14 @@ function getShortName(header: string): string {
 
         <!-- Package name -->
         <span
-          class="relative font-mono text-xs text-fg-muted truncate flex-shrink min-w-0"
+          class="relative font-mono text-xs text-fg-muted truncate flex-shrink-0"
           :title="headers[index]"
         >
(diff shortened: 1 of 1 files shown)


app/pages/package/[[org]]/[name].vue
```diff
@@ -40,13 +41,63 @@ function checkHeaderPosition() {
...
+    entries => {
+      const entry = entries[0]
+      if (!entry) return
+      navExtraOffset.value = entry.isIntersecting ? entry.intersectionRect.height : 0
+    },
+    { threshold: thresholdValues },
+  )
+  footerObserver.observe(footerEl)
+}
+
```
Comment (danielroe) [inline app/pages/package/[[org]]/[name].vue]: what about https://vueuse.org/core/useIntersectionObserver ?

Comment (ackzell) [reply app/pages/package/[[org]]/[name].vue]: duh 😅 

I'll update it.

PR #1353, opened 2026-02-10
fix: anonymous missing template or render function warning

jonathanyeong:
This PR also makes some other fixes:
- Fix embed not being loaded because of the change to `BLUESKY_API` which now omits the `xrpc` url.
- Refactor bluesky author profiles with comments and file name changes to clarify how it connects with the server side API call.

> [!NOTE]
> I'm adding a new lex type  so you will need to rerun `pnpm generate:lexicons` or `pnpm install` to pull in the new types. 

### Anonymous missing template warning

Before the change, console shows the missing template warnings:

<details><summary>Screenshot</summary>

console showing missing template warnings

</details>

After changes, no errors in the console:

<details><summary>Screenshot</summary>

console showing no warnings related to missing template

</details>

### Bluesky embeds

Before the change, my bluesky embeds were not loading with a 502 error:

<details><summary>Screenshot</summary>

broken bluesky embed on the blog post with a 502 error in the console

</details>

After the change, I can see the bluesky embed:
<details><summary>Screenshot</summary>

fixed bluesky embed on the blog post after changes

</details>

### Author profile images
Tested that [OG image still loads author profiles](https://npmxdev-git-fork-jonathanyeong-feat-blog-fe-fixes-npmx.vercel.app/__og-image__/static/blog/alpha-release/og.png)


app/components/EmbeddableBlueskyPost.vue
jonathanyeong, line 1: By removing `.client` this component will now appear during SSR for unplugin-vue-markdown to pick up when it's compiling the blog posts. But the `<ClientOnly>` tag should mean it's only rendered by the browser.

Kai-ros: - Feel free to correct my Nuxt/Vue ignorance on anything I say.
- My understanding is that Nuxt's `.client.vue` suffix is equivalent to the Nuxt `<ClientOnly>` wrapper component, so they're interchangeable in this regard.
https://nuxt.com/docs/4.x/directory-structure/app/components#client-components
- Is the only reason to move the component to server-side to ensure it is picked up by `unplugin-vue-markdown `?
- The `app/plugins/bluesky-embed.ts` below should already be doing the work of registering the component during the build process.

jonathanyeong: > My understanding is that Nuxt's .client.vue suffix is equivalent to the Vue <ClientOnly> wrapper component, so they're interchangeable in this regard.

They are very similar, but adding `.client` to the component means the entire component won't be available during SSR. While when we specify `ClientOnly` tag it's only applied at the usage site. So the component is still available during SSR with `ClientOnly`.

> Is the only reason to move the component to server-side to ensure it is picked up by unplugin-vue-markdown ?

Yup! 

> The app/plugins/bluesky-embed.ts below should already be doing the work of registering the component during the build process.

That's true, my change doesn't do anything regarding registering the component. But we still need to import the plugin during ssr rather than only client side.

Side note: I was looking at [unplugin-vue-components](https://github.com/unplugin/unplugin-vue-components) and this might be something we can add in the future so we don't have to explicitly import a global component.

Kai-ros: - I'm not sure I understand the distinction well enough to see the value add here, but if it satisfies a need or resolves a bug, then send it!
- I tried to make it work with `unplugin-vue-components` in an earlier attempt but it was unsuccessful so more power to you if you get it to.

whitep4nth3r:
Thank you!!!!!!!

whitep4nth3r:
I'm merging this. We can iron more things out if needed from the feature branch!

PR #1357, opened 2026-02-10
fix: show correct install size for native packages

gameroman:
## Before

image

image

## After

image

image


gameroman:
is github serious lol? why does it trim titles at the end of the title correctly automatically but not at the start

PR #1358, opened 2026-02-10
fix: search hydration errors

alexdln:
There was a hydration error that caused re-rendering and repeated requests.

There were three problems:

1. For Algolia search, if a query was processed on the server or retrieved from the cache, there was no place on the client to add suggestions, so they disappeared. Fixed it by returning them to `useLazyAsyncData` and, when ready, I re-assign them into suggestions ref.
2. For search method configuration - we simply didn't know the user's preferred method on the server. So the server always worked with algolia, and the client could then rehydrate with npm. Fixed by storing this section additionally in a cookie and reusing it from there
3. npm suggestions had a similar issue to the first one: the logic was only executed on the server, but never executed on the client and, as a result, server send suggestion, but client missed this information (*that's why this option doesn't work at all in production right now*). Refactored the npm suggestions retrieval to "useLazyAsyncData" as well

### DIFF
--- a/app/components/Header/SearchBox.vue
+++ b/app/components/Header/SearchBox.vue
@@ -16,13 +16,11 @@ const emit = defineEmits(['blur', 'focus'])
 const router = useRouter()
 const route = useRoute()
 const { searchProvider } = useSearchProvider()
-// The actual search provider (from URL, used for API calls)
-const searchProviderParam = computed(() => {
+const searchProviderValue = computed(() => {
   const p = normalizeSearchParam(route.query.p)
-  if (!p && searchProvider.value === 'npm') return 'npm'
-  return p === 'npm' ? 'npm' : 'algolia'
+  if (p === 'npm' || searchProvider.value === 'npm') return 'npm'
+  return 'algolia'
 })
-const searchProviderValue = computed(() => searchProviderParam.value || searchProvider.value)
 
 const isSearchFocused = shallowRef(false)
 
--- a/app/components/SearchProviderToggle.client.vue
+++ b/app/components/SearchProviderToggle.client.vue
@@ -1,12 +1,12 @@
 <script setup lang="ts">
 const route = useRoute()
 const router = useRouter()
-const searchProviderParam = computed(() => {
-  const p = normalizeSearchParam(route.query.p)
-  return p === 'npm' ? 'npm' : 'algolia'
-})
 const { searchProvider } = useSearchProvider()
-const searchProviderValue = computed(() => searchProviderParam.value || searchProvider.value)
+const searchProviderValue = computed(() => {
+  const p = normalizeSearchParam(route.query.p)
+  if (p === 'npm' || searchProvider.value === 'npm') return 'npm'
+  return 'algolia'
+})
 
 const isOpen = shallowRef(false)
 const toggleRef = useTemplateRef('toggleRef')
--- a/app/composables/npm/useOrgPackages.ts
+++ b/app/composables/npm/useOrgPackages.ts
@@ -11,12 +11,12 @@ import { mapWithConcurrency } from '#shared/utils/async'
  */
 export function useOrgPackages(orgName: MaybeRefOrGetter<string>) {
   const route = useRoute()
-  const searchProviderParam = computed(() => {
-    const p = normalizeSearchParam(route.query.p)
-    return p === 'npm' ? 'npm' : 'algolia'
-  })
   const { searchProvider } = useSearchProvider()
-  const searchProviderValue = computed(() => searchProviderParam.value || searchProvider.value)
+  const searchProviderValue = computed(() => {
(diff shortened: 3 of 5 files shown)


userquin:
why at search page with nuxt on input, pressing enter navigates to `/package/nuxt`? maybe because exact search entry found?

alexdln:
@userquin looks like the same behavior in production now already, I can look at it as the next issue (tomorrow 🫠)

app/composables/useSettings.ts
```diff
@@ -115,11 +115,18 @@ export function useAccentColor() {
...
+  })
   const { settings } = useSettings()
 
   const searchProvider = computed({
-    get: () => settings.value.searchProvider,
+    get: () => (cookie.value === 'npm' ? 'npm' : settings.value.searchProvider),
     set: (value: SearchProvider) => {
+      cookie.value = value
       settings.value.searchProvider = value
     },
```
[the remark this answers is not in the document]
alexdln: It is always default option, so this condition works better to us

app/composables/npm/useSearch.ts
```diff
@@ -306,121 +326,143 @@ export function useSearch(
...
+            availability = { name: trimmed, available: !exists }
+            packageAvailability.value = availability
+          }
+        })
+        .catch(() => {
+          availability = null
+        })
+    } else {
+      availability = null
+    }
```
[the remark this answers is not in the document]
alexdln: This is a copy of the previous behavior. Fixed, good catch

app/composables/useSettings.ts
```diff
@@ -115,11 +115,18 @@ export function useAccentColor() {
  * Composable for managing the search provider setting.
  */
 export function useSearchProvider() {
+  const cookie = useCookie('search-provider', {
+    secure: true,
+    sameSite: 'strict',
+    maxAge: 60 * 60 * 24 * 30,
+    path: '/',
+  })
```
danielroe: ideally we wouldn't have a cookie here, so that we could cache the search page (by query key).

I think we could show algolia data in initial load, but refresh on client-side with npm data if that's what they prefer.

wdyt?

alexdln: This will be a very slow behavior. It will first download the full HTML, then clean everything up and start loading npm...

It feels like in this case - this option will be an example of a bad experience and will rather harm the service than give the user an option

alexdln: But I had one pretty crazy idea of ​​adding the provider to the query before this approach. Is that suitable for us?

So, if the user has npm selected by default, every time they visit the search page, the provider will be added along with the query. But this won't work if they go directly to the search...

alexdln: Done. This setting is now just a preference. If the user accesses the search via any mechanism within the site, the provider will also be added to the URL. Otherwise, the page will continue to use algolia until the user switches to npm.

danielroe: hm. so I was thinking that algolia is enabled by default, so this is the edge case of the user preferring npm _and_ hard reloading the search page rather than using client-side navigation.

the other reason was that algolia is much faster and only requires one http request rather than three, so we are more performant by skipping the npm search.

(but I guess if you have npmx as a search engine in the browser, and prefer npm, you might hit this more often than I originally expected)

we could also prerender the search page and do the fetching entirely on the client side. this would mean no danger of mismatch, very fast initial response, no double fetching....

alexdln: Yes, definitely, rare case, but sometimes the results don't load and the user reloads the page and gets one more bad experience.

I think the current solution is the most optimal in terms of scenario coverage. What do you think about latest changes?


UPD: updated a bit, now it works as I planned and is unnoticeable for the user

app/composables/npm/useSearch.ts
```diff
@@ -306,121 +329,146 @@ export function useSearch(
...
+    [
+      () => asyncData.data.value?.packageAvailability,
+      () => npmSuggestions.data.value.packageAvailability,
+    ],
+    ([algoliaPackageAvailability, npmPackageAvailability]) => {
+      if (algoliaPackageAvailability || npmPackageAvailability) {
+        packageAvailability.value = algoliaPackageAvailability || npmPackageAvailability
+      }
+    },
+    { immediate: true },
```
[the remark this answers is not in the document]
alexdln: They will stay with default values

app/composables/useSettings.ts
alexdln: Another nice bonus of the current loading approach is that we now cache suggestions. I'd like to keep this logic anyway - it improves behavior for both providers.

And I think question now only in the routing behavior with the npm provider. I can remove/change it, but to prevent the server from returning results under algolia, we'll still need to let the server know it's npm

@danielroe check the current behavior please, perhaps it already solves the problems

PR #1366, opened 2026-02-11
feat: oauth server side and confidential client

### LINKED ISSUE #850: feat: atproto oauth confiedential client
Currently we have what is called a public OAuth client with the overall session lifetime of 2 weeks before logout. We can implement a confidential client for much longer. I've written a bit on that process [here](https://marvins-guide.leaflet.pub/3mbfvey7sok26/l-quote/38_0-39_547#38_0). It would also be nice if we could load in the JWKs from the multibase format since they can be generated with `goat key generate`. But the final solution may look more like [this](https://tangled.org/baileytownsend.dev/atproto-sveltekit-template/blob/main/bin/gen-jwk.js). The important thing is that the keys are kept private and out of the public's eye.

### LINKED ISSUE #870: feat: atproto oauth session refresh, testing, and edgecase handling

### LINKED ISSUE #1322: DPop key should never leave the server

fatfingers23:
This moves the OAuth session/state store back server-side. This should resolve #850, resolves #1322, and I think we can go ahead and close #870 and create new issues on new bugs

- Moves the oauth stores back to the backend to add more security and hopefully allow the oauth lock to do it's thing. The session stores holds a randomly generated UUID in the encrypted cookie that is combined with the users did to hopefully give each browser a unique session and does not cause collisions on token refresh.
- Added the confidential client since users will need to relog for this chaneg already. We will have to set the env variable `NUXT_OAUTH_JWK_ONE` on production before we do the next release, or find a better home for it. This value can be gotten from `pnpm run generate:jwk` and is a secret. This will hopefully allow oauth sessions to last much longer than 2 weeks
- One time redirect to relogin if the old oauth session store is being used to help those upgrade those logins to use the new store and conf client

@matthieusieben was kind enough to offer a review of this PR, so I am going to tag him. Thanks again for looking this over and for the other atproto help!

### DIFF
--- a/modules/cache.ts
+++ b/modules/cache.ts
@@ -4,8 +4,6 @@ import { provider } from 'std-env'
 
 // Storage key for fetch cache - must match shared/utils/fetch-cache-config.ts
 const FETCH_CACHE_STORAGE_BASE = 'fetch-cache'
-// Storage key for OAuth cache - must match server/utils/atproto/storage.ts
-const OAUTH_CACHE_STORAGE_BASE = 'atproto:oauth'
 
 export default defineNuxtModule({
   meta: {
@@ -37,11 +35,6 @@ export default defineNuxtModule({
         // Fetch cache storage (for SWR fetch caching)
         nitroConfig.storage[FETCH_CACHE_STORAGE_BASE] = {
           ...nitroConfig.storage[FETCH_CACHE_STORAGE_BASE],
-          driver: 'vercel-runtime-cache',
-        }
-
-        nitroConfig.storage[OAUTH_CACHE_STORAGE_BASE] = {
-          ...nitroConfig.storage[OAUTH_CACHE_STORAGE_BASE],
           driver: 'vercel-runtime-cache',
         }
       }


danielroe (approved):
I'll wait for @matthieusieben's review, but in the mean time I've created that key and configured it on vercel so that it's good to go from my pov.

fyi, I created two keys, one for production + a different one for preview deployments. if we should get rid of preview deployment key (maybe allow nuxt to regen a key per-preview, as we do with session passwords?) then let me know

matthieusieben:
Hey @fatfingers23,

While I wrote the `@atproto/oauth-client-*` libs, I am less familiar with Nuxt. So feel free to take my comments with a grain of salt. Before anything else, the version of the code here seems to work, and I don't see any security issues with it 👍. I do have some concerns, which have more to do with the way we intended the `@atproto/oauth-client-node` lib to work, than you implementation:

First, I am wondering if the use of `useSession` isn't a bit counter productive here. The `h3` cookie is defined as a `session` cookie, so it wont persist if the browser is closed. This defeats the purpose of using a confidential client, which is to allow having long lived sessions.

Secondly, by creating a new instance for every request, you are using the `NodeOAuthClient` in a way that was not foreseen. I can see the benefit of doing things that way, as having access to the `session` cookie allows to neatly "scope" the stored data to a particular browser. I have several concerns with it though: 1) it couples the cookie management logic (which might change due to my point above) to backend data management; 2) A full featured confidential client should proactively refresh the sessions stored in the backend to prevent them from becoming stale on the AS. Since the store logic is bound to availability of the browser session, the store would not be able to keep cookies in sync; 3) Because of the two previous points, you might encounter (rare) concurrency issues when multiple instances are trying to access the same data in parallel.

Thirdly, the implementation lacks a "cleanup" logic. That logic should ideally be implemented inside the `NodeOAuthClient` itself (I'm adding a TODO on my end to work on it). Until I get to it, you'll need to make sure that the state and session databases don't grow indefinitely. The state DB should be cleared from session created more than e.g. 30 minutes ago, and never completed. The session database can be cleared by refreshing sessions and removing inactive ones (another possibility would be not to refresh, and instead "signout" oauth sessions that were not refreshed in a while, because the user didn't recently visit).

Here is my recommendation (and again, this is based on my basic understanding of Nitro):

1) Create  a NitroPlugin so that there is a single `NodeOAuthClient` instance running.
2) The `NodeOAuthClient` instance can be used from the `oauth-client-metadata.json` and `jwks.json` endpoint (by returning `oauthClient.clientMetadata` and `oauthClient.keyset.publicJwks`), removing the need to expose a bunch of utils (`loadJWKs`, `getOauthClientMetadata`, ...)
3) Create a long lived "device ID" cookie (using either a strong random value, or a server encrypted value). This cookie can be used instead of the `atproto_oauth_req` I introduced in #1327 to bind a particular oauth request to a browser (simply by adding the "device ID" in the "state" data in `encodeOAuthState` and checking it in `decodeOAuthState`)
4) Create an additional "profile" backend storage, keyed by "device ID". This storage will basically contain what is currently in `UserServerSession['public']`.
5) Change `getOAuthSession()` to be using the "Device ID" cookie + the "profile" storage + static `oauthClient`
6) Change `useServerSession()` to be using the "Device ID" cookie + "profile" storage
8) Make sure that the `api/auth/session` endpoint triggers an oauth refresh (to keep the session alive), simply by calling `getOAuthSession()` (the `restore()` method will automatically trigger a refresh if the session is about to expire)

Here is a skeleton of what the plugin might look like:

```ts
/// Plugin definition
export default defineNitroPlugin(nitroApp => {

  // Store keeping pending OAuth authorization requests
  const stateStore = new SimpleStoreStorage<NodeSavedStateStore>("oauth:state")

  // Store keeping credentials from completed OAuth authorization requests
  const sessionStore = new SimpleStoreStorage<NodeSavedSessionStore>("oauth:session")
  
  const oauthClient = new NodeOAuthClient({
    stateStore,
    sessionStore,
    clientMetadata,
    requestLock: getOAuthLock(),
    handleResolver,
  })

  oauthClient.addEventListener('updated', ({ details: { sub, tokenSet: { aud } } }) => {
    // An OAuth session was refreshed. we might want to check if we need to update the
    // profile info (e.g. if the `aud` changed) in the [Device ID => sub] storage?
    // A better place to check this would probably be from the api/auth/session endpoint so that
    // we only perform outbound requests when a user actually requests data.
  })
  
  oauthClient.addEventListener('deleted', ({ details: { sub } }) => {
    // An OAuth session was deleted (for any reason: failure to refresh, user logout, etc.):
    // TODO: remove the the "profile" from the "profile" storage
  })
  
  nitroApp.oauthClient = oauthClient
  
  // Prevent the "oauth:state" store from growing indefinitely with un-completed oauth requests
  // @NOTE: This should eventually be implemented by the NodeOAuthClient itself !
  // Note that this can be implemented with a TTL on the store (if that is available)
  const stopCleanup = startAsyncTask(async (signal: AbortSignal) => {
    try {
      for await (const [key, { lastUpdatedAt }] of stateStore.list()) {
        if (signal.aborted) return
        
        // Remove every state that was created more than 30 minutes ago
        if (lastUpdatedAt < Date.now() - 30 * 60 * 1e3) {
          stateStore.del(key)
        }
      }
    } catch (err) {
      // "list" triggered an error, ignore it
      console.error('Failed to list active sessions', err)
    }
  })

  nitroApp.hooks.hook('close', stopCleanup)

  // Proactively keep OAuth sessions alive
  // @NOTE: This should eventually be implemented by the NodeOAuthClient itself !
  const stopRefresh = startAsyncTask(async (signal: AbortSignal) => {
    try {
      for await (const [sub, {lastUpdatedAt, value: savedSession }] of sessionStore.list()) {
        if (signal.aborted) return
  
        // Note that we can keep either one , or both, branches of the if/else-if bellow (depending on what we want)
        if (wasLastRefreshedTooLongAgo(savedSession, lastUpdatedAt)) {
          try {
            await oauthClient.revoke(sub)
          } catch (err) {
            // No biggie
          }
        } else if (shouldRefresh(savedSession, lastUpdatedAt)) {
          try {
            await oauthClient.restore(key, true)
          } catch (err) {
            console.warn(`Refreshing session ${key} failed`, err);
            // No need to remove from the store, the client will call sessionStore.del()
          }
        }
      }
    } catch (err) {
      // "list" triggered an error, ignore it
      console.error('Failed to list active sessions', err)
    }
  })

  nitroApp.hooks.hook('close', stopRefresh)

})
```

<details>
<summary>Implementation details</summary>

```ts
function shouldRefresh({ tokenSet }: NodeSavedSession, lastUpdatedAt: number) {
  if (tokenSet.expires_at) {
    // If the server returned an expiration date, use that as signal to refresh
    return new Date(tokenSet.expires_at).getTime() <= Date.now()
  } else {
    // Otherwise, refresh once a day (every 23.5H)
    return lastUpdatedAt <= Date.now() - 23.5 * 60 * 60e3
  }
}

export function startAsyncTask(
  task: (signal: AbortSignal) => void | Promise<void>,
  intervalMs = 60 * 60 * 1e3, // 1 hour
): () => Promise<void> {
  const abortController = new AbortController()
  let promise: Promise<void> | undefined
  let timer: ReturnType<typeof setTimeout> | undefined

  const runAndSchedule = async (signal: AbortSignal) => {
    await (promise = (async () => task(signal))())
    if (!signal.aborted) {
      timer = setTimeout(runAndSchedule, intervalMs, signal)
    }
  }

  runAndSchedule(abortController.signal)

  return async () => {
    abortController.abort()
    clearTimeout(timer)
    await promise // Await pending task run to finish
  }
}
```

Note, with this change, you can probably use a single implementation for the "store" instead of two:

```ts
/**
 * Wrapper around {@link Storage} that implements the {@link SimpleStore} interface, required by the AT Protocol OAuth client.
 */
class SimpleStoreStorage<T extends Record<string, unknown>> implements SimpleStore<string, T> {
  private readonly storage: Storage<{value: T, lastUpdatedAt: number }>

  constructor(storageName: string) {
    this.storage = useStorage(storageName)
  }

  async get(key: string, options?: GetOptions): Promise<undefined | T> {
    // TODO

    return data.value
  }
  
  async set(key: string, value: T): Promise<void> {
    // TODO
    const data = { value, lastUpdatedAt: Date.now() }
  }

  async del(key: string): Promise<void> {
    // TODO
  }

  // Note: not part of the simple store interface
  async *list(): AsyncGenerator<string, { value: T, lastUpdatedAt: number }> {
    const keys = await this.storage.getKeys()
    for (const key of keys) {
      try {
        yield await this.storage.get(key)
      } catch (err) {
        // the key was removed while async iterating
        if (isKeyNotFoundError(err)) continue
        throw err
      }
    }
  }
}
```
</details>

fatfingers23:
Thank you a ton for all the feedback @matthieusieben! Really appreciate it, and I think I'm on board with it all. It will probably be a bit before I can start digging in the code, but I did have a couple of things

- The oauth metadata and jwk endpoint are pre-rendered so they end up just being json files in our deployment. I don't think that really changes any thing from your review, but wanted to share. 
- This one may be a silly question. But what was your thinking for getting the "device id" to the stores? Just use `setCookie` and `getCookie` there? I think that makes sense to me, but concerned to do that without checking since that's what my understanding useSession is doing that already with it's encrypted cookie. 
- Also a bit afraid the key/value checks for the oauth session may run a bit longer and cost a bit more since we run mostly serverless and those reads could be pricey. But I think we can figure out something there

matthieusieben:
> This one may be a silly question. But what was your thinking for getting the "device id" to the stores? Just use setCookie and getCookie there? I think that makes sense to me, but concerned to do that without checking since that's what my understanding useSession is doing that already with it's encrypted cookie.

`useSession` uses a "session" cookie, ie. a cookie with no expiration date that the browser will expire when the browser is closed. Unless you can make that cookie a persistent cookie, I would indeed use `setCookie`/`getCookie`. A random value with enough entropy should be fine there. You can also encrypt that value to ensure authenticity (ie. that the device ID was indeed created by the server and not forged by a user).

> Also a bit afraid the key/value checks for the oauth session may run a bit longer and cost a bit more since we run mostly serverless and those reads could be pricey. But I think we can figure out something there

I see no problem with caching the current "profile" in `useSession`, as a way to have easy access to it (e.g. if it is needed from multiple endpoints). The biggest down side is that you could end-up in a situation where an oauth session was removed (e.g. from a background refresh), but the session cookie still contains the profile info. This means that the `useSession` content cannot be trusted to reflect the status of the session.

In that situation, any attempt to actually grab an oauth session (through `oauthClient.restore()`) may fail. This means that calls to `getOAuthSession()` should be expected to fail (which they probably should anyways), and when they *do* fail, the session cookie should be updated accordingly (which, again, is completely fine).

Now, I still recommend to proactively attempt to refresh the oauth session & `useSession` cookie from time to time. The `api/auth/session` seems like a great place to do that. We can detect that a user is re-visiting after being away for a longish time if the device ID cookie is present but the useSession data is empty, which might be a good time to "force refresh" (ie. by hitting in the "profile" storage) the status of both the oauth session and `useSession` cookie.

matthieusieben:
Now that I think about it, we could get away with storing the user did in a long lived cookie, instead of using a Device ID, and use `useSession` for caching the profile data.

Note that doing this **ABSOLUTELY REQUIRES** that the cookie is both long lived **AND** (**very importantly**) encrypted (to prevent stealing user sessions).

Note also that since `OAuthClientNode` only supports a single oauth session per did (this is a limitation I also need to address in the lib), we kinda *have* to account for users signing in the same account from different devices.

This means that whatever we do, we have to assume that the presence of both `useSession` data and/or the long lived cookie ("device ID"/"DID") is not enough to ensure that the oauth session is still active (every endpoint that needs to perform authenticated actions on the PDS using the `getOAuthSession` should expect an error).

If creating an encrypted version of the DID in a cookie is too hard (why would it be?), a random "device id" will provide a good workaround.

I guess the choice between "deviceID" vs. "DID cookie" boils down to how cleanly you want to model "active sessions" in you system. If you use "device ID", NPMX would be able to show "you have X active sessions (on <IP> using <UserAgent>)", while the "DID cookie" approach prevents that.

The "Device ID" approach also allows to better protect against session hijacking attacks (in particular if the device ID is rotated every now and then)

danielroe:
a few quick notes:
- you can pass any arbitrary options to `useSession`, including making it a long lived cookie, setting a custom maxAge, etc.
- as this is currently running on serverless infra, we are not guaranteed that there is a singleton server client - even parallel requests may be routed to different lambdas

matthieusieben:
> you can pass any arbitrary options to useSession, including making it a long lived cookie, setting a custom maxAge, etc.

We're golden then!

> as this is currently running on serverless infra, we are not guaranteed that there is a singleton server client - even parallel requests may be routed to different lambdas

That's fine, the redis based lock should handle concurrency

fatfingers23:
Thank you for all the feedback @matthieusieben! Helped a lot and I learned some new things as well. Was a bit more afk yesterday and just got around to working on it. If this works for you @danielroe I'm good with going with this.

- Moved the store to Redis so we can set TTL for state to 30mins and Session store TTL to 179 days to let Redis handle the clean up instead of a job
- `session.get.ts` now implements the `eventHandlerWithOAuthSession` handler so it can refresh those tokens when needed. Also added some handling where if the oauthsession is no longer valid it will clear the useSession as well
- left it using useSession for storing the users oauth sessionid, but set the max age to 179 days. I spent a bit of time seeing if I could trigger a way for that to be sliding, as in when the session updates, it updates that cookie. But no luck, so we may want to eventually move to setting a cookie directly. 
- left the creating a new instance of the oauth client. We should be able to move that to a nitro plugin later without disrupting logins if we want to

server/routes/oauth-client-metadata.json.get.ts
```diff
@@ -1,3 +1,7 @@
-export default defineEventHandler(() => {
-  return getOauthClientMetadata()
+export default defineEventHandler(async event => {
+  const config = useRuntimeConfig(event)
+  const keyset = await loadJWKs(config)
+  // @ts-expect-error Taken from statusphere-example-app. Throws a ts error
+  const pk = keyset?.findPrivateKey({ use: 'sig' })
```
matthieusieben: This has changed recently (and was not updated in the statusphere example)

```suggestion
  const pk = keyset?.findPrivateKey({ usage: 'sign' })
```

server/utils/atproto/oauth.ts
```diff
@@ -49,22 +69,50 @@ type EventHandlerWithOAuthSession<T extends EventHandlerRequest, D> = (
...
+export async function getNodeOAuthClient(
+  serverSession: SessionManager,
+  config: RuntimeConfig,
+): Promise<NodeOAuthClient> {
+  const { stateStore, sessionStore } = useOAuthStorage(serverSession)
+
+  // These are optional and not expected or can be used easily in local development, only in production
+  const keyset = await loadJWKs(config)
+  // @ts-expect-error Taken from statusphere-example-app. Throws a ts error
+  const pk = keyset?.findPrivateKey({ use: 'sig' })
```
matthieusieben:
```suggestion
  const pk = keyset?.findPrivateKey({ usage: 'sign' })
```

danielroe:
did you mean to tag this as closing #840?

modules/cache.ts
```diff
@@ -37,6 +39,11 @@ export default defineNuxtModule({
           ...nitroConfig.storage[FETCH_CACHE_STORAGE_BASE],
           driver: 'vercel-runtime-cache',
         }
+
+        nitroConfig.storage[OAUTH_CACHE_STORAGE_BASE] = {
+          ...nitroConfig.storage[OAUTH_CACHE_STORAGE_BASE],
+          driver: 'vercel-runtime-cache',
+        }
```
danielroe: I don't think this is needed as we have L50-52

```suggestion
```

modules/cache.ts
```diff
@@ -4,6 +4,8 @@ import { provider } from 'std-env'
 
 // Storage key for fetch cache - must match shared/utils/fetch-cache-config.ts
 const FETCH_CACHE_STORAGE_BASE = 'fetch-cache'
+// Storage key for OAuth cache - must match server/utils/atproto/storage.ts
+const OAUTH_CACHE_STORAGE_BASE = 'atproto:oauth'
```
danielroe:
```suggestion
```

fatfingers23:
@danielroe  Whoops meant 850. Fixed the description and cleaned up those useStorage leftovers

Issue #501, opened 2026-01-31
Discussion: changelogs button


WilcoSp:
in this issue we discuss, come up with ideas, give feedback & contribute to adding & evolving changelogs to npmx.
below is a roadmap and below that there is the original starting message of this discussion.

roadmaps in PRs (updated May 11th)
----
1. ✅ button, page & basic viewers (these will cover most people) (https://github.com/npmx-dev/npmx.dev/pull/1233 (ready for review & merge))
   - [x] changelog button at button page
   - [x] /package-changes/ page 
   - [x] a basic changelog.md viewer component with toc
   - [x] basic releases viewer component
   - [x] automatic scrolling to the as close as possible to the requested version with date+version (fallback to date)
   - [x] check whether the changelog.md from the latest release is in the repository directory specified in the npm package or the root of the package

1.5 some refactoring which isn't possible in a big branch (https://github.com/npmx-dev/npmx.dev/pull/2717 [preview](https://npmxdev-git-fork-wilcosp-feat-markdown-kit-npmx.vercel.app/))
   - [x] refactor the markdown renderer to better share code with the readme renderer
2. most of the other repositories services (if needed this can be spread across multiple PRs for each service)
   - [ ] add support for other git service provider documented [here](https://docs.npmx.dev/guide/features#supported-git-providers) for releases
   - [ ] add support for other git service providers documented [here](https://docs.npmx.dev/guide/features#supported-git-providers) for changelog.md
   - [ ] add support for for automatic links to pull/merge requests (#pr), issues (#issue) and users (@person)
       - parse from plain text to a link
       - parse from link to git provider style for the issue/pr/user
   - [x] improve experience when switching between package version's by using [useRouteParamss](https://vueuse.org/router/useRouteParams/#userouteparams) and only re render when the repository has been changed (example vue 3 <-> vue 2) (this isn't needed anymore, how now background resolving works it gives the same result)
   - [ ] have a button to the changelogs page in the package dependencies 
   - [ ] at the package page have buttons next to selectable package versions (similar to https://github.com/npmx-dev/npmx.dev/pull/1368)
* filters for releases
   - [ ] support for filtering for specific packages within releases (does need heading to be `package@version`)
   - [ ] support for filtering to specific version range (for example 3.5.x or 3.x.x)

* more advanced features for changelog.md
   - [ ] split the versions by heading (semver regex) 
* add npmx.changelog fields for package.json that can't be covered by the PRs above, this will be a separate issue when the time comes
[...]

PR #1368, opened 2026-02-11
feat: link to releases

antfu:
I don't know how common the case is for everyone, but I found that one of the most played actions for me to visit a package's npm page is when I found a new major version released, and I wanted to know what has changed. So my flow would be 

> `search for the package` -> `npm` -> `repo` -> `release page`

As many release notes are hosted in GitHub, I think it would be very useful for npmx to offer an easy entry for accessing the release notes.

CleanShot 2026-02-11 at 14 13 14@2x

---

A few open questions:

- [ ] Apparently, I use Claude Code to generate the `fetchReleases` for providers other than GitHub, I didn't test them, and actually don't really know how. Should we keep them as-is for future contribution to improve, or should we remove them entirely and add back one by one later?
- [ ] For improve the above experience for developers, maybe we should have a list of package example that we can go through (that are configured with different source hosting, that contributors can check them one-by-one). And also maybe a dev-time only page (debug page) for linking them easily (out of scope of this PR)
- [ ] At least for GitHub, we actually have the markdown and html content of the release note, should we render them in our app instead of linking them? (for future PR)
- [ ] As we have more and more items, the UI is getting a bit crowded, to keep things scoped, I plan to tweak the UI in another PR later.

--- 

Edit: I just notice it has some conflicts to the existing work #1233 (issue #501), linking for ack

### DIFF
--- a/app/components/Package/Versions.vue
+++ b/app/components/Package/Versions.vue
@@ -494,7 +494,7 @@ function majorGroupContainsCurrent(group: (typeof otherMajorGroups.value)[0]): b
                     })
                   "
                   class="inline-flex items-center text-fg-muted hover:text-fg transition-colors shrink-0"
-                ></LinkBase>
+                />
               </div>
               <div v-if="row.tags.length" class="flex items-center gap-1 mt-0.5 flex-wrap">
                 <span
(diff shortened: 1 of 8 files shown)


danielroe:
I love this!

I think we should have a page in npmx to show releases, if possible.

(the bar at the top: page/code/compare - I think should only link to internal pages within npmx, so maybe let's move it down to the 'repo meta' - stars, forks, etc.) --- or maybe render as a small icon button next to the 'provenance badge' next to the package name? (that way it's clear that it's about _that version_

can't wait to see the ui improvements you make 🙏

WilcoSp:
@danielroe at https://github.com/npmx-dev/npmx.dev/pull/1233 I'm working for page to display the gh releases and even later also the changelog.md

what may be an idea is to to temporarily have both this (link) PR & the [page](https://github.com/npmx-dev/npmx.dev/pull/1233) PR with link being the default behaviour and the page as a setting while it's under development ([planning](https://github.com/npmx-dev/npmx.dev/issues/501#issuecomment-3841185931))

when the releases/change log page is finished we have either that the link goes away or we can switch the default to the page and keep the setting


also @antfu maybe an idea is to also check for if releases are being used in general so if none of the tags are found that the link can still be able to send to `/releases`
for example any of the [graphql-codegen](https://npmx.dev/org/graphql-codegen) packages uses dates at the gh releases and [apollo](https://npmx.dev/package/apollo) does `{package}@{version}`.

PR #1369, opened 2026-02-11
feat: sort contributors list by role

shuuji3:
An attempt to sort contributors list by the governance role (steward > core > maintainer > contributor).

screenshot of contributor list in about page, daniel and patak is first two icons

I cannot find a public member list of the maintainers so kept it empty for now.

One option to avoid hardcoding is to create GitHub Team for each governance role, and create members list in GitHub Actions with command like `gh api /orgs/npmx-dev/teams/steward/members --jq 'map(.login)`.

/cc @patak-dev @userquin

(ref. original chat: https://bsky.app/profile/patak.dev/post/3mek245m7as26)

### DIFF
--- a/app/pages/about.vue
+++ b/app/pages/about.vue
@@ -1,4 +1,6 @@
 <script setup lang="ts">
+import type { Role } from '#server/api/contributors.get'
+
 const router = useRouter()
 const canGoBack = useCanGoBack()
 
@@ -27,6 +29,22 @@ const pmLinks = {
 }
 
 const { data: contributors, status: contributorsStatus } = useLazyFetch('/api/contributors')
+
+const governanceMembers = computed(
+  () => contributors.value?.filter(c => c.role !== 'contributor') ?? [],
+)
+
+const communityContributors = computed(
+  () => contributors.value?.filter(c => c.role === 'contributor') ?? [],
+)
+
+const roleLabels = computed(
+  () =>
+    ({
+      steward: $t('about.team.role_steward'),
+      maintainer: $t('about.team.role_maintainer'),
+    }) as Partial<Record<Role, string>>,
+)
 </script>
 
 <template>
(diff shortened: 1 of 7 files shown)


patak-cat:
Note: I thought I was looking at an issue 🤦🏼 
My mind couldn't wrap around you doing a PR so fast:
> Thanks for creating the issue! I think this is a good idea, we should surface maintainers and is a good way for some of them to flow up in line with their real contributions. The commits aren't telling the whole picture for a lot of us now.

patak-cat:
It would be good to also show for governance folk their github, npm, etc too. Maybe at one point we find a way to do it for everyone once we have more people connecting their accounts together.

image

One interesting discussion, we decided to remove the sponsors button for stewards in the main page of the repo. It didn't feel right that we were the only ones having the exposure there. But maybe we could add them back for everybody (or as many of us as possible, maybe first governance, later when people connect to github we could do better). I think we need to start trying to start working to help with OSS funding. I don't think people will give us a lot of money through this, I care more about the symbol that we should show that money is needed for npmx to work. It will also get better once we can start adding sponsors (soon!).

patak-cat:
Now that I know this is a PR, I checked how it looks :D 

So, we could merge this and iterated, but I would like we show the governance structure as a way to explain that npmx takes the foundation route seriously. So we should have Stewards first as a title, then Maintainers, then everyone else. Maybe with a short paragraph. And it could give us an option to do bigger cards for maintainers so we can show that they work in different companies and what other projects they participate in. Let's use this as a way to show that we are the result of uniting a lot of communities.

shuuji3:
It actually is a PR😄 Yes, this can be imrpoved iteratively. I also think about adding sections similar to https://github.com/npmx-dev/npmx.dev/blob/main/GOVERNANCE.md and we could add links to this docs later.

shuuji3:
Also, once npmx's profile page mature, we could link to the profile page `https://npmx.dev/profile/[user-name]`, instead of directly link to `https://github.com/[user-name]`

patak-cat:
I'm thinking that for the first iteration, until we have a core team, the best is if we only have two sections:

# Team

## Governance

Daniel Me [ ...maintainers ] <-- all with the same card size, just the role is diff

## Contributors

All other contributors, maybe including the above too as a way to show we are all part of the team

server/api/contributors.get.ts
```diff
@@ -46,8 +64,14 @@ export default defineCachedEventHandler(
...
 
-    // Filter out bots
-    return allContributors.filter(c => !c.login.includes('[bot]'))
+    return (
+      allContributors
+        // Filter out bots
+        .filter(c => !c.login.includes('[bot]'))
+
+        // Sort by role (steward > core > maintainer > contributor)
+        .sort((a, b) => getRoleOrder(a.login) - getRoleOrder(b.login))
```
userquin: can we change `roleMembers` to a `Set<string>` instead `string[]`? we have `O(N * M * log N)` (`N` is the number of contributors and `M` is the sum of stewards + core + maintainers => `O(M) + O(N log N)`

shuuji3: Updated to use `Set.has()` 👍🏻

app/pages/about.vue
```diff
@@ -138,17 +138,59 @@ const { data: contributors, status: contributorsStatus } = useLazyFetch('/api/co
...
+
+            <div class="grid grid-cols-1 sm:grid-cols-2 md:grid-cols-3 gap-3">
+              <a
+                v-for="person in contributors
+                  .filter(c => c.role !== 'contributor')
+                  .concat(contributors.filter(c => c.role !== 'contributor'))
+                  .concat(contributors.filter(c => c.role !== 'contributor'))
+                  .concat(contributors.filter(c => c.role !== 'contributor'))
+                  .concat(contributors.filter(c => c.role !== 'contributor'))"
+                :key="person.id"
```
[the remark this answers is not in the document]
shuuji3: no problem, this is just a stub for this WIP commit

shuuji3:
I've created a simple mini cards (icon + name + role) for governance section as a starting point. Should I use this same card format for all contributors, or keep the current contributors cloud?

I'll collect the maintainers GitHub handles from Discord and hardcode them (tomorrow).

Also, could you provide the short paragraphs for each sections? I'm not sure what the best wordings.

Screenshot from 2026-02-12 01-14-50

whitep4nth3r:
@shuuji3 I don't think we need the paragraph descriptions right now. Feel free to remove!

shuuji3:
Thanks for the update, Daniel🙂 It looks perfect to me.

about page screenshot, maintainers card has sponsor button now

Issue #1337, opened 2026-02-10
Date input in Trends cannot accept year input


rzzf:
When trying to enter 2026 in the year field of start or end, the input is immediately processed and further typing is blocked, making it impossible to complete the full year.

<details>
<summary>Video</summary>

https://github.com/user-attachments/assets/fd2bdc50-3b3f-46a7-a58b-f086b7bf6c95

</details>

PR #1370, opened 2026-02-11
fix: prevent trends date input from being disabled while typing

RYGRIT:
this pr:
+ fixes: #1337
+ add disabled option for Likes select

<details>
<summary>Video</summary>

https://github.com/user-attachments/assets/8fd10b0e-8475-4c25-81b9-9e2456a96225

</details>

**Question for Maintainers:** I noticed that the native HTML `<input type="date">` behavior allows for 6-digit years (e.g., `202612`) by default. Should I add `max="9999-12-31"` to restrict the input to 4-digit years? I'm happy to include that in this PR if you think it's better for the UX.

### DIFF
--- a/app/components/AppHeader.vue
+++ b/app/components/AppHeader.vue
@@ -180,7 +180,7 @@ onKeyStroke(
     for (const link of desktopLinks.value) {
       if (link.to && link.keyshortcut && isKeyWithoutModifiers(e, link.keyshortcut)) {
         e.preventDefault()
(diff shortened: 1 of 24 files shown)


danielroe:
I think restricting to 4-digit years would be safe 😆 (and good ux!)

RYGRIT:
Thanks! I'll update the PR right away.

RYGRIT:
done!

PR #1374, opened 2026-02-11
chore(i18n): add missing Japanese translations

morinokami:
I have added all missing Japanese translations. I confirmed that all keys are translated by running the `pnpm i18n:check ja-JP` command.

### DIFF
--- a/i18n/locales/en.json
+++ b/i18n/locales/en.json
@@ -295,7 +295,7 @@
       "recent_versions_only": "Recent versions only",
       "recent_versions_only_tooltip": "Show only versions published within the last year.",
       "show_low_usage": "Show low usage versions",
-      "show_low_usage_tooltip": "Include version groups with less than 1% of total downloads",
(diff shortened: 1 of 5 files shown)


antfu (approved):

i18n/locales/en.json
```diff
@@ -295,7 +295,7 @@
       "recent_versions_only": "Recent versions only",
       "recent_versions_only_tooltip": "Show only versions published within the last year.",
       "show_low_usage": "Show low usage versions",
-      "show_low_usage_tooltip": "Include version groups with less than 1% of total downloads",
+      "show_low_usage_tooltip": "Include version groups with less than 1% of total downloads.",
```
morinokami, line 298: This change is unrelated to the Japanese translations, but although both `recent_versions_only_tooltip` and `show_low_usage_tooltip` are sentences displayed inside a tooltip, only` show_low_usage_tooltip` was missing a comma, so I added one for consistency.

Image

Image

PR #1375, opened 2026-02-11
perf: optimize css chunks

### LINKED ISSUE #1239: CSS Size and icons
Image

We are currently transmitting ~150kb for our main CSS file. Which seems waaay to much given the complexity of the site. Below some analysis:

## Icons

By my rough estimates around ~80% of the CSS are just the icons. Half of the icons are only used by code by the file tree component (`i-vscode-icons`).

The icons itself are defined like this (indentation for readability):
```css
.i-vscode-icons-file-type-node{ 
    background:url("data:image/svg+xml;utf8,%3Csvg viewBox='0 0 32 32' display='inline-block' forced-color-adjust='preserve-parent-color' width='1.2em' height='1.2em' xmlns='http://www.w3.org/2000/svg' %3E%3Cpath fill='%2383cd29' d='M16 30a2.15 2.15 0 0 1-1.076-.288L11.5 27.685c-.511-.286-.262-.387-.093-.446a6.8 6.8 0 0 0 1.549-.7a.26.26 0 0 1 .255.019l2.631 1.563a.34.34 0 0 0 .318 0l10.26-5.922a.32.32 0 0 0 .157-.278V10.075a.33.33 0 0 0-.159-.283l-10.26-5.917a.32.32 0 0 0-.317 0L5.587 9.794a.33.33 0 0 0-.162.281v11.841a.32.32 0 0 0 .161.274L8.4 23.814c1.525.762 2.459-.136 2.459-1.038V11.085a.3.3 0 0 1 .3-.3h1.3a.3.3 0 0 1 .3.3v11.692c0 2.035-1.108 3.2-3.038 3.2a4.4 4.4 0 0 1-2.363-.642l-2.697-1.547a2.17 2.17 0 0 1-1.076-1.872V10.075A2.16 2.16 0 0 1 4.661 8.2l10.261-5.924a2.25 2.25 0 0 1 2.156 0L27.338 8.2a2.17 2.17 0 0 1 1.077 1.87v11.846a2.17 2.17 0 0 1-1.077 1.872l-10.26 5.924A2.15 2.15 0 0 1 16 30'/%3E%3Cpath fill='%2383cd29' d='M14.054 17.953a.3.3 0 0 1 .3-.3h1.327a.3.3 0 0 1 .295.251c.2 1.351.8 2.032 3.513 2.032c2.161 0 3.082-.489 3.082-1.636c0-.661-.261-1.152-3.62-1.481c-2.808-.278-4.544-.9-4.544-3.144c0-2.07 1.745-3.305 4.67-3.305c3.287 0 4.914 1.141 5.12 3.589a.3.3 0 0 1-.295.323h-1.336a.3.3 0 0 1-.288-.232c-.319-1.421-1.1-1.875-3.2-1.875c-2.36 0-2.634.822-2.634 1.438c0 .746.324.964 3.51 1.385c3.153.417 4.651 1.007 4.651 3.223c0 2.236-1.864 3.516-5.115 3.516c-4.495.006-5.436-2.055-5.436-3.784'/%3E%3C/svg%3E") 0 0/100% 100% no-repeat
}

.i-vscode-icons-file-type-node,.i-vscode-icons-file-type-npm{ 
    forced-color-adjust:preserve-parent-color;
    width:1.2em;
    height:1.2em;
    display:inline-block
}

.i-vscode-icons-file-type-npm{ 
    background:url("data:image/svg+xml;utf8,%3Csvg viewBox='0 0 32 32' display='inline-block' forced-color-adjust='preserve-parent-color' width='1.2em' height='1.2em' xmlns='http://www.w3.org/2000/svg' %3E%3Cpath fill='%23c12127' d='M2 2h28v28H2'/%3E%3Cpath fill='%23fff' d='M7.25 7.25h17.5v17.5h-3.5v-14H16v14H7.25'/%3E%3C/svg%3E") 0 0/100% 100% no-repeat
}

```

With two icons always sharing the same base style definition, even though that is repeated again in the inline svg definition.

Compare this to how FontAwesome transmits an icons:

```js
m = {
    prefix: "fasr",
    iconName: "magnifying-glass",
    icon: [576, 512, [128269, "search"], "f002", "M400.4 208a160 160 0 1 0 -320 0 160 160 0 1 0 320 0zM369.6 371.1c-35.5 28.1-80.3 44.9-129.1 44.9-114.9 0-208-93.1-208-208s93.1-208 208-208 208 93.1 208 208c0 48.8-16.8 93.7-44.9 129.1l133.9 133.9 17 17-33.9 33.9-17-17-133.9-133.9z"]
}
```

Especially for simpler icons, this saves a lot of data.

## Vue Data UI
Roughly another 25% are caused by Vue UI — I tried to solve in #1237 by lazy loading, but that did not work.

## Solutions

I am currently not really sure how to solve it. Uno CSS is supposed to only add classes, that are actually used, but this does not prevent all the CSS from Vue Data UI to end up in the final package.

I am also not sure, if the way the icons are currently handled is optimal, especially since (again) this prevents lazy loading.

antfu:
Resolves #1239 

This PR made the following two changes to optimize the css chunking:

- Make the component and styles of Vue Data UI async on-demand
- Separate the icons layer from the main UnoCSS entry and then it async
  - This would introduce a side-effect that the icons css become unblocking and might have flash-of-no-icons on first load. We could discuss if this would be worth it.

Deploy Preview: https://npmxdev-git-fix-css-optimization-npmx.vercel.app/

Before:
CleanShot 2026-02-11 at 18 18 07@2x

After:
CleanShot 2026-02-11 at 18 17 01@2x


danielroe:
this is amazing! I do think we need to move `inline-block` into the individual icons as otherwise there's now tiny layout shifts for every icon.

alexdln:
_So, our style utilities weren't working in tests, and this PR fixed it, which is why a11y tests started crashing. For example, opacity-50 isn't working in production tests right now, and we actually have a number of elements that aren't being covered correctly by our tests. I will fix what I can with low changes here or disable some of them tomorrow_

danielroe:
let's fix the styles-in-tests issue separately - probably the manual `uno.css` import is all that is needed.

(ie let's not split icon styles for now)

alexdln:
Tests crashed with this setting. I haven't yet determined why exactly, but it changes the styles in tests

```ts
  content: {
    pipeline: {
      exclude: [/\.(css|postcss|sass|scss|less|stylus|styl)($|\?)/, /[/\\]node_modules[/\\]/],
    },
  },
```

I've removed it for now and think I can try exploring it in the next issue.

---

Also, the PR originally had a transformer for icons that hardcoded extra classes. This was done so that after changes, icons are loaded lazily and later, and their styles weights become heavier. But this doesn't suit us, since we do a lot of customization on top of icons (sizes, colors, etc.), and it wasn't stable for pseudo-classes.

Fixed it via Layers - looks like it's the only stable option. Overall, I don't see any reason why we shouldn't use it - it seems like it's already become a standard properties with almost perfect support

---

Also enabled styles autoimport in storybook env

alexdln (approved):

uno.config.ts
```diff
@@ -26,7 +26,6 @@ export default defineConfig({
     presetWind4(),
     presetIcons({
       extraProperties: {
-        'display': 'inline-block',
```
danielroe, line 29: was there a reason we removed this?

alexdln: Hmm, I don’t know to be honest, but feels like it’s better without it anyway - we’ll always control it on our side and get more correct default styles (_f.e. if div - display:block, if span - inline_)

alexdln: But I found why it was removed. Because you made it as a separate rule and removed it here. And then I converted it to layers (_to avoid having a rule with !important sizes_) and hadn't returned this rule 😬

wdyt? should we return it (_it will be lazy loaded and can results in shifts if we haven't specified display on place, but it will have some default option_)?

serhalp:
hey friends, what's the status of this PR? 👀

alexdln:
@serhalp 
We've already pulled out several good optimizations from this PR, and essentially the last one left is separating icons and other CSS into different bundles. The latest solution generally works good, but the icons will appear a little later because they're loaded lazily. On the one hand, this is an optimization (and essentially means we now see other content earlier, rather than icons later), but on the other hand, we-re not sure whether it's worth it, as the user experience might be slightly worse

A solution might be to split icons into different sprites, making one blocking and the other non-blocking. But this adds complexity and is difficult to maintain in an open project

ghostdevv:
> @serhalp We've already pulled out several good optimizations from this PR, and essentially the last one left is separating icons and other CSS into different bundles. The latest solution generally works good, but the icons will appear a little later because they're loaded lazily. On the one hand, this is an optimization (and essentially means we now see other content earlier, rather than icons later), but on the other hand, we-re not sure whether it's worth it, as the user experience might be slightly worse
> 
> A solution might be to split icons into different sprites, making one blocking and the other non-blocking. But this adds complexity and is difficult to maintain in an open project

Is this still the case?

alexdln:
Well, the question here is more loke "Is saving ~~50 KB and ~~10 ms worth the slightly worse user experience?" And looks like not. We've already merged all the best optimizations from this PR separately, and it turns out that the latest part of it hasn't translated into sufficient profit

PR #1376, opened 2026-02-11
fix: filter search cannot be typed

rzzf:
The search input in Filters cannot accept any input.
https://npmx.dev/org/vue

### DIFF
--- a/app/components/Filter/Panel.vue
+++ b/app/components/Filter/Panel.vue
@@ -32,6 +32,11 @@ const { t } = useI18n()
 const isExpanded = shallowRef(false)
 const showAllKeywords = shallowRef(false)
 
+const filterText = computed({
+  get: () => props.filters.text,
+  set: value => emit('update:text', value),
+})
+
 const displayedKeywords = computed(() => {
   const keywords = props.availableKeywords ?? []
   return showAllKeywords.value ? keywords : keywords.slice(0, 20)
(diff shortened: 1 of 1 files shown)


app/components/Filter/Panel.vue
```diff
@@ -32,6 +32,11 @@ const { t } = useI18n()
 const isExpanded = shallowRef(false)
 const showAllKeywords = shallowRef(false)
 
+const filterText = computed({
+  get: () => props.filters.text,
+  set: value => emit('update:text', value),
+})
```
danielroe, line 38: I think we should use a `defineModel('text')` instead

rzzf: This property comes from `props.filters.text`, so is `defineModel('text')` unnecessary? 
If so, [ListToolbar.vue](https://github.com/npmx-dev/npmx.dev/blob/9aae0786f0a320648fdb7d9a36cb2a3143d55a17/app/components/Package/ListToolbar.vue#L236) would also need to be updated accordingly.

Referenced from
https://github.com/npmx-dev/npmx.dev/blob/9aae0786f0a320648fdb7d9a36cb2a3143d55a17/app/components/Package/ListControls.vue#L22-L25

danielroe: ahh, I see.

yes, this is awkward prop drilling we need to fix at some point but not in this PR

Issue #1346, opened 2026-02-10
"Compare" Search - Packages cant be selected via keyboard


Pascalmh:
Using the Keyboard the search results can't be reached.

Reproduce:
- go to https://npmx.dev/compare
- focus the search input
- type in "react"
- try selecting a search result using the keyboard

https://github.com/user-attachments/assets/7a5a61da-c9df-4e03-bed3-6425d58aadd9

PR #1377, opened 2026-02-11
feat: add keyboard navigation to compare page package search

iiio2:
most probably resolves https://github.com/npmx-dev/npmx.dev/issues/1346

In the compare page, we search packages to compare. Now, suppose we search "nuxt". So it provides you a list of suggestion. If we want to select any package from the suggestion list we need to use our mouse or touchpad to select a package from the suggestion list. It would be really nice if we can use keyboard page up and page down to select the packages from the suggestion list. That is the new feature of keyboard navigation. We can now select packages from the suggestion list using page up and page down in the keyboard. 

Before: We get suggestion of packages and if we want to select we need to use mouse or touchpad. 
wrong

After: We can now pick packages using page up or page down in the keyboard
correct


whitep4nth3r:
This works really nicely!

iiio2:
Thanks @whitep4nth3r . :blush:

Pascalmh:
@iiio2 just re-tested this - works great now! Thank you!

PR #1381, opened 2026-02-11
fix: improve collapsible section state detection

alexdln:
Collapsible sections (_in the sidebar of package pages_) sometimes showed content that weren't clickable.

The problem was that the logic simply checked for the presence of the current section's id in the dataset. For example - if `optional-dependencies` were collapsed, this check also worked for `dependencies` section, marking the wrong section as !isOpen.

At the same time, the content was visible, since the required style for hiding was not added (it is added by passing through exact ids)


alexdln:
> ?.split(' ') returns undefined

_Coderabbit definitely knows Vue a little better than JS_

PR #1384, opened 2026-02-11
fix: correct links to deps versions

alexdln:
If a range was specified in the dependencies, we had an invalid link. This caused 404 links, unnecessary prefetched, and just a bad experience.

I changed it so that we additionally clean up the versions, and if it results in valid version - the link will point to a specific version, otherwise - to the latest one. I was thinking about changing it to disabled, but it seems behavior with latest version is a bit more convenient (_the perfect option is to find the exact latest version under range, but it's to much requests - I'll try to look into this option later_)

### DIFF
--- a/app/components/Package/Dependencies.vue
+++ b/app/components/Package/Dependencies.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import { SEVERITY_TEXT_COLORS, getHighestSeverity } from '#shared/utils/severity'
 import { getOutdatedTooltip, getVersionClass } from '~/utils/npm/outdated-dependencies'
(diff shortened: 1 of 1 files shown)


danielroe:
hm. I think originally this was added deliberately so you just go to, e.g. https://npmx.dev/nuxt@^2

https://github.com/npmx-dev/npmx.dev/pull/103

alexdln:
Oh, oops... _Well, now I'm not sure where reality lies🫠_

Got it, thanks

What do you think would be better behavior for versions like ">= 18.0.0"? Should I repeat your logic or navigate to the latest version?

PR #1393, opened 2026-02-11
chore: add localias configuration for npmx.test

vinnymac:
Configures localias so developers can utilize https://npmx.test locally, which also makes it easier to test on mobile devices without having to get fancy with tunnels or wired connections.

image

See CONTRIBUTING.md for "Custom Local Domain" usage instructions.

----------------------

- includes a warning about AtProto redirect uris bringing you back to 127
- includes warnings for unsupported systems like windows with some recommendations, such as using WSL2
- native macOS / linux support of course
- configures nuxt to allow .test domains


ghostdevv (changes requested):
In lieu of any decision on this, personally I think this would make more sense in the documentation if you're interested in opening a PR for that? I will close this one for now though, but if other people want to upvote this or overrule me we can reopen!

vinnymac:
If I recall from Discord we wanted to explore if [portless](https://github.com/vercel-labs/portless) is ready as a replacement over localias. I haven’t tried it yet though, in any case I agree closing for now makes the most sense.

PR #1394, opened 2026-02-11
fix: missed input parts on quick search

### LINKED ISSUE #764: Initial seach improvements

### LINKED ISSUE #1211: Page auto-reload when typing in search bar causes keystroke miss
See video:

https://github.com/user-attachments/assets/e7f5e1cf-5303-4798-8ff5-9bf33952fa31

You can see in the overlay at the bottom what keys I'm pressing "qwerty".

However, npmx.dev refreshes the page after I start typing, causing the <kbd>w</kbd> to be missed by the input box.

alexdln:
## What

The user would type in one place, and several navigations would launch for each input. Then the route from the first navigation would change, and the user would continue typing—but navigation from the first page would still continue. They rewrote the route and the query listener updated the field value

## How

I created a global storage for search fields, and both fields write to it. The field doesn't actually listen to query now, but only uses it for default value and as the factor that `q` has been cleared (_when exiting the search_)

So now there's a single source of truth - the search store. And query parameter is simply synchronized with it for page reloads and for the server

resolves #1211 resolves #764

### DIFF
--- a/app/components/Header/SearchBox.vue
+++ b/app/components/Header/SearchBox.vue
@@ -28,8 +28,7 @@ const showSearchBar = computed(() => {
   return route.name !== 'index'
 })
 
-// Local input value (updates immediately as user types)
-const searchQuery = shallowRef(normalizeSearchParam(route.query.q))
+const searchQuery = useGlobalSearchQuery()
 
 // Pages that have their own local filter using ?q
 const pagesWithLocalFilter = new Set(['~username', 'org'])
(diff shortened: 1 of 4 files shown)


danielroe:
yes, this looks very promising as the source of the bug 👀

alexdln:
🤞🤞🤞

PR #1395, opened 2026-02-11
perf: add HTTP Link header (experimental)

userquin:
Right now only applied to prerendered routes.

**NOTE**: using `nitro.updateConfig` from nitro `prerender:done` since we're parsing the html (nitro config hot reloaded message), we can change this to use the manifest and use the logic at nuxt to build the Link header.

nitro config hot reloaded message after update routeRules


alexdln:
@userquin were there any noticeable improvements or can we close this PR?

UPD: discussed that we can close it

Issue #1354, opened 2026-02-10
Hide comments from description in search restults


onmax:
[https://npmx.dev/search?q=use-emai](https://npmx.dev/search?q=use-emai)

![Image](https://github.com/user-attachments/assets/c34fb98c-b2a0-4bef-b196-f2eee4069150)

PR #1397, opened 2026-02-11
fix: remove html comments from package description and deprecation notices

TheAlexLichter:
Resolves https://github.com/npmx-dev/npmx.dev/issues/1354

Not ideal as packages might have *no description* then but given that the description is fetched directly from NPM, we could only parse the README and take the first non-comment line.
This could be done in another iteration if worth it

### DIFF
--- a/app/composables/useMarkdown.ts
+++ b/app/composables/useMarkdown.ts
@@ -37,8 +37,8 @@ function stripAndEscapeHtml(text: string, packageName?: string): string {
   // Only match tags that start with a letter or / (to avoid matching things like "a < b > c")
   stripped = stripped.replace(/<\/?[a-z][^>]*>/gi, '')
 
-  // Strip HTML comments: <!-- ... -->
-  stripped = stripped.replace(/<!--[\s\S]*?-->/g, '')
+  // Strip HTML comments: <!-- ... --> (including unclosed comments from truncation)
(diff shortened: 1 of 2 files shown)


app/composables/useMarkdown.ts
```diff
@@ -37,6 +37,9 @@ function stripAndEscapeHtml(text: string, packageName?: string): string {
   // Only match tags that start with a letter or / (to avoid matching things like "a < b > c")
   stripped = stripped.replace(/<\/?[a-z][^>]*>/gi, '')
 
+  // Strip HTML comments: <!-- ... -->
+  stripped = stripped.replace(/<!--[\s\S]*?-->/g, '')
```
alexdln: As far as I understand, when there's no description in package.json, npm and algolia takes the description from Markdown. If the comment is long, only the opening tag will remain. What do you think about additional check so that in this case it's truncated to the end?




---
*p.p.s. just note - `[\s\S]*` is equal to `[\s\S]{0,}`, i.e. `*` means that there are more than 0 matches of any character. Therefore, the optional character (`?`) is not needed here. The optional character may be needed when testing with `+` instead of `*`, that equal to `[\s\S]{1,}` (_at least one any character_)*

TheAlexLichter: Ah, I *was* consider adding it but didn't know how the description is retrieved. Let me update the logic.

Note on the regex: `*?` is used to be non-greedy and not match more than wanted.

PR #1401, opened 2026-02-11
fix(a11y): use semantic html for keyword list

TheAlexLichter:
The `aria-label` is invalid on a `<div>` element. I seized the opportunity and used semantic html (ul + li) to describe the keyword list better, which also permits the `aria-label` attribute.

### DIFF
--- a/app/components/Package/Card.vue
+++ b/app/components/Package/Card.vue
@@ -161,7 +161,7 @@ const numberFormatter = useNumberFormatter()
     </div>
 
     <ul
-        role="list"  
+      role="list"
       v-if="result.package.keywords?.length"
       :aria-label="$t('package.card.keywords')"
       class="relative z-10 flex flex-wrap gap-1.5 mt-3 pt-3 border-t border-border list-none m-0 p-0 pointer-events-none items-center"


app/components/Package/Card.vue
```diff
@@ -160,30 +160,30 @@ const numberFormatter = useNumberFormatter()
...
-      >
-        {{ keyword }}
-      </ButtonBase>
-      <span
+      <li v-for="keyword in result.package.keywords.slice(0, 5)" :key="keyword">
+        <ButtonBase
+          class="pointer-events-auto"
+          size="small"
+          :aria-pressed="props.filters?.keywords.includes(keyword)"
+          :title="`Filter by ${keyword}`"
```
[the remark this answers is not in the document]
knowler: This is unrelated to the change that was made and only appears related due to formatting.

knowler (approved):
Nice catch!

app/components/Package/Card.vue
```diff
@@ -160,30 +160,32 @@ const numberFormatter = useNumberFormatter()
       </div>
     </div>
 
-    <div
+    <ul
```
knowler, line 163: Just remembered that we actually need to set `role=list` on this since it’s using `display: flex` and for Safari/VoiceOver, if a list uses flexbox (and maybe grid), it will remove the list semantics unless it’s a descendant of a `<nav>` element.

```suggestion
    <ul
        role="list"  
```

Further reading: <https://www.scottohara.me/blog/2019/01/12/lists-and-safari.html>

TheAlexLichter: TIL! Thanks for letting me know

PR #1402, opened 2026-02-12
fix: replace hardcoded npmx.dev & docs.npmx.dev with env-specific URLs

serhalp:
Extract a composable that returns URLs for this app itself (and docs, which lives in this codebase), so that URLs are always internally consistent within an environment, e.g. in dev links go to dev.

### DIFF
--- a/app/composables/useAppUrls.ts
+++ b/app/composables/useAppUrls.ts
@@ -1,0 +1,11 @@
+import { NPMX_DOCS_SITE_PROD } from '#shared/utils/constants'
+
+export function useAppUrls() {
+  const { env, siteUrl } = useAppConfig()
+  return {
+    siteUrl,
+    // TODO(serhalp): Handle preview environment. The docs site is a separate deployment, so we'd
+    // need to infer its preview URL from the site preview URL somehow...?
+    docsUrl: env === 'dev' ? 'http://localhost:3001' : NPMX_DOCS_SITE_PROD,
+  }
+}
--- a/app/pages/org/[org].vue
+++ b/app/pages/org/[org].vue
@@ -129,7 +129,8 @@ function handleClearFilter(chip: FilterChip) {
 const activeTab = shallowRef<'members' | 'teams'>('members')
 
 // Canonical URL for this org page
-const canonicalUrl = computed(() => `https://npmx.dev/@${orgName.value}`)
+const { siteUrl } = useAppUrls()
+const canonicalUrl = computed(() => `${siteUrl}/@${orgName.value}`)
 
 useHead({
   link: [{ rel: 'canonical', href: canonicalUrl }],
--- a/config/env.ts
+++ b/config/env.ts
@@ -2,6 +2,7 @@
 
 import Git from 'simple-git'
 import * as process from 'node:process'
+import { NPMX_SITE_PROD } from '../shared/utils/constants'
 
 export { version } from '../package.json'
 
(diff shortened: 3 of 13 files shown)


config/env.ts
```diff
@@ -87,6 +88,11 @@ export const getProductionUrl = () =>
         : undefined
     : undefined
 
+export const getSiteUrl = (isDevelopment: boolean) => {
+  if (isDevelopment) return 'http://localhost:3000'
```
knowler, line 92: wondering if we should or should not hardcode this? It is possible to pass flags with the dev command to change the port/host. Could we get that from somewhere else?

danielroe: unfortunately only 127.0.0.1 works for the atproto oauth callback

but yes, you can get the actual host and port in module space

app/composables/useAppUrls.ts
```diff
@@ -0,0 +1,11 @@
+import { NPMX_DOCS_SITE_PROD } from '#shared/utils/constants'
+
+export function useAppUrls() {
+  const { env, siteUrl } = useAppConfig()
+  return {
+    siteUrl,
+    // TODO(serhalp): Handle preview environment. The docs site is a separate deployment, so we'd
+    // need to infer its preview URL from the site preview URL somehow...?
```
sacrosanctic, line 8: https://vercel.com/docs/deployments/generated-urls#generated-from-git

the URL's are deterministic, so can we just construct them during the build?

config/env.ts
serhalp: as I have a feature PR stacked on this one and PRs get stale fast here 😅 and this PR leaves things no worse than before, how would you feel about following up on this later?

app/composables/useAppUrls.ts
serhalp: as I have a feature PR stacked on this one and PRs get stale fast here 😅 and this PR leaves things no worse than before, how would you feel about following up on this later?

app/composables/useAppUrls.ts
```diff
@@ -0,0 +1,11 @@
...
+    docsUrl: env === 'dev' ? 'http://localhost:3001' : NPMX_DOCS_SITE_PROD,
+  }
+}
```
danielroe, line 11: I'm not sure how useful this is - we can't get the preview url of a docs site, and we don't have a hard-coded port for the docs running locally (and this ends up being runtime code we don't really need in the site)

app/pages/org/[org].vue
```diff
@@ -129,7 +129,8 @@ function handleClearFilter(chip: FilterChip) {
 const activeTab = shallowRef<'members' | 'teams'>('members')
 
 // Canonical URL for this org page
-const canonicalUrl = computed(() => `https://npmx.dev/@${orgName.value}`)
+const { siteUrl } = useAppUrls()
+const canonicalUrl = computed(() => `${siteUrl}/@${orgName.value}`)
```
Comment (danielroe) [inline app/pages/org/[org].vue:133]: this ends up meaning that preview deployments get their own canonical url, and I would try to avoid that. just like with og images, you want to point to the canonical https://npmx.dev so that search engines don't end up indexing preview deployments accidentally

(I know we have a header instructing them not to, but this seems safe to hard-code. if we ever need to switch https://npmx.dev, we could easily do it throughout the codebase without much effort.)

serhalp:
> maybe you could explain a bit more about why?
> 
> I'm not sure I see a benefit here - it's straightforward enough in local development to open the docs in a new tab, and canonical urls or json-ld aren't really relevant either for previews or local development.

In https://github.com/npmx-dev/npmx.dev/pull/1405 I'm adding a link to one of our doc pages from within npmx. When you're in local dev, this link would otherwise go to prod, unlike every other internal link (I think of it as an "internal" link...).

I assume over time we'll be adding more and more deep links to npmx docs within npmx, and more and more contributors will waste time being confused by this (not seeing local changes reflected, 404 navigating to a new page, etc.).

Maybe I've just been bitten too many damn times by this during my career...

---

In any case, to keep things clean I'll rebase my feat PR to remove these commits and hardcode the prod docs domain for now. 👍🏼

PR #1409, opened 2026-02-12
docs: npm comparison/features

danielroe:
this might be too verbose e.g. for the readme, but this is almost more for my own brain to keep track of things.

wdyt?


whitep4nth3r:
What about rendering this on an actual web page on the front end? 🤓

serhalp:
@danielroe Is this one still relevant? Does it need a review?

PR #1417, opened 2026-02-12
feat: add clickable anchor links to readme headings

essenmitsosse:
Screenshot 2026-02-12 at 11 28 24

### DIFF
--- a/server/utils/readme.ts
+++ b/server/utils/readme.ts
@@ -378,7 +378,7 @@ export async function renderReadmeHtml(
     }
 
     /** The link href uses the unique slug WITHOUT the 'user-content-' prefix, because that will later be added for all links. */
-    return `<h${semanticLevel} id="${id}" data-level="${depth}"><a href="#${uniqueSlug}">${text}</a></h${semanticLevel}>\n`
+    return `<h${semanticLevel} id="${id}" data-level="${depth}"><a href="#${uniqueSlug}">${plainText}</a></h${semanticLevel}>\n`
   }
 
   // Syntax highlighting for code blocks (uses shared highlighter)


essenmitsosse:
There is a still a bug for certain headings, that were luckily caught by lighthouse. I am on it.

whitep4nth3r:
Just marking as draft as it's not ready yet and a couple of us are going down the list to see what we can review!

essenmitsosse:
@whitep4nth3r it is ready now.

whitep4nth3r:
Couple of things. I tested on mobile. 

1. When you click on a link, the top of the section scrolls out of view. 
2. Can you make the full heading clickable? Especially on mobile it's hard to know it's clickable, and it would make it easier to click rather than just clicking on the small icon on the left. (It's not obvious but I was trying to click on the headings.)

https://github.com/user-attachments/assets/2c7db955-d7af-46ee-a562-7b26d61a50db

danielroe:
> When you click on a link, the top of the section scrolls out of view.

we probably need to add scroll margin if you haven't already

danielroe:
honestly I thought this had already been done (including scroll-margin!) - maybe it was somehow reverted.

essenmitsosse:
@knowler 

1. Id be fine with that. Shall in update?
2. yes that is in theory possible. Nuxt for example has links in the headings. But without `href` 🤷‍♂️ they are pretty broken on GitHub as well. Otherwise I'd say let's copy GitHub's and/or npmjs.com behavior because that is what people will be testing against.

essenmitsosse:
Ok. Let's _not_ copy npmsjs. This is completely broken: https://www.npmjs.com/package/nuxt#local-development

essenmitsosse:
@whitep4nth3r 
The full heading should be clickable. 

I created an issue #1428 for the scroll margin problem because that also affects other things (like toc) and I have no idea why it is not working.

essenmitsosse:
I think GitHub does it fine: 

Anchor is before the heading, always visible on mobile and if the heading has a link only the anchor is clickable. 

Good example: https://github.com/facebook/react

knowler:
In https://github.com/npmx-dev/npmx.dev/pull/1417#issuecomment-3892801706, @essenmitsosse said,

> Anchor is before the heading, always visible on mobile and if the heading has a link only the anchor is clickable.

In [the React GitHub example](https://github.com/facebook/react), the full heading is never clickable though.

Visually, it’d be fine to put the anchor before, but for assistive tech discoverability (e.g. navigating by heading), it would be good to place it after in the DOM (GitHub also does this). This is an exception to the DOM order matches visual order rule.

In https://github.com/npmx-dev/npmx.dev/pull/1417#issuecomment-3892752523, @essenmitsosse said,

> yes that is in theory possible. Nuxt for example has links in the headings. But without href 🤷‍♂️ they are pretty broken on GitHub as well. Otherwise I'd say let's copy GitHub's and/or npmjs.com behavior because that is what people will be testing against.

Nuxt README is using [an old school technique](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#name). I’m less worried for [the `<a>` without `href` cases because they don’t implicitly get the `link` role](https://www.w3.org/TR/html-aria/#el-area-no-href).

It could be ok if we strip out the `<a>` elements, but I could see some cases where they serve a purpose. The React README example would be perfect… if they included _that_ one in the package (i.e. it looks like they swap it out for [a more package focused one](https://npmx.dev/package/react#readme)).

essenmitsosse:
I honestly struggle to find any package that has links in their readme heading. Would it be too farfetched to assume that is not a feature we need to support or should we support it just because it is possible?

If we ingore this case, I think the anchor after the heading is fine. It also makes it less tricky to place it on mobile without having to much spacing or indenting the headings.

In which case I would assume the current version is good enough for now?

PR #1419, opened 2026-02-12
chore: add `.node-version` and ignore `.nvmrc`

essenmitsosse:
This might be the single biggest change ever to npmx.

It allows to use `nvm use` or automatic switching of Node version.

We might also want to change `engine` in `package.json` to `24`, instead of `>=24`, because I wouldn't optimistically want to switch to `25` (which isn't even a stable release` or `26`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,7 +142,7 @@
     "vue-tsc": "3.2.4"
   },
   "engines": {
-    "node": ">=24"
+    "node": "24"
   },
   "simple-git-hooks": {
     "pre-commit": "npx lint-staged"


serhalp:
I don't want to bikeshed this, but could we use the standard `.node-version` instead? Pretty much all node version managers support it these days, don't they? (while `.nvmrc` is specific to nvm`)

essenmitsosse:
Unfortunately [nvm does not](https://github.com/nvm-sh/nvm/issues/794). The issue is over a decade old. I would be fine to also add `.node-version`. It's annoying that everybody uses `.node-version` except for the most common version manager. Its a little overhead, but I think, since this will only be updated once a year, while making everyones life easier it's a fine trade-off.

RYGRIT:
As far as I know, `.nvmrc` does not work on Windows.😭

ghostdevv:
What do you use instead @RYGRIT?

lorypelli:
> I've removed `.nvmrc` because I don't like non-standard config in the repo, and I do think there are better alternatives now, but I have added it to .gitignore so you can add your own .nvmrc without it getting committed automatically

Even `.node-version` shouldn't be a standard and `.nvmrc` is more standard (if I'm not wrong)

danielroe:
I don't believe `.nvmrc` is more standard - it's the config file for one tool called nvm.

essenmitsosse:
But unfortunately it's the one tool most people use. But the gitignore thing is fine for now. At least for me.

lorypelli:
> I don't believe `.nvmrc` is more standard - it's the config file for one tool called nvm.

.node-version shouldn't be a standard at all if I'm not wrong, nvmrc doesn't support it for example, but I totally agree with you, on a repo of mine, I replaced .nvmrc with .node-version

RYGRIT:
> What do you use instead @RYGRIT?

Currently, no alternative solution has been found. `nvm use` seems to only work properly with administrator privileges; I always switch manually. @ghostdevv

lorypelli:
> > What do you use instead @RYGRIT?
> 
> Currently, no alternative solution has been found. `nvm use` seems to only work properly with administrator privileges; I always switch manually. @ghostdevv 

https://github.com/Schniz/fnm

PR #1426, opened 2026-02-12
fix: blank page when javascript is disabled

### LINKED ISSUE #1291: package page doesn't load with javascript disabled
It doesn't work at all without javascript and before it was working

NandkishorJadoun:
resolves #1291 

## Description

This pr adds a nitro server plugin that injects a `<meta http-equiv="refresh">` tag specifically into the SPA fallback shell (`200.html`), wrapped in a `<noscript>` block.

### DIFF
--- a/app/utils/prehydrate.ts
+++ b/app/utils/prehydrate.ts
@@ -16,7 +16,9 @@ export function initPreferencesOnPrehydrate() {
     const validPMs = new Set(['npm', 'pnpm', 'yarn', 'bun', 'deno', 'vlt'])
 
     // Read settings from localStorage
-    const settings = JSON.parse(localStorage.getItem('npmx-settings') || '{}')
+    const settings = JSON.parse(
+      localStorage.getItem('npmx-settings') || '{}',
+    ) as Partial<AppSettings>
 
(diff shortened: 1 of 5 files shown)


NandkishorJadoun:
@danielroe , should i reduce the delay to 0 seconds? im concerned that might trigger an infinite loop, though I'm not entirely sure. i've set it to 1 second for now to be safe

ghostdevv:
cc: @LoryPelli

danielroe:
we may need to disable the validator on 200.html

on the other hand, I think setting to 0 should be safe.

server/plugins/fallback-ui.ts
```diff
@@ -0,0 +1,7 @@
+export default defineNitroPlugin(nitroApp => {
+  nitroApp.hooks.hook('render:html', (html, { event }) => {
+    if (event.path.includes('/200.html')) {
+      html.head.push('<noscript><meta http-equiv="refresh" content="1"></noscript>')
```
danielroe:
```suggestion
      html.head.push('<noscript><meta http-equiv="refresh" content="0"></noscript>')
```

PR #1434, opened 2026-02-12
fix:  JSON module should be marked as ESM type

### LINKED ISSUE #1420: JSON-only package should be marked as supporting ESM
https://npmx.dev/package/@tc39/ecma262-biblio

This package's `exports` do not contain any CommonJS code, only JSON. It should probably be marked as supporting ESM.

eryue0220:
resolve #1420


nicolo-ribaudo:
Thank you!

PR #1439, opened 2026-02-12
fix: consistent and likely safer regex escaping

ghostdevv:
Replaces random regex escaping with [`@li/regexp-escape-polyfill`](https://jsr.io/@li/regexp-escape-polyfill), a spec-compliant polyfill for [`RegExp.escape()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape). I searched for every `RegExp` call, and replaced the ones that needed them. Most didn't use any dynamic strings, but the couple that did are updated here or didn't need it

### DIFF
--- a/shared/utils/emoji.ts
+++ b/shared/utils/emoji.ts
@@ -1,3 +1,5 @@
+import { regExpEscape } from '@li/regexp-escape-polyfill'
+
 // copied from https://github.com/markdown-it/markdown-it-emoji/blob/master/lib/data/full.mjs
 const emojis = {
   '100': '💯',
@@ -1907,7 +1909,7 @@ const emojis = {
 
 const emojisKeysRegex = new RegExp(
   Object.keys(emojis)
-    .map(key => `:${key}:`)
+    .map(key => `:${regExpEscape(key)}:`)
     .join('|'),
   'g',
 )


ghostdevv:
> I'm not sure that the benefit is worth the cost in terms of browser compatibility.

yeaaa maybe, perhaps there is a minimal ponyfill that achieves the same result :thinking: (because we need something for this, and I'd like to avoid having multiple fns for it over the codebase)

ghostdevv:
some initial findings, need to step away for a bit so will take another look later

- https://jsr.io/@li/regexp-escape-polyfill
- https://github.com/zloirock/core-js/blob/master/packages/core-js/modules/es.regexp.escape.js

global.d.ts
```diff
@@ -0,0 +1,9 @@
+declare global {
+  interface RegExpConstructor {
+    escape(str: string): string
+  }
+}
+
+// required for the global type to work
+// oxlint-disable-next-line eslint-plugin-unicorn(require-module-specifiers)
+export {}
```
wojtekmaj: moduleDetection: force could be considered instead.

knip.ts
```diff
@@ -55,6 +55,7 @@ const config: KnipConfig = {
         'h3-next',
       ],
       ignoreUnresolved: ['#components', '#oauth/config'],
+      ignoreFiles: ['global.d.ts'],
```
wojtekmaj: Nit: we could just ignore all *.d.ts because we never expect them to be imported and yet they will be picked up by TS.

wojtekmaj:
> I'm not sure that the benefit is worth the cost in terms of browser compatibility.
> 
> wdyt @ghostdevv?

On the other hand, I don't think a large percentage of npmx audience is using IE or is 11 Chrome versions behind.

danielroe:
regexp.escape won't work on iOS 17 devices. I agree we need a canonical single escape function but it's a tiny one-liner, for our purposes. (https://github.com/lionel-rowe/regexp-escape-polyfill (linked by @ghostdevv above) is just [498 B](https://bundlejs.com/?q=jsr%3A%40li%2Fregexp-escape-polyfill&treeshake=%5B*%5D))

ghostdevv:
updated to use `@li/regexp-escape-polyfill`

ghostdevv:
For some reason the vercel preview is crashing when you access a package page directly :thinking: it doesn't happen in dev or local preview

43081j:
i caught up from main, and the preview seems to work now too FYI

PR #1440, opened 2026-02-12
fix: do not run lunaria when making a merge commit

alexdln:
I noticed during a meeting that when resolving merge-conflicts with `--continue`, Lunaria crashes (_it often can't find the file combination for large changes_). But when using `--continue` - you can't use `--no-verify` to bypass it.

I've added an option to bypass it via an env variable (`NO_VERIFY=1`) until we figure it out better

`NO_VERIFY=1 git merge --continue` 

or something similar if passing variables is not supported in the system/terminal

`cross-env NO_VERIFY=1 git merge --continue`

@ghostdevv fyi 

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,7 +145,7 @@
     "node": "24"
   },
   "simple-git-hooks": {
-    "pre-commit": "[ -n \"$NO_VERIFY\" ] && exit 0\nnpx lint-staged"
+    "pre-commit": "npx lint-staged"
   },
   "lint-staged": {
     "i18n/locales/*": [
--- a/lunaria/lunaria.ts
+++ b/lunaria/lunaria.ts
@@ -1,8 +1,15 @@
 import { createLunaria } from '@lunariajs/core'
-import { mkdirSync, readFileSync, writeFileSync } from 'node:fs'
+import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'node:fs'
 import { Page } from './components.ts'
 import { lunariaJSONFiles, prepareJsonFiles } from './prepare-json-files.ts'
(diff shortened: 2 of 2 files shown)


danielroe:
I've found that running `git commit --no-verify` works fine to complete a merge if you encounter this.

alexdln:
Yes, but some people strongly prefer `--continue` (_it seems like a better option_)

ghostdevv:
> Yes, but some people strongly prefer `--continue` (_it seems like a better option_)

yea it _works_, it'd just be nice if the hooks wouldn't break in that case since I think merge is "more correct" (and certainly easier)

package.json
```diff
@@ -145,7 +145,7 @@
     "node": "24"
   },
   "simple-git-hooks": {
-    "pre-commit": "npx lint-staged"
+    "pre-commit": "[ -n \"$NO_VERIFY\" ] && exit 0\nnpx lint-staged"
```
ghostdevv: is there any worry about this potentially not working on Windows? I don't have a windows machine to test it unfortunately

ghostdevv (changes requested):

package.json
alexdln: I use Windows, everything works 😅

It is standard POSIX shell, should work properly on all systems

danielroe:
passing an env variable seems like just as much work. in honesty, what if we update _lunaria_ not to break in this case? we could detect if we're in the middle of a merge commit, and bail? 💡

danielroe:
pushing something

danielroe:
wdyt?

alexdln:
Perfect

PR #1443, opened 2026-02-12
fix(a11y): include latest hash and params for skip button

### LINKED ISSUE #1437: Skip link clears query params
**Description**

Using the skip link clears out query parameters from the URL when it is activated. This prevents using the skip link to view search results.

**Reproduction Steps**

1. Navigate to npmx with a search query: <https://npmx.dev/search?q=intl-messageformat>
2. Tab once to show the skip link, then press enter
3. Focus moves to main content, but the search query gets scrubbed from the URL and search results disappear.
4. Resulting URL: `https://npmx.dev/search`

OrbisK:
fixes #1437 

ensures that we keep the query and params on skip link.

### DIFF
--- a/app/app.vue
+++ b/app/app.vue
@@ -121,9 +121,12 @@ if (import.meta.client) {
 <template>
   <div class="min-h-screen flex flex-col bg-bg text-fg">
     <NuxtPwaAssets />
-    <LinkBase to="#main-content" variant="button-primary" class="skip-link">{{
-      $t('common.skip_link')
-    }}</LinkBase>
+    <LinkBase
+      :to="{ hash: '#main-content', query: route.query, params: route.params }"
+      variant="button-primary"
+      class="skip-link"
+      >{{ $t('common.skip_link') }}</LinkBase
+    >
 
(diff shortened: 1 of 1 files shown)


OrbisK:
`useRouteQuery` has the same workaround: https://github.com/vueuse/vueuse/blob/7069eb4cc5ffd7d10ac07277db2d4b67967dfc3f/packages/router/useRouteQuery/index.ts#L93-L97

ben-basten:
Thank you for the quick response to the issue! I just tried the deployment preview - it looks like this PR fixes the query param issue, but now focus no longer moves to the main content and stays on the skip link. So the actual "skip" never happens.

danielroe:
merging as a hotfix but I think this is an issue with LinkBase

alexdln:
It seems that "NuxtLink" and HTML "a" behave differently in the context of hash anchors...

OrbisK:
@ben-basten Thank you! I have fixed this in #1460

ben-basten:
@OrbisK That looks great, thank you!!

PR #1444, opened 2026-02-12
feat(i18n): update Chinese (Simplified) translations.

craigary:
### 1. Added `vacations` section
* Added a complete set of translations for the **vacations** section.
* Covers the team's break from Feb 14–21 and Discord availability.

### 2. Punctuation & Key Completion
* **Punctuation**: Standardized the use of full-width parentheses `）` and two-em dashes `⸺` for better CJK typography.
* **Missing Keys**: Updated translation for `pkg.versions.y_axis_label`.

### 3. Terminology: Reverting `Copy` to `拷贝`
Updated 10 instances of `Copy` related actions to use `拷贝` instead of `复制`.

Regarding this point, I previously updated `Copy` to `拷贝` in a prior PR. However, it was reverted back to `复制` in PR #1108. I believe `拷贝` is a more appropriate translation for `Copy` than `复制` for the following reasons:

* **Semantic Clarity**: In Chinese, `复制` is more synonymous with **Duplicate** (creating an identical replica). `拷贝` better represents the action of placing content onto the clipboard for a "Copy and Paste" workflow.
* **Platform Standards**: Following the standard established by **Apple's Human Interface Guidelines**, "Copy" is localized as `拷贝`. While `复制` is common in some Windows-influenced contexts (which is clearly wrong imo 😈), translating "Copy" as `拷贝` is both accurate and unambiguous.
* **Phonetic & Technical Precision**: As a direct transliteration of `Copy`, `拷贝` is a well-established technical term in Chinese that eliminates ambiguity, maintaining a consistent professional tone throughout the project.

cc @danielroe @9romise @KazariEX

### DIFF
--- a/i18n/locales/ja-JP.json
+++ b/i18n/locales/ja-JP.json
@@ -59,9 +59,9 @@
     "exact_match": "完全一致",
     "suggestion": {
       "user": "ユーザー",
-      "org": "Organization",
+      "org": "organization",
       "view_user_packages": "このユーザーのパッケージを表示",
-      "view_org_packages": "このOrganizationのパッケージを表示"
+      "view_org_packages": "このorganizationのパッケージを表示"
     }
   },
   "nav": {
@@ -90,9 +90,9 @@
       "label": "データソース",
       "description": "npmxが検索データを取得する先を選択します。個別のパッケージページは常にnpmレジストリから直接取得します。",
       "npm": "npmレジストリ",
-      "npm_description": "公式npmレジストリから検索、Organization、ユーザー一覧を直接取得します。正確ですが、やや遅い場合があります。",
+      "npm_description": "公式npmレジストリから検索、organization、ユーザー一覧を直接取得します。正確ですが、やや遅い場合があります。",
       "algolia": "Algolia",
-      "algolia_description": "Algoliaを使用して検索、Organization、ユーザーページをより高速に取得します。"
+      "algolia_description": "Algoliaを使用して検索、organization、ユーザーページをより高速に取得します。"
     },
     "relative_dates": "日付を相対表記",
     "include_types": "インストール時に {'@'}types を含める",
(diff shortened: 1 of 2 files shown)


RYGRIT:
Thanks for the PR and detailed rationale!

I still prefer sticking with `复制` for these reasons:

- It matches what most users see in browsers (Chrome/Edge right-click → `复制`), which is the main way people access npmx.dev.

- In the Chinese dev community (B站, Zhihu, 掘金, etc.), `复制` is the overwhelming default for clipboard actions like copy install command or code snippets.

- While Apple uses `拷贝`, this is a cross-platform web app, not a macOS-native one. Aligning with browser/default Chinese usage feels more natural for our audience.

Happy to discuss more if you have other thoughts!

9romise:
Personally, I prefer "复制", just like my original translation. To me, it clearly reads as shorthand for "复制到剪贴板" (copy to clipboard).
In most mainstream Chinese apps (WeChat, Douyin, Xiaohongshu, etc.), "复制" is the standard translation for "copy". It’s already a very commonly used term.
I don’t think repeatedly toggling between "复制" and "拷贝" adds much value, both are perfectly understandable. All it really does is drain the maintainers’ energy.

i18n/locales/zh-CN.json
```diff
@@ -906,7 +907,7 @@
       "add_hint": "至少添加 2 个包以进行比较。"
     },
     "no_dependency": {
-      "label": "(不使用依赖)",
+      "label": "（不使用依赖）",
```
KazariEX: Here the half-width parentheses are used because the spacing of full-width parentheses would make the rendered visual appearance look unattractive.

| before | after |
| --- | --- |
| Image | Image |

craigary:
> I don’t think repeatedly toggling between "复制" and "拷贝" adds much value, both are perfectly understandable. All it really does is drain the maintainers’ energy.

That's totally valid! My initial reasoning was that since iOS translates "Duplicate" as "复制", the term has become ambiguous on digital platforms 1. Copying to the clipboard for a subsequent action, and 2. Creating an immediate duplicate. Translating "Copy" as "拷贝" effectively resolves this confusion.

I’ve unified the translation of 'Copy' to '复制'.

i18n/locales/zh-CN.json
craigary: Updated~

PR #1445, opened 2026-02-12
feat: add GitHub contributors graph

Tobbe:
I started working on this new graph. WDYT? 
I wanted to be able to see how "healthy", from a number-of-contibutors perspective a package is. 
Like, even if a package has hundreds of contributors, maybe those are all from years past, and the package is now basically only kept alive by one person, which isn't too good from a reliability/bus factor point of view
It uses the GitHub API, so only works for packages that have a github repo linked in the package metadata

image


graphieros:
When switching to monthly or yearly, the component is designed to display estimations for incomplete periods.
It was designed originally to chart downloads. However this KPI is not compatible with estimations. The component must be adapted to show estimations for compatible facets.

The facet does not appear to be working on the compare page.

I would also move the fetching to the useCharts composable

Tobbe:
> When switching to monthly or yearly, the component is designed to display estimations for incomplete periods. It was designed originally to chart downloads. However this KPI is not compatible with estimations. The component must be adapted to show estimations for compatible facets.

Yes, I did find it a bit weird that it was estimating how many contributors a package will have 😀  How should we handle the incomplete period though? 🤔

graphieros:
> > When switching to monthly or yearly, the component is designed to display estimations for incomplete periods. It was designed originally to chart downloads. However this KPI is not compatible with estimations. The component must be adapted to show estimations for compatible facets.
> 
> Yes, I did find it a bit weird that it was estimating how many contributors a package will have 😀 How should we handle the incomplete period though? 🤔

We are dealing with absolutes. n contributors today does not add up with n yesterday. So weekly, monthly, yearly should be averages.

Tobbe:
> The facet does not appear to be working on the compare page.

Yeah, I hadn't gotten around to fully testing that one yet.
I've made it work now (locally)
image

But I also need to find a package that doesn't have a GitHub repo in its package.json to test with.
You don't happen to know of one offhand, do you?

Tobbe:
I faked one not having a github link, just to be able to test how it'd work.
It now prints a small notice for packages not included

image

Tobbe:
> > > When switching to monthly or yearly, the component is designed to display estimations for incomplete periods. It was designed originally to chart downloads. However this KPI is not compatible with estimations. The component must be adapted to show estimations for compatible facets.
> > 
> > 
> > Yes, I did find it a bit weird that it was estimating how many contributors a package will have 😀 How should we handle the incomplete period though? 🤔
> 
> We are dealing with absolutes. n contributors today does not add up with n yesterday. So weekly, monthly, yearly should be averages.

Likes also isn't additive, right? So maybe likes and contributors should be handled differently compared to downloads.
But I don't want to touch the logic for existing graphs in this PR. So how about we leave it like it is for all three facets for now, and then have another PR to fix this logic?

Tobbe:
@graphieros Can we mark this as ready to review again?

graphieros:
I don't think it should be shipped with estimations for contributors.
I opened it for review again, would love to have more opinions.

Tobbe:
> I don't think it should be shipped with estimations for contributors.

I also agree that it's wrong.

You said
> We are dealing with absolutes. n contributors today does not add up with n yesterday. So weekly, monthly, yearly should be averages.

So just to make sure I understand what you mean, if I select "yearly" as the period, it should show the average so far for 2026 as the value for the full year? And same for "monthly", it should show the average so far for February as the value for this month?
Or did I misunderstand you @graphieros?

graphieros:
> > I don't think it should be shipped with estimations for contributors.
> 
> I also agree that it's wrong.
> 
> You said
> 
> > We are dealing with absolutes. n contributors today does not add up with n yesterday. So weekly, monthly, yearly should be averages.
> 
> So just to make sure I understand what you mean, if I select "yearly" as the period, it should show the average so far for 2026 as the value for the full year? And same for "monthly", it should show the average so far for February as the value for this month? Or did I misunderstand you @graphieros?

Yes that's right. An average of unique contributors, if this data is available of course.

Tobbe:
@graphieros I removed the extrapolation for Contributors.
I want each period in the chart be the total number of unique contributors for that period. So for Monthly "January" should be the total number of unique contributors for January. For February (which we're only halfway through now), the total number of unique contributors will not go down. So I just made it show the current total number of contributors (i.e. not try to guess any future value at all).

Tobbe:
@ghostdevv you requested changes. Is this good enough? I.e. no estimation, just current total

image

image

graphieros:
Looks good to me now !
And you are right, estimations should not be applied on likes.

Tobbe:
@graphieros Thanks for working with me on this one 🙏

graphieros:
@Tobbe I recommend you merge main in your branch, since vue-data-ui version was bumped and minor config details updated on the chart file

graphieros:
Ok, looks mergeable to me

Tobbe:
> Ok, looks mergeable to me

Yeah, I was just going to say it looks good :)

image

Tobbe:
@graphieros What's the setup like here, do we need to wait for @ghostdevv to review again?

graphieros:
Yep ;)

ghostdevv (approved):

Issue #1421, opened 2026-02-12
Add button to copy the version number


nicolo-ribaudo:
<img width="684" height="233" alt="Image" src="https://github.com/user-attachments/assets/96f47ae2-49a7-4e7a-a378-df3941101228" />

Somewhere here around the package name it'd be great to have a button to copy the version number, so that then I can paste it in my `package.json`. Right now I have to aim with my cursor to exclude the `v` and include up to the last digit but not more, which is a bit annoying.

Or maybe it could be that when I click on the version number it automatically selects it, and then I hit Ctrl+C by myself.

nicolo-ribaudo:
Adding a video showing my poor text selection skills to show what I mean by "I have to aim with the cursor" :)

https://github.com/user-attachments/assets/1ef60088-7677-4695-a04a-3e2ee09a51b6

colabottles:
I can take this.

agamyo168:
oh I forgot to write that I'm working on this ;-; can I post my work?

colabottles:
Go for it

PR #1447, opened 2026-02-12
feat: add copy button for package versions

agamyo168:
this is my attempt at resolving: https://github.com/npmx-dev/npmx.dev/issues/1421 
I'm not really good at front-end and I have no experience with vue so feel free to judge.

I have added a small button to copy the package version:
image

### DIFF
--- a/app/pages/package/[[org]]/[name].vue
+++ b/app/pages/package/[[org]]/[name].vue
@@ -712,9 +712,7 @@ const showSkeleton = shallowRef(false)
           >
             <!-- Version resolution indicator (e.g., "latest → 4.2.0") -->
             <template v-if="requestedVersion && resolvedVersion !== requestedVersion">
-              <span class="font-mono text-fg-muted text-sm" dir="ltr">
-                {{ requestedVersion }}
-              </span>
+              <span class="font-mono text-fg-muted text-sm" dir="ltr">{{ requestedVersion }}</span>
               <span class="i-carbon:arrow-right rtl-flip w-3 h-3" aria-hidden="true" />
             </template>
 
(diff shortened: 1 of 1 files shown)


app/pages/package/[[org]]/[name].vue
```diff
@@ -246,6 +246,12 @@ const { copied: copiedPkgName, copy: copyPkgName } = useClipboard({
   copiedDuring: 2000,
 })
 
+//copy package name
```
Comment (nicolo-ribaudo) [inline app/pages/package/[[org]]/[name].vue]:
```suggestion
//copy package version
```

app/pages/package/[[org]]/[name].vue
```diff
@@ -246,6 +246,12 @@ const { copied: copiedPkgName, copy: copyPkgName } = useClipboard({
...
+const { copied: copiedVersion, copy: copyVersion } = useClipboard({
+  source: resolvedVersion.value ?? '',
```
Comment (nicolo-ribaudo) [inline app/pages/package/[[org]]/[name].vue:251]: In which case there is no value? Maybe we should just not show the copy button when it happens.

nicolo-ribaudo:
Oh also, this copies the version _without_ the "v" prefix, right? (I'd expect it to do so)

app/pages/package/[[org]]/[name].vue
Comment (agamyo168) [reply app/pages/package/[[org]]/[name].vue:251]: I thought this was already the case? This is why I encapsulated it in the version span .

agamyo168:
> Oh also, this copies the version _without_ the "v" prefix, right? (I'd expect it to do so)

Yes, the button only copies the version without the v prefix

app/pages/package/[[org]]/[name].vue
Comment (nicolo-ribaudo) [reply app/pages/package/[[org]]/[name].vue:251]: Oh right

PR #1453, opened 2026-02-13
feat: improve scroll-to-top behavior

### LINKED ISSUE #1379: Back-to-top button on desktop
This perhaps can be discussed.

1. The back-top-top is currently only visible on mobile.
I think it would be fitting to also show it on desktop.

2. The default smooth scroll behavior could be normalized so that its timing be fixed, regardless of the scroll height. This would require JS, since no easing nor timing is available in the browser API.

RYGRIT:
Fixes #1379

Improves the ScrollToTop component behavior.

- **Desktop Support**: Enabled the button on desktop and fixed the visibility fallback logic for browsers without `scroll-state` support.
- **Better UX**: Normalized scroll duration (500ms) and added support for `prefers-reduced-motion`.
- **Interaction**: The auto-scroll animation now cancels immediately upon user interaction (wheel, touch, or click).

<details>

<summary>
Desktop
</summary>

https://github.com/user-attachments/assets/95f59dd1-16b1-4a72-940a-b2b81e68ac7f

</details>

<details>

<summary>
Mobile
</summary>

https://github.com/user-attachments/assets/755cb71f-4ef9-4a76-8d5e-3e8d81b3cb91

</details>

cc @graphieros @MatteoGabriele

### DIFF
--- a/app/components/Package/TrendsChart.vue
+++ b/app/components/Package/TrendsChart.vue
@@ -1410,7 +1410,7 @@ const chartConfig = computed<VueUiXyConfig>(() => {
     chart: {
       height: isMobile.value ? 950 : 600,
       backgroundColor: colors.value.bg,
-      padding: { bottom: displayedGranularity.value === 'yearly' ? 84 : 64, right: 100 }, // padding right is set to leave space of last datapoint label(s)
+      padding: { bottom: displayedGranularity.value === 'yearly' ? 84 : 64, right: 128 }, // padding right is set to leave space of last datapoint label(s)
       userOptions: {
         buttons: {
           pdf: false,
@@ -1682,7 +1682,11 @@ watch(selectedMetric, value => {
     </h2>
 
     <!-- Chart panel (active metric) -->
-    <div role="region" aria-labelledby="trends-chart-title" class="min-h-[260px]">
+    <div
+      role="region"
+      aria-labelledby="trends-chart-title"
+      :class="isMobile === false && width > 0 ? 'min-h-[567px]' : 'min-h-[260px]'"
+    >
       <ClientOnly v-if="chartData.dataset">
         <div :data-pending="pending" :data-minimap-visible="maxDatapoints > 6">
           <VueUiXy
--- a/app/components/Package/WeeklyDownloadStats.vue
+++ b/app/components/Package/WeeklyDownloadStats.vue
@@ -264,7 +264,7 @@ const config = computed<VueUiSparklineConfig>(() => {
 </script>
 
 <template>
-  <div class="space-y-8">
+  <div class="space-y-8 h-[110px] motion-safe:h-[140px]">
     <CollapsibleSection id="downloads" :title="$t('package.downloads.title')">
       <template #actions>
         <ButtonBase
@@ -306,6 +306,10 @@ const config = computed<VueUiSparklineConfig>(() => {
                   <div class="flex-1 flex items-end pe-3">
                     <SkeletonInline class="h-px w-full" />
                   </div>
+                </div>
+                <!-- Animation toggle placeholder -->
+                <div class="w-full hidden motion-safe:flex flex-1 items-end justify-end">
+                  <SkeletonInline class="h-[20px] w-30" />
                 </div>
               </div>
             </template>
@@ -351,10 +355,7 @@ const config = computed<VueUiSparklineConfig>(() => {
 
     <!-- This placeholder bears the same dimensions as the PackageTrendsChart component -->
     <!-- Avoids CLS when the dialog has transitioned -->
-    <div
-      v-if="!hasChartModalTransitioned"
-      class="w-full aspect-[390/634.5] sm:aspect-[718/622.797]"
-    />
+    <div v-if="!hasChartModalTransitioned" class="w-full aspect-[390/634.5] sm:aspect-[718/647]" />
   </PackageChartModal>
 </template>
 
--- a/app/composables/npm/useResolvedVersion.ts
+++ b/app/composables/npm/useResolvedVersion.ts
@@ -4,15 +4,17 @@ export function useResolvedVersion(
   packageName: MaybeRefOrGetter<string>,
   requestedVersion: MaybeRefOrGetter<string | null>,
 ) {
-  return useFetch(
-    () => {
+  return useAsyncData(
+    () => `resolved-version:${toValue(packageName)}:${toValue(requestedVersion) ?? 'latest'}`,
+    async () => {
       const version = toValue(requestedVersion)
-      return version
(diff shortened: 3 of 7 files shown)


graphieros:
> these look like good changes to the scroll composable 👌
> 
> my only question is about displaying the button on desktop. it feels like an affordance needed for mobile, but users with a keyboard will have a 'Home' key and an extra button clutters the UI

Would the button appear on tablets ?

danielroe:
it should probably be gated not by window width but by `isTouchDevice`
https://github.com/npmx-dev/npmx.dev/blob/8a70d6e124dc3c1560a59c76334ac64867dd07b7/app/utils/responsive.ts

RYGRIT:
> it should probably be gated not by window width but by `isTouchDevice` https://github.com/npmx-dev/npmx.dev/blob/8a70d6e124dc3c1560a59c76334ac64867dd07b7/app/utils/responsive.ts

Good point. I'll switch to using `isTouchDevice` to avoid cluttering the desktop UI.

> 'Home' key

Ah, TIL about the Home/End shortcuts! Thanks for the tip.

RYGRIT:
I've already implemented `isTouchDevice()` to determine if a button is hidden, but I've found that when simulating a mobile device in the browser, I have to refresh the browser for the button to appear.

app/components/ScrollToTop.client.vue
```diff
@@ -6,41 +6,44 @@ const excludedRoutes = new Set(['index', 'code'])
...
 const isMounted = useMounted()
+const isTouchDeviceClient = shallowRef(false)
 const isVisible = shallowRef(false)
 const scrollThreshold = 300
 const { isSupported: supportsScrollStateQueries } = useCssSupports(
   'container-type',
   'scroll-state',
   { ssrValue: false },
 )
+const shouldShowButton = computed(() => isActive.value && isTouchDeviceClient.value)
```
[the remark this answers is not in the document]
graphieros: This is ok, requirements changed to only show the button on touch decices

danielroe:
looking great!

but one more thought - we might want to update the implementation on the package page as we already have a bottom bar - maybe (on that page) build another button into that bottom bar and hide this one?

Screen Shot 2026-02-13 at 14 47 16

RYGRIT:
Yes, I also think they should be merged.

However, this raises another issue: when switching to other languages, the floating buttons at the bottom extend beyond the screen, and the display changes accordingly (for example, the floating button area becomes higher). I mentioned this issue in PR #1284 before. I'm unsure whether the text content should be hidden on small screens. I haven't thought of a better solution at the moment.


<details>
<summary>Device: iPhone 14 Pro Max - browser (simulated)</summary>

image
image
</details>

image

app/components/ScrollToTop.client.vue
```diff
@@ -6,41 +6,44 @@ const excludedRoutes = new Set(['index', 'code'])
...
   <!-- When CSS scroll-state is supported, use CSS-only visibility -->
   <button
-    v-if="isActive && supportsScrollStateQueries"
+    v-if="shouldShowButton && supportsScrollStateQueries"
     type="button"
-    class="scroll-to-top-css fixed bottom-4 inset-ie-4 z-50 w-12 h-12 bg-bg-elevated border border-border rounded-full shadow-lg md:hidden flex items-center justify-center text-fg-muted hover:text-fg transition-colors active:scale-95"
+    class="scroll-to-top-css fixed bottom-4 inset-ie-4 z-50 w-12 h-12 bg-bg-elevated border border-border rounded-full shadow-lg flex items-center justify-center text-fg-muted hover:text-fg transition-colors active:scale-95"
     :aria-label="$t('common.scroll_to_top')"
-    @click="scrollToTop"
+    @click="scrollToTop()"
```
OrbisK:
```suggestion
    @click="()=>scrollToTop()"
```

app/components/ScrollToTop.client.vue
```diff
@@ -56,11 +59,11 @@ onMounted(() => {
...
   >
     <button
-      v-if="isActive && isMounted && isVisible"
+      v-if="shouldShowButton && isMounted && isVisible"
       type="button"
-      class="fixed bottom-4 inset-ie-4 z-50 w-12 h-12 bg-bg-elevated border border-border rounded-full shadow-lg md:hidden flex items-center justify-center text-fg-muted hover:text-fg transition-colors active:scale-95"
+      class="fixed bottom-4 inset-ie-4 z-50 w-12 h-12 bg-bg-elevated border border-border rounded-full shadow-lg flex items-center justify-center text-fg-muted hover:text-fg transition-colors active:scale-95"
       :aria-label="$t('common.scroll_to_top')"
-      @click="scrollToTop"
+      @click="scrollToTop()"
```
OrbisK:
```suggestion
      @click="()=>scrollToTop()"
```

app/composables/useScrollToTop.ts
```diff
@@ -0,0 +1,96 @@
...
+    if (start <= 0) return
+
+    const startTime = performance.now()
+    const change = -start
+
+    const cleanup = [
+      useEventListener(window, 'wheel', cancel, { passive: true }),
+      useEventListener(window, 'touchstart', cancel, { passive: true }),
+      useEventListener(window, 'mousedown', cancel, { passive: true }),
+    ]
```
OrbisK: we should not add these inside of `scrollTop()`. This will lead to a memory leak, becuase the event listerners are not cleaned up after every call.
You should move them to the setup level.

app/composables/useScrollToTop.ts
```diff
@@ -0,0 +1,96 @@
...
+        rafId = requestAnimationFrame(animate)
+      } else {
+        cancel()
+      }
+    }
+
+    rafId = requestAnimationFrame(animate)
+  }
+
+  onBeforeUnmount(cancel)
```
OrbisK: We should prefer the `tryOnScopeDispose` method from Vueuse. This is because the component might be called in another scope, than the component (for example as a `sharedComposable`).

danielroe:
> However, this raises another issue: when switching to other languages, the floating buttons at the bottom extend beyond the screen, and the display changes accordingly

@RYGRIT maybe we can switch to icon-only for the bottom bar on mobile? and long-press on the button shows what it does?

RYGRIT:
> FYI: There is https://vueuse.org/core/useScroll

Thank you very much for your suggestion.

RYGRIT:
@danielroe Good call! I've updated the bottom bar to be icon-only on mobile (using sr-only to keep it a11y-friendly) and added title attributes.

Regarding the long-press: I'd prefer to skip it for now. The icons are quite self-explanatory, and users rarely long-press for hints on mobile. Plus, long-pressing a link triggers the native browser context menu (like 'Open in new tab'), which we probably shouldn't override. Let me know what you think!

PR #1457, opened 2026-02-13
chore: exclude prereleases from outdated packages

### LINKED ISSUE #1456: Ignore unstable version in outdated dependency logic
**Description**

In the list of dependencies, a gap in major versions is indicated with red colour and a warning symbol. This is a useful feature, but IMHO it should not check against unstable versions.

**Example**

Image

c12 latest stable version: v3.3.3
c12 latest version: 4.0.0-beta.2

**However**

What's problematic is that version 4.0.0-beta.2 uses the default dist-tag `latest`, so one might argue that the problem is on the developer side.

Image

Still, I wanted to bring it up and propose a fix. You can decide how to deal with it.

idleberg:
Fixes #1456. However, this might need some discussion because the problematic behaviour can be interpreted in two ways:

1. A developer intentionally chose to tag a version as `latest`, and we should honour this decision
2. A developer mistakenly tagged the version as `latest`, and we decided that `*-alpha` or `*-beta` should overrule the distTag

What are your thoughts on this?

### DIFF
--- a/app/composables/npm/useOutdatedDependencies.ts
+++ b/app/composables/npm/useOutdatedDependencies.ts
@@ -102,7 +102,8 @@ export function useOutdatedDependencies(
       let latestStable = data.distTags.latest
       if (!latestStable) continue
 
-      // If latest tag is a prerelease, find the latest stable version instead
+      // If latest tag is a prerelease, find the latest stable version instead. Take note that this
+      // overrides the latest distTag.
       if (prerelease(latestStable)) {
         const stableVersions = data.versions.filter(v => !prerelease(v))
         if (stableVersions.length > 0) {


danielroe:
I think we should honour the developer's decision to tag something as latest (but personally I would not tag an alpha or beta reease as latest)

idleberg:
> personally I would not tag an alpha or beta release as latest

Agreed, but I've seen experienced developers struggling (or being aware of) distTags. One final thought: as a middleground, packages with such discrepancies could be highlighted in yellow rather than red, to raise awareness.

Otherwise feel free to close this!

danielroe:
I understand developers might not understand, but I think that could be raised with them directly. after all, tagging a beta release as `latest` will have other consequences than on npmx - people will install it accidentally as well.

danielroe:
I think a good approach would be to change the messaging on the bit of the UI that displayed the updated dependency to make it clear that this isn't a bug, e.g. something like the new latest version is 1.0.0-beta.0

btw I checked with pooya (author of c12) and it's intentional - so don't bug him about it!

PR #1459, opened 2026-02-13
fix: use jsdelivr when npm has a truncated readme

### LINKED ISSUE #1458: fix: README content truncated for packages with large READMEs
## Problem

The npm registry truncates the `readme` field in packument JSON responses at exactly **65,536 characters** (2^16). This means packages with large READMEs (e.g., `vue-data-ui` at ~80KB) are rendered incomplete on npmx.dev — the content is silently cut off mid-document.

### Reproduction

```bash
curl -s "https://registry.npmjs.org/vue-data-ui" | jq '.readme | length'
# Returns: 65536
```

The actual README is 80,712 bytes, so ~19% of the content is lost.

## Root Cause

In `server/api/registry/readme/[...pkg].get.ts`, the handler reads `packageData.readme` from the npm packument as the primary source. The jsDelivr CDN fallback (`fetchReadmeFromJsdelivr`) only triggers when the packument readme is **missing** or the filename is non-standard — not when it's **truncated**.

## Fix

Prefer fetching the README directly from jsDelivr CDN (which serves the actual file from the npm tarball, untruncated), and fall back to the packument `readme` field only when jsDelivr doesn't have the file.

lukeocodes:
## Summary

Large READMEs are being cut off because the npm registry silently truncates the `readme` field in packument responses at exactly 65,536 characters. For example, `vue-data-ui` has an 80KB README but only 64K makes it through.

This switches the README handler to fetch the actual file from jsDelivr CDN (which serves from the npm tarball without truncation) as the primary source, keeping the packument `readme` field as a fallback for packages where jsDelivr doesn't have the file.

Closes #1458

## Test plan

- [x] Verified locally with `vue-data-ui` — full README now renders
- [x] Spot check other packages with large READMEs
- [x] Confirm packages with non-standard readme filenames still render correctly (fallback to packument)

### DIFF
--- a/server/utils/readme-loaders.ts
+++ b/server/utils/readme-loaders.ts
@@ -79,10 +79,11 @@ export const resolvePackageReadmeSource = defineCachedFunction(
 
     const hasValidNpmReadme = readmeContent && readmeContent !== NPM_MISSING_README_SENTINEL
 
-    const isLikelyTruncated =
-      hasValidNpmReadme && readmeContent!.length >= NPM_README_TRUNCATION_THRESHOLD
-
-    if (!hasValidNpmReadme || !isStandardReadme(readmeFilename) || isLikelyTruncated) {
+    if (
+      !hasValidNpmReadme ||
+      !isStandardReadme(readmeFilename) ||
+      readmeContent!.length >= NPM_README_TRUNCATION_THRESHOLD
+    ) {
       const resolvedVersion = version ?? packageData['dist-tags']?.latest
       const jsdelivrReadme = await fetchReadmeFromJsdelivr(
         packageName,


server/utils/readme-loaders.ts
```diff
@@ -58,35 +58,34 @@ export const resolvePackageReadmeSource = defineCachedFunction(
...
-      if (versionData) {
-        readmeContent = versionData.readme
-        readmeFilename = versionData.readmeFilename
+    // Prefer jsDelivr (actual file from npm tarball) because the npm registry
+    // truncates the packument readme field at 65,536 characters.
+    let readmeContent = await fetchReadmeFromJsdelivr(
+      packageName,
+      standardReadmeFilenames,
+      resolvedVersion,
+    )
```
danielroe: how about we only do this if the readme approaches 64K in size? for smaller (= most) readmes this will be a needless extra request

lukeocodes: that's not a bad idea too. do we want to limit our reliance on jsDelivr if we can use the registry packument?

danielroe (approved):
👌

server/utils/readme-loaders.ts
```diff
@@ -75,11 +79,15 @@ export const resolvePackageReadmeSource = defineCachedFunction(
 
     const hasValidNpmReadme = readmeContent && readmeContent !== NPM_MISSING_README_SENTINEL
 
-    if (!hasValidNpmReadme || !isStandardReadme(readmeFilename)) {
+    const isLikelyTruncated =
+      hasValidNpmReadme && readmeContent!.length >= NPM_README_TRUNCATION_THRESHOLD
```
danielroe: this comparison is likely costly so we should probably refactor it so we don't calculate readme length for every readme

PR #1462, opened 2026-02-13
fix: extract label from image-only playground links

rzzf:
https://npmx.dev/package/element-plus

Fixes an issue where playground links containing only an image (e.g. `[alt](url)`) resulted in a null label. Now uses the image's alt text as the label in such cases.

### DIFF
--- a/server/utils/readme.ts
+++ b/server/utils/readme.ts
@@ -422,7 +422,8 @@ ${html}
       plainText = tokens[0].text
     }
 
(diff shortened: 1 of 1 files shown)


danielroe (approved):

danielroe:
it could further fall back to the provider name, e.g. 'codesandbox' in the case above

rzzf:
done

PR #1470, opened 2026-02-13
fix: localization for "go back home" button on error page

Adebesin-Cell:
### Changes

* ~~Added `cursor-pointer` to the **"Go back home"** button in `app/error.vue`.~~
* Replaced hardcoded text with `$t('common.go_back_home')` (if available).
* ~~Added `cursor-pointer` to `NuxtLink` in `Base.vue` to ensure consistent button-like link behavior.~~

### Verification

**Before (current behavior – not working as expected):**

* Navigate to `npmx.dev/non-existent-page`
* Hover over the **"Go back home"** button
* Cursor does **not** change to `pointer`

**After (with this fix):**

* Navigate to `npmx.dev/non-existent-page`
* Hover over the **"Go back home"** button
* Cursor changes to `pointer` and navigation works correctly

### DIFF
--- a/app/components/Link/Base.vue
+++ b/app/components/Link/Base.vue
@@ -83,7 +83,7 @@ const isButtonMedium = computed(() => props.size === 'medium' && !isLink.value)
         !isLinkAnchor && isLink && !noUnderline,
       'justify-start font-mono text-fg hover:(decoration-accent text-accent) focus-visible:(decoration-accent text-accent) transition-colors duration-200':
         isLink,
-      'justify-center font-mono border border-border rounded-md transition-all duration-200 cursor-pointer':
+      'justify-center font-mono border border-border rounded-md transition-all duration-200':
         isButton,
       'text-sm px-4 py-2': isButtonMedium,
       'text-xs px-2 py-0.5': isButtonSmall,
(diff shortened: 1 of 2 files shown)


ghostdevv (approved):
I removed the cursor changes since in #1074 I believe we've decided to only use `cursor-pointer` on links which is the browser default behavior 🙏 If there is a link that is missing `cursor-pointer` you could make a PR to fix the underlying cause?

Adebesin-Cell:
> I removed the cursor changes since in #1074 I believe we've decided to only use `cursor-pointer` on links which is the browser default behavior 🙏 If there is a link that is missing `cursor-pointer` you could make a PR to fix the underlying cause?

Do you mean to add `cursor-pointer` to links that might have been missed?

ghostdevv:
> > I removed the cursor changes since in #1074 I believe we've decided to only use `cursor-pointer` on links which is the browser default behavior 🙏 If there is a link that is missing `cursor-pointer` you could make a PR to fix the underlying cause?
> 
> Do you mean to add `cursor-pointer` to links that might have been missed?

Ah, I mean like we can add it if we find it's missing I think? I would've merged it tbf if I was more sure, but I think that the i18n was important to get in soon. Maybe you can make another PR to add the cursor-pointer to linkbase if we want to be sure? I'll then get a review from Alex maybe

PR #1476, opened 2026-02-13
refactor: add radio and checkbox component

essenmitsosse:
This adds a button and checkbox component and replaces all things I could find that were `button` before but should be a radio or checkbox. It doesn't touch the `SettingsToggle` component, which is also a checkbox.

The radio button also replaces the previous `TagRadioButton`

For the checkboxes I added a variant, that looks like a toggable button. For the radio buttons I didn't. I am going to leave this PR as a draft, because I am sure there is some discussion to be had.

The goal of this PR should be to move all things that should be radio/checkboxes to an abstract component. I don't think these components need to be functionally or visually perfect just now. Having only one place to maintain them should be a benefit in itself. So I think we should try not to get lost in detailed "how round should the focus be" discussions. I also think surfacing that something is a radio/checkbox in the UI by actually displaying it can be good intermediate step to make everybody think more consciously about UX dicisions. What I mean by that: 'Maybe if too many radio buttons look to crowded, the problem is not showing them, but having so many of them' — maybe.

Screenshot 2026-02-13 at 20 26 14

Screenshot 2026-02-13 at 20 30 48

Screenshot 2026-02-13 at 20 30 23 Screenshot 2026-02-13 at 20 37 06


essenmitsosse:
@knowler @alexdln could you take a look if you like the general direction of this or have any remarks, especially regarding regressions? This would move the majority of inputs and buttons into a few abstract components. If you think this generally looks fine I can fix the tests and rebase.

alexdln:
The checkbox in the columns is currently 14x14 and looks a bit small. Maybe let's make it 1.2em

Also, let's make the color `accent-accent`, otherwise the blue looks too default

image

Regarding the facets logic - I like it and wanted to talk about this too. But it seems like the logic is broken somewhere now. Also, I'm a bit worry that this is a very complex logic, maybe we should postpone it until post-launch (or at least to separate PR)?

Based on the tag style, making them brightly colored looks better, but it's a bit too bright when there are a lot of them, and when active, it requires a different hover effect (_at the end of the video, when you hover over a tag, it's unclear whether it's activated or not_)

https://github.com/user-attachments/assets/c6970d94-9f61-495b-9e4f-3cb67cc455f4

Regarding the versions, I think this version of toggle is very specific for us. They've already been updated into a unified look in production, but I'm in favor of a redesigning it anyway, but to something like for filters toggle ([PR](https://github.com/npmx-dev/npmx.dev/pull/1659), [discord](https://discord.com/channels/1464542801676206113/1476511478671736883/1476529617895555082)). Maybe the design of the active tag can also be taken from here

Preview/prod/mentioned toggle
image
image
image

I like the idea of ​​radio filters, but not for the search category. It's like the search category is more like a select or a toggle-group. It also seems better to reduce the indents to gap-6 instead of gap-8. And let's use `accent-accent` as well
image

PR #1484, opened 2026-02-13
fix(ui): fix missing icons

wojtekmaj:
In #1454, Carbon icons were removed. This caused a conflict when later that day #1417 was merged. Additionally, in #1454 some non-existing icons were used, like x-filled, or check-filled.

### DIFF
--- a/app/components/Org/OperationsQueue.vue
+++ b/app/components/Org/OperationsQueue.vue
@@ -164,9 +164,9 @@ function getStatusIcon(status: string): string {
     case 'running':
       return 'i-svg-spinners:ring-resize'
     case 'completed':
-      return 'i-lucide:check-filled'
+      return 'i-lucide:check'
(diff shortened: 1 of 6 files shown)


wojtekmaj:
@essenmitsosse FYI :)

alexdln:
Goddess, we're still updating too fast 😅

@wojtekmaj you are great, thanks for checking and fixing so quickly ❤️

essenmitsosse:
@wojtekmaj oh shit thank you for fixing this. I kind of worried something like this would happen but hopped it wouldn't. Some of the fixes after the initial PR were too rushed to get it out before the break. Thanks for fixing my mess!

wojtekmaj:
@essenmitsosse Absolutely no worries. That's why we're a team :) I'll mess something up next week 😁

PR #1486, opened 2026-02-13
fix: improve badge text measurement accuracy

wojtekmaj:
This PR improves badge appearance by using a more accurate method to measure text width. Instead of relying on a predefined character width, it utilizes the Canvas API to measure the actual width of the text, resulting in better alignment and spacing for badges with varying character widths.

Speaking in code, this is how the old and new methods compare: 
```
{ label: 'iiiiiiiiii', leftWidthOld: 86, leftWidth: 41 }
{ label: 'wwwwwwwwww', leftWidthOld: 86, leftWidth: 97 }
```

Actual images before:
iiiold
wwwold

After:
iiinew
wwwnew

### DIFF
--- a/server/api/registry/badge/[type]/[...pkg].get.ts
+++ b/server/api/registry/badge/[type]/[...pkg].get.ts
@@ -1,4 +1,5 @@
 import * as v from 'valibot'
+import { createCanvas, type SKRSContext2D } from '@napi-rs/canvas'
 import { hash } from 'ohash'
 import { createError, getRouterParam, getQuery, setHeader } from 'h3'
 import { PackageRouteParamsSchema } from '#shared/schemas/package'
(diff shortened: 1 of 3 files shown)


server/api/registry/badge/[type]/[...pkg].get.ts
```diff
@@ -34,15 +35,47 @@ const COLORS = {
...
+  }
+
+  try {
+    cachedCanvasContext = createCanvas(1, 1).getContext('2d')
+  } catch {
+    cachedCanvasContext = null
+  }
+
+  return cachedCanvasContext
 }
```
Comment (ghostdevv) [inline server/api/registry/badge/[type]/[...pkg].get.ts:59]: is reusing canvas context safe? like we're not going to run into issues with methods mutating the context and causing weird behaviour?

Comment (wojtekmaj) [reply server/api/registry/badge/[type]/[...pkg].get.ts:59]: Yes, were not drawing anything, we're just using a method you call "what IF you drew on that canvas"?

ghostdevv:
actually will wait for current renovate PRs to go in, then merge

PR #1487, opened 2026-02-14
feat: add shields.io-style badges

wojtekmaj:
This PR implements a new, optional parameter in badges API: `style`. By passing `style=shieldsio` we change the appearance of the generated shields from the default one to custom one.

Default:
default

Default with custom color:
defaultcolor

shields.io-style:
shieldsio

shields.io-style with custom color:
shieldsiocolor


server/api/registry/badge/[type]/[...pkg].get.ts
```diff
@@ -34,15 +35,141 @@ const COLORS = {
...
+    return cachedCanvasContext
+  }
+
+  try {
+    cachedCanvasContext = createCanvas(1, 1).getContext('2d')
+  } catch {
+    cachedCanvasContext = null
+  }
+
+  return cachedCanvasContext
```
Comment (ghostdevv) [inline server/api/registry/badge/[type]/[...pkg].get.ts:61]: is reusing canvas context safe? like we're not going to run into issues with methods mutating the context and causing weird behaviour?

Comment (ghostdevv) [reply server/api/registry/badge/[type]/[...pkg].get.ts:61]: I meant to add this comment in the other PR https://github.com/npmx-dev/npmx.dev/pull/1486#discussion_r2806750845

danielroe (approved):

PR #1493, opened 2026-02-14
fix(i18n): german translation for "hit" in a11y section + missing keys

TheDevMinerTV:
stößt was missing an ö.

Also added missing German translation keys for the trends section (untranslated version below):

image


TheDevMinerTV:
@trueberryless Can you check if the singular / plural translations make sense?

~~Do we need to use the `{count}` variable in there for plural stuff?~~

EDIT: Seems fine

image

PR #1497, opened 2026-02-14
fix: dialog title switch with select

zzxming:
Dialog title doesn't match facet

image

Fixed

image


ghostdevv (approved):
Hey! Thank you for your PR - I've made a change to use `package.trends.items.downloads` instead of `package.downloads.modal_title` since that seems to be more correct as `package.downloads.modal_title` was translated as "Weekly Downloads" which isn't the case for the modal anymore as you can change the period that the downloads/likes/contributors are shown over

PR #1503, opened 2026-02-14
fix: prevent duplicate items in column selection

### LINKED ISSUE #1474: Columsn selection on org search show items twice
If you scroll down the list of selectable columns, all items repeat twice. Example: https://npmx.dev/org/prisma

Image

rzzf:
fix #1474

When a configuration exists in localStorage, defu’s default merge strategy for arrays is concatenation.

### DIFF
--- a/test/nuxt/composables/use-preferences-provider.spec.ts
+++ b/test/nuxt/composables/use-preferences-provider.spec.ts
@@ -49,7 +49,7 @@ describe('usePreferencesProvider', () => {
     })
   })
 
-  it('handles array merging by replacement (preventing duplicates)', () => {
+  it('handles array merging by replacement', () => {
     mountWithSetup(() => {
       const defaults = { cols: ['name', 'version', 'date'] }
       const stored = { cols: ['name', 'version'] }


rzzf:
done.

test/nuxt/composables/use-preferences-provider.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+      const stored = { theme: 'dark' }
+      setLocalStorage(stored)
+      const { data } = usePreferencesProvider(defaults)
+      onMounted(() => {
+        expect(data.value).toEqual(stored)
+      })
+    })
+  })
+
+  it('handles array merging by replacement (preventing duplicates)', () => {
```
ghostdevv: I think here, it's not so much preventing duplicates but overriding the defaults? What's the current behaviour? Would it result in `['name', 'version', 'date', 'name', 'version']`? What do we desire here, the overriden defaults or a deduplication of that merged array?

ghostdevv (changes requested):

app/composables/usePreferencesProvider.ts
```diff
@@ -13,6 +13,13 @@ interface StorageProvider<T> {
   remove: () => void
 }
 
+const defu = createDefu((object, key, value) => {
+  if (Array.isArray(object[key]) && Array.isArray(value)) {
+    object[key] = value
+    return true
+  }
+})
```
ghostdevv, line 21:
```suggestion
const defu = createDefu((object, key, value) => {
  if (Array.isArray(object[key]) && Array.isArray(value)) {
    object[key] = Array.from(new Set([...object[key], ...value]))
    return true
  }
})
```

If we want them merged and deduplicated (see other comment) then I think this would work? at least for literals

test/nuxt/composables/use-preferences-provider.spec.ts
rzzf: Ah, I think the expected behavior is to override rather than deduplicate. (I rely too much on translation.)

app/composables/usePreferencesProvider.ts
rzzf: This doesn’t resolve #1474. From what I’ve reviewed, the original design behavior was to override.

https://github.com/npmx-dev/npmx.dev/pull/1286/changes#diff-61a5eaf8df87acb523d928c18640ae5ac303f06d631ba312c3ae357e3d13f6a5R73

ghostdevv (approved):

rzzf:
thanks merged

PR #1508, opened 2026-02-14
fix: trends modal should show the facet in the title

graphieros:
- use the expected `modal-title` prop  instead of `title` (typical false positive)
- use proper translation for versions chart modal title


ghostdevv:
oh! I just merged #1497 :sweat_smile: - could you rebase to make any of the changes that it I've missed?

ghostdevv (approved):
Merge when ready! @graphieros

PR #1511, opened 2026-02-14
feat(i18n): add Arabic font support

abaza738:
Added a font for Arabic language because the current font does not support Arabic letters (and I guess many other languages?)😅.  
Tested this locally and found it to preserve the original font Geist when English text is written in the middle of Arabic text, I suppose this is `unicode-range` working properly.

## Changes Preview
Before:
image

After:
image


abaza738:
I updated the PR description to include the screenshots, sorry I missed it at first!

FYI, Daniel had mentioned on Discord that there's extra fonts being downloaded and wanted to investigate further before merging.

ghostdevv:
> FYI, Daniel had mentioned on Discord that there's extra fonts being downloaded and wanted to investigate further before merging.

Do you have a link :eyes:

abaza738:
> [I saw an extra woff font being downloaded at one point and needed to figure out why](https://discord.com/channels/1464542801676206113/1465734070817259601/1475627302510399488)

ghostdevv:
ah, it does seem to download two files with firefox only being able to show that preview thingy for one :thinking: 

image

ghostdevv:
also tagging Alex for a review since we're changing fonts - not sure if we should add this to the brand page? @alexdln

alexdln:
@ghostdevv I think nope. We use Geist in our logo, banners and og-images. While this one is only for internal content and therefore isn't important for the press

ghostdevv:
After syncing with main and checking the [latest preview](https://npmx-e1oph8szm-npmx.vercel.app) (the first comment with the link seems to be broken on this pr lol), the duplicate font download _seems_ to be gone - so will merge this for now and we can make any other changes in a follow up :saluting_face:

PR #1513, opened 2026-02-14
feat(i18n): update Tamil (ta-IN) translations

Meganathanrbm:
Add complete Tamil (ta-IN) locale translation (798 keys)
- Register ta-IN locale in i18n config with proper metadata
- Add lunaria tracking file for translation progress monitoring


danielroe:
were these translations done with Claude? do you speak Tamil?

it's fine if it was done with Claude, but you speak Tamil and were able to check them.

but if you don't speak Tamil and just used Claude to generate the translations, I'd prefer to wait for a native speaker.

Meganathanrbm:
Tamil is my mother tounge and I'm Native tamil speaker. So, I added it here.

PR #1529, opened 2026-02-16
chore: remove vite-plus (for now)

wojtekmaj:
It seems like we're currently using vite-plus solely for running tests. It caused issues in the past that forced us to migrate away from it in most places (vite fmt, vite lint running postinstall and causing severe delays). It currently crashes on Windows arm64 machines. It is also explicitly said:

> Vite+ is not yet stable and should not be used in production.

With that in mind, vite-plus does not provide a lot of value to the project at the moment and therefore I propose removing it for now.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -26,7 +26,7 @@ importers:
         version: 0.0.19
       '@atproto/oauth-client-node':
         specifier: ^0.3.15
-        version: 0.3.16
+        version: 0.3.17
       '@deno/doc':
         specifier: jsr:^0.189.1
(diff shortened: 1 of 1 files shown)


43081j:
caught up the branch for you.

i agree we should use vitest until the void zero branch is stable enough.

danielroe:
if we can itemise issues we can pester @TheAlexLichter to fix them 👀

TheAlexLichter:
In all cases, please send issues! This is super valuable for the upcoming alpha and to improve Vite+ further.

danielroe:
I think https://github.com/voidzero-dev/vite-plus-discussions/issues/9 is the most pesky one right now.

alexdln:
@wojtekmaj, it looks like this was fixed in #1668, and now everything works for me on Windows. Please reopen if you still have issues

PR #1532, opened 2026-02-16
chore: add tests for mobile menu

### LINKED ISSUE #1505: Add a test that checks if the mobile menu is closed by default
We need to add a check that the menu is closed by default and opens on request. Also, we need to check logic isolation and reusability (related - #1481)

colinscz:
Closes #1505

- [ ] Extract utilities in separate file
- [ ] Revisit tests

### DIFF
--- a/test/e2e/app-header.spec.ts
+++ b/test/e2e/app-header.spec.ts
@@ -1,0 +1,457 @@
+import type { Page } from '@playwright/test'
+import { expect, test } from './test-utils'
+
+/**
+ * Mobile viewport configuration for testing
+ * Matches the sm: breakpoint (640px) - tests below this width
+ */
+const MOBILE_VIEWPORT = { width: 375, height: 667 }
+
+/**
+ * Helper: Check if current viewport is mobile
+ */
+async function isMobileViewport(page: Page): Promise<boolean> {
+  const viewport = page.viewportSize()
+  return viewport ? viewport.width < 640 : false
+}
+
+/**
+ * Helper: Verify menu is closed
+ */
+async function expectMenuClosed(page: Page): Promise<void> {
+  const menuButton = page
+    .locator(
+      'button[aria-label*="open_menu"], button[aria-label*="Buka"], button[aria-label*="Apri"], button[aria-label*="Otevřít"], button[aria-label*="Ouvrir"]',
+    )
+    .first()
+  await expect(menuButton).toHaveAttribute('aria-expanded', 'false')
+
+  // Menu panel should not be visible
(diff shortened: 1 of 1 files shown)


test/e2e/app-header.spec.ts
```diff
@@ -0,0 +1,457 @@
...
+  test.use({ viewport: MOBILE_VIEWPORT })
+
+  test.describe('Default State', () => {
+    test('menu is closed by default on page load', async ({ page, goto }) => {
+      await goto('/', { waitUntil: 'hydration' })
+
+      const isOnMobile = await isMobileViewport(page)
+      if (!isOnMobile) {
+        test.skip()
+      }
```
thomasbritton: As you're setting `test.use({ viewport: MOBILE_VIEWPORT })` you do not need to be specifying these skips in each test case.

I'd recommend removing all cases of this check.

test/e2e/app-header.spec.ts
```diff
@@ -0,0 +1,457 @@
...
+
+      // Menu button should be visible
+      const menuButton = page
+        .locator(
+          'button[aria-label*="open_menu"], button[aria-label*="Buka"], button[aria-label*="Apri"], button[aria-label*="Otevřít"], button[aria-label*="Ouvrir"]',
+        )
+        .first()
+      await expect(menuButton).toBeVisible()
```
thomasbritton, line 82: This is repeated a lot, it'll be worth moving this and other DOM selectors into a helper function which you can reference in the tests instead of repeating it each time.

test/e2e/app-header.spec.ts
```diff
@@ -0,0 +1,457 @@
...
+      await expectMenuOpen(page)
+    })
+
+    test('isMobileViewport helper works correctly', async ({ page, goto }) => {
+      await goto('/', { waitUntil: 'hydration' })
+
+      const isMobile = await isMobileViewport(page)
+      expect(isMobile).toBe(true)
+    })
+  })
```
thomasbritton: These “helper functions work independently” tests don’t give you user value, all it's doing is validating test code. It's better to test behaviours only.

I'd recommend removing these tests.

test/e2e/app-header.spec.ts
colinscz: hi @thomasbritton thanks for catching these things, the PR is still in draft state so please wait with further reviews as some things definitely will change until I convert the PR to ready. I addressed your already raised comments. 👍🏽

thomasbritton: Hey @colinscz,

It's all good 👍 

(I didn't realise it was in draft state until after submitting the comments)

PR #1535, opened 2026-02-17
fix: make social cards clickable and remove z-index trickery

### LINKED ISSUE #1517: CTA cards not fully clickable
Hovering anywhere in a CTA card triggers a color change, so I would assume that I can click it to go to the link. However, if my cursor happens to be above another text block, it covers the link so the click doesn't work.

I would just fix this myself, but I noticed in #799, this seemed to be an intentional decision by @knowler:

> using `z-index` to keep heading and paragraph text on top so it can be directly selected

It seems unusual to want to have selectable text within a clickable card, and IMO this is very confusing to the user since they can click in some parts of the card but not others. Is there a reason to have this behavior?

https://github.com/user-attachments/assets/b9c109e0-e716-4f20-a450-b491915557b1

pauliecodes:
## Description
This PR addresses the usability issue where **clicking on text** within social cards did not trigger navigation.

Previously, the text elements were positioned above the link overlay (using `z-index`) to allow for text selection, but this created "dead zones" where clicks were not registered.

## Changes
- **Removed CSS Tricks:** Deleted the `absolute inset-0` link overlay and `z-1` classes, simplifying the DOM structure as suggested by @knowler.
- **Added Click Handler:** Implemented a `handleCardClick` function that programmatically clicks the CTA link when the card is clicked.
- **Preserved Native Behavior:** The handler checks for:
  - Nested links/icons (using `.closest('a')`)
  - Text selection (using `window.getSelection()`)
  - Modifier keys (Ctrl/Cmd/Shift/Alt)

## Demo

| Before | After |
| :---: | :---: |
| npmx_card_before | npmx_card_after |

Fixes #1517

### DIFF
--- a/app/components/CallToAction.vue
+++ b/app/components/CallToAction.vue
@@ -27,7 +27,7 @@ const socialLinks = computed(() => [
 ])
 
 function handleCardClick(event: MouseEvent) {
-  if ((event.target as HTMLElement).closest('a')) return
+  if ((event.target as HTMLElement).closest(':any-link')) return
   if (event.ctrlKey || event.metaKey || event.shiftKey || event.altKey) return
 
   const selection = window.getSelection()


app/components/CallToAction.vue
```diff
@@ -25,6 +25,18 @@ const socialLinks = computed(() => [
     ctaKey: $t('about.get_involved.follow.cta'),
   },
 ])
+
+function handleCardClick(event: MouseEvent) {
+  if ((event.target as HTMLElement).closest('a')) return
```
knowler: Oh, nice! ~~This probably handles if there was something inside of the link that got clicked? Like the SVG?~~ I see in the PR description it’s for icon clicks—awesome!

- - -

This is definitely me being a bit of a stickler, but I would prefer `:any-link` over `a` as `a:not([href])` is not a link… but at the same time it’s not like we’re applying this broadly and we know the links will have `href` so it probably doesn’t matter (`:any-link` also matches `area[href]`, but that’s so uncommon, I never feel the need to use `a:any-link`). I’ll leave it up to your own discretion—what you have works!

pauliecodes: Thanks for the suggestion! I just added the change :)

PR #1536, opened 2026-02-17
fix(a11y): improve facet selector accessibility

Dayifour:
## Summary
This PR improves the accessibility of the compare facet selector by moving away from complex radio semantics towards standard action buttons, ensuring better focus management and clarity for screen readers.

## Changes
* **Simplified Semantics:** Reverted "All / None" controls to plain buttons (removing `role="radio"` and `role="radiogroup"`).
* **Improved Focus Management:** Removed the roving tabindex; controls now follow the natural browser tab order.
* **Enhanced Discoverability:** Replaced the HTML `disabled` attribute with `aria-disabled`. This keeps the buttons in the focus order while indicating they are currently inactive.
* **Clearer Labelling:** Implemented specific `aria-label` attributes (e.g., "Select all health facets") to provide unambiguous context for screen reader users.
* **Individual Facets:** Maintained individual facets as checkboxes with proper `role="checkbox"` and `aria-checked` attributes.
* **Testing:** Updated `FacetSelector` tests to validate the new button-based logic and ARIA attributes.

## Testing
```bash
pnpm test:nuxt -- --runTestsByPath test/nuxt/components/compare/FacetSelector.spec.ts
```

### Notes
* **Visuals:** No visual changes or CSS regressions.
* **Compliance:** This refactor prioritizes standard accessibility patterns as discussed with the maintainers to ensure a robust and predictable experience.

### DIFF
--- a/i18n/schema.json
+++ b/i18n/schema.json
@@ -3391,6 +3391,18 @@
             "none": {
               "type": "string"
             },
+            "select_all_category_facets": {
+              "type": "string"
+            },
+            "deselect_all_category_facets": {
+              "type": "string"
+            },
+            "selected_all_category_facets": {
+              "type": "string"
+            },
+            "deselected_all_category_facets": {
+              "type": "string"
+            },
             "coming_soon": {
               "type": "string"
             },


Dayifour:
Ready for review

app/components/Compare/FacetSelector.vue
```diff
@@ -28,16 +28,20 @@ function isCategoryNoneSelected(category: string): boolean {
   <div class="space-y-3" role="group" :aria-label="$t('compare.facets.group_label')">
     <div v-for="category in categoryOrder" :key="category">
       <!-- Category header with all/none buttons -->
-      <div class="flex items-center gap-2 mb-2">
+      <div
+        class="flex items-center gap-2 mb-2"
+        role="radiogroup"
+        :aria-label="getCategoryLabel(category)"
+      >
         <span class="text-3xs text-fg-subtle uppercase tracking-wider">
```
knowler: Nice catch with the labels for the `group`/`radiogroup` roles. 

Let’s set an `id` on the `<span>` with the visual label and use `aria-labelledby` to reference it.

Dayifour:
Hi @knowler ! I've completed the refactor. Here is a summary of the latest changes:

A11y & Keyboard: Implemented a proper roving tabindex and arrow-key navigation for the radiogroup.

Refined Semantics: Replaced aria-pressed with aria-checked and ensured aria-disabled is only used where appropriate.

Clean Labels: Linked groups via aria-labelledby and removed redundant aria-label attributes on checkboxes.

Tests & CI: Updated all component tests to match the new ARIA roles. All 17 checks are now green, including i18n schema and TypeScript.

I've also addressed the latest feedback regarding decorative separators and event guards

knowler:
@Dayifour I’ll review this in-depth again once myself and the team come back from [“vacation”](https://npmx.dev/recharging) on Saturday.

Upon seeing the changes though, I’m having doubts about whether radio controls the correct tool for the job here as when interacting with the “all” or “none” control, there’s no way to clear that input other than using the individual facet controls and that feels a bit strange to me, especially when using a screen reader. I do agree the existing production implementation does need change as setting `disabled` to effectively do the same thing in a worse way than a radio control is also not great.

I might open an issue where a proper discussion can take place with regard to the design of these controls.

Dayifour:
Thanks for the feedback, @knowler! I completely understand the concern. By implementing the strict WAI-ARIA radio pattern, it indeed highlights that the 'All/None' actions behave more like command buttons than persistent states, which can be tricky for screen reader users to mentally map.
I'm totally open to exploring a different pattern if the team decides to move away from the radio approach.

Enjoy your vacation, and I'll look forward to the team's feedback on Saturday! No rush at all

serhalp:
hey @Dayifour @knowler what's the status of this PR? is it blocked on anything?

Dayifour:
Hi @serhalp! The PR is technically ready, but i were having a discussion with @knowler about whether the radio pattern is the best UX for 'All/None' controls, or if we should switch to standard buttons.

That said, I've noticed the branch now has conflicts due to recent changes in the lunaria files. I'm going to rebase/resolve those right now to make the PR mergeable again.

@knowler, let me know if the team has reached a decision on the radio vs button approach!

knowler:
@Dayifour Sorry for the delay. Without totally redesigning the feature, this is how I think we can improve the semantics and usability of the current design:

- [x] For the “all” and “none” controls, just use semantically plain buttons (e.g. our `<ButtonBase>`) with no added `role` or ARIA attributes.
  - As I pointed out in a prior comment, radio buttons are weird because there’s no in-between or unchecking everything. An indeterminate checkbox might be a better control altogether, but for the sake of keeping the design the same, we can just use stateless action buttons.
- [x] Get rid of the radiogroup semantics and behaviour for the “all” or “none” controls
  - We don’t need to group the buttons (semantically).
  - This means we don’t need a roving tabindex either. Standard navigating by tab is expected.
- [x] Keep those buttons enabled, even when they won’t do anything.
  - The visual feedback of the items being selected is enough.
  - We can use a live region to provide feedback for screen reader users (e.g. “Selected all health facets”; “Deselected all health facets”).
  - **If** we really do want to keep the visual of the buttons not having a relevant effect, then we should use `aria-disabled`. This allows the button to remain in the focus order for screen reader users and prevents loss of focus when pressed. 
- [x] Make sure the accessible names (i.e. `aria-label` attributes) are indicative of their action without any ambiguity considering the entire context.
  - Since there are multiple controls on the same page that visually are labelled the same way, we’ll want to make sure their accessible names are specific about what they’re doing (e.g. “Select all health facets”). We might have already made this change.
  
The individual facets are fine to remain as checkboxes.
  
Some of these might have already been taken care of. I’ll review and update the list to reflect that.

Dayifour:
Hi @knowler! I’ve refactored the component based on your latest feedback:

Simpler semantics: Reverted 'All/None' to plain <ButtonBase> (removed role="radio" and radiogroup).

Standard Focus: Removed roving tabindex; controls now follow the natural tab order.

Better a11y state: Replaced disabled with aria-disabled so the buttons remain in the focus order even when inactive.

Clear Labels: Ensured unique aria-label names for every action.

Regarding the live region, I haven't implemented it yet to keep this refactor focused, but I'm happy to add it if you feel the state change on the checkboxes isn't descriptive enough for screen readers.

app/components/Compare/FacetSelector.vue
```diff
@@ -28,48 +28,55 @@ function isCategoryNoneSelected(category: string): boolean {
   <div class="space-y-3" role="group" :aria-label="$t('compare.facets.group_label')">
     <div v-for="category in categoryOrder" :key="category">
       <!-- Category header with all/none buttons -->
-      <div class="flex items-center gap-2 mb-2">
-        <span class="text-3xs text-fg-subtle uppercase tracking-wider">
+      <div
+        class="flex items-center gap-2 mb-2"
+        :aria-labelledby="`facet-category-label-${category}`"
```
knowler: We can remove the `aria-labelledby` as well since the element no longer has a role.
```suggestion
```

app/components/Compare/FacetSelector.vue
```diff
@@ -28,48 +28,55 @@ function isCategoryNoneSelected(category: string): boolean {
...
-        <span class="text-2xs text-fg-muted/40">/</span>
+        <span class="text-2xs text-fg-muted/40" aria-hidden="true">/</span>
         <ButtonBase
-          :aria-label="
-            $t('compare.facets.deselect_category', { category: getCategoryLabel(category) })
-          "
-          :aria-pressed="isCategoryNoneSelected(category)"
-          :disabled="isCategoryNoneSelected(category)"
-          @click="deselectCategory(category)"
+          :aria-disabled="isCategoryNoneSelected(category)"
```
knowler:
```suggestion
```

Can we either get rid of `aria-disabled` here or apply the disabled styles (i.e. the utility classes) used by `ButtonBase` when `aria-disabled=true` is set? Those classes would be `aria-disabled:(opacity-40 cursor-not-allowed border-transparent)`.

app/components/Compare/FacetSelector.vue
```diff
@@ -28,48 +28,55 @@ function isCategoryNoneSelected(category: string): boolean {
...
         </span>
-        <!-- TODO: These should be radios, since they are mutually exclusive, and currently this behavior is faked with buttons -->
         <ButtonBase
-          :aria-label="
-            $t('compare.facets.select_category', { category: getCategoryLabel(category) })
-          "
-          :aria-pressed="isCategoryAllSelected(category)"
-          :disabled="isCategoryAllSelected(category)"
-          @click="selectCategory(category)"
+          :aria-disabled="isCategoryAllSelected(category)"
```
knowler: See https://github.com/npmx-dev/npmx.dev/pull/1536/changes#r2963673409 for possible options.

```suggestion
```

app/components/Compare/FacetSelector.vue
```diff
@@ -28,48 +28,55 @@ function isCategoryNoneSelected(category: string): boolean {
...
         <ButtonBase
           v-for="facet in facetsByCategory[category]"
           :key="facet.id"
           size="small"
           :title="facet.comingSoon ? $t('compare.facets.coming_soon') : facet.description"
-          :disabled="facet.comingSoon"
-          :aria-pressed="isFacetSelected(facet.id)"
-          :aria-label="facet.label"
-          class="gap-1 px-1.5 rounded transition-colors focus-visible:outline-accent/70"
+          :aria-disabled="facet.comingSoon"
```
knowler: These checkboxes are fine to use `disabled` when coming soon.

```suggestion
          :disabled="facet.comingSoon"
```

Dayifour:
Hi **@knowler**! I’ve just pushed the final adjustments based on your last review:

* **Header Clean-up**: Removed the unnecessary `aria-labelledby` from the category div.
* **Stateless Styles**: Applied the manual utility classes (`opacity-40 cursor-not-allowed`) to the All/None buttons to match the `aria-disabled` state without losing focus.
* **Checkbox Logic**: Reverted to the native `disabled` attribute for "Coming Soon" facets.
* **i18n Fix**: Removed the unused `group_label` key to satisfy the CI validation.

Thanks again for the detailed guidance on this refactor. I’ve learned a lot about balancing custom UI with standard A11y patterns!

All CI checks are now green. Ready for your final look!

knowler (approved):
Great work, @Dayifour! Thanks for sticking with it through my slow review. I made a few changes:

1. I added a live region so that there is feedback for the buttons when pressed. We can replace this with `useAnnouncer()` when Nuxt gets updated.
2. I’ve refactored the styles to use `aria-` variants where possible.
3. I made a few styling tweaks as I noticed a few things were out of sync with what’s live.
4. I removed the styling test as really we should be styling based on ARIA state (like I updated this PR to do) — so correct ARIA state is what we should be testing rather than specific class name which can change frequently and doesn’t actually guarantee styles are applied (visual regression testing is a better choice for that kind of testing).
5. There are some localization changes that will need to be made, but that can be done as a follow up.

I’m going to merge this once it’s all passing.

Dayifour:
Thanks a lot @knowler! I really appreciate the final polishes and the explanations about the live region and ARIA-based styling. It was a great learning experience. Looking forward to the next one!

PR #1540, opened 2026-02-18
feat: calculate directory sizes in code view

wojtekmaj:
Before/After:

Zrzut ekranu 2026-02-18 o 10 21 43

npm reference: 

Zrzut ekranu 2026-02-18 o 10 22 56

### DIFF
--- a/app/components/Code/DirectoryListing.vue
+++ b/app/components/Code/DirectoryListing.vue
@@ -120,10 +120,7 @@ const bytesFormatter = useBytesFormatter()
               </svg>
               <span class="w-full flex justify-self-stretch items-center gap-2">
                 <span class="flex-1">{{ node.name }}</span>
-                <span
-                  v-if="node.type === 'file' && node.size"
-                  class="text-end text-xs text-fg-subtle"
-                >
+                <span v-if="typeof node.size === 'number'" class="text-end text-xs text-fg-subtle">
                   {{ bytesFormatter.format(node.size) }}
                 </span>
               </span>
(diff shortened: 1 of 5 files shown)


alex-key:
Hi @wojtekmaj , very nice feature)
What is the reason for having a difference between the sum of file sizes (and it seems that npmjs calculates it correctly) and your implementation?

image
image

wojtekmaj:
> Hi @wojtekmaj , very nice feature) What is the reason for having a difference between the sum of file sizes (and it seems that npmjs calculates it correctly) and your implementation?

@alex-key Heh, looks like you've found a bug - but not in my PR, but in npmx in general. Taking /fast-xml-parser/lib/fxbuilder.min.js as an example, which is 8414 bytes long, npm reports its size as 8.41 kB, we as 8.2 kB. So we give size as KiB while we say it's kB. Will fix in a separate PR cause it has nothing to do with that one.

There you go: https://github.com/npmx-dev/npmx.dev/pull/1541

alex-key:
> > Hi @wojtekmaj , very nice feature) What is the reason for having a difference between the sum of file sizes (and it seems that npmjs calculates it correctly) and your implementation?
> 
> @alex-key Heh, looks like you've found a bug - but not in my PR, but in npmx in general. Taking /fast-xml-parser/lib/fxbuilder.min.js as an example, which is 8414 bytes long, npm reports its size as 8.41 kB, we as 8.2 kB. So we give size as KiB while we say it's kB. Will fix in a separate PR cause it has nothing to do with that one.
> 
> There you go: #1541

Cool, thank you for investigating this! Do you want me to create an issue or you gonna do by yourself?

danielroe (approved):
nice! ❤️

Issue #1563, opened 2026-02-22
Reduce anchor icon size in packages


alexdln:
We need to reduce the size of the anchor icon in package headers. I think a font-size of ~0.8 would be a good idea.

_Feel free to experiment and check on different heading levels_

https://npmx.dev/package/nuxt
<img width="376" height="104" alt="Image" src="https://github.com/user-attachments/assets/31c422e0-fbdc-4209-8321-22c553656ad3" />

PR #1566, opened 2026-02-22
fix(ui): reduce size of link icon

radosvet93:
Fixes: https://github.com/npmx-dev/npmx.dev/issues/1563

image

### DIFF
--- a/app/components/Readme.vue
+++ b/app/components/Readme.vue
@@ -154,11 +154,11 @@ function handleClick(event: MouseEvent) {
   /* I don't know what kind of sorcery this is, but it ensures this icon can't wrap to a new line on its own. */
   content: '__';
   @apply inline i-lucide:link rtl-flip ms-1 opacity-0;
+  font-size: 0.75em;
 }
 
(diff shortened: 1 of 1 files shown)


app/components/Readme.vue
```diff
@@ -158,6 +158,7 @@ function handleClick(event: MouseEvent) {
 
 .readme :deep(a[href^='#']:hover::after) {
   @apply opacity-100;
+  font-size: 0.75em;
```
danielroe: I think this should be set not just on hover, but in general on the `::after` pseudo element

also a question: do we need to add some padding to increase tap target?

radosvet93: Ahh, yes, it should be on the `::after`!

Hmm, the whole heading is the tap target, so I don't think it needs padding in that case, unless we want to add functionality to copy to clipboard only when the icon is clicked, then probably it's good to have a bigger tap target specifically for the icon

danielroe: good point!

PR #1567, opened 2026-02-22
chore: add minor effect for home search input

alexdln:
### 🧭 Context

_Experimenting_

### DIFF
--- a/app/pages/index.vue
+++ b/app/pages/index.vue
@@ -100,7 +100,7 @@ function onMouseMove(event: MouseEvent) {
                     :placeholder="$t('search.placeholder')"
                     no-correct
                     size="large"
-                    class="w-full ps-8 pe-24"
+                    class="w-full ps-8 pe-24 bg-bg-subtle/90! backdrop-blur-sm"
                     @focus="isSearchFocused = true"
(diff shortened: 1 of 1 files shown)


app/pages/index.vue
```diff
@@ -24,6 +25,16 @@ defineOgImageComponent('Default', {
...
+
+function onMouseMove(event: MouseEvent) {
+  const inputWrapper = inputWrapperRef.value
+  if (!inputWrapper) return
+  const rect = inputWrapper.getBoundingClientRect()
+  const x = event.clientX - rect.left
+  const y = event.clientY - rect.top
+  inputWrapper.style.setProperty('--mouse-x', `${x}px`)
+  inputWrapper.style.setProperty('--mouse-y', `${y}px`)
+}
```
OrbisK, line 37: FYI: https://vueuse.org/core/useMouseInElement/#usemouseinelement has `elementX` and `elementY`

(+https://vueuse.org/core/useCssVar/#usecssvar to bind the cssvar)

alexdln: And I even touched this somewhere. Thanks for the tip ❤️

Let's see if we need this effect and if so I'll update it then

Issue #1534, opened 2026-02-17
Linking to specific version of package returns 404


dkazmer:
Linking the following URL from Bluesky yields 404 error and the weird transformed URL thereafter.
https://npmx.dev/package/obsidium/v/3.1
https://npmx.dev/package/[name]/v/[version]-isr

npmx has a lot of flaws and they've blocked me there because I dared to expose them (even constructively).
I'll probably go back to using regular npmjs.com

mikouaji:
Looks like a ISR regression, npmx tries to cache very aggressively so its possible with working on it this issue was introduced. Thanks for reporting.

danielroe:
the issue here is that 3.1 is not a valid version.

you can use `https://npmx.dev/package/obsidium/v/^3.1`

we should probably improve the error you get when you link to an invalid/incomplete version (or redirect to a valid one)

> npmx has a lot of flaws and they've blocked me there because I dared to expose them (even constructively).

no idea what you're talking about but we very much welcome the reporting of bugs 🙂

dkazmer:
@mikouaji thanks. Yeah, I noticed the aggressive caching. I have to open up dev tools and then refresh the browser to load the new data.

@danielroe 
> no idea what you're talking about

I'm sure you don't but nonetheless...
The error message could include a note about proper version URLs (with the little cap), or better yet, just cache the same way as npmjs.com

danielroe:
this isn't related to caching

danielroe:
you can see npm doesn't work either, because it is an invalid version:

https://www.npmjs.com/package/obsidium/v/3.1

dkazmer:
Test links for comparison
https://www.npmjs.com/package/obsidium/v/3.1.0
https://npmx.dev/package/obsidium/v/3.1.0

dkazmer:
@danielroe the first in above test works as expected, while the second doesn't..
Test w/ cap: https://npmx.dev/package/obsidium/v/^3.1
This one kinda works, but omits the README

alex-key:
Some more details for reproduction:

- Go to https://npmx.dev/package/obsidium
- Click `3.1.0` (latest version)
- See URL is correct (https://npmx.dev/package/obsidium/v/3.1.0)
- Refresh the page
- Url is broken, page is broken

I will try to investigate. It should be some incorrect logic with redirection for the latest version (where `"package/{name}"` and `"package/{name}/v/{latest}"` should be the same)

alex-key:
[...]

Issue #1530, opened 2026-02-16
"500 - Internal server error" as a title?


wojtekmaj:
https://npmx.dev/package/vite-plus/v/0.0.1

<img width="1146" height="737" alt="Image" src="https://github.com/user-attachments/assets/defd75dc-f16b-4d92-b9fa-cfe372ced065" />

rzzf:
This seems to be caused by the same issue as #1483.

Issue #1267, opened 2026-02-09
Page loads with blank layout


sybers:
When loading with hard refresh (Cmd+shift+R) the page is "empty", happens not that often (I'd say 10% of the time).

Happened to me with latest Chrome version on the [vitest package page](https://npmx.dev/package/vitest).

<img width="1181" height="1065" alt="Image" src="https://github.com/user-attachments/assets/5c703e29-55fe-4801-ac41-a0bc038fccad" />

Issue #1266, opened 2026-02-09
Page loads but page title indicates : "500 Internal Server Error"


sybers:
Sometimes the page loads correctly but the page title is "500 Internal Server Error".

Happened to me on the [vite package page](https://npmx.dev/package/vite) on the current production version.

<img width="989" height="801" alt="Image" src="https://github.com/user-attachments/assets/d76577fb-e3fe-4416-b0b9-df9a56d564ae" />

PR #1575, opened 2026-02-22
fix: refactor routing to fix errors

### LINKED ISSUE #1558: Hydration mismatch on package pages
Image

alexdln:
### 🔗 Linked issue

Resolves #1558 #1530 #1534 #1267 #1266 #1194

### 🧭 Context

* Version pages almost always displayed a 500 - server error as the title;
* Version pages often failed to load, with a redirect to segmentPath in half the cases;
* In many cases, the readme didn't appear, or the content disappeared after hydration.

From the [initial research](https://github.com/npmx-dev/npmx.dev/issues/1534#issuecomment-3941287347):
* Server always sees the wrong path for version pages. But sometimes it returns a blank page, and once on the client, it manages to work correctly and displays the page as expected
* Sometimes it redirects immediately, apparently on the client. That is, the request path is correct, but the response immediately shows that the path is segmented, not injected parameters
* Definitely server routing issue. But when there's a trailingSlash at the end of the path, the problem doesn't reproduce (https://npmxdev-git-fork-alexdln-chore-test-route-overriting-npmx.vercel.app/package/obsidium/v/3.1.0/)
* But if we enable trailingSlash in the vercel settings, then for versions it will always redirect to a page without it (?!)
* It seems to redirect when we request for the version failed for some reason (f.e. rate limit). That is, if our request failed, we get a server error and a redirect. And if it somehow went through, we get a blank page from server, since the [name] package doesn't exist and then we trying to load it client-side

### 📚 Description

* Rewrote the version loading logic, as the previous variant often lost data from server and failed to load on the client;
* Enabled trailingSlash - without it, navigation errors in Nitro often occur
  - Enabled it via the internal middleware, as the Vercel setting often didn't work as expected. And the nuxt.config setting didn't work at all
* Removed certain settings in nuxt.config for payload paths. They aren't rendered independently of the HTML, but they could be cached. This caused mismatches, and when the initial payload was returned, an additional hydration error occurred

Some prefetch requests now return an 404 error - this isn't a degradation; it's just that previously, in the event of a failure, a fallback was returned with an empty object that could be cached and processed. Now the behavior is more correct.

🤞🤞🤞

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -135,10 +135,6 @@ export default defineNuxtConfig({
       },
     },
     // pages
-    '/package/:name': getISRConfig(60),
-    '/package/:name/v/:version': getISRConfig(3600),
-    '/package/:org/:name': getISRConfig(60),
-    '/package/:org/:name/v/:version': getISRConfig(3600),
     '/package/:name/': getISRConfig(60),
     '/package/:name/v/:version/': getISRConfig(3600),
     '/package/:org/:name/': getISRConfig(60),


alexdln:
I was so close to the cause, but I chose the wrong conclusion.

It turns out the trailing slash only worked because ISR is strict about it, so pages with slashes were essentially ignored by nitro. And as a result, I mostly solved the problems that had been added because of this

The problem is rather that our condition for isr was incomplete (again due to strictness)

reopened #1604

PR #1578, opened 2026-02-22
feat: add animation toggle for sparkline chart

### LINKED ISSUE #1539: Weekly Downloads animation should include a way to pause or stop it (WCAG Level A requirement)
We probably should include a mechanism to stop or pause the weekly downloads animation on the package page.

https://github.com/user-attachments/assets/ca27a7b5-18aa-464f-9eb6-d6d120d96069

[WCAG 2.2 Success Criterion 2.2.2 Pause, Stop, Hide (Level A)](https://www.w3.org/TR/WCAG22/#pause-stop-hide) states:

> For moving, [blinking](https://www.w3.org/TR/WCAG22/#dfn-blinking), scrolling, or auto-updating information, all of the following are true:
>
> <dl><dt>Moving, blinking, scrolling</dt><dd>
>
> For any moving, blinking or scrolling information that (1) starts automatically, (2) lasts more than five seconds, and (3) is presented in parallel with other content, there is a [mechanism](https://www.w3.org/TR/WCAG22/#dfn-mechanism) for the user to [pause](https://www.w3.org/TR/WCAG22/#dfn-pause), stop, or hide it unless the movement, blinking, or scrolling is part of an activity where it is [essential](https://www.w3.org/TR/WCAG22/#dfn-essential); and
> </dd><dt>Auto-updating</dt><dd>
>
> For any auto-updating information that (1) starts automatically and (2) is presented in parallel with other content, there is a mechanism for the user to pause, stop, or hide it or to control the frequency of the update unless the auto-updating is part of an activity where it is essential.
> </dd></dt>

For **“Moving, blinking, scrolling,”** the weekly downloads animation applies for all cases (i.e. it starts automatically, lasts more than five seconds, and is present in parallel with other content).

It’s a little iffy whether or not respecting reduced motion—which this animation does—is sufficient to conform to this success criterion (see Hidde de Vries’ article [“Meeting ‘2.2.2 Pause, Stop, Hide’ with prefers-reduced-motion”](https://hidde.blog/meeting-2-22-pause-stop-hide-with-prefers-reduced-motion/)). Conformance isn’t the purpose of accessibility—providing a better user experience for those with various need is. For example, folks with some cognitive conditions might find it distracting for an animation like this to be continually playing, however, they might positively benefit from the effects of UI animations as it can help with context switching.

graphieros:
Resolves #1539 

This adds a toggle button under the sparkline chart to play/pause the pulse animation.
The button is not displayed when prefers-reduced-motion is enabled, since the chart does not animate in this case.
The state of the animation state is saved in the application settings.

| Animation enabled | Animation disabled |
| -------- | ---------|
|image|image|

### DIFF
--- a/app/components/Package/WeeklyDownloadStats.vue
+++ b/app/components/Package/WeeklyDownloadStats.vue
@@ -311,15 +311,7 @@ const config = computed(() => {
           </ClientOnly>
 
           <div v-if="hasWeeklyDownloads" class="hidden motion-safe:flex justify-end">
-            <ButtonBase
-              size="small"
-              :aria-label="
-                hasSparklineAnimation
-                  ? $t('package.trends.pause_animation')
-                  : $t('package.trends.play_animation')
-              "
-              `@click`="toggleSparklineAnimation"
-            >
+            <ButtonBase size="small" @click="toggleSparklineAnimation">
               {{
                 hasSparklineAnimation
                   ? $t('package.trends.pause_animation')


knowler (approved):
Awesome—LGTM!

RYGRIT:
@graphieros  I found a small problem: when activated with the Tab key, some content is not fully displayed.

image

graphieros:
@RYGRIT thanks for spotting this :)
Fixed in #1589

PR #1586, opened 2026-02-22
feat: add package download button

### LINKED ISSUE #1528: Proposal: Add a button to direct download a dependency
It shouldn't be too hard but it would be interesting if you added a button to download the package and optionally it's dependencies.

Adebesin-Cell:
### 🔗 Linked issue

resolves https://github.com/npmx-dev/npmx.dev/issues/1528

### 🧭 Context

There was previously no way to directly download a package tarball or fetch all dependencies from the package detail page. 

This PR introduces a Download button to make that happen.

### 📚 Description

This change adds a new **Download** button to the package detail page. The button includes a dropdown menu with two options:

* Download the package `.tgz` tarball directly.
* Generate and download a `.jh` script to fetch all dependencies.

### Screenshot

-> OLD UI
Screenshot 2026-02-22 at 21 36 23

-> NEW UI
Screenshot 2026-03-17 at 04 41 54


serhalp:
@Adebesin-Cell 👋🏼 I'd love to get this great feature reviewed and shipped! Would you mind rebasing it first? 🙏🏼

Adebesin-Cell:
@serhalp Resolved. I spent the entire day trying to figure out why browser testing kept failing, but I couldn’t get it to work. Any ideas?

app/components/Package/DownloadButton.vue
```diff
@@ -0,0 +1,253 @@
...
+  },
+)
+
+const triggerRef = useTemplateRef('triggerRef')
+const listRef = useTemplateRef('listRef')
+const isOpen = shallowRef(false)
+const highlightedIndex = shallowRef(-1)
+const dropdownPosition = shallowRef<{ top: number; left: number } | null>(null)
+
+const { t } = useI18n()
```
serhalp: We just use the `$t` auto-import
```suggestion
```

app/components/Package/DownloadButton.vue
```diff
@@ -0,0 +1,253 @@
...
+    const link = document.createElement('a')
+    link.href = tarballUrl
+    link.download = `${props.packageName.replace(/\//g, '__')}-${props.version.version}.tgz`
+    document.body.appendChild(link)
+    link.click()
+    document.body.removeChild(link)
+  }
+}
+
+function downloadDependenciesScript() {
```
serhalp: I'm having a hard time imagining the value of this feature. What use cases are you thinking of?

My suggestion would be to punt on this part for this PR and revisit this in a follow-up. Does that seem ok?

app/components/Package/DownloadButton.vue
```diff
@@ -0,0 +1,253 @@
...
+  if (!props.installSize) return
+
+  const lines = [
+    '#!/bin/bash',
+    `# Download dependencies for ${props.packageName}@${props.version.version}`,
+    'mkdir -p node_modules',
```
serhalp: This script will only work on Linux and macOS machines, so we'd need to either generate a Windows version conditionally or warn that it isn't Windows compatible.

But see my comment above – I'm not sure this feature is worthwhile 🤔.

app/components/Package/DownloadButton.vue
```diff
@@ -0,0 +1,253 @@
...
+const listRef = useTemplateRef('listRef')
+const isOpen = shallowRef(false)
+const highlightedIndex = shallowRef(-1)
+const dropdownPosition = shallowRef<{ top: number; left: number } | null>(null)
+
+const { t } = useI18n()
+const menuId = 'download-menu'
+const menuItems = computed(() => {
+  const items = [{ id: 'package', label: t('package.download.package'), icon: 'i-lucide:package' }]
+  if (props.installSize) {
```
serhalp: nit: Since this doesn't have anything to do with "install size" per se, it feels like this is misusing the field a bit?

serhalp (commented):
🙌🏼 This is looking good to me and works well.

However I'd vote to remove the deps `.sh` feature for now and open a new issue to discuss it.

app/components/Package/DownloadButton.vue
Adebesin-Cell: That’s fair, I’m happy to revisit later.

The main use case I had in mind is allowing users to download dependency tarballs directly so they can inspect or audit them before installing. It’s not something everyone would use, but it can be helpful in cases where people want more visibility into what’s being pulled in. It aligns with the original author of the feature as well https://github.com/npmx-dev/npmx.dev/issues/1528#issuecomment-4088851180 

But we can talk about it if we want to implement it, I'm happy to work on that and make it cross-platform as well 👍

app/components/Package/DownloadButton.vue
Adebesin-Cell: Good catch! Fixed here [`ba7281d` (this PR)](https://github.com/npmx-dev/npmx.dev/pull/1586/commits/ba7281d24ffa118abc22ea0f89ce0f0c1b2aee47)

i18n/locales/en.json
```diff
@@ -595,6 +595,11 @@
       "b": "{size} B",
       "kb": "{size} kB",
       "mb": "{size} MB"
+    },
+    "download": {
+      "button": "Download",
+      "package": "Download Package (.tgz)",
+      "dependencies": "Download Dependencies (.sh)"
```
ghostdevv: this says `.sh` but it downloads a JS script?

ghostdevv (changes requested):

i18n/locales/en.json
Adebesin-Cell: Good catch! I changed the implementation a bit to run a script instead of downloading an sh file to make it adaptable to any platform.

ghostdevv (changes requested):
What if we downloaded the tarballs for the user instead of having them run a script?

Adebesin-Cell:
> What if we downloaded the tarballs for the user instead of having them run a script?

**Interesting idea, though my concern is the number of files**, a package like `express` has ~60 deps. Serving all those tarballs individually would be a lot.

We could zip them into a single archive server-side (e.g., using something like `archiver` / `zip-stream`), fetching the tarballs on demand and streaming them into a zip as the user requests the download, rather than pre-fetching everything. That keeps it manageable.

That said, @serhalp also had second thoughts on the dependency download part of this PR and suggested punting on it to revisit in a follow-up. The single-package `.tgz` download works well as-is.

Maybe we land the single-package download now and open a separate issue to discuss the bulk dependency download approach (zipping, streaming, etc.)?

ghostdevv:
> Interesting idea, though my concern is the number of files, a package like express has ~60 deps. Serving all those tarballs individually would be a lot.

But you'd need to download the files anyway? Is the idea that the script just gives you the option to do it later?

> We could zip them into a single archive server-side (e.g., using something like archiver / zip-stream), fetching the tarballs on demand and streaming them into a zip as the user requests the download, rather than pre-fetching everything. That keeps it manageable.

I'd be worried about the memory usage

Adebesin-Cell:
> > Interesting idea, though my concern is the number of files, a package like express has ~60 deps. Serving all those tarballs individually would be a lot.
> 
> But you'd need to download the files anyway? Is the idea that the script just gives you the option to do it later?
> 
> > We could zip them into a single archive server-side (e.g., using something like archiver / zip-stream), fetching the tarballs on demand and streaming them into a zip as the user requests the download, rather than pre-fetching everything. That keeps it manageable.
> 
> I'd be worried about the memory usage

Yeah, you're right! I've removed the download dependencies script option for now in [5e0882f](https://github.com/npmx-dev/npmx.dev/pull/1586/commits/5e0882f2a3e920b17ae803df55a594195a49989e).
                                                                                                                           
We can open a separate issue to discuss the best approach for bulk dependency downloads if it's something we want to revisit.

ghostdevv:
@Adebesin-Cell I removed the dropdown for now, as well as simplified the subtle button code

We should do a follow up for subtle button because there are a few places that use button base but have some subtle styles on them. The package manager dropdown uses the old button and can probably be updated to use button base with a new subtle style. Here are some examples of how the newer button style has a subtle hint to it:

image

ghostdevv (approved):

Adebesin-Cell:
> @Adebesin-Cell I removed the dropdown for now, as well as simplified the subtle button code
> 
> We should do a follow up for subtle button because there are a few places that use button base but have some subtle styles on them. The package manager dropdown uses the old button and can probably be updated to use button base with a new subtle style. Here are some examples of how the newer button style has a subtle hint to it:
> 
> image

Hey! Thanks for this, it looks neat 👏. I was actually going to ask if we should declutter the dropdown into a button. Glad you went ahead and did it!                                                                                          

Thanks for staying on top of the review and helping improve the PR 🙏

PR #1594, opened 2026-02-23
feat: show recently viewed packages/orgs/users on homepage

serhalp:
### 🔗 Linked issue

N/A but see https://discord.com/channels/1464542801676206113/1464542802565140506/1475323380226854953

### 🧭 Context

This replaces the hardcoded list of quick links to frameworks, which provides limited value and isn't scalable.

### 📚 Description

Instead, we now track up to 5 most recently viewed entities (packages, orgs, users) in localStorage and show them on the homepage. This allows users to quickly navigate back to previously viewed pages. If no recently viewed items are known, the section is hidden.

https://github.com/user-attachments/assets/55a7a98e-4b1c-4b55-8c25-8c7639bf1358

Note: There's potential for this to eventually be powered by atproto and tied to a user (and therefore to sync across devices), but we can tackle that later (among other things, there are data privacy implications).


app/composables/useRecentlyViewed.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export function trackRecentView(item: Omit<RecentItem, 'viewedAt'>) {
+  if (import.meta.server) return
+  const items = getRecentRef()
+  const filtered = items.value.filter(
+    existing => !(existing.type === item.type && existing.name === item.name),
+  )
+  filtered.unshift({ ...item, viewedAt: Date.now() })
+  items.value = filtered.slice(0, MAX_RECENT_ITEMS)
+}
```
OrbisK: What is the reason for separating `trackReventView` and `items`? 

If we move `trackRecentView` into `useRecentlyViewed` we dont need `recentRef` as a shared state. `useLocalStorage` [should sync](https://github.com/vueuse/vueuse/blob/main/packages/core/useStorage/index.ts#L207-L211) across all refs. If we want to share the exact same ref we could use https://vueuse.org/shared/createSharedComposable/#createsharedcomposable

```suggestion
export function useRecentlyViewed() {
  const items = useLocalStorage<RecentItem[]>(STORAGE_KEY, [])
  
  function trackRecentView(item: Omit<RecentItem, 'viewedAt'>) {
    if (import.meta.server) return
    const filtered = items.value.filter(
      existing => !(existing.type === item.type && existing.name === item.name),
    )
    filtered.unshift({ ...item, viewedAt: Date.now() })
    items.value = filtered.slice(0, MAX_RECENT_ITEMS)
  }
  return { items: computed(() => items.value), trackRecentView }
}
```

serhalp: Great feedback, thank you! I'm not a vue/nuxt expert so this is helpful 😃.

Issue #1562, opened 2026-02-22
New tab icon in governance card


alexdln:
Right now in governance cards, clicking the new-tab icon (_on top right corner of cards_) doesn't do anything. We need to set the icon's `pointer-events: none` property

https://npmx.dev/about

<img width="190" height="83" alt="Image" src="https://github.com/user-attachments/assets/73f52794-8d64-443b-8d11-3f40caaaef76" />

PR #1596, opened 2026-02-23
feat: add contributors popup card

### LINKED ISSUE #1564: feat: core team in /about, with description and identity claims
See [Vitest team page](https://deploy-preview-9663--vitest-dev.netlify.app/team)

Image

We don't need our cards to be as big as these ones, but we could expand them a bit so we can have the identity claims to github, bsky, discord, etc. And allow for a short description line. Mine would be: "Open Source Adventurer", other folks would normally add their position at the company they work for, specially if the company let them spend some of their paid time on npmx.

userquin:
### 🔗 Linked issue

resolves #1564

### 🧭 Context

This PR includes a single popup for all contributors avatar including sponsor link.

**NOTE**: this PR dont't use `dir="auto"` nor `bdi` tags for RTL locales in the popup content, the browser interprets that as LTR and, by default, the text will align to the physical left of the popup. In a purely RTL (Real-Time) UI, this creates a "zigzag" effect:
- The popup appears on the right (good).
- The name is stuck to the left of the popup (because it's ASCII/Latin).
- The bio, if it starts with an emoji or Arabic text, is stuck to the right.
- unocss icons (like the company logo) remain in their original "start" position (on the right).

I cannot test this on Safari, so any help on this is welcome 🙏 .

**NOTE**: we need `NUXT_GITHUB_ORG_TOKEN` at Vercel (for PR preview environment), without the token only popup shown for `steward` contributors (default ones)

<details>
<summary>some screenshots</summary>

danielroe's popup card

<br/>

antfu's popup card

</details>

### 📚 Description

This PR uses GH graphql to get the following info to be displayed in the popup (sponsor was already there):
- name
- bio
- company and companyName (companyHTML when present is used, it is sanitized from the graphql endpoint)
- location
- website url
- twitter link
- bluesky and mastodon links using `socialAccounts` from `graphql` endpoint

The card uses purple colors for the sponsor anchor colors and includes also an anchor for the GH account. The card won't work without JavaScript, too complex to handle focus state, the user can just press enter on the avatar and will be sent to the GH account page.

~~This PR also includes a change to `Link/Base.vue` since there is no way to remove the external link icon (added back `noExternalIcon` boolean prop).~~ Now we're using the new `no-new-tab-icon` prop.

**NOTE**: we'll need to paginate GH `graphql` endpoint call. DONE ✅ 

To test this PR in your local you MUST be a maintainer here, and create a PAT with `read:org` at `admin:org`:
- go to your [tokens](https://github.com/settings/tokens)
- click on `Generate new token` and select `Generate new token (classic)`
- add a note to identify the token and check `read:org` at `admin:org`
- click on `Generate token` button: don't close the page, copy the token before leaving the page
- you can also use any token you have with the `read:org`

PAT with `read:org` at `admin:org`

Once created, you MUST copy and paste it at `.env` file at root (included at `.gitignore` and generated by one module) using:
`NUXT_GITHUB_ORG_TOKEN=<your_token>`:

.env file containing the NUXT_GITHUB_ORG_TOKEN

Stop dev server, delete `.nuxt/cache` folder and start dev server.

Used Gemini PRO to change the popup position and the arrow at the bottom/top, I did some manual changes.


server/api/contributors.get.ts
```diff
@@ -61,15 +75,48 @@ async function fetchTeamMembers(token: string): Promise<TeamMembers | null> {
...
+    html
+      // 1. Remove GitHub-specific tracking and metadata attributes
+      .replace(/\s(data-hovercard-[a-z-]+|data-octo-[a-z-]+|aria-keyshortcuts)="[^"]*"/gi, '')
+      // 2. Inject security and target attributes to all <a> tags
+      .replace(/<a /gi, '<a target="_blank" rel="noopener noreferrer" ')
+      // 3. Clean up any resulting double spaces
+      .replace(/\s{2,}/g, ' ')
+      .trim()
+  )
+}
```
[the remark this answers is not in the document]
userquin: upps, my fault, but we want to render the `companyHTML`, GH graphql should escape the content.

app/components/Link/Base.vue
```diff
@@ -36,9 +36,16 @@ const props = withDefaults(
 
       /** should only be used for links where the context makes it very clear they are clickable. Don't just use this, because you don't like underlines. */
       noUnderline?: boolean
+
+      /**
+       * should external link icon be displayed?.
+       *
+       * @default false
+       */
+      noExternalIcon?: boolean
```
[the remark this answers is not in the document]
userquin: fixed

app/pages/about.vue
```diff
@@ -252,43 +303,175 @@ const roleLabels = computed(
...
+          <div class="flex flex-col gap-1 min-w-0">
+            <span class="w-full font-sans font-bold text-fg leading-tight truncate block">
+              {{ activeContributor.name || activeContributor.login }}
+            </span>
+            <div
+              v-if="activeContributor.role"
+              class="font-mono text-3xs uppercase tracking-wider text-accent font-bold"
+            >
+              {{ roleLabels[activeContributor.role] }}
+            </div>
```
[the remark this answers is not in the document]
userquin: fixed

app/pages/about.vue
```diff
@@ -252,43 +303,175 @@ const roleLabels = computed(
...
+            >
+              <div
+                class="i-lucide:building-2 size-3 shrink-0 mt-0.5 text-accent/80"
+                aria-hidden="true"
+              />
+              <div
+                class="company-content leading-relaxed break-words min-w-0 [&_a]:(text-accent no-underline hover:underline transition-all)"
+                v-html="activeContributor.companyHTML || activeContributor.company"
+              />
+            </div>
```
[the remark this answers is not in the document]
userquin: fixed

server/api/contributors.get.ts
```diff
@@ -90,15 +137,31 @@ async function fetchSponsorable(token: string, logins: string[]): Promise<Set<st
...
+        (GitHubUserData & { login: string; hasSponsorsListing: boolean }) | null
+      >
     }
 
     const sponsorable = new Set<string>()
     if (json.data) {
       for (const user of Object.values(json.data)) {
-        if (user?.hasSponsorsListing) {
+        if (!user) continue
+        console.log(JSON.stringify(user))
```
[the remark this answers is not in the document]
userquin: fixed

server/api/contributors.get.ts
```diff
@@ -90,15 +137,31 @@ async function fetchSponsorable(token: string, logins: string[]): Promise<Set<st
...
+        // --- SERVER-SIDE SANITIZATION AND BATCHING ---
+        usersData.set(user.login, {
+          name: cleanString(user.name),
+          bio: cleanString(user.bio),
+          company: cleanString(user.company),
+          // Rich HTML sanitization for company mentions/orgs
+          companyHTML: sanitizeGitHubHTML(user.companyHTML),
+          location: cleanString(user.location),
+          websiteUrl: cleanString(user.websiteUrl),
+          twitterUsername: cleanString(user.twitterUsername),
```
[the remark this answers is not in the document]
userquin: fixed

server/api/contributors.get.ts
```diff
@@ -61,15 +75,52 @@ async function fetchTeamMembers(token: string): Promise<TeamMembers | null> {
...
+ * Prevents UI issues with empty icons or broken conditional logic.
+ */
+function cleanString(val: string | null, url = false): string | null {
+  if (!val || val === 'undefined' || val.trim() === '') return null
+  val = val.trim()
+  if (!url) {
+    return val
+  }
+  return val.startsWith('https://') || val.startsWith('http:') ? val : null
+}
```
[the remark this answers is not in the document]
userquin: fixed

server/api/contributors.get.ts
```diff
@@ -167,22 +233,24 @@ export default defineCachedEventHandler(
...
     return filtered
       .map(c => {
         const { role, order } = getRoleInfo(c.login, teams)
+        const userInfo = userData.get(c.login) || {}
         const sponsors_url = sponsorable.has(c.login)
           ? `https://github.com/sponsors/${c.login}`
           : null
-        Object.assign(c, { role, order, sponsors_url })
+        Object.assign(c, { role, order, sponsors_url, ...userInfo })
         return c as GitHubContributor & { order: number; sponsors_url: string | null; role: Role }
```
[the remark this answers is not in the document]
userquin: fixed

app/pages/about.vue
```diff
@@ -1,9 +1,20 @@
...
+// SSR & Validation Fix
+const isMounted = shallowRef(false)
+const activeContributor = shallowRef<GitHubContributor | null>(null)
+const openTimer = shallowRef<any>(null)
+const closeTimer = shallowRef<any>(null)
+const isFlipped = shallowRef(false)
+
+onMounted(() => {
+  isMounted.value = true
+})
```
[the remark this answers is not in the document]
userquin: uppps, final cleanup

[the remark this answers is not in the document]
userquin: fixed

server/api/contributors.get.ts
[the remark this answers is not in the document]
userquin: How about `sanitize-html` (we have this at dependencies: `"sanitize-html": "2.17.0"`)?. We already use it in readme rendering: https://github.com/npmx-dev/npmx.dev/blob/54aa86a47d064b11291dcfc75b2b1273ebbe7eaf/server/utils/readme.ts#L2

thx @shuuji3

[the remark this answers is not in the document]
userquin: @coderabbitai we may use `sanitize-html` at `server/api/contributors.get.ts`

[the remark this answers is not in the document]
userquin: @coderabbitai shouldnt handle some `aria-**` attrs (like aria-label) for the anchors?

RYGRIT:
You seem to have cited the wrong issue link; it seems more relevant to #1562.

userquin:
> You seem to have cited the wrong issue link; it seems more relevant to #1562.

Upps, no, I pick the wrong issue, it is #1564, updated thx ❤️

app/pages/about.vue
```diff
@@ -252,43 +307,192 @@ const roleLabels = computed(
...
                     loading="lazy"
                   />
-                  <span
-                    class="pointer-events-none absolute -top-9 inset-is-1/2 -translate-x-1/2 whitespace-nowrap rounded-md bg-gray-900 text-white dark:bg-gray-100 dark:text-gray-900 text-xs px-2 py-1 shadow-lg opacity-0 scale-95 transition-all duration-150 group-hover:opacity-100 group-hover:scale-100"
-                    dir="ltr"
-                    role="tooltip"
-                  >
-                    @{{ contributor.login }}
-                  </span>
                 </LinkBase>
```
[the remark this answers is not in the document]
userquin: This is intentional, we may add some long description (via aria-label), the popop requires JavaScript enabled and won't work with keyboard (too complex to add aria-controls, aria-expanded, etc.. to the anchor and handle focusin, focusout an escapce key); it is a visual/mouse enhancement (the click on the anchor will send the user to the GH page with all the info at GH).

We may need to add some description in the aria-label (company...).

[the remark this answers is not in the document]
userquin: The role is at `npmx.dev` not at the user's company! Check the logic at `server/api/contributors.get.ts`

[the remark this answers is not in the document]
userquin: @coderabbitai if you show the popup on focus, you need a way to close it, since the focus is trap by the popup, a user will need to navigate every link in the card (via Tab key) to jump to the next contributor; that's why I don't want to add support for keyboard (maybe via shortcut), here my tests with Space and esc when adding a popup per contributor (Firefox was taking 3/4 seconds to switch between Top Context) instead a single popup => in this PR):

<details>
<summary>focusin/focusout/keydown</summary>

```vue
<script setup lang="ts">
import type { Role, GitHubContributor } from '#server/api/contributors.get'

const router = useRouter()
const canGoBack = useCanGoBack()

const isMounted = shallowRef(false)
onMounted(() => {
  isMounted.value = true
})

useSeoMeta({
  title: () => `${$t('about.title')} - npmx`,
  ogTitle: () => `${$t('about.title')} - npmx`,
  twitterTitle: () => `${$t('about.title')} - npmx`,
  description: () => $t('about.meta_description'),
  ogDescription: () => $t('about.meta_description'),
  twitterDescription: () => $t('about.meta_description'),
})

defineOgImageComponent('Default', {
  primaryColor: '#60a5fa',
  title: 'about npmx',
  description: 'a fast, modern browser for the **npm registry**',
})

const pmLinks = {
  npm: 'https://www.npmjs.com/',
  pnpm: 'https://pnpm.io/',
  yarn: 'https://yarnpkg.com/',
  bun: 'https://bun.sh/',
  deno: 'https://deno.com/',
  vlt: 'https://www.vlt.sh/',
}

const { data, status: contributorsStatus } = useLazyFetch('/api/contributors')

/**
 * Checks if a contributor should show the expanded card.
 * Stewards and Maintainers always get a card to show their project role.
 */
function isExpandable(c: GitHubContributor) {
  const isGovernance = c.role === 'steward' || c.role === 'maintainer'
  return (
    isGovernance ||
    !!c.bio ||
    !!c.company ||
    !!c.name ||
    !!c.location ||
    !!c.twitterUsername ||
    !!c.websiteUrl
  )
}

/**
 * Clean computed for production data
 */
const contributors = computed(() => data.value ?? [])

const roleLabels = computed(
  () =>
    ({
      steward: $t('about.team.role_steward'),
      maintainer: $t('about.team.role_maintainer'),
    }) as Partial<Record<Role, string>>,
)

// --- Centralized Popover Management (Optimized) ---
const timers: Record<number, any> = {}
const openTimer = shallowRef<any>(null) // Timer para evitar aperturas accidentales
const activeId = shallowRef<number | null>(null)

/**
 * Optimized close: Only target the specific active popover
 */
function closeActivePopover() {
  if (activeId.value !== null) {
    const popover = document.getElementById(`popover-${activeId.value}`)
    if (popover) {
      try { (popover as any).hidePopover() } catch (e) {}
    }
    activeId.value = null
  }
}

/**
 * Handles opening with a small "grace period" to prevent lag in Firefox
 */
function openPopover(e: Event, id: number) {
  clearTimeout(openTimer.value)
  if (timers[id]) clearTimeout(timers[id])

  // Grace period: Only open if user hovers for 80ms
  openTimer.value = setTimeout(() => {
    if (activeId.value === id) return

    // Close the previous one efficiently
    closeActivePopover()

    const popover = document.getElementById(`popover-${id}`) as HTMLElement
    const anchor = document.getElementById(`anchor-${id}`)

    if (popover && anchor) {
      activeId.value = id
      const rect = anchor.getBoundingClientRect()

      const cardWidth = 256
      const padding = 16
      const idealLeft = rect.left + rect.width / 2
      const minLeft = cardWidth / 2 + padding
      const maxLeft = window.innerWidth - cardWidth / 2 - padding
      const finalLeft = Math.max(minLeft, Math.min(idealLeft, maxLeft))

      popover.style.left = `${finalLeft}px`
      popover.style.top = `${rect.top + 2}px`

      const arrow = popover.querySelector('.popover-arrow') as HTMLElement
      if (arrow) {
        const delta = idealLeft - finalLeft
        arrow.style.left = `calc(50% + ${delta}px)`
      }

      if (!popover.matches(':popover-open')) popover.showPopover()
    }
  }, 80) // 80ms debounce is the "sweet spot" for performance
}

function closePopover(id: number) {
  clearTimeout(openTimer.value) // Cancel opening if we leave early
  timers[id] = setTimeout(() => {
    const popover = document.getElementById(`popover-${id}`)
    if (popover && !popover.matches(':hover') && !popover.matches(':focus-within')) {
      if (activeId.value === id) {
        try { (popover as any).hidePopover() } catch (e) {}
        activeId.value = null
      }
    }
  }, 120)
}

function handleKeyDown(e: KeyboardEvent, id: number) {
  if (e.code === 'Space') {
    e.preventDefault()
    openPopover(e, id)
  } else if (e.code === 'Escape' && activeId.value === id) {
    e.preventDefault()
    closeActivePopover()
    document.getElementById(`anchor-${id}`)?.focus()
  }
}
</script>

<template>
  <main class="container flex-1 py-12 sm:py-16 overflow-x-hidden">
    <article class="max-w-2xl mx-auto">
      <header class="mb-12">
        <div class="flex items-baseline justify-between gap-4 mb-4">
          <h1 class="font-mono text-3xl sm:text-4xl font-medium">
            {{ $t('about.heading') }}
          </h1>
          <button
            v-if="canGoBack"
            type="button"
            class="cursor-pointer inline-flex items-center gap-2 font-mono text-sm text-fg-muted hover:text-fg transition-colors duration-200 rounded focus-visible:outline-accent/70 shrink-0"
            @click="router.back()"
          >
            <span class="i-lucide:arrow-left rtl-flip w-4 h-4" aria-hidden="true" />
            <span class="hidden sm:inline">{{ $t('nav.back') }}</span>
          </button>
        </div>
        <p class="text-fg-muted text-lg">
          {{ $t('tagline') }}
        </p>
      </header>

      <section class="prose prose-invert max-w-none space-y-8">
        <div>
          <h2 class="text-lg text-fg-subtle uppercase tracking-wider mb-4">
            {{ $t('about.what_we_are.title') }}
          </h2>
          <p class="text-fg-muted leading-relaxed mb-4">
            <i18n-t keypath="about.what_we_are.description" tag="span" scope="global">
              <template #betterUxDx>
                <strong class="text-fg">{{ $t('about.what_we_are.better_ux_dx') }}</strong>
              </template>
              <template #jsr>
                <LinkBase to="https://jsr.io/">JSR</LinkBase>
              </template>
            </i18n-t>
          </p>
          <p class="text-fg-muted leading-relaxed">
            <i18n-t keypath="about.what_we_are.admin_description" tag="span" scope="global">
              <template #adminUi>
                <strong class="text-fg">{{ $t('about.what_we_are.admin_ui') }}</strong>
              </template>
            </i18n-t>
          </p>
        </div>

        <div>
          <h2 class="text-lg text-fg-subtle uppercase tracking-wider mb-4">
            {{ $t('about.what_we_are_not.title') }}
          </h2>
          <ul class="space-y-3 text-fg-muted list-none p-0">
            <li class="flex items-start gap-3">
              <span class="text-fg-subtle shrink-0 mt-1">&mdash;</span>
              <span>
                <strong class="text-fg">{{
                  $t('about.what_we_are_not.not_package_manager')
                }}</strong>
                {{ ' ' }}
                <i18n-t
                  keypath="about.what_we_are_not.package_managers_exist"
                  tag="span"
                  scope="global"
                >
                  <template #already>{{ $t('about.what_we_are_not.words.already') }}</template>
                  <template #people>
                    <LinkBase :to="pmLinks.npm" class="font-sans">{{
                      $t('about.what_we_are_not.words.people')
                    }}</LinkBase>
                  </template>
                  <template #building>
                    <LinkBase :to="pmLinks.pnpm" class="font-sans">{{
                      $t('about.what_we_are_not.words.building')
                    }}</LinkBase>
                  </template>
                  <template #really>
                    <LinkBase :to="pmLinks.yarn" class="font-sans">{{
                      $t('about.what_we_are_not.words.really')
                    }}</LinkBase>
                  </template>
                  <template #cool>
                    <LinkBase :to="pmLinks.bun" class="font-sans">{{
                      $t('about.what_we_are_not.words.cool')
                    }}</LinkBase>
                  </template>
                  <template #package>
                    <LinkBase :to="pmLinks.deno" class="font-sans">{{
                      $t('about.what_we_are_not.words.package')
                    }}</LinkBase>
                  </template>
                  <template #managers>
                    <LinkBase :to="pmLinks.vlt" class="font-sans">{{
                      $t('about.what_we_are_not.words.managers')
                    }}</LinkBase>
                  </template>
                </i18n-t>
              </span>
            </li>
            <li class="flex items-start gap-3">
              <span class="text-fg-subtle shrink-0 mt-1">&mdash;</span>
              <span>
                <strong class="text-fg">{{ $t('about.what_we_are_not.not_registry') }}</strong>
                {{ $t('about.what_we_are_not.registry_description') }}
              </span>
            </li>
          </ul>
        </div>

        <div>
          <h2 class="text-lg text-fg-subtle uppercase tracking-wider mb-4">
            {{ $t('about.team.title') }}
          </h2>
          <p class="text-fg-muted leading-relaxed mb-6">
            {{ $t('about.contributors.description') }}
          </p>

          <section aria-labelledby="contributors-heading">
            <h3
              id="contributors-heading"
              class="text-sm text-fg-subtle uppercase tracking-wider mb-4"
            >
              {{
                $t(
                  'about.contributors.title',
                  { count: $n(contributors.length) },
                  contributors.length,
                )
              }}
            </h3>

            <div
              v-if="contributorsStatus === 'pending'"
              class="text-fg-subtle text-sm"
              role="status"
            >
              {{ $t('about.contributors.loading') }}
            </div>
            <div
              v-else-if="contributorsStatus === 'error'"
              class="text-fg-subtle text-sm"
              role="alert"
            >
              {{ $t('about.contributors.error') }}
            </div>

            <ul
              v-else-if="contributors.length"
              class="grid grid-cols-[repeat(auto-fill,48px)] justify-center gap-2 list-none p-0 overflow-visible pt-16"
            >
              <li
                v-for="contributor in contributors"
                :key="contributor.id"
                class="relative h-12 w-12 list-none"
              >
                <LinkBase
                  :id="`anchor-${contributor.id}`"
                  :to="contributor.html_url"
                  no-underline
                  no-external-icon
                  class="group relative block h-12 w-12 rounded-lg transition-all outline-none focus-visible:(ring-2 ring-accent z-20)"
                  :class="[
                    !isExpandable(contributor)
                      ? 'hover:scale-125 focus-visible:scale-125'
                      : 'hover:scale-110 focus-visible:scale-110',
                  ]"
                  :aria-haspopup="isMounted ? 'dialog' : undefined"
                  :aria-expanded="isMounted ? activeId === contributor.id : undefined"
                  :aria-controls="isMounted ? `popover-${contributor.id}`: undefined"
                  :aria-label="isMounted
                    ? $t('about.contributors.view_profile_template', { name: contributor.login, press_space: $t('about.contributors.press_space') })
                    : $t('about.contributors.view_profile', { name: contributor.login })"
                  @mouseenter="openPopover($event, contributor.id)"
                  @mouseleave="closePopover(contributor.id)"
                  @keydown="handleKeyDown($event, contributor.id)"
                >
                  <img
                    :src="`${contributor.avatar_url}&s=64`"
                    :alt="$t('about.contributors.avatar', [contributor.login])"
                    width="48"
                    height="48"
                    class="w-12 h-12 rounded-lg ring-2 ring-transparent transition-all duration-200 hover:ring-accent"
                    loading="lazy"
                  />

                  <span class="absolute -bottom-2 -right-2 hidden group-focus-visible:block z-30">
                    <kbd class="flex items-center justify-center min-w-8 h-4 px-1 rounded border border-border-subtle bg-bg-elevated text-[8px] font-sans font-bold text-fg-subtle shadow-sm uppercase tracking-tighter">
                      <i18n-t keypath="about.contributors.press_space_template" tag="span" scope="global">
                        <template #press>
                          <span class="sr-only">{{ $t('about.contributors.press') }}</span>
                        </template>
                        <template #space>
                          {{ $t('about.contributors.space') }}
                        </template>
                      </i18n-t>
                    </kbd>
                  </span>
                </LinkBase>

                <ClientOnly>
                  <div
                    v-if="isExpandable(contributor)"
                    :id="`popover-${contributor.id}`"
                    popover="manual"
                    class="m-0 border-none bg-transparent p-0 overflow-visible inset-auto fixed translate-x-[-50%] translate-y-[-100%] pb-4 opacity-0 scale-95 invisible translate-y-[-97%] transition-all duration-300 ease-[cubic-bezier(0.23,1,0.32,1)] allow-discrete popover-open:(opacity-100 scale-100 blur-0 visible translate-y-[-100%])"
                    @mouseenter="clearTimer(contributor.id)"
                    @mouseleave="closePopover(contributor.id)"
                    @keydown="handleKeyDown($event, contributor.id)"
                  >
                    <div
                      class="relative z-10 w-64 overflow-hidden rounded-xl border border-border-subtle bg-bg-elevated p-4 shadow-[0_20px_50px_rgba(0,0,0,0.3)] dark:shadow-[0_20px_50px_rgba(0,0,0,0.5)] ring-1 ring-fg/10 max-w-[calc(100vw-32px)] text-start"
                    >
                      <div class="flex flex-col gap-1">
                        <span
                          class="font-sans font-bold text-fg leading-tight truncate block max-w-full"
                          :title="contributor.name || contributor.login"
                        >
                          {{ contributor.name || contributor.login }}
                        </span>
                        <div
                          v-if="roleLabels[contributor.role]"
                          class="font-mono text-3xs uppercase tracking-wider text-accent font-bold"
                        >
                          {{ roleLabels[contributor.role] }}
                        </div>

                        <p
                          v-if="contributor.bio"
                          class="mt-1 font-sans text-xs italic text-fg-subtle line-clamp-3 leading-relaxed"
                        >
                          "{{ contributor.bio }}"
                        </p>

                        <div
                          v-if="contributor.companyHTML || contributor.company"
                          class="mt-1 flex items-center gap-1 font-sans text-2xs text-fg-muted text-start min-w-0"
                        >
                          <div
                            class="i-lucide:building-2 size-3 shrink-0 mt-0.5 text-accent/80"
                            aria-hidden="true"
                          />
                          <div
                            class="company-content leading-relaxed break-words min-w-0 [&_a]:(text-accent no-underline hover:underline transition-all)"
                            v-html="contributor.companyHTML || contributor.company"
                          />
                        </div>
                      </div>

                      <div class="mt-3 flex flex-col gap-1 text-3xs text-fg-subtle font-sans">
                        <div v-if="contributor.location" class="flex items-center gap-1">
                          <div class="i-lucide:map-pin size-3" />
                          <span>{{ contributor.location }}</span>
                        </div>

                        <a
                          v-if="contributor.websiteUrl"
                          :href="contributor.websiteUrl"
                          target="_blank"
                          class="flex items-center gap-1 hover:text-accent transition-colors"
                        >
                          <div class="i-lucide:link size-3" />
                          <span class="truncate">{{
                            contributor.websiteUrl.replace(/^https?:\/\//, '')
                          }}</span>
                        </a>

                        <a
                          v-if="contributor.twitterUsername"
                          :href="`https://x.com/${contributor.twitterUsername}`"
                          target="_blank"
                          class="flex items-center gap-1 hover:text-accent transition-colors"
                        >
                          <div class="i-simple-icons:x size-2.5" />
                          <span>{{ contributor.twitterUsername }}</span>
                        </a>
                      </div>

                      <div
                        class="mt-3 flex items-center justify-between border-t border-t-gray-400/65 dark:border-t-gray-300 pt-3"
                      >
                        <a
                          :href="contributor.html_url"
                          target="_blank"
                          rel="noopener noreferrer"
                          class="text-3xs text-fg-subtle font-mono hover:text-accent transition-colors duration-200 focus-ring rounded px-0.5"
                        >
                          @{{ contributor.login }}
                        </a>

                        <a
                          v-if="contributor.sponsors_url"
                          :href="contributor.sponsors_url"
                          target="_blank"
                          class="flex items-center gap-1 rounded border border-purple-700/30 bg-purple-700/5 text-purple-700 dark:border-purple-300/30 dark:bg-purple-300/5 dark:text-purple-300 px-2 py-0.5 text-4xs font-bold uppercase tracking-wider transition-colors hover:bg-purple-700/15 dark:hover:bg-purple-300/15"
                        >
                          <div class="i-lucide:heart size-3" aria-hidden="true" />
                          <span class="leading-none">
                            {{ $t('about.team.sponsor') }}
                          </span>
                        </a>
                      </div>
                    </div>

                    <div
                      class="popover-arrow absolute bottom-[11px] inset-is-1/2 z-0 h-3 w-3 -translate-x-1/2 rotate-45 border-be border-ie border-border-subtle bg-bg-elevated transition-transform duration-300"
                    />
                  </div>
                </ClientOnly>
              </li>
            </ul>
          </section>
        </div>

        <CallToAction />
      </section>
    </article>
  </main>
</template>

<style scoped>
[popover]:not(:popover-open) { display: none; }
[popover] {
  /* GPU acceleration for smoother performance in Firefox */
  backface-visibility: hidden;
  transform: translateZ(0);
  /* Simplify transitions: Firefox struggles with complex allow-discrete in rapid succession */
  transition: opacity 0.2s, transform 0.2s, visibility 0.2s;
}

/* Styling for the KBD hint */
kbd {
  box-shadow: 0 1px 0 rgba(0,0,0,0.2);
}
</style>
```

app/pages/about.vue
```diff
@@ -45,6 +70,97 @@ const roleLabels = computed(
...
+  closeTimer.value = setTimeout(() => {
+    const popover = document.getElementById('shared-contributor-popover')
+    if (popover && !popover.matches(':hover')) {
+      try {
+        ;(popover as any).hidePopover()
+      } catch (e) {}
+      activeContributor.value = null
+    }
+  }, 120)
+}
```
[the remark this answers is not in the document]
userquin: fixed

app/pages/about.vue
```diff
@@ -45,6 +70,97 @@ const roleLabels = computed(
...
+
+  // 1. Wait for Vue to update the reactive state
+  await nextTick()
+
+  // 2. Open popover so it enters Top Layer (Firefox dimensions fix)
+  if (!popover.matches(':popover-open')) {
+    try {
+      ;(popover as any).showPopover()
+    } catch (e) {}
+  }
```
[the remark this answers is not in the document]
userquin: fixed

app/pages/about.vue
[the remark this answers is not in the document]
userquin: fixed with a detailed aria label

server/api/contributors.get.ts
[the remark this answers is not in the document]
userquin: fixed

app/pages/about.vue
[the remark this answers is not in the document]
userquin: @coderabbitai this should be fixed

app/pages/about.vue
```diff
@@ -45,6 +70,129 @@ const roleLabels = computed(
...
+  // --- Inside positionPopover function ---
+  const arrow = popover.querySelector('.popover-arrow') as HTMLElement
+  if (arrow) {
+    const delta = idealLeft - finalLeft
+    // We use a CSS variable to pass the value safely to the template
+    arrow.style.setProperty('--arrow-delta', `${delta}px`)
+  }
+}
+
+function onMouseEnter(contributor: GitHubContributor) {
```
danielroe: note - we should also do this when focusing these cards, otherwise the content is inaccessible to non-mouse-users.

same for tap devices - maybe if on a device with touch support then tapping should open the popup rather than go directly to the link

app/pages/about.vue
```diff
@@ -252,43 +339,196 @@ const roleLabels = computed(
...
     </article>
+
+    <ClientOnly>
+      <div
+        id="shared-contributor-popover"
+        popover="manual"
+        class="fixed top-0 force-left-0 m-0 border-none bg-transparent p-0 overflow-visible opacity-0 invisible transition-[opacity,visibility] duration-150 ease-out allow-discrete popover-open:(opacity-100 visible) z-40"
+        @mouseenter="cancelClose"
+        @mouseleave="onMouseLeave"
+      >
```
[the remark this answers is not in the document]
userquin: it s a custom rule at preset-rtl.ts and it is there, same as left-0, we use logical selectors (inset-is-0)

app/pages/about.vue
userquin: please check my discussion with coderabbitai here https://github.com/npmx-dev/npmx.dev/pull/1596#discussion_r2839092018 it is a complex problem we handle as a progressive enhacement.

We will need another approach, like an activator but ouside the link/anchor. Check some code I did on first approach in the previous coderabbitai comment.

app/pages/about.vue
```diff
@@ -252,43 +339,196 @@ const roleLabels = computed(
...
+            <a
+              v-if="activeContributor.twitterUsername"
+              :href="`https://x.com/${activeContributor.twitterUsername}`"
+              target="_blank"
+              rel="noopener noreferrer"
+              class="flex items-center gap-1 hover:text-accent transition-colors"
+            >
+              <div class="i-simple-icons:x size-2.5 shrink-0" />
+              <span>@{{ activeContributor.twitterUsername }}</span>
+            </a>
```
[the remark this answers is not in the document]
userquin: fixed

app/pages/about.vue
```diff
@@ -252,43 +339,196 @@ const roleLabels = computed(
               {{ $t('about.contributors.error') }}
             </div>
             <ul
-              v-else-if="communityContributors.length"
-              class="grid grid-cols-[repeat(auto-fill,48px)] justify-center gap-2 list-none p-0"
+              v-if="contributors.length"
+              class="grid grid-cols-[repeat(auto-fill,48px)] justify-center gap-2 list-none p-0 overflow-visible"
             >
```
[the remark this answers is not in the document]
userquin: fixed

app/pages/about.vue
userquin: As alternative we can switch the BaseLink to a BaseButton or just a custom button and use ENTER to go to GH profile or SPACE to show the popup or just prevent this logic and use the lnk in the popup to go to the GH profile if the user needs more info or just want.

userquin: We can't use an anchor as both a link and a button; this is semantically incorrect and a screen reader will go crazy: tapping the link will open a pop-up window, so what? It's an anchor/link.

userquin: Another problem, without JavaScript we'll need to use an anchor via noscript inside a list: maybe I can explore css alternatives (CSS grid)

app/pages/about.vue
```diff
@@ -252,43 +339,196 @@ const roleLabels = computed(
...
+              <div
+                class="i-lucide:building-2 size-3 shrink-0 mt-0.5 text-accent/80"
+                aria-hidden="true"
+              />
+              <div
+                class="company-content leading-relaxed break-words min-w-0 [&_a]:(text-accent no-underline hover:underline transition-all)"
+              >
+                {{ activeContributor.company }}
+              </div>
+            </div>
```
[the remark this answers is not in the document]
userquin: remove mt-0.5 from icon adding gap-1 to the container

app/pages/about.vue
danielroe: how about we just _never_ have the cards work as a link? they always popup a dialog...

userquin: Yeah, but patak wants same layout for any contributor but we cannot add such ammount of cards, we'll need to add an upper limit (the card size may vary, some empty some full, some with long bio or location)...

knowler: Since the “hovercard” includes contents that would be available if a user were to following the link and its not essential for the contents to be displayed on our page, I think it could be worth thinking of this feature as an enhancement. So, I don’t think it’s worth making it work for mobile—they can just tap through and find the same information.

I think we could just start off by copying GitHub here:

- Use a link as the base, but don’t alter its normal activation behaviour.
- Show the card on hover (for mouse users).
- We can set a shortcut to show the card for keyboard/screen reader users. GitHub uses `Alt+ArrowUp` which it sets in `aria-keyshortcuts`. We can include whatever we decide on in our keyboard shortcuts legend (I don’t think it’s worth showing the keys like we do for other shortcuts).
	- When the shortcut is used, the popover opens with the first focusable item focused.
- GitHub uses the `region` landmark here, but I might prefer the `article` role for the hovercard instead.

I’ll take a look around for more and do some more testing (I only tested VoiceOver on macOS), but I think this is an acceptable start.

knowler: Ok, I’m looking at the updated design and since it does include the name and role, that would be considered essential content, so it might be better for these to be buttons that open an anchored popover. I wouldn’t make the buttons themselves links. Links would need to be within the popover.

A rough example: https://knowler.dev/demos/r3QNDGk?codepen (Firefox and Safari don’t quite work as expected)

userquin: @knowler can you check latest changes?

knowler: @userquin The earliest I can get to this would be this evening or tomorrow morning, but I can review then.

server/api/contributors.get.ts
```diff
@@ -90,15 +167,56 @@ async function fetchSponsorable(token: string, logins: string[]): Promise<Set<st
...
+          for (const account of user.socialAccounts.nodes) {
+            if (account.url.includes('bsky.app')) {
+              // Extract handle from URL: https://bsky.app/profile/handle.bsky.social
+              const match = account.url.match(/bsky\.app\/profile\/([^/?]+)/)
+              if (match) {
+                blueskyHandle = match[1] as string
+              }
+            } else if (account.provider === 'MASTODON') {
+              mastodonUrl = account.url
+            }
```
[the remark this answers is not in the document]
userquin: fixed

app/pages/about.vue
```diff
@@ -252,43 +292,244 @@ const roleLabels = computed(
...
+                  ]"
+                  @mouseenter="
+                    isExpandable(contributor) &&
+                    open(contributor, $event.currentTarget as HTMLElement)
+                  "
+                  @mouseleave="isExpandable(contributor) && scheduleClose(contributor)"
+                  @click="
+                    isExpandable(contributor) &&
+                    toggle(contributor, $event.currentTarget as HTMLElement)
+                  "
```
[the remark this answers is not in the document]
userquin: fixed

app/pages/about.vue
```diff
@@ -45,6 +59,93 @@ const roleLabels = computed(
...
+}
+
+function onDocumentPointerDown(e: PointerEvent) {
+  if (!activeContributor.value) return
+  const t = e.target as HTMLElement
+  if (
+    !t.closest('[data-popover-panel]') &&
+    !t.closest(`[data-cid="${activeContributor.value.id}"]`)
+  ) {
+    activeContributor.value = null
```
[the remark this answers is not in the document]
userquin: fixed

server/api/contributors.get.ts
```diff
@@ -61,15 +109,44 @@ async function fetchTeamMembers(token: string): Promise<TeamMembers | null> {
...
  */
-async function fetchSponsorable(token: string, logins: string[]): Promise<Set<string>> {
+async function fetchGitHubUserData(
+  token: string,
+  logins: string[],
+  usersData: Map<string, GitHubUserData>,
+): Promise<Set<string>> {
   if (logins.length === 0) return new Set()
 
   // Build aliased GraphQL query: user0: user(login: "x") { hasSponsorsListing login }
```
Tobbe: Nitpick: The code comment is now outdated

app/pages/about.vue
```diff
@@ -47,6 +61,108 @@ const roleLabels = computed(
...
+  const separator = $t('about.contributors.separator')
+  const role = roleLabels.value[c.role]
+    ? $t('about.contributors.role', { separator, role: roleLabels.value[c.role] })
+    : ''
+  const works_at = c.company
+    ? $t('about.contributors.works_at', { separator, company: c.company })
+    : ''
+  const location = c.location
+    ? $t('about.contributors.location', { separator, location: c.location })
+    : ''
```
userquin: I think these entries are not required, and so we can remove them: `getAriaLabel` used only with LinkBase (non-expandable item) we can just use `$t('about.contributors.role', { separator, role: roleLabels.value[c.role] })` (any entry about the layout and the separator)

userquin: Daniel if the PR is fine for you I can add graphql pagination and remove previous translations later

app/pages/about.vue
```diff
@@ -265,44 +323,257 @@ const roleLabels = computed(
...
+                  />
+                </LinkBase>
+                <button
+                  v-else
+                  type="button"
+                  :data-cid="contributor.id"
+                  :aria-expanded="
+                    isMounted && activeContributor?.id === contributor.id ? 'true' : undefined
+                  "
+                  :aria-haspopup="isMounted && isExpandable(contributor) ? 'true' : undefined"
```
knowler:
```suggestion
```

We don’t need [`aria-haspopup`](https://www.w3.org/TR/wai-aria-1.3/#aria-haspopup) here as it was deprecated for [tooltips](https://www.w3.org/TR/wai-aria-1.3/#tooltip) in ARIA 1.2.

[“aria-haspopup might not do what you think it does” by Manuel Matuzović](https://matuzo.at/blog/2026/aria-haspopup-menu)

app/pages/about.vue
```diff
@@ -265,44 +323,257 @@ const roleLabels = computed(
...
-                  <span
-                    class="pointer-events-none absolute -top-9 inset-is-1/2 -translate-x-1/2 whitespace-nowrap rounded-md bg-gray-900 text-white dark:bg-gray-100 dark:text-gray-900 text-xs px-2 py-1 shadow-lg opacity-0 scale-95 transition-all duration-150 group-hover:opacity-100 group-hover:scale-100"
-                    dir="ltr"
+                </button>
+
+                <Transition name="pop">
+                  <div
+                    v-if="isExpandable(contributor) && activeContributor?.id === contributor.id"
+                    data-popover-panel
                     role="tooltip"
```
knowler:
```suggestion
```

We’ll also want to remove the tooltip semantics here since this isn’t a tooltip—it has a bunch of interactive content. I kinda think that an `<article>` would work well with a heading set at the appropriate level for the context. It’s fine that these headings don’t show up in the heading structure for screen readers since I think they would be noise anyway.

knowler (commented):
This is a partial review. I’ll take a look at a bit more soon…

app/pages/about.vue
userquin: ok, will check/change it tmr, thx

app/pages/about.vue
userquin: we're using it at the button

app/pages/about.vue
userquin: done

app/pages/about.vue
userquin: done

userquin: @knowler I added you as co-author in last commit, thx ❤️ for the help on this (wrong email at https://github.com/npmx-dev/npmx.dev/pull/1596/commits/1e8088e4dd7427bd2d5028250b55cc65f4cc3d99 🤦‍♂️ )

userquin:
I'm fixing key, mouse and tap events, right now article losing the scoped style... 🤔 and FF is (too) slow again

userquin:
I still need to fix tap events, the img doesn't scale...

userquin:
This should be ready now (fixed Firefox performance issues), here the summary:

## Firefox

- **Performance**: With 189+ contributors, Firefox can still degrade over time. Mitigated via event delegation on `<ul>`, `v-once` on each `<li>`, `contain: layout style` per item, and a single global popover node outside the `v-for`. May need revisiting as the list grows.
- **Popup outline on programmatic focus**: `outline: none !important` on `.contributor-popover` is a workaround — ideally `:focus:not(:focus-visible)` alone should suppress the outline, but Firefox ignores it on elements with `tabindex="-1"` when focused programmatically via JS.

## Images

- **`transition-all` caused massive reflows in Firefox**: replaced with `transition: box-shadow 100ms ease, transform 200ms ease` scoped to `[data-cid] img`. `transition-all` forces Firefox to recalculate every CSS property on every hover frame across all 189 items, collapsing the main thread. Always prefer explicit property transitions on large lists.

## Layout

- **`flex-wrap` vs `grid auto-fill`**: switched to `flex-wrap` for performance, but the last row centers its items instead of aligning left as `grid` with `auto-fill` would. `justify-center` on the `<ul>` mitigates the visual imbalance, though it is not as precise as a proper grid alignment.
- **`v-once` + `contain: layout style`**: each `<li>` is rendered once and treated as static. If a contributor's data changes (avatar, name, etc.), the node will not re-render without a page reload.

## Accessibility

- **Panel is outside its trigger's DOM subtree**: `role="group"` panel lives outside the button's `<li>` for performance reasons. Screen readers do not automatically associate the panel with its trigger button. Ideally the button should have `aria-controls="panel-id"` and the panel a matching `id`, but `v-once` prevents reactive bindings on the button. In practice, the semantic relationship is maintained implicitly through focus management and `aria-expanded` — and `aria-controls` itself has limited and inconsistent support across screen readers, so its absence is not critical.
- **`aria-expanded` managed via direct DOM manipulation**: moved out of Vue template bindings to avoid triggering reactive updates across all 189 buttons on every hover. Side effect: Vue DevTools will not reflect the `aria-expanded` state.

## Safari (untested) (asked to Claude AI)

- **`color-mix()` support**: `--un-ring-color` relies on `color-mix(in srgb, ...)` for the accent ring. Supported from Safari 16.2+. Older versions will silently ignore it — no ring will appear on hover, focus, or expanded state.
- **`contain: layout style` with `position: fixed`**: Safari may handle stacking context differently for elements inside contained ancestors. This risk applies to the buttons inside each `<li>` (which have `contain: layout style`), not to the popover itself — the popover is a direct child of `<main>`, outside the `<ul>` and its contained items, so its `position: fixed` placement should not be affected.
- **`padding-top` / `margin-top` hover bridge**: the 8px invisible area bridging the gap between button and popup (preventing premature close on mouse move) may behave differently in Safari regarding hit-testing.

app/pages/about.vue
```diff
@@ -265,44 +427,269 @@ const roleLabels = computed(
...
+        </div>
+      </div>
+    </Transition>
   </main>
 </template>
+
+<style scoped>
+[data-cid] img {
+  transition:
+    box-shadow 100ms ease,
```
knowler: A `box-shadow` transition will cause a whole lot of repaints. Set this on a pseudo element and then use `opacity` to fade it instead.

ghostdevv:
@userquin is this ready for review?

userquin:
Yes, it should be ready...

app/pages/about.vue
```diff
@@ -47,6 +61,212 @@ const roleLabels = computed(
...
+    popoverPos.align = 'left'
+    popoverPos.left = r.left
+  } else if (center + POP_W / 2 > vw - GAP) {
+    popoverPos.align = 'right'
+    popoverPos.left = r.right
+  } else {
+    popoverPos.align = 'center'
+    popoverPos.left = center
+  }
+}
```
knowler, line 117: Should we be using `@floatingui/vue` here instead? We use it elsewhere (e.g. the tooltips).

knowler (commented):
I still need to review some specifics, but I pulled down the repo with a token set up with the correct permissions, here is some initial feedback:

- I think we need to reduce how fast the hovercard shows up, because if you were to try to get to someone in the second or third row, a card in the first role shows up blocking those avatar.
- If you scroll when a card is hovered, it scrolls with you which makes it seem like its been unanchored from the user it’s for. See the video below.

https://github.com/user-attachments/assets/da7e740a-f69c-49d5-83c4-5306bf9566da

app/pages/about.vue
userquin: I need to check floating-vue in Firefox, I use it at vitest ui but with virtual scroll and it is working fine, but here we have 200+ contributors, all my changes have been to not block main thread in Firefox. I will check the scroll behavior in your video (I guess it is related to allow use tap or maybe even missing scroll logic)

userquin:
> * If you scroll when a card is hovered, it scrolls with you which makes it seem like its been unanchored from the user it’s for. See the video below.

Should be fixed with last commit (missing logic)

app/pages/about.vue
```diff
@@ -47,6 +61,244 @@ const roleLabels = computed(
...
+    activeBtnDom.removeAttribute('aria-expanded')
+  }
+  activeBtnDom = btn
+  if (btn) {
+    if (value) {
+      btn.setAttribute('aria-expanded', 'true')
+    } else {
+      btn.removeAttribute('aria-expanded')
+    }
+  }
```
knowler, line 134:
```suggestion
  if (btn) btn.ariaExpanded = value
```

We should set `aria-expanded` to `false` when not expanded so the button is conveyed as a expand/collapse button.

app/pages/about.vue
```diff
@@ -47,6 +61,244 @@ const roleLabels = computed(
...
+  } else {
+    popoverPos.align = 'center'
+    popoverPos.left = center
+  }
+}
+
+// DON'T MOVE aria-expanded to the template, Firefox performance issues
+function setActiveBtnExpanded(btn: HTMLElement | null, value: boolean) {
+  if (activeBtnDom && activeBtnDom !== btn) {
+    activeBtnDom.removeAttribute('aria-expanded')
```
knowler:
```suggestion
    activeBtnDom.ariaExpanded = false
```

This should be `false` and we can use the ARIAMixin property for `Element` here (which allows for boolean values — update: I guess `setAttribute()` does allow for booleans for `aria-expanded` too, but I imagine TypeScript would complain).

app/pages/about.vue
```diff
@@ -47,6 +61,244 @@ const roleLabels = computed(
...
+  const location = c.location
+    ? $t('about.contributors.location', { separator, location: c.location })
+    : ''
+  return $t('about.contributors.view_profile_detailed', {
+    name: c.name || c.login,
+    role,
+    works_at,
+    location,
+  })
+}
```
knowler: Just their name is sufficient for the ARIA label. We don’t even need “View” because the fact its a link or expand/collapse button is indicative of that. We can leave out the “GitHub profile” bit, too. I think this works fine for the link variant as well.

knowler (commented):
I need to do another pass on the focus management, but here are a few comments with regard to some ARIA attributes and labels.

app/pages/about.vue
userquin: changing this...

userquin: FF and Chrome removing the aria-expaned with `btn.ariaExpanded = false` or `btn.ariaExpanded = 'false'`

knowler: This must be Vue or Vite messing with things… in a minimal example, setting the `ariaExpanded` property with booleans is interoperable across browsers: https://knowler.dev/demos/yuWmNjT?codepen

userquin: weird

userquin: we're removing the attribute 🤦‍♂️

PR #1598, opened 2026-02-23
feat(ui): add versions modal permalink

### LINKED ISSUE #1588: Versions modal should have a url just like chart does
Image

Currectly only for charts, for example: https://npmx.dev/package/astro?modal=chart

Should also be for versions

MatteoGabriele:
### 🔗 Linked issue

resolves #1588 

### 🧭 Context

Permalinks have been added to the downloads modal, but this feature has not been included in the versions modal.

Adds a new query parameter when the versions modal opens or closes. This follows the same pattern of the downloads modal.

### 📚 Description

Adds the ability to link the versions modal. Along with the permalink, I also implemented a UI refactor to improve the positioning of elements, date ranges, and toggles. I used a combination of ButtonGroup and ButtonBase, but a custom ToggleGroup component will be needed to reduce code repetition and enhance accessibility.


graphieros:
Looks good to me !
Also fixes the CLS when toggling options 🤗

Edit: I reviewed a few details

app/components/Package/VersionDistribution.vue
```diff
@@ -107,23 +107,90 @@ const sanitise = (value: string) =>
...
+
+const hasMinimap = computed<boolean>(() => {
+  const series = xyDataset.value[0]?.series ?? []
+  return series.length > 6
+})
+
 const chartConfig = computed(() => {
   return {
-    theme: isDarkMode.value ? 'dark' : 'default',
+    theme: 'dark',
```
graphieros: we need to keep the ternary we had before based on isDarkMode, if not, the chart will have a dark theme in all cases

MatteoGabriele: default doesn't seem to exist as a theme, though, which is why I got rid of it. 
Also, themes seem to change only on a full re-render. It happens to the other modal as well.

app/components/Package/VersionDistribution.vue
```diff
@@ -676,4 +645,8 @@ const endDate = computed(() => {
     left: calc(100% + 4rem) !important;
   }
 }
+
+#version-distribution .atom-title {
```
graphieros, line 650: I recommend adding a comment to explain this override (absence of configurable css for the built-in title wrapper)

app/components/Package/VersionDistribution.vue
graphieros: You can set
`theme: isDarkMode.value ? 'dark' : ''`, instead which will be correct.
internally, a named theme that does not exist will default to the 'default' theme. Empty string or `undefined` works, 'default' works, I put it initially as documentation. But empty string or `undefined` is better

graphieros: And it does matter, if you load the chart with light theme, you will notice label colors are for dark mode

graphieros (approved):

gameroman:
I think instead of `All | Low` it should be `Exclude low | All` or something like that

image


Also it currently only adds `?modal=versions`. I think we should also deep link the options, similar to `&granularity` option in chart modal

graphieros:
@gameroman 

I think you can create issues(or PRs 🙂) for this:
- better translations for labels
- including options in the router query

PR #1523, opened 2026-02-15
feat: add TanStack Start to framework list

farisaziz12:
## Context

[TanStack Start](https://tanstack.com/start/latest) as a framework has been gaining popularity (13.5K GH Stars). Based on that, it makes sense to add it to the homepage framework list.

I've picked the `oklch` color based on the `--color-cyan-600` used on the TanStack Start page of the TanStack website.

Screenshot 2026-02-15 at 23 23 32

### Desktop
Screenshot 2026-02-15 at 23 27 09

### Mobile
Screenshot 2026-02-15 at 23 26 56

CC: @tannerlinsley, @tkdodo, @schiller-manuel

PR #1610, opened 2026-02-23
fix: tanstack start package

farisaziz12:
### 🧭 Context

In [this PR](https://github.com/npmx-dev/npmx.dev/pull/1523), we added TanStack to the list of frameworks. In [this comment](https://github.com/npmx-dev/npmx.dev/pull/1523#issuecomment-3941863072), @schiller-manuel pointed out that the referenced package is now considered stale.

### 📚 Description

Updating the referenced package to the current most popular one, `@tanstack/react-start`

### DIFF
--- a/.github/workflows/semantic-pull-requests.yml
+++ b/.github/workflows/semantic-pull-requests.yml
@@ -34,5 +34,10 @@ jobs:
             The subject "{subject}" found in the pull request title "{title}"
             didn't match the configured pattern. Please ensure that the subject
             doesn't start with an uppercase character.
+
+            Examples:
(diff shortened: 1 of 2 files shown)


farisaziz12:
> I already fixed it. the package currently in the code includes more downloads than the react version (as it is depended on by both react and solid versions)

But do we want to reference a sub-dependency of the frameworks or the framework itself?

on the page for `@tanstack/start-server-core`:

> This package is not meant to be used directly. It is a dependency of the TanStack Start framework-specific packages.

imho it could confuse someone to point to something they should not use but references what they should use. WDYT?

PR #1616, opened 2026-02-24
feat: replace homepage frameworks with fun rotating "n p m x" picks

serhalp:
### 🔗 Linked issue

😶 discussed on Discord

### 🧭 Context

- The hardcoded framework package quick list has an unclear purpose. More and more folks are requesting to add their link. It isn't a scalable feature.
- The below is a subtle homage to the OG npmjs.com (they used to show [a random "definition" of "npm"](https://github.com/npm/npm-expansions) on page load).
- It's a fun little easter egg that may lead curious users to wonder how the packages are selected... encouraging them to look at our repo, learn about our atproto features, and join our community 😈.
- It's somewhat of a metaphor for npmx being borne out of the ecosystem.

### 📚 Description

Show 4 randomly-selected packages whose names contain the letters n, p, m, x, with the matching letter highlighted in bold + accent colour ~~(one distinct colour per letter when no user accent is set)~~.

Screenshot 2026-02-26 at 22 13 32
Screenshot 2026-02-26 at 22 14 52
Screenshot 2026-02-26 at 22 16 38

Selection algorithm:
1. Fetch top 500 popular packages from the Algolia search index (empty query, default popularity ranking). Algolia doesn't support filtering by name substring (and doesn't know about our social likes), so we filter these results down after the fact (not deprecated, >=10k downloads/30d).
2. For each letter (n, p, m, x):
    1. Take 30 random candidates whose name contains that letter and check their social like count.
    2. If there are candidates with >=5 community likes, keep only those; otherwise, keep all.
    3. Remove any remaining candidates already assigned to another letter in "n p m x".
    4. Randomly pick one remaining candidate.
    5. If there are no remaining candidates, pick the hardcoded default for this letter (nuxt, pnpm, module-replacements, oxfmt).

Results are cached for 1 hour via with SWR, so all users see the same picks for about an hour, and no user ever experiences a cache miss (and Algolia/constellation slowness).


graphieros:
Since 4 packages are displayed, I think the original width of the wrapper could be restored (https://github.com/npmx-dev/npmx.dev/pull/1591 reduced max-w to wrap the long list)

server/api/picks.get.ts
```diff
@@ -0,0 +1,106 @@
...
+  LIKES_SAMPLE_SIZE,
+  selectPicks,
+  type NpmxLetter,
+  type PickCandidate,
+} from '../utils/picks'
+
+const PICKS_MAX_AGE_MS = 60 * 60 * 1000
+
+/** Pick `n` random items from `arr` (Fisher-Yates on a copy, sliced). */
+function randomSample<T>(arr: T[], n: number): T[] {
```
serhalp, line 15: Didn't seem worth pulling in a dependency for this. AFAIK this is as simple as it gets without caring about cryptographically secure randomness or anything like that.

test/fixtures/mock-routes.cjs
```diff
@@ -549,6 +568,7 @@ const routes = [
     pattern: 'https://constellation.microcosm.blue/**',
     match: matchConstellationApi,
   },
+  { name: 'Algolia API', pattern: 'https://*-dsn.algolia.net/**', match: matchAlgoliaApi },
```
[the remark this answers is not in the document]
serhalp: I think this is fine for a test mock...

test/unit/server/utils/picks.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    // Run multiple times — liked should always win for 'n'
+    for (let i = 0; i < 20; i++) {
+      const picks = selectPicks(candidates)
+      expect(picks[0]!.name).toBe('liked-n')
+    }
+  })
+
+  it('falls back to non-liked when no liked candidates exist', () => {
+    const candidates: Record<NpmxLetter, PickCandidate[]> = {
+      n: [{ name: 'nodemon', totalLikes: 2 }],
```
trueberryless, line 59: Small nitpick here: But if `MIN_LIKES` is equal to 2, then this test case seems misleading. I'm not sure if it would be cleaner, but maybe `MIN_LIKES - 1` could make this test case more robust? Open for ideas 💡

server/api/picks.get.ts
trueberryless: But what if npmx.dev gets hacked because of those picks 🤯 
JK of course 😆 

Awesome PR btw 🚀

alexdln:
image

image

On home page

Also I always see next, why? :(

server/api/picks.get.ts
```diff
@@ -0,0 +1,109 @@
...
+  const copy = [...arr]
+  for (let i = copy.length - 1; i > 0; i--) {
+    const j = Math.floor(Math.random() * (i + 1))
+    const tmp = copy[i]!
+    copy[i] = copy[j]!
+    copy[j] = tmp
+  }
+  return copy.slice(0, n)
```
alexdln, line 23: I have a strong feeling there are some errors here. Can't we just do

for i =0; i < 4 {
get a random number based on the array length -> remove it from the array -> add it to the resulting array

server/api/picks.get.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    return { picks }
+  },
+  {
+    maxAge: PICKS_MAX_AGE_MS / 1000,
+    swr: true,
+    name: 'npmx-picks',
+    getKey: () => 'picks',
+  },
+)
```
alexdln, line 109: So complicated. Maybe we should generate it once per build? It will also fix hydration errors

PR #1619, opened 2026-02-24
fix: address CLS in downloads modal

### LINKED ISSUE #1342: CLS when switching facets in trends modal
Follow-up to #1310

graphieros:
Resolves #1342

- Update the placeholder's aspect ratio
- Set a min-h for the chart wrapper on desktop

Also fixes CLS introduced with the addition of the animation toggle:
- Add a skeleton (visible when prefers-reduced-motion has no-preference)
- Set a fixed height on the weekly downloads section of the side bar (shorter when prefers-reduced-motion is reduced, since the toggle is not displayed in this case)

### DIFF
--- a/app/components/Package/TrendsChart.vue
+++ b/app/components/Package/TrendsChart.vue
@@ -1685,7 +1685,7 @@ watch(selectedMetric, value => {
     <div
       role="region"
       aria-labelledby="trends-chart-title"
-      :class="isMobile ? 'min-h-[260px]' : 'min-h-[567px]'"
+      :class="isMobile === false && width > 0 ? 'min-h-[567px]' : 'min-h-[260px]'"
     >
       <ClientOnly v-if="chartData.dataset">
         <div :data-pending="pending" :data-minimap-visible="maxDatapoints > 6">


app/components/Package/WeeklyDownloadStats.vue
```diff
@@ -264,7 +264,7 @@ const config = computed<VueUiSparklineConfig>(() => {
 </script>
 
 <template>
-  <div class="space-y-8">
+  <div class="space-y-8 h-[110px] motion-safe:h-[140px]">
```
RYGRIT, line 267: Hi @graphieros, I've noticed that this change causes a large blank area after the "Weekly Downloads" section collapses.

Image

I'm considering moving the height limit (`min-h-[110px] motion-safe:min-h-[140px]`) into the inner container of `CollapsibleSection`, but I'm unsure if this complies with the CLS specification.

If it does, please let me know, and I'd be happy to submit a follow-up PR!

graphieros: Yes please :)

Issue #1510, opened 2026-02-14
Non-text contrast is not sufficient (almost everywhere)


knowler:
We’re not sufficiently meeting [Success Criterion 1.4.11 (Level AA) of WCAG 2.2](https://www.w3.org/TR/WCAG22/#non-text-contrast) which states:

> The visual [presentation](https://www.w3.org/TR/WCAG22/#dfn-presentation) of the following have a [contrast ratio](https://www.w3.org/TR/WCAG22/#dfn-contrast-ratio) of at least 3:1 against adjacent color(s):
> 
> <dl><dt>User Interface Components</dt>
>   <dd>
>
>   Visual information required to identify [user interface components](https://www.w3.org/TR/WCAG22/#dfn-user-interface-components) and [states](https://www.w3.org/TR/WCAG22/#dfn-states), except for inactive components or where the appearance of the component is determined by the [user agent](https://www.w3.org/TR/WCAG22/#dfn-user-agents) and not modified by the author;
> 
> </dd>
> <dt>Graphical Objects</dt>
>   <dd>
>
>   Parts of graphics required to understand the content, except when a particular presentation of graphics is [essential](https://www.w3.org/TR/WCAG22/#dfn-essential) to the information being conveyed.
> 
> </dd>

This issue has come up in the audit (#1259) and I wanted to surface it early (and in a little more detail). We’re failing for “User Interface Components” as the visual elements such as borders or backgrounds used to distinguish user interface components from their surrounding visual context or to indicate a particular state is not sufficient.

Eric Eggert has a very good blog post on both how to interpret this success criterion and how to practically meet this requirement: [“In detail: 1.4.11 Non-Text Contrast (User Interface Components)”](https://yatil.net/blog/non-text-contrast-in-detail-ui-components).

## What’s affected (work-in-progress)

My audit has only covered the home page so far, but I do know from glancing at other pages, they are affected. I’ll fill out this list as I find specific changes we need to make.

- [ ] Hover state for the “compare” and “settings” links, as well as, the “connect” menu button.
  - Dark theme contrast ratio: 1.274:1 (`#272727` / `#101010`)
  - Light theme contrast ratio: 1.248:1 (`#E6E6E6` / `#FFFFFF`)
[...]

PR #1625, opened 2026-02-24
fix(a11y): improve non-text (border) colour contrast

knowler:
A step towards resolving #1510, but it does have a wide effect on a lot of components (e.g. search boxes, buttons, etc.) that would be failing (and causes them to pass).

As stated in #1510, we are currently failing [WCAG 2.2 SC 1.4.11 (Level AA)](https://www.w3.org/WAI/WCAG22/quickref/#non-text-contrast) which requires a minimum non-text colour ratio of 3:1 for user interface components and graphical objects. This was most notable for the search field and buttons.

This PR increases the lightness channels of the `--border` and `--border-hover` values for the light and dark colour schemes which in effect causes their contrast on the main background to just pass 3:1. The original issue did include the ratios of the borders and the inner surface (i.e. the background colour of the component itself). For buttons, I don’t think this is necessary. For text fields, this maybe necessary unless there are other visual elements within the border (e.g. our main search field has the leading `./`).

Admittedly, just increasing the border contrast alone isn’t a great design decision. Adding other visual effects could help, but it would be a bit of a departure from the current design paradigm. There’s also the issue of pairing these border colours with different background colours—there could be failures.

| Before | After |
|-|-|
| The original, dark mode, cropped view of the main logo and search on the homepage. Contrast ratio: 1.257:1 (`#262626` / `#101010`) | The increased-contrast, dark mode, cropped view of the main logo and search on the homepage. Contrast ratio: 3.167:1 (`#636363` / `#101010`) |
| The original, light mode, cropped view of the main logo and search on the homepage. Contrast ratio: 1.574:1 (`#CECECE` / `#FFFFFF`) | The increased-contrast, light mode, cropped view of the main logo and search on the homepage. Contrast ratio: 3.033:1 (`#949494` / `#FFFFFF`) |

<details><summary>Full page comparison of home and package views</summary>

Ignore other non-border-contrast visual differences. I’ve got screenshots of different builds. This is tedious work.

| Before | After |
|-|-|
| The original, dark-mode, home page, full screenshot | The increased-contrast, dark-mode, home page, full screenshot |
| The original, light-mode, home page, full screenshot | The increased-contrast, light-mode, home page, full screenshot |
| The original, dark-mode, package page, full screenshot | The increased-contrast, dark-mode, package page, full screenshot |
| The original, light-mode, package page, full screenshot | The increased-contrast, light-mode, package page, full screenshot |

</details>

### DIFF
--- a/app/assets/main.css
+++ b/app/assets/main.css
@@ -19,9 +19,9 @@
   --fg-subtle: oklch(0.673 0 0);
 
   /* border, separator colors */
-  --border: oklch(0.269 0 0);
+  --border: oklch(0.5 0 0);
   --border-subtle: oklch(0.239 0 0);
-  --border-hover: oklch(0.371 0 0);
+  --border-hover: oklch(0.6 0 0);
 
   /* accent color, set by user from settings */
   --accent: var(--accent-color, oklch(1 0 0));
@@ -90,9 +90,9 @@
   --fg-muted: oklch(0.398 0 0);
   --fg-subtle: oklch(0.48 0 0);
 
-  --border: oklch(0.8514 0 0);
+  --border: oklch(0.666 0 0);
   --border-subtle: oklch(0.922 0 0);
-  --border-hover: oklch(0.715 0 0);
+  --border-hover: oklch(0.6 0 0);
 
   --accent: var(--accent-color, oklch(0.145 0 0));
   --accent-muted: var(--accent-color, oklch(0.205 0 0));


alexdln:
I think we definitely need to work on the contrast and visibility of inputs and buttons (_and all interactive elements_). But I don't think we need to extend this to the entire interface. Unfortunately, I can't understand the wording of their expectations for background elements:

> User Interface Components: Visual information required to identify user interface components and states, except for inactive components or where the appearance of the component is determined by the user agent and not modified by the author;
> 
> Graphical Objects: Parts of graphics required to understand the content, except when a particular presentation of graphics is essential to the information being conveyed.

Now the interface looks like a terminal from the 2000s or, as some would say, like a tech newspaper...

How about to add a `border-contrast` variable and apply it only to the interactive elements themselves?

alexdln:
@knowler wdyt? Or can we close this PR for now?

knowler:
@alexdln:

> Now the interface looks like a terminal from the 2000s or, as some would say, like a tech newspaper...

ya, it doesn’t look great without additional changes.

> How about to add a `border-contrast` variable and apply it only to the interactive elements themselves?

Something like this or a semantic token like `border-ui`.

> wdyt? Or can we close this PR for now?

Let’s close this for now. I no longer feel a sense of urgency in addressing this, as the alpha launch has come to pass. This was mostly meant to be an accessibility hotfix that we could have made a few iterative tweaks upon before launch. I think we can factor these changes into a broader design discussion.

Issue #1416, opened 2026-02-12
Add possibility to search for specific version of a package


chronoB:
When using pnpm in a project it outputs the packages that had ignored build scripts with specific versions:
```
╭ Warning ────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                 │
│   Ignored build scripts: @parcel/watcher@2.5.6, esbuild@0.25.12, esbuild@0.25.9, less@4.5.1,    │
│   lmdb@3.4.2, msgpackr-extract@3.0.3.                                                           │
│   Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.        │
│                                                                                                 │
╰─────────────────────────────────────────────────────────────────────────────────────────────────╯
```

I tried to copy the package including the version into the search field but it didn't find the package.

**What I would like to happen:**
When adding a version to the search string and it is found for the corresponding package I want to open the package page with the specified version preselected

Codefoxdev:
It is currently not possible to search for version directly using the npm or algolia search provider. However, we could add client-side filtering for the specific version to the search results?

For fetching the results, we would just query the package itself, and then at the client side filter the results to only include packages that satisfy the specified version.

It would at least already be an improvement if we strip the version info from the search results, as currently the search gives completely inaccurate results when searching with the version info (https://npmx.dev/search?q=nuxt@3.4.1)

chronoB:
Yeah I think the minimal improvement of stripping the version would be a good starting point.

I also thought about just stripping the version for the search and then appending the version when opening the package page e.g.:
- Searching for nuxt@3.4.1
- Search will be done for nuxt
- When opening the package page the version will be appended: (https://npmx.dev/package/nuxt/v/3.4.1 instead of https://npmx.dev/package/nuxt)

There would be a problem if the version does not exist because it only shows a 404 and does not fallback to latest (which is probably a good thing but sad to see in this specific use case).

I don't know if it's easily possible to filter the results client side because I don't exactly know how much information is send to the client when searching for a package. But if the information is sufficient my workaround would also work, because we made sure that the version does exist so no 404 is possible.

Codefoxdev:
Exactly, that is why I proposed client side filtering. Version info is definitely present, however, I don't know if that requires another fetch call for EACH package yet, but I am currently looking into it.

Codefoxdev:
[...]

PR #1626, opened 2026-02-24
feat: strip version info from the search query

Codefoxdev:
### 🔗 Linked issue

Partially addresses #1416

### 🧭 Context

Currently, when searching for a package along with version info (e.g. when copying a package name from the output of a package manager) it just returns random results, instead of useful ones. (e.g. https://npmx.dev/search?q=nuxt@3.4.1)

This pr strips the version info from the query to improve search results.

### 📚 Description

This pr implements a new regex query to extract essential date from the search query on the search page, such as package scope (which replaces an existing method), package specifier and (optionally) version info. The query is then reconstructed without version info to support the existing search api. It can currently parse a package name in the following format:
```
@scope/specifier@version
```

An object containing the separate parts of the query is exposed as the `parsedQuery` ref and it is able to be used for other features. (Such as implementing the other part of the linked issue)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -452,7 +452,7 @@ The command palette is a first-class navigation surface. When you add a new user
 
 ### Cursor and navigation
 
-**npmx** uses `cursor: pointer` only for links to match users’ everyday experience. For all other interactive elements, including buttons, use the default cursor (_or another appropriate cursor to indicate state_).
+**npmx** uses `cursor: pointer` for links and buttons to match users’ everyday experience. For all other interactive elements, use the default cursor (_or another appropriate cursor to indicate state_).
 
 > [!NOTE]
 > A link is any element that leads to another content (_go to another page, authorize_)
--- a/app/components/Button/Base.vue
+++ b/app/components/Button/Base.vue
@@ -44,7 +44,7 @@ defineExpose({
 <template>
   <button
     ref="el"
-    class="group gap-x-1 items-center justify-center font-mono border border-border rounded-md transition-all duration-200 disabled:(opacity-40 cursor-not-allowed border-transparent)"
+    class="group gap-x-1 items-center justify-center font-mono border border-border rounded-md transition-all duration-200 cursor-pointer disabled:(opacity-40 cursor-not-allowed border-transparent)"
     :class="{
       'inline-flex': !block,
       'flex': block,
--- a/app/components/Compare/FacetBarChart.vue
+++ b/app/components/Compare/FacetBarChart.vue
@@ -152,6 +152,7 @@ const config = computed<VueUiHorizontalBarConfig>(() => {
           })
         },
       },
+      useCursorPointer: true,
     },
     skeletonDataset: skeletonDataset.value,
     skeletonConfig: {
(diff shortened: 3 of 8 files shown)


app/pages/search.vue
```diff
@@ -33,6 +33,29 @@ const updateUrlPage = debounce((page: number) => {
...
+  if (!match) return { scope: null, name: q, version: null, strippedQuery: q }
+
+  const { scope, specifier, version, trailing } = match.groups ?? {}
+  // Reconstruct the query without the version info, essentially stripping the version data:
+  // anything directly after the @ for the version specifier is stripped.
+  const name = scope ? `@${scope}/${specifier}` : (specifier ?? '')
+  const strippedQuery = `${name} ${trailing ?? ''}`.trim()
+
+  return { scope: scope ?? null, name: name, version: version || null, strippedQuery }
+})
```
gameroman: Might be a good idea to extract this into a util and add a unit test for it

Codefoxdev: I have extracted this into a util file and added some basic unit tests in the new commits

Codefoxdev:
There are maybe some places where this util can be used instead of existing implementations. 
The most prominent example is in `modules/runtime/server/cache.ts` with the `parseScopedPackageWithVersion` function, which implements a part of this method's functionality.

app/pages/search.vue
```diff
@@ -33,6 +33,16 @@ const updateUrlPage = debounce((page: number) => {
...
 const query = computed(() => searchQuery.value)
 
+const {
+  scope: packageScope,
+  name: packageName,
+  trailing: queryTrailing,
+} = useParsedSearchQuery(query)
+
+const versionStrippedQuery = computed(() =>
+  `${packageName.value} ${queryTrailing.value ?? ''}`.trim(),
```
gameroman: Looking at the tests, this probably doesn't need an extra space since the `queryTrailing` seems to have it

```suggestion
  `${packageName.value}${queryTrailing.value ?? ''}`.trim(),
```

gameroman (approved):
Looks good overall 👍

app/pages/search.vue
Codefoxdev: Yep, already fixed it

Codefoxdev:
@gameroman Should I also update the similar existing implementations to use this new method?

gameroman:
> @gameroman Should I also update the similar existing implementations to use this new method?

If there is something that you think could use this util, that would be good👍

Codefoxdev:
Actually, I dont think that anything should be changed. The example I mentioned lives outside of the app scope, and the behaviour is slightly different from this util. So I think this pr can be merged.

gameroman:
> Actually, I dont think that anything should be changed. The example I mentioned lives outside of the app scope, and the behaviour is slightly different from this util. So I think this pr can be merged.

Thank you, could you resolve the merge conflict, please?

Codefoxdev:
Sorry, I accidentally messed up trying to resolve the conflicts on the go. I will try to fix it later, but I currently do not have access to my pc.

Codefoxdev:
I fixed the removed commits on my fork, but I cannot seem to reopen the pull request, can someone help?

gameroman:
@Codefoxdev you can probably make a new branch, re-apply the changes, and make a new PR

PR #1629, opened 2026-02-24
fix: experimentally swap napi-rs/canvas to canvas

wojtekmaj:
This is an EXPERIMENTAL pull request to see if swapping the libs would fix the issue where badges text is not measured correctly.

Continuation of #1584


wojtekmaj:
Crashing, no logs :(

PR #1633, opened 2026-02-24
fix(i18n): add missing French UI strings + update some others

ArmandPhilippot:
### 🔗 Linked issue

None

### 🧭 Context

Some UI strings were missing in French. Some others were using English words, at best "Frenchified"... might be better to use a proper translation.

### 📚 Description

* Translates the missing French UI strings... Well, I don't know where `distribution_range_date_same_year` or `grouping_usage_low` are used in the UI, so I'm not 100% sure this is correct.
* Fixes a few French UI strings:
  * `meta_description`: this was grammatically wrong ("votre... préférences", ie. singular/plural)
  * `asset` is not a French word
  * `refs`... might seem strange alongside the others (full word vs English abbreviation) so I replaced it with the full word
  * `scopé` is not a word... I'm not sure everyone understand what this means. I think the appropriate translation here would be "portée limitée".

### DIFF
--- a/i18n/locales/fr-FR.json
+++ b/i18n/locales/fr-FR.json
@@ -79,7 +79,7 @@
   "settings": {
     "title": "paramètres",
     "tagline": "personnalisez votre expérience npmx",
-    "meta_description": "Personnalisez votre expérience npmx.dev avec votre thème, langage et préférences d'affichage.",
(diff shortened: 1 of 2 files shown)


graphieros (approved):
Perfect

PR #1635, opened 2026-02-25
chore: update play/pause button of trend to be an icon

jycouet:
### 🔗 Linked issue

Maybe a follow up on https://github.com/npmx-dev/npmx.dev/issues/1539

### 🧭 Context

"Pause animation" & "Play animation" is rather big on this screen, maybe icons with these alt text can work better.

### 📚 Description

image

Feel free to discard.


RYGRIT:
I prefer the effect of the icons.

wojtekmaj:
Much better. @jycouet Could you please rebase?

app/components/Package/WeeklyDownloadStats.vue
```diff
@@ -282,46 +282,61 @@ const config = computed<VueUiSparklineConfig>(() => {
...
+              size="small"
+              class="hidden motion-safe:inline-flex absolute top-0 inset-ie-0 !p-1 !border-0 !bg-transparent hover:!bg-transparent text-fg-subtle hover:text-fg transition-colors duration-200 focus-visible:outline-accent/70 rounded"
+              :title="
                 hasSparklineAnimation
                   ? $t('package.trends.pause_animation')
                   : $t('package.trends.play_animation')
-              }}
+              "
+              :classicon="hasSparklineAnimation ? 'i-lucide:pause' : 'i-lucide:play'"
+              :aria-pressed="hasSparklineAnimation"
```
knowler:
```suggestion
```

We don’t need to set `aria-pressed` since we’re already updating the label (and updating the label is generally better for this specific kind of button). From [Sarah Higley’s ”Playing with state”](https://sarahmhigley.com/writing/playing-with-state/):

> As a final note, never change both the name and `aria-pressed` state in tandem. That way, confusion lies. Just imagine trying to parse the meaning of "play button, on" vs. "pause button, off".

app/components/Package/WeeklyDownloadStats.vue
```diff
@@ -282,46 +282,61 @@ const config = computed<VueUiSparklineConfig>(() => {
...
+
+            <ButtonBase
+              v-if="hasWeeklyDownloads"
+              type="button"
+              size="small"
+              class="hidden motion-safe:inline-flex absolute top-0 inset-ie-0 !p-1 !border-0 !bg-transparent hover:!bg-transparent text-fg-subtle hover:text-fg transition-colors duration-200 focus-visible:outline-accent/70 rounded"
+              :title="
                 hasSparklineAnimation
                   ? $t('package.trends.pause_animation')
                   : $t('package.trends.play_animation')
```
knowler: (Optional for this PR… eventually we’ll fix globally)

A tooltip whose contents are the label for this button (i.e. associated with `aria-labelledby`) would be better. This is a pattern I’d like to see us change across the board (eventually I’ll file an issue about removing `title` pretty much everywhere). I’ll note that I think our tooltip implementation is also flawed (i.e. I’m not sure if we add `role=tooltip` and it doesn’t meet the dismissable requirement of [1.4.13 Content on Hover or Focus Level AA](https://www.w3.org/WAI/WCAG22/quickref/#content-on-hover-or-focus)).

Don’t worry about solving that here. If it’s easier to just leave `title` than use our existing flawed Tooltip solution, we can always circle back to fix it when we address the more pervasive problem (I also don’t have time to work out what the exact solution we need to use everywhere quite yet).

Further reading: 
- [“The Trials and Tribulations of the Title Attribute” by Scott O’Hara](https://www.24a11y.com/2017/the-trials-and-tribulations-of-the-title-attribute/)
- [“Tooltips in the time of WCAG 2.1” by Sarah Higley](https://sarahmhigley.com/writing/tooltips-in-wcag-21/)

app/components/Package/WeeklyDownloadStats.vue
```diff
@@ -282,46 +282,61 @@ const config = computed<VueUiSparklineConfig>(() => {
...
-
-          <div v-if="hasWeeklyDownloads" class="hidden motion-safe:flex justify-end p-1">
-            <ButtonBase size="small" @click="toggleSparklineAnimation">
-              {{
+              </template>
+            </ClientOnly>
+
+            <ButtonBase
+              v-if="hasWeeklyDownloads"
+              type="button"
```
knowler: This is redundant, we already add `type=button` in the component.

```suggestion
```

knowler (changes requested):
I’m fine with using icons for this. A few thoughts on the icons and then I have some other suggestions as comments…

- Could we increase the hit target to at least be 24 by 24 pixels as per [2.5.8 Target Size (Minimum) Level AA](https://www.w3.org/WAI/WCAG22/quickref/#target-size-minimum)? Right now it’s 20 by 20 pixels.
- It’s pretty close to the button above which might make it easy to hit the wrong thing.
- Could we add a border to the button? I’m finding that the outlined pause icon is hard to understand that it means pause and that it’s a button. A filled one would be better if it’s available, in my opinion.

jycouet:
Thx for all these info @knowler 
I took care of some points that were easy to action and believe that it's an improvement from today's state

graphieros:
After a compromise found in chat, it was decided to run the animation only once, and remove the play/pause button altogether (#1818 ).

jycouet:
I like a lot the animation. :/
But I guess some people know better

graphieros:
> I like a lot the animation. :/ But I guess some people know better

I know, me too !
But don't worry, we'll probably be able to get it back on a loop, when settings will be improved to have this kind of animations opt-in.

knowler:
Thank you for your work on this, @jycouet!

PR #1636, opened 2026-02-25
feat: fix known download anomalies with interpolation

### LINKED ISSUE #1241: clean up npm downloads data for graphs
## removing outliers

See for example [npmx.dev/package/vite](https://npmx.dev/package/vite)

Image

Probably a rogue CI created the spike you see here in the graph. This isn't significant information, and destroys the scale of the graph.

We should do a clean step so we can remove these kind of spikes. There are several packages that are ruined by them, more when doing full-history graphs. It may require some tweaking, and we may need to reach to some data experts here to help us but the big majority of these spikes are clearly outliers that should be easy to remove.

We could do a few things after the cleanup:
1. remove that data points and not show them
2. extrapolate in the missing region so we get a full graph
3. extrapolate using a dotted line so users know that this part of the graph is estimated

I think we should go with the last option.

## remove zeroed data

Another issue in npm data is that sometimes it gets borked for a few days. During these periods, sometimes lasting a few days, npm returns zero. See for example the [Nov 9, 2024 incident](https://bsky.app/profile/francoisbest.com/post/3lahwietghc2q). We can extrapolate when the data is zerod out.

jycouet:
### 🔗 Linked issue

Closes #1241

### 🧭 Context

Having nicer charts in general without data going off!

Right now, this PR is the Option 4 with these defaults: 
```json
chartFilter: {
  averageWindow: 0,
  smoothingTau: 1,
  anomaliesFixed: true,
},
```

FYI, I split commits to be able to go back to other options, if you want to try on some packages :)

#### 1/ Hampel

https://github.com/user-attachments/assets/8700621e-965d-44f1-9a99-aeb4b9c0e5ef

Was good, be it's a lot of variables... And it's hard to get all situations correct

#### 2/ Outlier Sensitivity

https://github.com/user-attachments/assets/25f9bf7b-4d4c-422c-b190-c37d9943366c

Same thing... I tried with some linear reduction & some tweaks to keep first an last point. Good for some... not the bast for others...

#### 3/ 2 Passes Avg or Smoothing

https://github.com/user-attachments/assets/c0498a87-cdfb-41fb-8835-cfc018410951

Std Average / Smoothing are not so good because they change the start or end. But here in the 2 passes process we minimise this effect.

#### 4/ 2 Passes Avg or Smoothing & Manual tuning

https://github.com/user-attachments/assets/19b40385-3057-451f-8d48-4a02362cff0b

This is actually the best! Manually declaring where it's wrong. In the vite example it's obvious, so we can just add it. And the community is soooo into it that we could encourage people to enter things there _(we could even offer an API at some point)_.

Keeping the option for Avg & Smooth to have a good mix.

---

What do you think?!

### DIFF
--- a/app/components/Package/TrendsChart.vue
+++ b/app/components/Package/TrendsChart.vue
@@ -1637,11 +1637,14 @@ const packageAnomalies = computed(() => getAnomaliesForPackages(effectivePackage
 const hasAnomalies = computed(() => packageAnomalies.value.length > 0)
 
 function formatAnomalyDate(dateStr: string) {
-  return new Date(dateStr).toLocaleDateString(locale.value, {
+  const [y, m, d] = dateStr.split('-').map(Number)
+  if (!y || !m || !d) return dateStr
+  return new Intl.DateTimeFormat(locale.value, {
     year: 'numeric',
     month: 'short',
     day: 'numeric',
-  })
+    timeZone: 'UTC',
+  }).format(new Date(Date.UTC(y, m - 1, d)))
 }
 
 // Trigger data loading when the metric is switched
@@ -1781,7 +1784,11 @@ watch(selectedMetric, value => {
             >
               {{ $t('package.trends.known_anomalies') }}
               <TooltipApp interactive :to="inModal ? '#chart-modal' : undefined">
-                <span class="i-lucide:info w-3.5 h-3.5 text-fg-muted cursor-help" tabindex="0" />
+                <button
+                  type="button"
+                  class="i-lucide:info w-3.5 h-3.5 text-fg-muted cursor-help"
+                  :aria-label="$t('package.trends.known_anomalies')"
+                />
                 <template #content>
                   <div class="flex flex-col gap-3">
                     <p class="text-xs text-fg-muted">
@@ -1794,16 +1801,22 @@ watch(selectedMetric, value => {
                       <ul class="text-xs text-fg-subtle list-disc list-inside">
                         <li v-for="a in packageAnomalies" :key="`${a.packageName}-${a.start}`">
                           {{
-                            $t('package.trends.known_anomalies_range', {
-                              start: formatAnomalyDate(a.start),
-                              end: formatAnomalyDate(a.end),
-                            })
+                            isMultiPackageMode
+                              ? $t('package.trends.known_anomalies_range_named', {
+                                  packageName: a.packageName,
+                                  start: formatAnomalyDate(a.start),
+                                  end: formatAnomalyDate(a.end),
+                                })
+                              : $t('package.trends.known_anomalies_range', {
+                                  start: formatAnomalyDate(a.start),
+                                  end: formatAnomalyDate(a.end),
+                                })
                           }}
                         </li>
                       </ul>
                     </div>
                     <p v-else class="text-xs text-fg-muted">
-                      {{ $t('package.trends.known_anomalies_none') }}
+                      {{ $t('package.trends.known_anomalies_none', effectivePackageNames.length) }}
                     </p>
                     <div class="flex justify-end">
                       <LinkBase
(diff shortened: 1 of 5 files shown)


graphieros:
@jycouet 

This looks awesome :)
I do like the last option a lot.

A few remarks before I take a look at the code, regarding the labels for the range inputs. 
I think the labels should either be more explicit about what they do:

- Summary label: 'Filters' is  probably too broad. Maybe something more precise like "Data correction controls" could be more explicit ?
- Same for the labels of the range inputs, "Average window" perhaps ?  "Smoothing" is fine, maybe something like "Noise reduction" ?

Also, perhaps an info icon with a tooltip could explain what the checkbox does (and only showing the checkbox when the 'database' contains corrections ?)

Finally, I could not test on a newly created package that has only one download datapoint.
Edit: did test, works fine :)

app/utils/chart-data-correction.ts
graphieros, line 1: I would suggest renaming this file to something like 'chart-data-correction'

app/utils/chart-filters.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export interface ChartFilterSettings {
+  averageWindow: number
+  smoothingTau: number
+}
+
+/**
+ * Applies moving average then smoothing in sequence.
+ */
+export function applyDownloadFilter<T extends { value: number }>(
```
graphieros: I would suggest renaming this function to something like `applyDataCorrection`

app/utils/download-anomalies.ts
```diff
@@ -0,0 +1,109 @@
...
+    case 'weekly':
+      return weeklyValue
+    case 'monthly':
+      return Math.round((weeklyValue / 7) * 30)
+    case 'yearly':
+      return Math.round((weeklyValue / 7) * 365)
+  }
+}
+
+export function applyBlocklistFilter(
```
graphieros: I would suggest renaming this function to something like `applyBlocklistCorrection`

app/components/Package/WeeklyDownloadStats.vue
```diff
@@ -177,8 +179,18 @@ watch(
   () => loadWeeklyDownloads(),
 )
 
+const filteredDownloads = computed<WeeklyDataPoint[]>(() => {
```
graphieros: I would suggest renaming this var to something like `correctedDownloads`

app/utils/download-anomalies.ts
```diff
@@ -0,0 +1,109 @@
...
+  data: EvolutionData,
+  packageName: string,
+  granularity: ChartTimeGranularity,
+): EvolutionData {
```
graphieros: It would probably be better to have the 3 params as an object, to have call sites more readable

danielroe (approved):
this looks amazing!

@graphieros I'll let you press the merge button when you're happy 🙏

jycouet:
I'll try to address comments tomorrow morning 🤞

app/components/Package/TrendsChart.vue
```diff
@@ -1598,6 +1630,20 @@ const chartConfig = computed<VueUiXyConfig>(() => {
...
+
+const packageAnomalies = computed(() => getAnomaliesForPackages(effectivePackageNames.value))
+const hasAnomalies = computed(() => packageAnomalies.value.length > 0)
+
+function formatAnomalyDate(dateStr: string) {
+  return new Date(dateStr).toLocaleDateString(locale.value, {
+    year: 'numeric',
+    month: 'short',
+    day: 'numeric',
+  })
```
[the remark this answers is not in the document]
graphieros: @jycouet wdyt ?

jycouet:
> Looks awesome to me ! You can probably address the coderabbit critical notification. One last detail, regarding the tooltip content when the chart is used for multiple series in the compare page, the translation could use plural management.

haha, I didn't know this chart! 
Thx for letting me know, I'm on it now.

I'll look at the rabbit as well. _(I have to say that this is a bit overwhelming)_

PR #1643, opened 2026-02-25
perf: cache rendered payloads

### LINKED ISSUE #1194: Repeated reloadings shortly shows error page
This is not critical issue but when you hold down the F5 (or Crtl + R etc.) to reload page repeatedly, the app shows 500 internal error page only for a short of time for some reason. 🤔 

screenshot of error - 500 Internal server error - error loading dynamically imported module: https://npmx.dev/_nuxt/zBfOPwis.js

### LINKED ISSUE #1534: Linking to specific version of package returns 404

### LINKED ISSUE #1558: Hydration mismatch on package pages

danielroe:
### 🔗 Linked issue

resolves #1194
resolves #1534
resolves #1558

### 📚 Description

this aims to implement a payload cache so we don't have to render 2x - once for HTML and once for JSON.

### DIFF
--- a/server/plugins/payload-cache.ts
+++ b/server/plugins/payload-cache.ts
@@ -119,9 +119,10 @@ export default defineNitroPlugin(nitroApp => {
 
     if (isPayloadRequest) {
       // This was a _payload.json render — cache the response body directly
+      if (typeof response.body !== 'string') return
       const routePath = getRouteFromPayloadUrl(ctx.event.path)
       cachePayload(ctx.event, routePath, {
-        body: response.body as string,
+        body: response.body,
         statusCode: response.statusCode ?? 200,
         headers: {
           'content-type': 'application/json;charset=utf-8',
@@ -133,7 +134,8 @@ export default defineNitroPlugin(nitroApp => {
       // stashed a serialized payload on the event context
       const cachedPayload = ctx.event.context._cachedPayloadResponse
       if (cachedPayload) {
-        const routePath = ctx.event.path === '/' ? '/' : ctx.event.path.replace(/\/$/, '')
+        const pathWithoutQuery = ctx.event.path.replace(/\?.*$/, '')
+        const routePath = pathWithoutQuery === '/' ? '/' : pathWithoutQuery.replace(/\/$/, '')
         cachePayload(ctx.event, routePath, cachedPayload)
         // Clean up the stashed payload
         delete ctx.event.context._cachedPayloadResponse
--- a/app/pages/package/[[org]]/[name].vue
+++ b/app/pages/package/[[org]]/[name].vue
@@ -271,7 +271,7 @@ const isHydratingWithServerContent = shallowRef(
 // HTML as a static snapshot while data refetches, avoiding any visual flash.
 const serverRenderedHtml = shallowRef<string | null>(
   isHydratingWithServerContent.value
-    ? (document.querySelector('article')?.innerHTML ?? null)
+    ? (document.getElementById('package-article')?.innerHTML ?? null)
     : null,
 )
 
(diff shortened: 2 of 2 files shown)


server/plugins/payload-cache.ts
```diff
@@ -0,0 +1,177 @@
...
+      }
+    }
+  })
+
+  // -------------------------------------------------------------------------
+  // render:response — Cache payloads after rendering
+  // -------------------------------------------------------------------------
+  nitroApp.hooks.hook('render:response', (response, ctx) => {
+    // Don't cache error responses
+    if (response.statusCode && response.statusCode >= 400) return
```
alexdln: In what situation do we not know the statusCode? Maybe we should:


```suggestion
    if (!response.statusCode || response.statusCode >= 400) return
```

server/plugins/payload-cache.ts
```diff
@@ -0,0 +1,177 @@
...
+      // This was an HTML render for an ISR route — check if the Nuxt plugin
+      // stashed a serialized payload on the event context
+      const cachedPayload = ctx.event.context._cachedPayloadResponse
+      if (cachedPayload) {
+        const routePath = ctx.event.path === '/' ? '/' : ctx.event.path.replace(/\/$/, '')
+        cachePayload(ctx.event, routePath, cachedPayload)
+        // Clean up the stashed payload
+        delete ctx.event.context._cachedPayloadResponse
+      }
+    }
```
alexdln, line 143: What if one user requests the page directly, while the other one prefetches the JSON?

It seems there's a chance we could write different values ​​in this scenario, thereby replicating the current problem. We could use a map with active promises, but I don't know if we have the ability to write to it at the start of a request (and whether it's worth it)

server/plugins/payload-cache.ts
```diff
@@ -0,0 +1,177 @@
...
+        console.log(`[payload-cache] HIT: ${routePath} (age: ${age.toFixed(1)}s)`)
+      }
+
+      // Set the response — this completely skips the Nuxt render function
+      ctx.response = {
+        body: cached.body,
+        statusCode: cached.statusCode,
+        statusMessage: 'OK',
+        headers: cached.headers,
+      }
```
alexdln, line 100: Can we set the Cache Control header to the remaining time until the next ISR (-5 seconds) or it's not worth it?

server/plugins/payload-cache.ts
```diff
@@ -0,0 +1,177 @@
+import type { H3Event } from 'h3'
+
+/**
+ * Runtime payload cache for ISR-enabled routes.
+ *
+ * Mirrors Nuxt's pre-render `payloadCache` behavior at runtime:
+ * - When an HTML page is rendered, the payload is cached (serialized by the
+ *   Nuxt app plugin `payload-cache.server.ts` and stashed on event.context)
+ * - When a `_payload.json` request arrives, the cache is checked first.
+ *   If a cached payload exists, it's served immediately — completely skipping
```
alexdln, line 10: So, the ISR won't run in the background [on prefetch]? And will it only run when another user accesses the page directly?

danielroe: this ISR will run in the background, yes.

server/plugins/payload-cache.ts
danielroe: I don't think the current issue is caused by mismatching values, but by the total absence of a value - resolving atm.

server/plugins/payload-cache.ts
danielroe: not sure it's worth it

server/plugins/payload-cache.ts
alexdln: It works much [much] more reliably, but I get an error sometimes.

I don't think it's related to what I described, but probably that there's might be a one-second gap between the HTML request and the payload request. So if the request arrives 59.5 seconds after the start, the HTML will be returned from the cache, and the JSON requests (60.5 secs age) will be considered outdated

<details>
<summary>
Screenshots
</summary>
image
image
image

</details>

Maybe sth like stale/expired states will fix it. Trigger the update after 60 seconds, but return the last response up to 120 seconds. But it seems like there would be risks of other artifacts

alexdln: @danielroe even better now, feels stable for 99% of reloads. It seems like you've set up a fallback logic for this case, but it seems like for some reason it didn't work for the readme

<details>
image
image
image
</details>

PR #1644, opened 2026-02-25
feat: generate alt text in charts

### LINKED ISSUE #1264: feat: generate alt text in charts
Alt text may be important for my users and systems. It's needed both for immediate access to content and for sharing later in an A11Y-friendly environment (_like bluesky_).

Content needs to be generated based on the chart. Localization is possible if we can localize the entire chart as well.

It's usually standard practice to add the ability to copy alt text in the options.

Services typically have a character limit of 1000-2000. Bluesky has 2000 characters, so I think it's worth sticking to it.

graphieros:
Resolves #1264 

- Bump vue-data-ui to 3.15.8 (TS types fine tuning)
- Add data analysis utilities which correct for outliers before generating analytics that are injected into translation keys
- Extend the useCharts composable with the alt generation & copy feature

The feature is only applied on the trends charts (modal & compare page) and works for single or multiple series.
We can iterate on other charts (versions).

A new 'Copy alt text' button is added in the chart's contextual menu:

https://github.com/user-attachments/assets/b3378bed-5816-4921-a648-7102484e4df2

| Chart | Alt text |
|------|---------|
|image|Downloads line chart for the vue package. The Y axis represents the number of downloads. The X axis represents the date range, from 2025-03-04 to 2026-02-24, with a weekly time period. vue starts at 6.5M and ends at 9.1M, showing a mostly flat trend with a slope of 38.5K downloads per time interval, and an overall 32.3% growth. At the bottom, a watermark reads "./npmx a fast, modern browser for the npm registry".|
|image|Packages download comparison line chart between: vue, svelte. The Y axis represents the number of downloads. The X axis represents the date range, from 2025-03-04 to 2026-02-24, with a weekly time period. vue starts at 6.5M and ends at 9.1M, showing a mostly flat trend with a slope of 38.5K downloads per time interval, and an overall 32.3% growth, svelte starts at 2.2M and ends at 3M, showing a mostly flat trend with a slope of 12K downloads per time interval, and an overall 35.2% growth. At the bottom, a watermark reads "./npmx a fast, modern browser for the npm registry".|

### DIFF
--- a/app/utils/charts.ts
+++ b/app/utils/charts.ts
@@ -306,7 +306,9 @@ export function computeLineChartAnalysis(values: Array<number | null>): LineChar
 
   let trend: LineChartAnalysis['interpretation']['trend'] = 'undefined'
 
-  if (rSquared !== null) {
+  if (standardDeviation === 0) {
+    trend = 'none'
+  } else if (rSquared !== null) {
     if (rSquared > 0.75) {
       trend = 'strong'
     } else if (rSquared > 0.4) {
--- a/test/unit/app/utils/charts.spec.ts
+++ b/test/unit/app/utils/charts.spec.ts
@@ -449,7 +449,7 @@ describe('computeLineChartAnalysis', () => {
     expect(result.slope).toBe(0)
     expect(result.rSquared).toBeNull()
     expect(result.interpretation.volatility).toBe('undefined')
-    expect(result.interpretation.trend).toBe('undefined')
+    expect(result.interpretation.trend).toBe('none')
   })
 
   it('handles two values where the first value is 0', () => {
@@ -521,7 +521,7 @@ describe('computeLineChartAnalysis', () => {
     expect(result.mean).toBe(5)
     expect(result.standardDeviation).toBe(0)
     expect(result.rSquared).toBeNull()
-    expect(result.interpretation.trend).toBe('undefined')
+    expect(result.interpretation.trend).toBe('none')
   })
 
   it('handles nulls within a perfect linear trend using original indexes', () => {
@@ -666,6 +666,6 @@ describe('computeLineChartAnalysis', () => {
 
     const undefinedTrend = computeLineChartAnalysis([0, 0, 0, 0])
     expect(undefinedTrend.rSquared).toBeNull()
-    expect(undefinedTrend.interpretation.trend).toBe('undefined')
+    expect(undefinedTrend.interpretation.trend).toBe('none')
   })
 })


app/composables/useCharts.ts
```diff
@@ -615,11 +623,103 @@ export function useCharts() {
...
+      daily: 'package.trends.granularity_dayly',
+      weekly: 'package.trends.granularity_weekly',
+      monthly: 'package.trends.granularity_monthly',
+      yearly: 'package.trends.granularity_yearly',
+    }
+
+    const granularityKey =
+      granularityKeyByGranularity[config.granularity as unknown as string] ??
+      'package.trends.granularity_day'
+
```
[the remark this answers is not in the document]
graphieros: done

app/components/Package/TrendsChart.vue
```diff
@@ -1458,10 +1461,22 @@ const chartConfig = computed<VueUiXyConfig>(() => {
...
           },
-          // altCopy: ({ dataset: dst, config: cfg }: { dataset: Array<VueUiXyDatasetItem>; config: VueUiXyConfig}) => {
-          //   // TODO: implement a reusable copy-alt-text-to-clipboard feature based on the dataset & configuration
-          //   console.log({ dst, cfg})
-          // }
+          altCopy: ({ dataset: dst, config: cfg }) =>
+            copyAltTextForTrendLineChart({
+              dataset: dst,
+              config: {
+                ...cfg,
```
alexdln, line 1468: The weekly chart [here](https://npmxdev-git-1264-feat-generate-alt-text-in-charts-npmx.vercel.app/compare?packages=react,react-dom) shows that

> with a weekly time period. react starts at 32.1M and ends at 96.4M, showing a weak trend with a slope of 642.4K downloads per time interval, and an overall 107.7% growth, react-dom starts at 29.3M and ends at 89.5M, showing a weak trend with a slope of 559.9K downloads per time interval, and an overall 97.5% growth

But the chart doesn't show this growth (it's about 200% here) and this is a difference over time (probably correct for "overall"), but I think it either shouldn't be specified for the other options (daily/weekly/monthly) at all or should be adjusted to what the user sees on the chart

Looks like we can count it right here, sth like `(end_value / start_value) * 100 - 100`

graphieros: I though using a linear regression to compute the growth for the whole dataset would be more accurate than using first and last values, which can be misleading if you have for example [10, 100, 100, 100, 12], the 20% growth does not mean anything. Do you think we should remove this datapoint entirely ? (I would rather have nothing than the end / start growth, imo).

alexdln: Maybe we should show the maximum and minimum instead?
> With a minimum value of 2M on 10.06.2025, and a maximum value of 5M on 23.01.2026.

The alt text should just reflect what's visible, and the different methods might be a bit complicated to understand (_or maybe this particular method is not suitable, since the growth here is different_)

Or this all just redundant information and better simply removing it, since overall there's enough data 🤔

graphieros: You may be right, perhaps the slope unit and the trend interpretation are enough. I'll remove the % :)

PR #1645, opened 2026-02-25
fix(ui): prevent scroll jump during pagination

### LINKED ISSUE #1640: Scrolling doesn't work reliably and jumps back to top
### Describe the bug

Scrolling down on the package list is jumping around and regularly jump back to top for me.

In this screen capture I continuously scroll down on my mouse wheel:

https://github.com/user-attachments/assets/0ce08333-b440-46b6-bbc8-02bdae867ab9

As a side note: all those animations on the list are very distracting imho. 🐈 

### Additional context

Zen 1.18.10b (Firefox 147.0.4)

marlonwq:
### 🔗 Linked issue
Resolves #1640

### 🧭 Context
Fixed the bug where scrolling jumps back to the top when the infinite pagination updates the URL.

### 📚 Description
- Replaced `router.replace` with `window.history.replaceState` to update the page number without triggering Nuxt's default scroll-to-top behavior.
- Prevented `currentPage` from resetting if only the pagination changes, not the search text.
- Used `v-show` and CSS (`overflow-anchor: none`) to keep the results list in the DOM and prevent the page from collapsing while loading.

### DIFF
--- a/app/pages/package/[[org]]/[name].vue
+++ b/app/pages/package/[[org]]/[name].vue
@@ -106,7 +106,15 @@ const { data: readmeData } = useLazyFetch<ReadmeResponse>(
     const version = requestedVersion.value
     return version ? `${base}/v/${version}` : base
   },
-  { default: () => ({ html: '', mdExists: false, playgroundLinks: [], toc: [] }) },
+  {
+    default: () => ({
+      html: '',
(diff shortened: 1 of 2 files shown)


danielroe:
this was an annoying bug - thank you ❤️

PR #1647, opened 2026-02-25
fix(ui): visual improvements compare page 1627

### LINKED ISSUE #1627: Visual improvements for Compare page
### Describe the bug

1. Make page viewport wider (just the same as other pages =1200px) - it was discussed in Discord
2. Shrink longer package versions to the maxwidth value that 4 packages will not trigger horizontal scroll
3. Check that no titles overlap with 1st package stats column (see screenshot)

> It is necessary not to break the layout of the rest of the page and create a consistent view when adding columns (without shifts)

Image

### Additional context

_No response_

### Logs

```shell-script

```

Sybren-crwk:
### 🔗 Linked issue
Resolves #1627

### 🧭 Context

The Compare page needed some visual improvements.

Make page viewport wider (just the same as other pages =1200px) - it was discussed in Discord
Shrink longer package versions to the maxwidth value that 4 packages will not trigger horizontal scroll
Check that no titles overlap with 1st package stats column (see screenshot)

### 📚 Description

I removed the wrapper around the compare content to allow more width (container) `<div class="max-w-2xl mx-auto">`.
Added minmax for grid columns to be all same width.
Fixed already truncate by making it `inline-block` and give it a full width.

Before:
image

After:
Screenshot 2026-02-25 at 18 07 26

### DIFF
--- a/app/components/Compare/ComparisonGrid.vue
+++ b/app/components/Compare/ComparisonGrid.vue
@@ -39,11 +39,10 @@ function getReplacementTooltip(col: ComparisonGridColumn): string {
 
         <!-- Package columns -->
         <div v-for="col in columns" :key="col.name" class="comparison-cell comparison-cell-header">
-          <span class="inline-flex items-center gap-1.5 truncate">
+          <span class="inline-flex items-center justify-center gap-1.5 w-full">
             <LinkBase
               :to="packageRoute(col.name, col.version)"
-              class="text-sm truncate"
-              block
+              class="text-sm inline-block truncate"
               :title="col.version ? `${col.name}@${col.version}` : col.name"
             >
               {{ col.name }}<template v-if="col.version">@{{ col.version }}</template>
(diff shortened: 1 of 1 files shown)


alexdln:
@carwack 
As humans, we have many habits and challenges that we address through design. One of them is that we have difficulty processing long lines, as we get lost when navigating through too much content. Therefore, a line in a book contains 60-70 characters (_which further reinforces this habit_)

There are two resulted points from this: it is not recommended to make the number of characters more than 75 characters, and it is recommended to make the content container width no more than 768px with a standard font.

(_Also, because of how we perceive and read content, there are rules to place the most important things in the upper left corner and to make the text left-aligned_)

What's the point of all this... Because of these habits, such a huge field and the back button (_which are so far off to the side_), look and feel cumbersome

And the text in the pre-table area below is far from the previous content, which causes additional cognitive load for user


_I hope it's not too long and boring, and I hope you find this stream of thoughts useful_ ❤️

graphieros:
I think the chart container width should remain identical to the previous implementation

Sybren-crwk:
@alexdln @graphieros thanks for the feedback!
I do understand what you do mean. And love to learn more from UX perspective. 
But what confuses me is what was in the issue https://github.com/npmx-dev/npmx.dev/issues/1627
_Make page viewport wider (just the same as other pages =1200px) - it was discussed in Discord_

How could I best implement this? Only get the comparison grid to be wider to the 1200px?

alexdln:
@carwack Yeah, sorry, I think we just didn't fully described the idea in the issue.

The idea is to make the page container wider. But at the same time, we want the interface to not jump around and everything to be visually smooth and correct. The text also should remain at a comfortable width for reading.

One of the ideas I had:

image

We haven't quite decided on it yet. I think the layout should be more or less like this, it's just a matter of details. If you have any ideas, please share them in the issue or on Discord (we're more active there and can discuss them faster), but unfortunately, I can't give you a definitive design yet

graphieros:
I like @alexdln's mock up, I think it will probably work too when there are only 2 compared packages.
I would recommend centering the chart horizontally, but would not be offended if it is not^^

alexdln:
Regarding the chart's centering - it will look a bit different depending on the number of columns \:(

But two smaller charts, one below the other, definitely seem appropriate on the right ☺️

alexdln:
@carwack I'm closing this PR for now. If you would like to return to this issue - let me know and I will share the design. It also would be better to ping us on discord and we will polish all the details in the fastest way there

PR #1656, opened 2026-02-26
chore: use `nuxt-security`

iiio2:


danielroe:
let's avoid this for now - happy to revisit later.

iiio2:
@danielroe , Ok. Thanks.

PR #1666, opened 2026-02-26
fix: prevent suggestion dropdown from closing when clicking scrollbar

iiio2:
On the /compare page (https://npmx.dev/compare), the package search suggestion dropdown has a scrollbar when there are many results, but clicking on the scrollbar causes the dropdown to close immediately. This makes it impossible to scroll through results using the scrollbar. This issue is fixed.


danielroe (approved):

iiio2:
Thanks @danielroe . :blush:

PR #1667, opened 2026-02-26
docs: various improvements

atinux:
- use same logo
- correct edit this page url
- use same font
- add link to npmx.dev in header


danielroe:
would you have a look at the `knip` test failures? ~~do we need `@nuxt/ui` explicitly installed, for example?~~

maybe sufficient to add `app/assets/css/main.css` to knip.... 🤔

atinux:
tried multiple ways, should be good now @danielroe

PR #1672, opened 2026-02-26
feat(ui): add multi-select and bulk actions for packages

### LINKED ISSUE #1509: Select packages to compare from the search results view
Often, when i wanna install a package for something, but don't know any packages yet, i search by functionalty, and I end up with multiple packages that seem to fit what I need

I'd love to have a way to select some of them and then compare them, without having to go to any package's page

Kinda like "shopping for packages" if that makes sense ^^

Image

MatteoGabriele:
### 🔗 Linked issue

resolves #1509

### 🧭 Context
Added a multi-select feature to the search page that allows users to select multiple packages and perform bulk actions on them. Currently supports comparing selected packages, with the possibility of adding more actions in the future.

### 📚 Description
- **Selection UI**: Added checkboxes to package cards that appear on hover. Selected cards are visually indicated with a border highlight and checkbox state.
- **Persistent counter**: New "View selected (X)" button in the toolbar shows active selections and navigates to a dedicated view for managing them.
- **Floating action bar**: When items are selected, a floating action bar appears at the bottom with the selection count, primary action (Compare), and clear button.
- **Selection state management**: Uses composable to maintain selections across view changes (card/table/selections view), allowing users to continue browsing while keeping their selections.
- **Selection view**: It's a separate component view that retrieves each saved component similar to the Compare page request logic.
- **Accessibility**: Includes aria-live announcements for selection changes and keyboard shortcuts ("b" key) to focus the action bar.

At the moment, the logic is locked at a maximum of 4 selectable items. This won't scale, but for now, to reduce complexity, it will mimic what's needed by the Compare page, which is the only current functionality available in the multi-select.
My take is to re-think this along the way, when and if another action gets added.

https://github.com/user-attachments/assets/0f1aa5f6-6ad8-4a22-a261-d9e700ee59b8

### DIFF
--- a/app/pages/search.vue
+++ b/app/pages/search.vue
@@ -7,6 +7,10 @@ import { isValidNewPackageName } from '~/utils/package-name'
 import { isPlatformSpecificPackage } from '~/utils/platform-packages'
 import { normalizeSearchParam } from '#shared/utils/url'
 
+definePageMeta({
+  preserveScrollOnQuery: true,
+})
+
 const route = useRoute()
 
 const { selectedPackages, showSelectionView, openSelectionView, closeSelectionView } =


MatteoGabriele:
@coderabbitai review

app/components/Package/TableRow.vue
```diff
@@ -196,6 +200,15 @@ const allMaintainersText = computed(() => {
...
     </td>
+
+    <td class="pe-2">
+      <PackageSelectionCheckbox
+        :package-name="result.package.name"
+        :disabled="isMaxSelected && !isSelected"
+        :checked="isSelected"
+        @change="togglePackageSelection"
+      />
+    </td>
```
knowler: I would consider moving this to the first column. It’s a lot of work for some users to trace from the start to the end of a table row to match selection with a package name.

app/components/Package/ActionBar.vue
```diff
@@ -0,0 +1,84 @@
...
+    <div
+      v-if="selectedPackages.length"
+      class="fixed bottom-10 inset-is-0 w-full flex items-center justify-center z-36 pointer-events-none"
+    >
+      <div
+        ref="actionBarRef"
+        tabindex="-1"
+        aria-keyshortcuts="b"
+        class="pointer-events-auto bg-bg shadow-2xl shadow-accent/20 border-2 border-accent/60 p-3 min-w-[300px] rounded-xl flex gap-3 items-center justify-between animate-in ring-1 ring-accent/30"
+      >
```
knowler, line 38: Since this container is focusable (even if programmatically) and it currently is not within a landmark ([best practice is to put everything in a landmark](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA11.html#:~:text=It%20is%20a%20best%20practice%20to%20include%20all%20content%20on%20the%20page%20in%20landmarks%2C%20so%20that%20screen%20reader%20users%20who%20rely%20on%20them%20to%20navigate%20from%20section%20to%20section%20do%20not%20lose%20track%20of%20content)), I’m thinking that we should make this a `region` landmark. We can do so with a named `<section>` element (either using `aria-label` or `aria-labelledby`), since when named that is implicitly a `region` landmark. Since single character keys often conflict with screen reader shortcuts, this provides another means of navigating to the bar.

app/components/Package/ActionBar.vue
```diff
@@ -0,0 +1,84 @@
...
+        <div aria-live="polite" aria-atomic="true" class="sr-only">
+          {{ $t('action_bar.selection', selectedPackages.length) }}.
+          {{ $t('action_bar.shortcut', { key: shortcutKey }) }}.
```
knowler, line 41: I worry that including the search shortcut message each time could get annoying for screen reader users. It is at least at the end. I might need to think more on this.

MatteoGabriele:
Thanks a lot for the feedback @knowler 
The accessibility part was difficult for me to understand because, at least for the cards view, the action bar is pretty much unreachable.

Open to making changes for better accessibility. In the meantime, I will apply some of your suggestions and see where we land.

app/components/Package/ActionBar.vue
knowler: @MatteoGabriele We’ll also need to move the `tabindex=-1` and `aria-keyshortcuts` attributes to the `<section>` element instead of on the `<div>` element.

We do this because when something is focused, assistive technology needs to be able to identify what has been focused to a user. If we focus a `<div>` element then it’ll likely infer it as an unnamed `role=group` which isn’t very helpful. If we focus the _labelled_ `<section>` element, then it will announce it with the name and role (e.g. “action bar, region” and then it’ll probably announce it has a shortcut).

MatteoGabriele: Done! I focused the section, but still visually highlighted the div; otherwise, it would have shown a big outline on a transparent block.

app/components/Package/ActionBar.vue
knowler: We can always tweak this one later.

knowler (approved):
Looks good to me from an accessibility perspective. We can always iterate/tweak things later. Others should review the Vue/JS before we merge.

app/router.options.ts
```diff
@@ -1,19 +1,29 @@
...
+
+    // Scroll to top when main search query changes
+    if (to.path === '/search' && to.query.q !== from.query.q) {
+      return { left: 0, top: 0 }
+    }
+
+    // Don't scroll for other param changes (filters, pagination, etc.)
+    if (to.path === from.path) {
+      return false
+    }
```
serhalp: I believe this breaks the hash anchor navigation logic just below this, by returning too early before we can check it.

My hunch is always to add a test for the behaviour that almost broke when this occurs, as it's a sign we're missing valuable coverage!

app/composables/usePackageSelection.ts
```diff
@@ -0,0 +1,77 @@
+export const MAX_PACKAGE_SELECTION = 4
+
+export function usePackageSelection() {
+  const selectedPackagesParam = useRouteQuery<string>('selection', '', { mode: 'push' })
+  const showSelectionViewParam = useRouteQuery<string>('view', '', { mode: 'push' })
+
+  // Parse URL param into array of package names
```
serhalp, line 7: 🤔 I'm all for the URL-as-source-of-truth-for-state school of thought, but I wonder if selection within search results falls into the ephemeral state bucket that should _not_  be in the URL and consequently in the browser navigation history?

When I play around with it, it admittedly _looks_ cool to hit back and forward across those selections/deselections, but it feels surprising.

I'd be curious to hear other opinions, but my hunch would be not to include it in the URL.

app/composables/usePackageSelection.ts
```diff
@@ -0,0 +1,77 @@
...
+        .slice(0, MAX_PACKAGE_SELECTION)
+    },
+    set(pkgs: string[]) {
+      // Ensure all items are strings before joining
+      const validPkgs = (Array.isArray(pkgs) ? pkgs : []).map(p => String(p).trim()).filter(Boolean)
+      selectedPackagesParam.value = validPkgs.length > 0 ? validPkgs.join(',') : ''
+    },
+  })
+
+  const isMaxSelected = computed(() => selectedPackages.value.length >= MAX_PACKAGE_SELECTION)
```
serhalp: nit: this is super nitpicky but (especially since it's exported) this could be more expressive of the intent rather than the implementation, something like:
```suggestion
  const shouldDisableSelection = computed(() => selectedPackages.value.length >= MAX_PACKAGE_SELECTION)
```

app/composables/usePackageSelection.ts
```diff
@@ -0,0 +1,77 @@
...
+      showSelectionViewParam.value = isOpen ? 'selection' : ''
+    },
+  })
+
+  function closeSelectionView() {
+    showSelectionView.value = false
+  }
+
+  function isPackageSelected(packageName: string): boolean {
+    return selectedPackages.value.includes(String(packageName).trim())
```
serhalp: nit: the type says `packageName` is a string but we're casting it to a string. Is the type wrong or the cast unnecessary?

app/composables/usePackageSelection.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  function togglePackageSelection(packageName: string) {
+    const safeName = String(packageName).trim()
```
serhalp: same nit here

app/components/Package/SelectionView.vue
```diff
@@ -0,0 +1,71 @@
...
+const { selectedPackages, clearSelectedPackages, selectedPackagesParam, closeSelectionView } =
+  usePackageSelection()
+
+const { data, pending } = useAsyncData(
+  async () => {
+    const results = await Promise.all(
+      selectedPackages.value.map(name =>
+        $fetch(`/api/registry/package-meta/${encodeURIComponent(name)}`)
+          .then(response => ({ package: response }))
+          .catch(() => []),
```
serhalp: This fallback's type is incorrect.
```suggestion
          .catch(() => ({})),
```

This may be another good case for test coverage!

btw we can type this a little closer to the fetch response by passing a generic: https://npmx.dev/package/ofetch#user-content-type-friendly

app/components/Package/ActionBar.vue
```diff
@@ -0,0 +1,87 @@
...
+        </div>
+
+        <div class="flex items-center gap-2">
+          <span class="text-fg font-semibold text-sm flex items-center gap-1.5">
+            {{ $t('action_bar.selection', selectedPackages.length) }}
+          </span>
+          <button
+            @click="clearSelectedPackages"
+            class="flex items-center ms-1 text-fg-muted hover:(text-fg bg-accent/10) p-1.5 rounded-lg transition-colors"
+            aria-label="Close action bar"
```
serhalp: ⚠️ this string needs to be internationalized

app/composables/usePackageSelection.ts
MatteoGabriele: I can change it, but I disagree with this one. This is a composable property that tells the component that the maximum selectable items have been reached (perhaps the name could be tweaked), but it is not the job of the composable to tell you that it is disabled, since other parts of the UI might be bound to a completely different functionality or state. Ex: one component disables a checkbox, another component simply displays a tooltip.

app/composables/usePackageSelection.ts
MatteoGabriele: Good catch! Those are leftovers from when I was trying to hack some stuff around 👍

app/composables/usePackageSelection.ts
MatteoGabriele: I need to include it in the URL; otherwise, I won't be able to refresh and get the list from your selection. This is especially helpful if you refresh, navigate to other pages, and then come back. It also makes the selection page shareable. I agree that perhaps it shouldn't be pushed into history, but it can be replaced. Let me know what you think.
I have already made this change.

app/composables/usePackageSelection.ts
serhalp: I can see that. Maybe `canSelectMore`?

app/composables/usePackageSelection.ts
serhalp: Ah yeah, the selection page feels different to me than the existing search results page with selections.

MatteoGabriele: Is that a good thing or a bad thing?

serhalp: Sorry, that wasn't very clear. I meant that on the new selection view I agree selections should be part of the URL and navigation history, but on the normal full search results view selection state feels more ephemeral. 🤔

MatteoGabriele: I tried using useState in a previous version, but the saved state felt wrong, and the data would appear when you didn't expect it to. I believe this one is best suited to a URL: I don't think this functionality fits any other storage options that don't require _complex babysitting._ URL, in a way, is straightforward, managed by navigation, and predictable. You go back and find what you left; you arrive there from another route and see a clean slate. That said, if you have another idea, I'm happy to try it out and see how it feels.

app/router.options.ts
MatteoGabriele: Navigation with hashes works. I can add tests for the scroll behavior: should I go with e2e or unit tests?

app/composables/usePackageSelection.ts
```diff
@@ -11,18 +11,18 @@ export function usePackageSelection() {
...
     set(pkgs: string[]) {
       // Ensure all items are strings before joining
-      const validPkgs = (Array.isArray(pkgs) ? pkgs : []).map(p => String(p).trim()).filter(Boolean)
+      const validPkgs = (Array.isArray(pkgs) ? pkgs : []).map(p => p.trim()).filter(Boolean)
       selectedPackagesParam.value = validPkgs.length > 0 ? validPkgs.join(',') : ''
     },
   })
 
-  const isMaxSelected = computed(() => selectedPackages.value.length >= MAX_PACKAGE_SELECTION)
+  const canSelectMore = computed(() => selectedPackages.value.length >= MAX_PACKAGE_SELECTION)
```
serhalp: This is semantically backwards:

```suggestion
  const canSelectMore = computed(() => selectedPackages.value.length < MAX_PACKAGE_SELECTION)
```

(We'll also need to negate the logic in the two places we read it)

app/router.options.ts
serhalp: I must admit I don't understand why navigation from /foo#bar to /foo#baz still works based on reading this code, but it does seem to work!

---

I would vote for some sort of a test with a browser. (We can add the test in a follow-up PR, doesn't need to block this!)

app/composables/usePackageSelection.ts
MatteoGabriele: oh you're right! didn't think about that. I'll change it

MatteoGabriele:
@serhalp I made the changes, and someone just added a `preserveScrollOnQuery` feature to the router, so I replaced my logic with that to prevent the page from scrolling up on query update. Perfect timing.

serhalp (approved):
Thanks for all your work on this! Can't wait to see all the bulk actions we come up with later

mbtools:
very cool feature. thx! 🏆

PR #1677, opened 2026-02-26
fix: filter ux to keep toggle on right

jellydeck:
stop layout jump when toggling b/w card & table view.
- compare options open correctly as per window size
image

image

---

all the discussion + images are [here on discord](https://discord.com/channels/1464542801676206113/1476511478671736883/1476571369952120923)

https://github.com/user-attachments/assets/5bdaf348-2193-4cc0-ac62-c23b2c0c007d


danielroe:
I was thinking we needed to do this! thank you ❤️

danielroe (approved):

PR #1682, opened 2026-02-27
chore: remove vacation page

serhalp:
### 🔗 Linked issue

N/A

### 🧭 Context

We had a vacation page recently at /recharging

### 📚 Description

Next vacation, we can build a new page. That'll be part of the fun. 😁

We hadn't actually translated the i18n keys anyway.


danielroe:
indeed, I thought the same, but @patak-cat requested we keep it for now

patak-cat:
@serhalp we modified the wording to be in the past, i still use the /recharging url to discuss in social and people may have linked it from articles like the one [from the register](https://bsky.app/profile/theregister.com/post/3mfjiycejm32d). I would also like to make /recharging a stable url, where we could talk about our 2 weeks summer vacation starting around April. let's keep it if there isn't a good reason against that.

PR #1693, opened 2026-02-27
fix(i18n): update Russian translations

dragomano:


wojtekmaj:
@dragomano Please don't put "Resolves #xxxx" because in this particular case, this issue is about many many other locales as well.

i18n/locales/ru-RU.json
```diff
@@ -1,37 +1,60 @@
 {
-  "$schema": "../schema.json",
```
danielroe, line 2: this should not be removed

danielroe (changes requested):

Issue #1689, opened 2026-02-27
Update npmx tagline translation


shuuji3:
This is a tracking issue to update the npmx tagline for each language.

We updated the tagline from "a better browser for the npm registry" to "a fast, modern browser for the npm registry" on Feb 6. We want to make a good npm browser, but we don't want to say that we are replacing npmjs.com or that it is not as good. "Better" might be misunderstood by some npmx users, so we chose "fast, modern" to better describe our intent.

We also need to update `seo.home.description` along with the `tagline`. These two are the minimum. If you have time, it would be great if you could update new strings as well.

The alpha release is on March 3rd. We hope to have the updates by then. I tagged those who translated the `tagline` before the update (the date is when the `tagline` was last updated). Thank you!

- [x] 2026-01-27 [fr-FR.json](https://github.com/npmx-dev/npmx.dev/blame/main/i18n/locales/fr-FR.json) @jycouet #1701
- [x] 2026-01-27 [it-IT.json](https://github.com/npmx-dev/npmx.dev/blame/main/i18n/locales/it-IT.json) @MatteoGabriele #1691
- [x] 2026-01-28 [zh-CN.json](https://github.com/npmx-dev/npmx.dev/blame/main/i18n/locales/zh-CN.json) @9romise #1733
- [ ] 2026-01-30 [ar.json](https://github.com/npmx-dev/npmx.dev/blame/main/i18n/locales/ar.json) @skaldebane
- [x] 2026-01-30 [es.json](https://github.com/npmx-dev/npmx.dev/blame/main/i18n/locales/es.json) @userquin #1703
- [x] 2026-01-30 [ru-RU.json](https://github.com/npmx-dev/npmx.dev/blame/main/i18n/locales/ru-RU.json) @dragomano #1693
- [x] 2026-01-31 [az-AZ.json](https://github.com/npmx-dev/npmx.dev/blame/main/i18n/locales/az-AZ.json) @chz #1692
- [x] 2026-01-30 [uk-UA.json](https://github.com/npmx-dev/npmx.dev/blame/main/i18n/locales/uk-UA.json) @thevrus #1722
- [x] 2026-01-31 [pl-PL.json](https://github.com/npmx-dev/npmx.dev/blame/main/i18n/locales/pl-PL.json) @wojtekmaj #1716
[...]

PR #1696, opened 2026-02-27
feat(i18n): update Spanish description for SEO

userquin:
### 🧭 Context
Check https://github.com/npmx-dev/npmx.dev/issues/1689 (Spanish)

### 📚 Description

Check https://github.com/npmx-dev/npmx.dev/issues/1689 (Spanish)


userquin:
This PR only changes the SEO entry and we should change also the home page

PR #1702, opened 2026-02-27
fix: strip HTML tags from plaintext search result descriptions

### LINKED ISSUE #1681: Raw HTML rendering in search results
### Describe the bug

Compare page:
Raw HTML rendering in package search results instead of parsed or sanitized content.

Image

serhalp:
### 🔗 Linked issue

Fixes #1681

### 🧭 Context

Probably a regression from https://github.com/npmx-dev/npmx.dev/pull/1582

### 📚 Description

After decoding HTML entities, descriptions containing entity-encoded HTML (e.g. `&lt;a&gt;`) were displaying reconstructed tags as raw text. Apply `stripHtmlTags` after `decodeHtmlEntities` in `PackageSelector` and `TableRow` components.

Screenshot 2026-02-27 at 09 00 42

### DIFF
--- a/shared/utils/html.ts
+++ b/shared/utils/html.ts
@@ -12,6 +12,18 @@ export function decodeHtmlEntities(text: string): string {
   return text.replace(/&(?:amp|lt|gt|quot|apos|nbsp|#39);/g, match => htmlEntities[match] || match)
 }
 
+/**
+ * Strip all HTML tags from a string, looping until stable to prevent
+ * incomplete sanitization from nested/interleaved tags
+ * (e.g. `<scr<script>ipt>` → `<script>` after one pass).
+ */
 export function stripHtmlTags(text: string): string {
-  return text.replace(/<[^>]*>/g, '')
+  const tagPattern = /<[^>]*>/g
+  let result = text
+  let previous: string
+  do {
+    previous = result
+    result = result.replace(tagPattern, '')
+  } while (result !== previous)
+  return result
 }
--- a/server/utils/readme.ts
+++ b/server/utils/readme.ts
@@ -3,7 +3,7 @@ import sanitizeHtml from 'sanitize-html'
 import { hasProtocol } from 'ufo'
 import type { ReadmeResponse, TocItem } from '#shared/types/readme'
 import { convertBlobOrFileToRawUrl, type RepositoryInfo } from '#shared/utils/git-providers'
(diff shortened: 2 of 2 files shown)


shared/utils/html.ts
```diff
@@ -11,3 +11,7 @@
 export function decodeHtmlEntities(text: string): string {
   return text.replace(/&(?:amp|lt|gt|quot|apos|nbsp|#39);/g, match => htmlEntities[match] || match)
 }
+
+export function stripHtmlTags(text: string): string {
+  return text.replace(/<[^>]*>/g, '')
```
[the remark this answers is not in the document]
serhalp: I think this is not worth addressing. Using sanitize-html in the client would add 50 KB to the bundle(?) and vue's sanitization already protects us here. Yeah?

shared/utils/html.ts
```diff
@@ -11,3 +11,7 @@ const htmlEntities: Record<string, string> = {
 export function decodeHtmlEntities(text: string): string {
   return text.replace(/&(?:amp|lt|gt|quot|apos|nbsp|#39);/g, match => htmlEntities[match] || match)
 }
+
+export function stripHtmlTags(text: string): string {
```
danielroe, line 20: I believe we already have this code somewhere in the code base. We don't need to create a new function.

app/components/Package/TableRow.vue
```diff
@@ -69,7 +69,7 @@ const allMaintainersText = computed(() => {
       v-if="isColumnVisible('description')"
       class="py-2 px-3 text-sm text-fg-muted max-w-xs truncate"
     >
-      {{ decodeHtmlEntities(pkg.description || '-') }}
+      {{ stripHtmlTags(decodeHtmlEntities(pkg.description || '-')) }}
```
vmrjnvc, line 72: now with this changes there could be full empty descriptions displayed, but there should be dash if there is no description

Image

so correct should be `{{ stripHtmlTags(decodeHtmlEntities(pkg.description || '')) || '-'}}` because if there is just html in the description, all will be removed and we don't have fallback

serhalp: good catch! would you like to open a quick PR for that? 🙏🏼

vmrjnvc: sure, no problem!

here https://github.com/npmx-dev/npmx.dev/pull/1763

PR #1720, opened 2026-02-27
feat(i18n): add missing German translations and pluralization

tomvoet:
### 📚 Description
- Added all missing german translations 
- Improved download trends y-label translation
- Added pluralization to `files_count`, `lines_hidden`, and  `deps_count`
- Added a `title` and small-screen-`sr-only` to the diff button on the package page, to match the other buttons

The pluralization fixes and especially the diff button change go slightly beyond i18n. I'd be happy to split them into a separate PR if that's preferred.


userquin (approved):
LGTM

PR #1723, opened 2026-02-27
feat: route `pkg@version` from deps, direct nav, and search for any valid specifier

### LINKED ISSUE #1120: Dependency links with version union ("1.0 || 2.0") resolves to 404 page
Links on dependenices with **version union** (defined with OR operator like `1.0 || 2.0 || 3.0`) lead to 404 page. Such version union should be split into 3 different links.

Example page: https://npmx.dev/package/@ngxi/unjs (see "Peer Dependency" section on the right). Clicking on version lands on https://npmx.dev/package/@angular/core/v/%5E18.0.0%20%7C%7C%20%5E19.0.0%20%7C%7C%20%5E20.0.0

Image
Image

### LINKED ISSUE #1416: Add possibility to search for specific version of a package

serhalp:
### 🔗 Linked issue

Closes https://github.com/npmx-dev/npmx.dev/issues/1120
Closes https://github.com/npmx-dev/npmx.dev/issues/1416

### 🧭 Context

We have two seemingly separate issues:
- Dependency links with version ranges (e.g. `^18.0.0 || ^19.0.0`, `>15 <=16.0.2`) currently navigate to a non-existent version page and result in a 404.
- _Searching_ for `nuxt@4.0.1` or `nuxt@>=4` or `nuxt@>=4<6` (convenient if pasting from somewhere) does not result in a match. There's a proposed temporary solution in https://github.com/npmx-dev/npmx.dev/pull/1626 to strip the version segment from the query.

### 📚 Description

- Keep links to a _single_ package version as is, but for all other _version specifiers_, link to the package page with the "Filter by semver" input pre-populated! (In the future if/when we have a dedicated versions page we could link to that instead.)
- Rather than implementing this logic at the "where to link to" layer, I implemented this in the routing layer, i.e. `https://npmx.dev/package/nuxt/v/:version` will route to the specific versioned package page if `version` is a specific version string and route to the package page with `?semver=<version>` otherwise.
- ... which means we get a fix to https://github.com/npmx-dev/npmx.dev/issues/1416 almost for free! This PR makes it so that searching for `nuxt@4.0.1` or `nuxt@>=4` or `nuxt@>=4<6` and pressing Enter now navigates directly to the package version page for a specific version or to the semver-filtered package page for a semver specifier, instead of returning no results. This just works for both exact versions and semver ranges, because of the above routing change.

https://github.com/user-attachments/assets/3d0bc107-c00c-455c-8769-9c15719ad6e1


trueberryless:
Awesome idea, I really like this one. Two birds with one stone 🚀

chronoB:
I like that! ❤️  I just checked it out locally and jumping to the correct version by pressing enter is amazing. 

The problem I see is I do not know about that functionality if I don't press enter and the search results do not indicate that this is even possible (because the package is not shown). Maybe a combination with stripping the version for the search as in #1626 could help so that the correct package will also be shown in the results?

Us3r-gitHub:
> npmx.smart.pkg@v.routing.mp4

Hi! Sorry if this is a bit out of context 😅
Just curious—what app did you use to record the demo?
Thanks!

serhalp:
> > npmx.smart.pkg@v.routing.mp4
> 
> Hi! Sorry if this is a bit out of context 😅 Just curious—what app did you use to record the demo? Thanks!

😁 @Us3r-gitHub https://screen.studio/

modules/runtime/server/cache.ts
```diff
@@ -65,42 +66,6 @@ function getFixturePath(type: FixtureType, name: string): string {
...
-  const atIndex = input.indexOf('@')
-  if (atIndex === -1) {
-    return { name: input }
-  }
-  return {
-    name: input.slice(0, atIndex),
-    version: input.slice(atIndex + 1),
-  }
-}
-
```
gameroman: Was this intentional? 👀 

[CI](https://github.com/npmx-dev/npmx.dev/actions/runs/24140209656/job/70439181082?pr=1723#step:4:15) seems to be failing now

gameroman:
Hi @serhalp would you like to continue working on this? ❤️

trueberryless:
I took the liberty to update the PR with changes that happened on main in the meantime (mainly consolidating the `parsePackageSpecifier` and `parsePackageSpec` methods). But I haven't looked at the other CI failures yet. Hope it helps 🫡

modules/runtime/server/cache.ts
```diff
@@ -102,6 +102,56 @@ function getMockForUrl(url: string): MockResult | null {
...
+          version: parsed.version || 'latest',
+          files: [
+            { name: 'package.json', hash: 'abc123', size: 1000 },
+            { name: 'index.js', hash: 'def456', size: 500 },
+            { name: 'README.md', hash: 'ghi789', size: 2000 },
+          ],
+        },
+      }
+    }
+  }
```
gameroman, line 153: Is this related to this PR or some merge artifact?

trueberryless: It was already present before my merge commits: https://github.com/npmx-dev/npmx.dev/pull/1723/changes/BASE..dffa4f2d185b93d6107e20589a219328efcf3407#diff-352c8868c0ceed7c416a532e448dc5865ca568d25b33f6f911099d9e885f4af0

gameroman: @serhalp (do you know) what it's for?

PR #1727, opened 2026-02-27
feat(i18n): add browser locale support for formatting

userquin:
### 🧭 Context

Add support for browser locale formatting via new setting for:
- dates
- datetimes
- sizes
- numbers

### 📚 Description

This PR adds a new composable (`useUserLocale`) to use the new setting `useSystemLocaleForFormatting` (false by default, current behavior) or the app locale.

This will allow use correct formatting for country variants locales where we use the generic one (ar-EG or es-419): date, datetime, size or numbers can vary (decimal separator, thousand separator, date formats can change some style => es-CO and es-ES dates with short style for example)

TODO:
- delete unused `package.size.b` , `package.size.k` and `package.size.m`
- review a lot of vue and ts files
- remove svg ltr hack at `app/components/Package/WeeklyDownloadStats.vue`: we need to check VueUi RTL support (flex-start)
- fix tests


app/pages/settings.vue
```diff
@@ -7,6 +7,8 @@ const colorMode = useColorMode()
 const { currentLocaleStatus, isSourceLocale } = useI18nStatus()
 const keyboardShortcutsEnabled = useKeyboardShortcuts()
 
+const rtl = new Map<string, boolean>(locales.value.map(l => [l.code, l.dir === 'rtl']))
```
userquin, line 10: don't remove this, maybe we'll need it

ghostdevv:
is this related to #699?

Issue #1731, opened 2026-02-28
Link component missing storybook documentation


cylewaitforit:
The link component is not currently document in the Storybook.

PR #1732, opened 2026-02-28
docs(ui): add stories for Link

cylewaitforit:
### 🔗 Linked issue

Closes: #1731

### 🧭 Context

Adds stories for the Link component in the storybook. 

### 📚 Description

- Created Link.stories.ts file

> [!NOTE]
> One tricky thing I ran into with this one is that for the "WithIcon" story I had to pick an Icon that was already *for sure* bundled by unocss in the build. Otherwise Storybook wasn't displaying the icon. (Was trying to use `i-lucide:scale` but switched to `i-lucide:check` instead)

### DIFF
--- a/app/components/Link/Base.vue
+++ b/app/components/Link/Base.vue
@@ -10,10 +10,14 @@ const props = withDefaults(
      * `type` should never be used, because this will always be a link.
      * */
     type?: never
+    /** Visual style of the link */
     variant?: 'button-primary' | 'button-secondary' | 'link'
+    /** Size (only applicable for button variants) */
     size?: 'small' | 'medium'
+    /** Makes the link take full width */
     block?: boolean
 
+    /** Keyboard shortcut hint */
     ariaKeyshortcuts?: string
 
     /**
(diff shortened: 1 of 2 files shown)


sacrosanctic:
This would be a good time to discuss how to structure stories for UI Library Components aka atomic components. I've left `Input.stories.ts` and `Button.stories.ts` in the original PR as a reference.

I prefer `Input.stories.ts`, where the first story showcases the variations and the subsequent stories are for tests

cylewaitforit:
> This would be a good time to discuss how to structure stories for UI Library Components aka atomic components. I've left `Input.stories.ts` and `Button.stories.ts` in the original PR as a reference.
> 
> I prefer `Input.stories.ts`, where the first story showcases the variations and the subsequent stories are for tests

@sacrosanctic 

Great points. 

I think when we are able to use autodocs the docs page becomes a sort of showcase. (Obviously there’s the option of using manual doc pages for now but I think those are a lot to manage.) 

In my experience having the default/primary component usage as the first story makes the autodoc organization flow well. 

That said, I believe for reducing the number of regression test in chromatic I have seen it suggested to only run those test on a combined story like those. So I think we will still find value in a snapshot or showcase story even after we can use autodocs.

PR #1744, opened 2026-02-28
feat: show registered anomalies as dashed segments on download charts

graphieros:
Follow up to #1636 

- When a package has registered anomalies (e.j. vite), and when the data correction is applied, shows corrections as dashed segments on the trends charts

| Without correction | With correction |
|--------------------|----------------|
|image|image|

- This is also applied on the chart of the compare page
- The estimation legend item is also displayed in this case

### DIFF
--- a/i18n/locales/zh-TW.json
+++ b/i18n/locales/zh-TW.json
@@ -33,7 +33,8 @@
     "navigate_results": "瀏覽搜尋結果",
     "go_to_result": "前往搜尋結果",
     "open_code_view": "開啟程式碼檢視",
-    "open_docs": "開啟文件"
+    "open_docs": "開啟文件",
+    "disable_shortcuts": "你可以在 {settings} 中停用鍵盤快速鍵。"
   },
   "search": {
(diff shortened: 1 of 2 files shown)


app/components/Package/TrendsChart.vue
```diff
@@ -1014,15 +1035,23 @@ const chartData = computed<{
...
+    const series = dates.map(t => valueByTimestamp.get(t) ?? 0)
+    const dashIndices = dates
+      .map((t, index) => (anomalyTimestamps.has(t) ? index : -1))
+      .filter(index => index !== -1)
 
     const item: VueUiXyDatasetItem = {
       name: pkg,
       type: 'line',
       series,
+      dashIndices,
```
alexdln, line 1054: Shouldn't it be "`Indixes`" (X, not C) here?

graphieros: indices is a valid form of plural for index
(and it's part of the vue-data-ui api)

alexdln:
I really love this change 🔥

PR #1753, opened 2026-02-28
fix: configure client side limit for comparison

alexdln:
### 🧭 Context

When requesting files that were too large for comparison, after a while the error simply occurred.

image
https://npmx.dev/diff/vue/v/3.6.0-alpha.7...3.5.29?file=dist/vue.runtime-with-vapor.esm-browser.js

### 📚 Description

Configured the limits similar to how they are set up on the code viewing page (250kb on comparison page)

I also noticed that file size calculation was different with preview page. Reused the logic on both pages (_there's also a difference with the server values and comparisons - I'll create an additional PR for unification_)

image

### DIFF
--- a/app/composables/useNumberFormatter.ts
+++ b/app/composables/useNumberFormatter.ts
@@ -16,8 +16,8 @@ export const useBytesFormatter = () => {
   const decimalNumberFormatter = useNumberFormatter({
     maximumFractionDigits: 1,
   })
-  const KB = 1024
-  const MB = 1024 * 1024
+  const KB = 1000
+  const MB = 1000 * 1000
 
   return {
     format: (bytes: number) => {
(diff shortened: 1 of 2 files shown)


app/components/diff/ViewerPanel.vue
```diff
@@ -69,13 +83,6 @@ const changeRatioPercent = computed(() => calcPercent(maxChangeRatio.value, 0, 1
 const diffDistancePercent = computed(() => calcPercent(maxDiffDistance.value, 1, 60))
 const charEditPercent = computed(() => calcPercent(inlineMaxCharEdits.value, 0, 10))
 
-function formatBytes(bytes: number | undefined): string {
-  if (bytes === undefined) return ''
-  if (bytes < 1024) return `${bytes} B`
-  if (bytes < 1024 * 1024) return `${(bytes / 1024).toFixed(1)} KB`
-  return `${(bytes / (1024 * 1024)).toFixed(1)} MB`
-}
```
serhalp, line 77: oh, great, this fixes i18n of the size string too!

PR #1756, opened 2026-02-28
chore: fix harcoded oauth dev port

serhalp:
### 🔗 Linked issue

N/A - just found this while working on something else

### 🧭 Context

When the dev server runs on another port, you can't auth because it redirects back to to port 3000.

### 📚 Description

Use the nuxt dev server port instead of hardcoding

### DIFF
--- a/modules/oauth.ts
+++ b/modules/oauth.ts
@@ -12,8 +12,8 @@ export default defineNuxtModule({
   async setup() {
     const nuxt = useNuxt()
 
-    const { previewUrl, productionUrl } = await getEnv(nuxt.options.dev)
-    const clientUri = productionUrl || previewUrl || 'http://127.0.0.1:3000'
+    const { devUrl, previewUrl, productionUrl } = await getEnv(nuxt.options.dev)
+    const clientUri = productionUrl || previewUrl || devUrl || 'http://127.0.0.1:3000'
 
     // bake it into a virtual file
     addServerTemplate({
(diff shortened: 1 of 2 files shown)


modules/oauth.ts
```diff
@@ -12,8 +12,8 @@ export default defineNuxtModule({
   async setup() {
     const nuxt = useNuxt()
 
-    const { previewUrl, productionUrl } = await getEnv(nuxt.options.dev)
-    const clientUri = productionUrl || previewUrl || 'http://127.0.0.1:3000'
+    const { devUrl, previewUrl, productionUrl } = await getEnv(nuxt.options.dev)
+    const clientUri = productionUrl || previewUrl || devUrl || 'http://127.0.0.1:3000'
```
serhalp, line 16: We still need this final fallback for the test env.

danielroe:
have you tested this when running the dev server on another port?

serhalp:
> have you tested this when running the dev server on another port?

😓 great point. It doesn't work. and I just tried about six other approaches and none of them worked either. I _believe_ we need to use a `.hook('listen')` to capture the actual dev server url, but I think the `oauth` module needs to be refactored, because it does all its work in `setup`, which runs too early(?).

I'll change this to a draft and focus on something else for now 😶.

ghostdevv:
is this still relevant? @serhalp

serhalp:
Yes but I couldn't come with a proper solution 🤷🏼. I'll just close.

Issue #1747, opened 2026-02-28
Automate `en.json` change detection and PR creation


shuuji3:
### Describe the feature

> I come up with a meta file format and workflows that can track changes and is ML-translatable:
> 
> ## 1. update `en.meta.json`
> we'll update this file during PR autocheck when `en.json` is changed (so current text change workflow should be same). it records the latest commit for each updated key. (I guess Lunaria has some helper function to find change detection in en.json (before/after pr))
> 
> before:
> ```
> { "tagline": { "text": "old text", "commit": "sha-a" } }
> ```
> after (by autocheck):
> ```
> { "tagline": { "text": "new text", "commit": "sha-b" } }
> ```
> 
> ## 2. ml translation PR with `ja.meta.json`
> in a separate github actions, we'll compare `en.meta.json` and other `[lang].meta.json`. when a change is detected, apply ml translation as a draft translation.
> 
> before (outdated):
> ```
> {
>   "tagline": {
>     "translation": "old japanese text",
>     "en_text": "old text",
>     "en_commit": "sha-a"
>   }
> }
> ```
> after (ML generated draft):
> ```
> {
>   "tagline": {
>     "translation": "new translated text", // updated by ML
>     "en_text": "new text", 
>     "en_commit": "sha-b"
>   }
> }
> ```
> I think this allows reviewers to see the source text and the translation in a single github pr diff view, and can modify translations as needed in pr review.

Originally posted at https://discord.com/channels/1464542801676206113/1465734070817259601/1477234340688564308

See also #1689

Based on the above idea, I'm going to try implementing a script to generate, `en.meta.json` genearaion as a first step.

### Tasks
- [ ] write a script to generate `en.meta.json` #1761

- [ ] setup GitHub Actions to run `en.meta.json` generation script #1761

- [ ] write a script to update other languages' `*.meta.json` files
- [ ] setup GitHub Actions to run `*.meta.json` update script
- [ ] automated PR creation based on meta json files
- [ ] integrate (optional) ML translation

### Additional information

- [x] Would you be willing to help implement this feature?

### Final checks

- [x] Read the [contribution guide](https://github.com/npmx-dev/npmx.dev/blob/main/CONTRIBUTING.md).
- [x] Check existing [issues](https://github.com/npmx-dev/npmx.dev/issues).

PR #1761, opened 2026-02-28
chore(i18n): add `en.meta.json` generation script

shuuji3:
### 🔗 Linked issue

#1747

### 🧭 Context

Currently, there is no mechanism to detect changes in `en.json`. Previously, we had to choose one of the following approaches when English text needed an update:
- Using a different key: Even a small change requires re-translation for all other languages.
- Updating the existing text: It is difficult to track which keys were updated, making it hard to sync with other languages.

This first step records metadata, tracking which commit updated each English key, which allows us to monitor changes in the English locale file effectively.

### 📚 Description

Implemented the first two tasks for automated `en.json` change detection:
- Added a new generation script:
  - This script generates or updates `en.meta.json` alongside `en.json`.
  - If `en.json` is changed (added, removed, or modified) since the previous update, the corresponding values in `en.meta.json` are updated with the latest commit hash and the new string value.
- Added a GitHub Actions workflow:
  - The script is triggered when a new commit is pushed to the `main` branch, only when the commit modifies `i18n/locales/en.json`.
  - This is needed to record a commit hash in `main` branch history.
    - I also considered and experimented to add auto-commit during the PR or local branch, but it cannot find a reliable commit hash if PR is squash-merged.
    - It could record PR number but I wanted to be git platform agnostic.

The example auto-commit of `en.meta.json` can be found in this CI test branch: https://github.com/shuuji3/npmx.dev/commits/feat/generate-en-meta-json-ci-test 
(Note: In the test branch, I pushed directly to `main` branch and `lunaria/files/*` is included. But they are usually merged along with PR, so they should not be included in auto-commit)

You can also check the Vitest file to see how this script records commit hash and text.


trueberryless:
Hi, @shuuji3!

Sorry to be that guy, but it seems this PR is quite outdated. Do you know the current state of it? Do we still aim to provide more info for translation strings or is the current setup with schema and `en.json` enough?

I took the liberty to help you out and resolve the merge conflicts, so this is up-to-date again 👍

PR #1766, opened 2026-02-28
fix(i18n): update French UI strings

ArmandPhilippot:
### 🔗 Linked issue

None

### 🧭 Context

New English UI strings have been added.

### 📚 Description

Adds missing UI strings to the French translation.

Note that I'm using non breaking spaces before punctuation (`:`, `?`) because I always find this horrible when a colon or an interrogation mark end up alone on a new line (this shouldn't happen according to the typographic rules...). It doesn't seem those kinds of spaces are used elsewhere, so let me know if using regular spaces is a requirement.

I'll also leave a comment where I had a doubt ("vous" vs "tu").

### DIFF
--- a/i18n/locales/fr-FR.json
+++ b/i18n/locales/fr-FR.json
@@ -33,7 +33,8 @@
     "navigate_results": "Parcourir les résultats",
     "go_to_result": "Aller au résultat",
     "open_code_view": "Ouvrir la vue code",
-    "open_docs": "Ouvrir la doc"
+    "open_docs": "Ouvrir la doc",
+    "disable_shortcuts": "Vous pouvez désactiver les raccourcis clavier dans {settings}."
   },
   "search": {
     "label": "Rechercher des paquets npm",
@@ -84,7 +85,8 @@
       "appearance": "Apparence",
       "display": "Affichage",
       "search": "Source de données",
-      "language": "Langue"
+      "language": "Langue",
+      "keyboard_shortcuts": "Raccourcis clavier"
     },
     "data_source": {
       "label": "Source de données",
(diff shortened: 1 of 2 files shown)


i18n/locales/fr-FR.json
```diff
@@ -140,7 +144,28 @@
...
+    "website_placeholder": "https://exemple.com",
+    "likes": "Mentions j'aime",
+    "seo_title": "{handle} - npmx",
+    "seo_description": "Profil npmx par {handle}",
+    "not_found": "Profil introuvable",
+    "not_found_message": "Le profil de {handle} est introuvable.",
+    "invite": {
+      "message": "Il semblerait qu'ils n'utilisent pas encore npmx. Vous voulez leur en parler ?",
+      "share_button": "Partager sur Bluesky",
+      "compose_text": "Salut {'@'}{handle} ! As-tu déjà testé npmx.dev ? C'est un navigateur pour le registre npm : rapide, moderne et open source.\nhttps://npmx.dev"
```
ArmandPhilippot, line 167: In the English version, this begins with "Hey" which suggests a certain familiarity. So, I think "salut"/"tu" is appropriate here? But, this is the only place where we use it so I'm not sure if this should be "Bonjour... vous" instead.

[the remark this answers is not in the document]
ArmandPhilippot: Yeah, I asked a reviewer opinion about informal vs formal for the second sentence.
But, anyway, I don't think your concern is necessarily valid: we can address to the user using a formal approach and the user can use an informal approach to contact someone (the `compose_text` is what should be sent by the user on Bluesky).

graphieros: I think "tu" fits indeed

PR #1654, opened 2026-02-26
feat: nuxt og image v6 + new og images

harlan-zw:
### 🔗 Linked issue

 https://github.com/npmx-dev/npmx.dev/pull/1118 continued

### 🧭 Context

https://github.com/nuxt-modules/og-image/issues/435

### 📚 Description

Nuxt OG Image brings significant improvements to the npmx.dev OG Image implementation:
- Switch to [Takumi](https://takumi.kane.tw/) renderer, which is significantly lighter and faster than Satori (can drop resvg binary workarounds)
- Zero config UnoCSS integration
- Zero config Nuxt Fonts integration
- Better caching
- Fixes type circular dependency issues

This brings the Nitro bundle down 11.3% (8.67 → 7.69 MB) as well as a small reduction in the client bundle.

With the migration, we look to clean up the templates for a more cohesive look and feel. 

_Below are slightly outdated see the PR snapshots_

<details>
<summary>splash</summary>

image

</details>

<details>
<summary>generic page</summary>

image

</details>

<details>
<summary>package without org</summary>

image

</details>

<details>
<summary>package with org</summary>

image

</details>

<details>
<summary>package code - no org</summary>

image

</details>

<details>
<summary>package API docs - no org</summary>

image

</details>

PR #1770, opened 2026-02-28
fix(ui): improve logo

vladh:
### 🔗 Linked issue

It was easier to just create the PR to demonstrate the fix — hope that's okay.

### 🧭 Context

The logo, on the homepage, and in the navigation, had various issues. Most significantly, the kerning and spacing seemed wrong to me.

### 📚 Description

I've create a new logo SVG and applied it throughout the website. Here are the benefits, which are visible in the below screenshots. Please look at the screenshots in full size to be able to see the differences properly.

1. The logo is now an SVG made up of `<path>`s, ensuring the logo is rendered identically across all browsers.
2. **The kerning is now consistent between the letter pairs of “npmx”. Previously, the kerning for the {np} pair was too wide, and the kerning for the other pairs was too tight.**
3. The slash is now less tall, leaving more space for the “npmx” text to stand out, especially as small sizes.
4. The environment text (“dev”, “canary” etc) is now a little bit more spaced out from the “npmx”, avoiding visual clashing.
5. The space between the “./” and “npmx” is now smaller, tying the “./” mark into the text more, and also making the logo more consistent with the OpenGraph image (see below).
6. On mobile, the left edge of the “./” logo is now correctly aligned with the content underneath.
7. On non-homepage pages, the “npmx” text and the environment text are now larger and more readable.
8. On non-heomapge pages, the logo is now properly spaced away from the search box.
9. There is now both an `AppLogo` and an `AppMark` component, making it easier to use either the full logo, or just the “./” mark.
10. `AppLogo` now contains the full logo, allowing it to be more easily used across the website.
11. `AppLogo` previously incorrectly had a background colour and rounded corners, which has been fixed.
12. The `logo.svg` file has been renamed to `logo-icon.svg`, creating space for the new `logo.svg` which contains the full logo, and `logo-mark.svg` which contains just the “./” mark on a transparent background. This should be handy for future use, in eg marketing materials.

shapes at 26-02-28 22 42 24
shapes at 26-02-28 22 42 27
shapes at 26-02-28 22 42 30

### DIFF
--- a/app/components/AppHeader.vue
+++ b/app/components/AppHeader.vue
@@ -224,12 +224,12 @@ onKeyStroke(
           :to="{ name: 'index' }"
           :aria-label="$t('header.home')"
           dir="ltr"
-          class="relative inline-flex items-center gap-1 py-1 header-logo font-mono text-lg font-medium text-fg hover:text-fg/90 transition-colors duration-200 me-4"
+          class="relative inline-flex items-center gap-1 py-2 header-logo font-mono text-lg font-medium text-fg hover:text-fg/90 transition-colors duration-200 me-4"
         >
-          <AppLogo class="h-5 w-auto" />
+          <AppLogo class="h-4.5 w-auto" />
           <span
             aria-hidden="true"
-            class="scale-35 transform-origin-br font-mono tracking-wide text-accent absolute -bottom-0.5 -inset-ie-1"
+            class="scale-35 transform-origin-br font-mono tracking-wide text-accent absolute bottom-0.75 -inset-ie-1"
           >
             {{ env === 'release' ? 'alpha' : env }}
           </span>


danielroe:
would you mind resolving the conflict? 🙏

vladh:
@danielroe Fixed!

7a0fc21b06252ebb5441186da04cbfbb1629bcfc updated `public/logo.svg` but not `public-staging/logo.svg` or `public-dev/logo.svg`. I'm assuming the intent was to also update the latter two, so I updated those to be the same as the `public/logo.svg` that was created in 7a0fc21b06252eb.

Then, as before, I renamed `public{-dev,-staging}/logo.svg` to `public{-dev,-staging}/logo-icon.svg`. Now, `public/logo.svg` is the new full logo that I created.

vladh:
The test failures seem unrelated?

```
malloc_consolidate(): invalid chunk size
```

danielroe:
seems a takumi/og image flakiness (cc: @harlan-zw)

danielroe:
I'm sorry! reverted the og image change due to the flakiness (was also apparent per-route and on other PRs ...)

vladh:
> seems a takumi/og image flakiness (cc: @harlan-zw)

Assuming it is an issue with [takumi](https://github.com/kane50613/takumi) as @danielroe said, I investigated the aforementioned message: `malloc_consolidate(): invalid chunk size`.

Here's what the dependency chain looks like:

* npmx.dev (JS)
  * @takumi-rs/core (JS) 0.69.5 ([source](https://github.com/npmx-dev/npmx.dev/blob/55ff2870279a8dafd3c51d76d83b8a4ce98b8e67/package.json#L77))
    * takumi (Rust) 0.69.5
      * mimalloc_rust (Rust) 0.1.48 ([source](https://github.com/kane50613/takumi/blob/ac748643707bcc5bed19dba2738ebe65733adafa/Cargo.lock#L1539))
        * libmimalloc-sys (Rust) 0.1.44 ([source](https://github.com/purpleprotocol/mimalloc_rust/blob/7daf67f9c464240889a3a8963d22f279ca39ef72/Cargo.toml#L28))
          * mimalloc (C) v3 ([dfa50c3](https://github.com/microsoft/mimalloc/tree/dfa50c37d951128b1e77167dd9291081aa88eea4), [source](https://github.com/purpleprotocol/mimalloc_rust/blob/7daf67f9c464240889a3a8963d22f279ca39ef72/libmimalloc-sys/Cargo.toml#L43))

We use takumi to generate images, and takumi uses mimalloc_rust as its memory allocator. But there's a recent issue on mimalloc_rust https://github.com/purpleprotocol/mimalloc_rust/issues/153 where people are reporting:

> Since we upgrade mimalloc crate from 0.1.47 to 0.1.48, we started seeing Unhandled signal SIGSEGV error from our error logs.
> …
> `… in malloc_consolidate (av=0x7e3a6c000030) at ./malloc/malloc.c:4876`

And in fact, the transition from mimalloc_rust 0.1.47 to 0.1.48 seems significant, because mimalloc_rust commit [747b5b](https://github.com/purpleprotocol/mimalloc_rust/commit/747b5b151432ffb671e99856874e2a7e34414fb3) bumps the version of the underlying C library (Microsoft's mimalloc) from v2 to v3.

Could this be the cause of our problem? Maybe, but it doesn't look like a recent dependency change would have caused it.

The version of mimalloc we depend on is based on the version that takumi depends on, and takumi bumped to mimalloc 0.1.48 in commit [be28f9f](https://github.com/kane50613/takumi/commit/be28f9f), made on 31 Aug 2025. (cc @andrew, this was a great use of [git-pkgs](https://git-pkgs.dev/), thanks)

So it might be that the above mimalloc change is causing our issue, but if so, it's either happening intermittently, or only with certain inputs.

@danielroe Should I just create a new issue for this, so we can keep track of it?

vladh:
@danielroe No problem, rebased!

vladh:
> **Minor note (optional):** The `stroke-width=".26458"` attributes on all paths have no effect since no `stroke` colour is defined. These appear to be leftover from SVG export tooling and could be removed for cleaner output, but they cause no functional issues.

This originates from the original `logo.svg` (the one with just the “./”). Let's leave it, it's just a few extra bytes, won't do any harm, and it saves me some testing.

alexdln:
@vladh Wow, excellent work, thank you for all the changes and unifications

A bit about details:

image

I'm aware that the slash is too low at the top. All the fonts create a different experience and habit, and I felt the difference here - you get \*that\* feeling that something is wrong, but you can't explain what

The stage text looks too far away when it's in the header. It's like it's at a different distance on the home page _where it looks comfortable_, and in the header - where it's a bit too far away

vladh:
Hey @alexdln, thanks for the feedback!

> I'm very aware that the slash is too low at the top. All the fonts create a different experience and habit, and I felt the difference here - you get *that* feeling that something is wrong, but you can't explain what

Am I correct in understanding that you would like the slash to be a little bit taller? Initially, I thought it was too tall and a bit distracting — of course, the font designer will have chosen an appropriate height for normal text, but it felt a bit tall in a “display” scenario like this logo. But, I may have overdone the change, making it too short. Maybe I can find a middle ground — does that sound good?

> The stage text looks too far away when it's in the header. It's like it's at a different distance on the home page _where it looks comfortable_, and in the header - where it's a bit too far away

No problem, I'll make the spacing of the environment text a bit tighter in the header.

alexdln:
> Maybe I can find a middle ground — does that sound good?

Yes, I think a good option would probably be somewhere in between. Maybe make the distance between the top and bottom of the lowercase roughly equal. Maybe just a liiittle bit more at the top

vladh:
> Yes, I think a good option would probably be somewhere in between. Maybe make the distance between the top and bottom of the lowercase roughly equal. Maybe just a liiittle bit more at the top

On it! 🫡

public-staging/logo-mark.svg
alexdln, line 1: Can you please make the same distance at the top and bottom, otherwise it will always be sth like `mt-0.125`

vladh:
@alexdln Changes done!

* Made the slash taller, and a bit taller on the top than the bottom relative to lowercase letter size
* Centered contents of `logo-mark.svg`
* Reduced space above environment text in the navbar
* Cleaned up SVG to remove miscellaneous extra bits, like the stroke info mentioned above

Here are screenshots, though you might want to open them in separate tabs and flip back/forth between them.

Of course, let me know if you have any additional feedback.

<details>
<summary>Before/after screenshots</summary>

## Homepage

### Before

Screen Shot 2026-03-01 at 23 09 53

### After

Screen Shot 2026-03-01 at 23 09 51

## Header

### Before

Screen Shot 2026-03-01 at 23 10 24

### After

Screen Shot 2026-03-01 at 23 10 23

</details>

vladh:
> `9-17`: **Align SVG accessibility semantics.**
> 
> `aria-hidden="true"` means assistive tech will ignore this SVG, so the `<title>` on Line 16 is effectively unused. Either remove the `<title>` or expose the SVG semantics intentionally (e.g. no `aria-hidden`, with explicit labelling).

Looks like both `AppMark` and `AppLogo` have `aria-hidden`, but this was there previously, so I'm setting it aside for the purposes of this PR.

vladh:
> FWIW this PR will require updating all OG image snapshots (or templates), as they were using `logo.svg`. I rasterized the `/` in the PR #1654?

I see the following templates:

* `app/components/OgImage/BlogPost.vue`
* `app/components/OgImage/Default.vue`
* `app/components/OgImage/Package.vue`

None of them seem to refer to any `.svg` files.

On the other hand, I do see that `Default.vue` has a logo SVG embedded within it, and `Package.vue` has the logo “./” embedded, too. I think we can update those files in a different PR, because the logo in the OG image looked different from the logo on the homepage to begin with.

So I don't immediately see any additional OG image work that needs to be done as part of this PR specifically, but @harlan-zw please let me know if I missed something! 🙏🏻

harlan-zw:
Sorry, I was going based on the previous PR being merged, didn't realize it got reverted https://github.com/npmx-dev/npmx.dev/pull/1654. I can swap around the paths when I recreate it so please ignore me :smile:

vladh:
> Sorry, I was going based on the previous PR being merged, didn't realize it got reverted #1654. I can swap around the paths when I recreate it so please ignore me 😄

No problem! 🙌🏻

kane50613:
> Assuming it is an issue with [takumi](https://github.com/kane50613/takumi) as @danielroe said, I investigated the aforementioned message: `malloc_consolidate(): invalid chunk size`.

Appreciate the analysis, but mimalloc is only used for [`takumi-server`](https://github.com/kane50613/takumi/blob/ac748643707bcc5bed19dba2738ebe65733adafa/takumi-server/Cargo.toml#L10), and theres no explicit allocator override for `@takumi-rs/core` & `@takumi-rs/wasm` which means the system's underlying allocator is used (maybe mimalloc for bun?).

I can try to experiment with explicit use of latest `mimalloc` for `@takumi-rs/core` and `wee_alloc` for `@takumi-rs/wasm` then see how it goes.

alexdln:
@danielroe lgtm

I like the updated logo on the main page, but I'm a bit not so sure about it in the header. It might just be a habit, as everything looks correct overall (_before imo it was too big, I made it smaller and now it doesn't take so much space and attention_)

The rest of the fixes are really cool and useful

vladh:
Thanks again for the great feedback @alexdln, this looks way better after your suggestions.

> I like the updated logo on the main page, but I'm a bit not so sure about it in the header. It might just be a habit, as everything looks correct overall (_before imo it was too big, I made it smaller and now it doesn't take so much space and attention_)

We can always keep an eye on it and tweak it if needed!

PR #1777, opened 2026-03-01
feat(i18n): add native Lunaria key merging

yanthomasdev:
### 🧭 Context

Before, we were using a custom system to add key merging from base locale to variants (e.g. en -> en-US -> en-GB, etc). This PR uses the equivalent feature recently added to Lunaria to do it instead.

### 📚 Description

It solves the issue we were having with the confusing setup of both `i18n` and `lunaria/files` that was creating confusion among contributors.

This PR completely removes `lunaria/files` and changes the other i18n scripts to adapt to the new structure.

Note: there's a small issue in the dashboard that is making en-US and en-GB be considered complete in the dashboard's progress by locale but not in the table (i.e. the outdated icon -> 🔄), I'll have to look into this yet, but if someone is able to look before me, I appreciate it.

We'll also need to look into every i18n file and see if there's nothing we are missing from other PRs.

### DIFF
--- a/knip.ts
+++ b/knip.ts
@@ -48,6 +48,9 @@ const config: KnipConfig = {
         /** Some components import types from here, but installing it directly could lead to a version mismatch */
         'vue-router',
 
+        /** Required by @nuxtjs/i18n at runtime but not directly imported in production code */
+        '@intlify/shared',
+
         /** Oxlint plugins don't get picked up yet */
         '@e18e/eslint-plugin',
         'eslint-plugin-regexp',


trueberryless:
The inconsistency between list and table seems to be this expression: https://github.com/yanthomasdev/npmx.dev/blob/3f2dd7a329c97f4c984f02bb0938489c56ab1d95/lunaria/components.ts#L267

Because the list view just uses raw JSON files to determine whether keys are missing, but the table views uses the raw Lunaria status, which includes git timestamps. So alto `missingKeys` is false because `en-US` inherits everything, the git timestamps tell the table that the `en.json` is newer, and therefore `localization.status` is `"outdated"`.

I'm not totally sure how Lunaria is supposed to use all the available info in the different views, but I would suggest extracting the logic for the correct status into a separate function, so it can consistently be reused.

danielroe:
thank you so much @yanthomasdev ❤️❤️❤️

Issue #1793, opened 2026-03-01
bug: CLI connector server does not allow preview deployments access due to CORS restrictions


neutrino2211:
The CLI connector does not allow preview deployments of npmx access via CORS headers.

<img width="464" height="695" alt="Image" src="https://github.com/user-attachments/assets/880ef632-c6c5-472f-a1e7-68a20c60bbcc" />

<img width="610" height="113" alt="Image" src="https://github.com/user-attachments/assets/0f9766ce-6b34-4cd7-a68a-f64c7fd22c9e" />

I have the simple fix implemented in [neutrino2211:fix/allow-vercel-connector-access](https://github.com/neutrino2211/npmx.dev/tree/fix/allow-vercel-connector-access)

danielroe:
~> https://github.com/npmx-dev/npmx.dev/pull/1794#issuecomment-3980592228

PR #1794, opened 2026-03-01
fix: allow vercel preview builds to access the local connector

neutrino2211:
### 🔗 Linked issue

#1793 

### 🧭 Context

The CLI connector uses CORS headers to secure access to it, the headers however do not allow access from vercel preview deployments.

### 📚 Description

This PR adds the following regex to allowed origins

```
/^https:\/\/(.*)\.vercel\.app$/
```

Regex itself might need some work 😬

### DIFF
--- a/cli/src/server.ts
+++ b/cli/src/server.ts
@@ -82,7 +82,12 @@ function generateOperationId(): string {
 }
 
 const corsOptions: CorsOptions = {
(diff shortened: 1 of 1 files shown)


danielroe:
I deliberately did not do this as it allows _any_ person to create a spoof of npmx that can access the connector (in other words, we might as well not even have CORS origins). having a reviewer run the preview build locally is an acceptable tradeoff for the better security.

neutrino2211:
Okay, that makes sense. Should have waited for clarification before opening PR, my apologies.

PR #1797, opened 2026-03-01
fix: add neutral to accentColorIds in prehydration

### LINKED ISSUE #1792: npmx has an accent style even when turned off in settings
Image

maxchang3:
### 🔗 Linked issue

Fixes #1792

### 🧭 Context

`neutral` was missing from `accentColorIds`, so the `has()` check skipped it on page load, and `--accent-color` was not applied, falling back to the CSS default.

### 📚 Description

- Added `neutral` to `accentColorIds`
- Reordered entries to match `ACCENT_COLORS`

A better way to enforce exhaustiveness of `accentColorIds` against `AccentColorId` may be worth exploring.

### DIFF
--- a/app/utils/prehydrate.ts
+++ b/app/utils/prehydrate.ts
@@ -10,7 +10,15 @@ export function initPreferencesOnPrehydrate() {
   // All constants must be hardcoded inside the callback.
   onPrehydrate(() => {
     // Valid accent color IDs (must match --swatch-* variables defined in main.css)
-    const accentColorIds = new Set(['coral', 'amber', 'emerald', 'sky', 'violet', 'magenta'])
+    const accentColorIds = new Set([
+      'sky',
+      'coral',
+      'amber',
+      'emerald',
+      'violet',
+      'magenta',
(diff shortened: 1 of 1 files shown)


maxchang3:
> (💭 I think this has regressed a couple times, so we're probably due for an integration or e2e test of some sort.)

This is what I came up with, though I wasn't sure about a better approach. Maybe worth trying [in-source testing](https://main.vitest.dev/guide/in-source.html) to directly access `accentColorIds`. 

Another approach that comes to mind is using `satisfies` operator.

```typescript
// I randomly copied `UnionToTuple` from https://github.com/type-challenges/type-challenges/issues/737,
// A simpler alternative is to export an `ACCENT_COLOR_IDS` array as a `const`.

// Valid accent color IDs (must match --swatch-* variables defined in main.css)
const accentColorIds = new Set([
  'sky',
  'coral',
  'amber',
  'emerald',
  'violet',
  'magenta',
  'neutral',
] as const satisfies UnionToTuple<AccentColorId>)
```

PR #1799, opened 2026-03-01
fix: update vlt logo to current wordmark version

vltbaudbot:
Updates the vlt sponsor logo from the old molecule-only icon to the current vōlt wordmark version (molecule + text).


serhalp:
Screenshot 2026-03-01 at 15 28 21

it's quite small — are you okay with that?

danielroe:
yes, it would be nice to make it bigger - I think the issue is the svg has built-in padding whereas we would ideally like it not to.

I'll fix.

danielroe:
a bit better I hope 🙏 

image

PR #1812, opened 2026-03-02
fix: debounce search results & centralized live region

### LINKED ISSUE #1803: Debounce search results live region and remove “updating” messages
The existing live region on the search result page needs a few improvements for screen reader users:

1. We should debounce the live region updates to avoid both clobbering other messages and spamming screen reader users. After an input change if we can delay the live region updates by 250ms on desktop and around 700ms on mobile that would likely improve the experience for screen reader users in both situations.
2. We don’t need to include “Updating” in the live region for the search results page. This is just extra noise.

Thank you to [James Scholes](https://dragonscave.space/@jscholes) for the feedback here.

RYGRIT:
### 🔗 Linked issue

resolves #1803 

### 📚 Description

- **Centralized Live Region:** Removed the `role="status"` tags scattered throughout the user interface. Now, all content is handled by a single `sr-only` element at the bottom.

- **Debouncing Updates:** Added debouncing for live region (250ms on desktop, 700ms on mobile).

- **Less Noise:** Hidden the "Updating..." message from screen readers

### DIFF
--- a/app/pages/search.vue
+++ b/app/pages/search.vue
@@ -756,6 +756,9 @@ onBeforeUnmount(() => {
                   $t('search.found_packages_sorted', { count: $n(effectiveTotal) }, effectiveTotal)
                 }}
               </template>
+              <span aria-hidden="true" v-if="status === 'pending'" class="text-fg-subtle">
+                {{ $t('search.updating') }}
+              </span>
             </p>
             <p
               v-if="viewMode === 'table' || paginationMode === 'paginated'"


RYGRIT:
thank you very much for review!

danielroe:
when the next nuxt version is out we might be able to use https://github.com/nuxt/nuxt/pull/34318.

RYGRIT:
> when the next nuxt version is out we might be able to use [nuxt/nuxt#34318](https://github.com/nuxt/nuxt/pull/34318).

Okay, I will continue to follow up.

PR #1818, opened 2026-03-02
fix: run sparkline pulse animation only once

graphieros:
To address concerns regarding the pulse animation expressed in chat:

- pulse animation runs only once
- toggle button and related translations are removed

### DIFF
--- a/app/components/Package/WeeklyDownloadStats.vue
+++ b/app/components/Package/WeeklyDownloadStats.vue
@@ -249,7 +249,7 @@ const config = computed<VueUiSparklineConfig>(() => {
         color: colors.value.borderHover,
         pulse: {
           show: true, // the pulse will not show if prefers-reduced-motion (enforced by vue-data-ui)
-          loop: false, // runs only once if false
+          loop: false,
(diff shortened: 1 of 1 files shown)


graphieros:
Note: this would require to close #1635

app/components/Package/WeeklyDownloadStats.vue
```diff
@@ -254,8 +248,8 @@ const config = computed<VueUiSparklineConfig>(() => {
       line: {
         color: colors.value.borderHover,
         pulse: {
-          show: hasSparklineAnimation.value, // the pulse will not show if prefers-reduced-motion (enforced by vue-data-ui)
-          loop: true, // runs only once if false
+          show: true, // the pulse will not show if prefers-reduced-motion (enforced by vue-data-ui)
+          loop: false, // runs only once if false
```
jhroemer: fwiw `loop: false` is descriptive enough for me to not need the comment

graphieros: You are right, removing this comment

PR #1829, opened 2026-03-02
feat: publish draft posts (with badge + noindex)

danielroe:
### 📚 Description

CleanShot 2026-03-02 at 12 53 09@2x

this allows us to preview in prod - whilst also preventing indexing from search engines (and adding a disclaimer)

inspired by https://opensourcepledge.com/blog/npmx-a-lesson-in-open-source-collaboration-feedback-loops/

### DIFF
--- a/modules/blog.ts
+++ b/modules/blog.ts
@@ -8,6 +8,7 @@ import { read } from 'gray-matter'
 import { safeParse } from 'valibot'
 import { BlogPostSchema, type BlogPostFrontmatter } from '../shared/schemas/blog'
 import { globSync } from 'tinyglobby'
+import { isProduction } from '../config/env'
 
 /**
  * Scans the blog directory for .md files and extracts validated frontmatter.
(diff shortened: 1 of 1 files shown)


whitep4nth3r:
@danielroe can you provide some context about why we want to show draft posts to everyone in the world please?

Also you'll need to remove the hello world blog post to merge this I think.

danielroe:
the blog posts shouldn't be visible on the index (that's a bug!) or crawlable by search engines, so you have to know the exact link.

but it's possibly helpful for sharing links in advance, and getting review from people. (like the oss pledge does)

whitep4nth3r:
We could still share the direct link even when it was a draft, but I do like the draft banner and the noindex, good additions.

Fix up the bug and I'll look again :)

danielroe:
ah - before merging, I actually filtered them out so the drafts weren't even accessible by direct link (I didn't want it looking like a final article)... but as I reflected I realised it was a useful thing to be able to share.

hopefully bug now fixed 🙏

app/components/BlogPostListCard.vue
```diff
@@ -30,7 +32,15 @@ defineProps<{
     >
       <!-- Text Content -->
       <div class="flex-1 min-w-0 text-start gap-2">
-        <span class="text-xs text-fg-muted font-mono">{{ published }}</span>
+        <div class="flex items-center gap-2">
+          <span class="text-xs text-fg-muted font-mono">{{ published }}</span>
+          <span
```
whitep4nth3r, line 37: I'm not seeing this badge on the preview

whitep4nth3r: oh it's the list card, if we're not showing drafts on the blog index we don't need this

danielroe: I think it would be useful when previewing on preview URLs and locally.

whitep4nth3r: But locally we don't show drafts on the index?

I'm confused, but you know more about this than I.

danielroe: it does now show drafts on the `/blog` index locally and on preview deploys (but with badge) and in production, only allows accessing them directly (and displays a banner at the top of the page).

likely you checked before I pushed (yet another) fix 🤦

whitep4nth3r: Ah perfect.

PR #1849, opened 2026-03-02
fix(a11y): use a roving tabindex for connect menu

knowler:
A [`menu`](https://www.w3.org/TR/wai-aria-1.3/#menu) (`composite`) widget is [navigated with arrow keys](https://www.w3.org/TR/wai-aria-1.3/#managingfocus) as opposed to the <kbd>Tab</kbd> key. This means we need to implement a roving tabindex for our connect menu.

Here is the expected behaviour:

- When a user activates the “connect” menu button, it should open the menu and focus the first item.
- When the user presses the down arrow key, it navigates to the next menu item.
- When the user presses the up arrow key, it navigates to the previous menu item.
- When the user presses the <kbd>Home</kbd> or <kbd>End</kbd> keys, they navigate to the first or last menu item respectively (optional).
- When a user presses <kbd>Escape</kbd>, tabs, or clicks away from the menu, the menu should close. In the first case, the focus should return to the menu button.

Since `role=menu` has an implicit `aria-orientation=vertical` property, I’m just implementing the up/down arrow keys. It could be worth just doing up/left and down/right in all cases as the direction isn’t often exposed to users.

I’ve done my best with the Vue and TypeScript code, however, I imagine I’m doing a few things a bit strange, so feel free to edit or make suggestions.

### DIFF
--- a/app/components/Header/AccountMenu.client.vue
+++ b/app/components/Header/AccountMenu.client.vue
@@ -13,6 +13,9 @@ const {
 
 const { user: atprotoUser } = useAtproto()
 
+const menuButtonRef = useTemplateRef('menuButtonRef')
+const menuRef = useTemplateRef('menuRef')
+
 const isOpen = shallowRef(false)
 
 /** Check if connected to at least one service */
@@ -32,6 +35,7 @@ onClickOutside(accountMenuRef, () => {
 
 useEventListener('keydown', event => {
   if (event.key === 'Escape' && isOpen.value) {
+    menuButtonRef.value?.focus()
     isOpen.value = false
   }
 })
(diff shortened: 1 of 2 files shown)


userquin:
Looks like using mouse click at connect doesn't add "focus" styles to first item on chrome, works pressing enter.

knowler:
@userquin I wouldn’t expect the focus styles to appear on mouse click since focus styles only appear for focus visible.

I did notice an issue that if you click the menu button while it’s open, since the menu is checking for loss of focus to close itself, it will cause the menu to immediately open again.

userquin:
I mean, the first item in the list without focus styles (tabindex=0) when using mouse click, I'm not talking about focusing the "selector/menu". Mouse click and pressing enter should have the same behavior (? ), I'm just asking.

_pressing enter_
first item focused when pressing enter

knowler:
@userquin `:focus-visible` styles are only applied based on the heuristics the browser decides. If you use a keyboard the browser will display them, if you use a mouse the browser won’t. The value of `tabindex` is not relevant for mouse/touch users in this situation—it’s purely for keyboard users and assistive tech. This is the case for any control. When you open a model with a click/touch, you won’t see a focus ring around the dismiss button even though that’s where the initial focus is set.

serhalp (approved):
This LGTM and seems to work well.

This is a *wild* amount of boilerplate for basic a11y though. We should be using a headless component library or something, right??? https://reka-ui.com/?

knowler:
I might eventually make a roving tabindex composable since that's a very common pattern.

- - -

Please no one merge until I fix the light dismiss double trigger issue.

danielroe (approved):

knowler:
I’m making this one a draft. I won’t have time to finish it before I turn in for the night.

There are a few problems here:

1. When the menu is open and the user clicks the menu button again, it causes the popup to dismiss and then reopen. To me that’s enough to block this from merging.
2. Upon further reflection, I think we also need to implement a focus trap for this. Luckily, there is already one included in [@vueuse/integrations](https://vueuse.org/integrations/useFocusTrap/), but I think there might need to be a bit of refactoring to get it working.

My code is also a bit messy and trying to add the focus trap on top only makes it messier. The roving tabindex bit is pretty self-contained (i.e. the `onMenuKeyDown` function), but handling the light dismiss is kind of all over the place.

PR #1861, opened 2026-03-03
fix: pds page style bug

RYGRIT:
### 📚 Description

https://github.com/user-attachments/assets/f73967c8-76da-486f-982e-f5fb00b2fcb0

### DIFF
--- a/app/pages/about.vue
+++ b/app/pages/about.vue
@@ -275,29 +275,24 @@ const roleLabels = computed(
                 :key="contributor.id"
                 class="block group relative"
               >
-                <a
-                  :href="contributor.html_url"
-                  target="_blank"
-                  rel="noopener noreferrer"
-                  :aria-label="$t('about.contributors.view_profile', { name: contributor.login })"
-                  class="block rounded-lg"
-                >
-                  <img
-                    :src="`${contributor.avatar_url}&s=64`"
-                    :alt="`${contributor.login}'s avatar`"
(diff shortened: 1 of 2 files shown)


danielroe:
have you checked that tooltips display properly with this change?

RYGRIT:
Sorry, the member list wasn't displayed in my local environment; I overlooked this. Thanks for pointing it out.

I'll fix it right away.

RYGRIT:
@danielroe  Are we willing to increase the left and right padding to solve this problem? 🤔 (But we still can't determine how long the user's name is) The current method causes tooltips to be truncated; I've seen a similar issue on the "About" page.

image

danielroe:
I think for some reason this doesn't occur on the about page (at least to the left) 🤔

increasing padding doesn't look right

RYGRIT:
> I think for some reason this doesn't occur on the about page (at least to the left) 🤔

Perhaps it's because each contributor's username is relatively short.

However, when I visit npmx.dev, the "About" page also has a clipping issue because it adds "overflow-x-hidden" to the main box. 
See video

https://github.com/user-attachments/assets/75a6b3d7-7864-4c42-9dc0-03eb23a96a03

alexdln:
@RYGRIT can we use TooltipApp in both pages? It should update position automatically

alexdln:
_We can_ 😅
Thanks for contribution🩵

PR #1890, opened 2026-03-03
fix: improve sponsors layout on the about page for mobile users

iiio2:
Before: 

wrong

After: 

correct

### DIFF
--- a/app/pages/about.vue
+++ b/app/pages/about.vue
@@ -164,7 +164,7 @@ const roleLabels = computed(
           </h2>
           <AboutLogoList
             :list="SPONSORS"
-            class="grid grid-cols-2 gap-4 md:flex md:flex-row md:items-center"
+            class="grid grid-cols-2 md:flex md:flex-row md:items-center"
           />
         </div>
 


app/pages/about.vue
```diff
@@ -164,7 +164,7 @@ const roleLabels = computed(
           </h2>
           <AboutLogoList
             :list="SPONSORS"
-            class="flex-col gap-6 items-start md:flex-row md:items-center md:gap-4"
+            class="grid grid-cols-2 gap-4 md:flex md:flex-row md:items-center md:gap-4"
```
43081j: looks like this fails ci because `gap-4` is in there twice

the `md:gap-4` is probably redundant now? since `gap-4` is across all viewports, right?

iiio2:
@43081j , I've done it.

43081j:
im not sure whats going on with CI

it still thinks there are two gap-4's 🤔

ghostdevv:
> im not sure whats going on with CI
> 
> it still thinks there are two gap-4's 🤔

seems to be because the underlying component has `gap-4` - fixed in 2ed88bc5ea7409cf3c9d09fcd2c9267ef9204822

iiio2:
Thanks @43081j , @ghostdevv .

PR #1901, opened 2026-03-03
fix: standard.site fixes

fatfingers23:
### 🧭 Context

Our blog posts should now show on readers like https://pckt.blog/read. Will need to clear all previous site.standard.* records  from npmx.dev. Can use pdsls for this https://pdsls.dev/at://did:plc:u5zp7npt5kpueado77kuihyz

- Moved to atcute for tids and dropped atproto/common. Works a bit better if needed client side and smaller package size if i'm not mistaken
- Added the .well-known endpoint/ and at:// links on blog posts that the standard.site indexers need
- change the tid offset for blog posts from file path to slug so it works better on the BlogPostWrapper
- site.standard.documents now have a published time for when they're uploaded so they surface better

### DIFF
--- a/modules/standard-site-sync.ts
+++ b/modules/standard-site-sync.ts
@@ -86,6 +86,8 @@ export default defineNuxtModule({
                 rkey: possiblePublication.tid,
               },
             )
+            // Wait for the firehose and indexers to catch up if we create a publication
+            await new Promise(sleepResolve => setTimeout(sleepResolve, 2_000))
           }
           if (documentsToSync.length > 0) {
             await syncsiteStandardDocuments(authenticatedClient, documentsToSync)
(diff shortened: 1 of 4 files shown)


app/components/global/BlogPostWrapper.vue
```diff
@@ -14,6 +15,15 @@ useSeoMeta({
...
 })
 
+useHead({
+  link: [
+    {
+      rel: 'site.standard.document',
+      href: `at://${NPMX_DEV_DID}/site.standard.document/${generateBlogTID(props.frontmatter.date, props.frontmatter.slug)}`,
+    },
+  ],
+})
```
danielroe, line 25: this should probably be done only for production

fatfingers23: `import { isProduction } from '../config/env'` the best check for that?

PR #1906, opened 2026-03-03
fix: reorganize footer elements and BuildEnvironment behavior

j0u1:
### 🔗 Linked issue

n/a

### 🧭 Context

The footer layout was visually inconsistent when `BuildEnvironment` was hidden, breaking spacing and structure.

### 📚 Description

Moved elements in the footer so that the commit version is now on the left and always visible. Added a "LINKS" label to the right-side links to match the design from the settings page.

`before`:
image

`after`:
image


j0u1:
Current UI state
image

danielroe (commented):
in honesty I prefer the current layout - the site's tagline is more important to show first than the npm disclaimer, and I don't think we need a title for the links

j0u1:
> in honesty I prefer the current layout - the site's tagline is more important to show first than the npm disclaimer, and I don't think we need a title for the links

I agree, but the footer seems to have a bald spot when I see the commit number

j0u1:
> in honesty I prefer the current layout - the site's tagline is more important to show first than the npm disclaimer, and I don't think we need a title for the links

You can do this. The last commit will appear at the bottom
image

danielroe:
I feel it isn't balanced.

I feel like the tagline + links are both 'related to this site' and should be aligned.

it's possible we might move links left, so everything is in a column, or maybe something else. I think this is a case of 'I'll know it when I see it'.

I'm sorry - I think going back and forth on this might be frustrating for you!

might be worth looping in @alexdln who thought this area might need further work and possibly already has some ideas

j0u1:
okay, no worries!

alexdln:
I haven't worked on the footer design yet, but the general idea is that we already have a lot of links (_and we're not displaying them all here yet_) - so we can already categorize them. Also, the keyboard here confuses me that this is a dialog disguised as a link.

There's an imbalance in suggested changes for a bit - if you try to quickly scan the footer, your gaze will go zig-zag (_while we're always used to read information left to right and that's why Daniel immediately felt that something was wrong_)

I'll try to prioritize footer ideas. If you wanna take focus on it - feel free to ping me on Discord (@alexdln.com) and I'll share in the thread there and discuss a little faster

j0u1:
Thanks for feedback, @alexdln. I wrote you in Discord

alexdln:
We've discussed designs on Discord and still a bit unsure about details. For now I'm converting PR to draft

<details>
<summary>Design options</summary>

image

---

image

---

image
</details>

PR #1910, opened 2026-03-04
fix(ui): don't show state before profile/likes finish loading

### LINKED ISSUE #1781: Profile page: empty state glitches on Atmosphere account loading
### Describe the bug

When user has 0 likes and Profile page reloads **empy state is shown for less than 1 second** and disappears after user data is loaded.

- [ ] Add loading state to prevent this glitch.
- [ ] Improve _Like_ widget rendering by adding **loading** state, so it does not glitch when user has liked packages: 0 -> X (see recording below)

Steps to reproduce:
- Go to https://npmx.dev/profile/[YOUR_HANDLE]
- Remove all your likes (sic!)
- Reload page
- Empty state glitches (<1s)

### Additional context

Image

Image

serhalp:
### 🔗 Linked issue

Closes #1781

### 🧭 Context

See the issue. There were 2 visual glitches on the profile page.

#### fix: show skeleton instead of 0 in like cards while loading

Like cards showed the "zero likes" default before the client-side fetch resolved, causing a visible 0 -> X jump. This changes it to show a skeleton loader with a neutral-ish heart icon during the pending state.

#### fix: wait for auth before showing profile invite section

The invite empty state briefly flashed on page load because the auth session (`server: false`) hadn't resolved yet, making the "not own profile" state flash briefly.

https://github.com/user-attachments/assets/71ceb110-310f-4abc-9790-0ea6fca064dc

> [!TIP]
> We could probably avoid the slight layout shift when the number of likes comes in, but I don't see an obvious solution since the number has an unpredictable string length and the heart icon is to the left of it... We can fix that later.

### DIFF
--- a/app/components/Package/LikeCard.vue
+++ b/app/components/Package/LikeCard.vue
@@ -76,6 +76,7 @@ const likeAction = async () => {
       <div class="flex items-center gap-4 justify-between shrink-0">
         <ClientOnly>
           <TooltipApp
+            v-if="likesStatus !== 'pending'"
             :text="likesData?.userHasLiked ? $t('package.likes.unlike') : $t('package.likes.like')"
             position="bottom"
           >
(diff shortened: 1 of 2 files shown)


danielroe:
to avoid the extra layout shift, let's not show the skeleton _or_ heart at all until the data loads.

PR #1912, opened 2026-03-04
fix(ui): load all versions when semver range filter is entered

### LINKED ISSUE #1828: No results matched when filtering by semver
### Describe the bug

https://npmx.dev/package/vue

The following filter conditions return no results:

- `3.5.1`
- `=3.5.1`
- `>=3.5.0 <3.5.8`
- `~3.4.0`
- `3.4.x`

Image

### Additional context

_No response_

### Logs

```shell-script

```

serhalp:
### 🔗 Linked issue

Closes #1828

### 🧭 Context

See #1828 — in some deterministic cases matching versions aren't shown when filtering a package's versions by a semver range.

### 📚 Description

It turns out there were two layered causes here:

1. We were only loading a subset of versions. There was a code path to load all versions but we weren't triggering it. Now, entering a valid semver range triggers loading all versions and auto-expands the "Other versions" section. A guard prevents repeated fetches when typing subsequent valid ranges in the semver input.
2. Tag rows were hidden when their "primary" version didn't match the filter, even if child versions did. For example, filtering for `~3.4.0` on vue hid the `latest` row (pointing to `3.5.13`), making `3.4.0` invisible since it was "claimed" by (nested under) that tag and not in "Other versions". This just needed a fix to the filtering logic to take this into account.

The new integration test covers both.

Try it: https://npmx-g9g3y6w15-npmx.vercel.app/package/vue

Screenshot 2026-03-04 at 07 41 06

Screenshot 2026-03-04 at 07 42 03

(The new auto-expand behaviour is quite nicer, IMO!)

### DIFF
--- a/app/components/Package/Versions.vue
+++ b/app/components/Package/Versions.vue
@@ -457,6 +457,14 @@ function getExpandedTagVersions(tag: string, primaryVersion: string): VersionDis
   return versions.filter(v => filteredVersionSet.value.has(v.version))
 }
 
+// Check if a tag row's children are expanded (manually or via active filter)
+function isTagExpanded(tag: string, primaryVersion: string): boolean {
+  return (
+    expandedTags.value.has(tag) ||
+    (isFilterActive.value && getExpandedTagVersions(tag, primaryVersion).length > 0)
+  )
+}
+
 function findClaimingTag(version: string): string | null {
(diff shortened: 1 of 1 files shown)


ghostdevv:
is this ready for review/merge? @serhalp

PR #1918, opened 2026-03-04
perf: only load `vue-data-ui` css in components that use it

danielroe:
### 📚 Description

this is an experiment to decrease blocking time of initial load (from css size) by only loading vue-data-ui css when we need it

it seems to drop ~7kb from the initial load:
- https://pagespeed.web.dev/analysis/https-npmx-a2mkdwsw7-npmx-vercel-app/jx3cn5ov8d?form_factor=mobile


alexdln (approved):

alexdln:
I even got 17kb diff locally - it went into a separate bundle and only loaded where there are charts 🔥

PR #1937, opened 2026-03-05
fix(lunaria): underline anchor tags and lighten consistently on hover like `summary`

20jasper:
### 🔗 Linked issue

N/A. I asked on Discord if I should make an issue, and was told this is small enough not to require an issue, as any discussion can happen in this PR!

### 🧭 Context

It's difficult to tell anchors apart from non interactive text in the lunaria translations tracker, which makes it harder for new contributors to understand where to find translation work

I updated the CSS to underline anchor tags and lighten the color consistently like with `summary`'s

### 📚 Description

- styles all anchors equivalently across the page
- adds underline to differentiate from non interactive text

| State | Before |
|--------|--------|
| Unhovered Before | Translation anchors look the same as other text. There is no underline |
| Unhovered After | Translation anchors are underlined |
| Hovered Before|  Hovered links are underlined and do not change color |
| Hovered After| Hovered links become lighter, consistent with `summary`'s hover colors |

I did not see any tests for lunaria, and @yanthomasdev is working on a new dashboard to replace this one, so I think it's reasonable to forgo tests for this PR. Glad to add some if anyone feels strongly about adding them though!

### DIFF
--- a/server/utils/npm.ts
+++ b/server/utils/npm.ts
@@ -1,9 +1,11 @@
 import type { Packument, NpmSearchResponse } from '#shared/types'
 import { encodePackageName, fetchLatestVersion } from '#shared/utils/npm'
 import { maxSatisfying, prerelease } from 'semver'
-import { CACHE_MAX_AGE_FIVE_MINUTES, CACHE_MAX_AGE_ONE_DAY } from '#shared/utils/constants'
-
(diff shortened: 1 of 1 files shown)


20jasper:
@coderabbitai I've merged main now if you'd still like to review

PR #1940, opened 2026-03-05
chore: dark mode iconify logo

cyberalien:
### 📚 Description

In `logos/oss-partners` Iconify logo is for light mode, this replaces it with a logo for dark mode.

### DIFF
--- a/app/assets/logos/oss-partners/iconify-light.svg
+++ b/app/assets/logos/oss-partners/iconify-light.svg
@@ -32,10 +32,10 @@
 			<circle cx="24" cy="23" r="5" fill="#000" />
 		</mask>
 	</defs>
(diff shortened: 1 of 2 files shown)


cyberalien:
Logo contains elements that act as masks, drawing background color on top of few shapes, so if icon has the same background as website it all looks seamless.

Though I just noticed there is an option to change background color for website. In that case it might be better to quickly redo icon to have proper masks. I'll test that, will reply in 5 minutes.

userquin:
You need to add the dark variant at index.ts. uhmm, you just change it with transparent bg, we support both variant.

cyberalien:
Oh, I see. I'll fix that!

userquin:
Maybe you can add both with transparent bg color, we have both svg at icons-set nuxt app.

cyberalien:
Done. Also removed previous commits to clean up pull request

danielroe (approved):

PR #1952, opened 2026-03-05
fix: address settings-related hydration issues using prehydrate

### LINKED ISSUE #1948: address settings related hydration issues using prehydrate
### Describe the bug

According to `useSettings.ts`, now all settings are stored in localStorage, it is a client only storage which means server side should never read the settings or it will just get default settings.

https://github.com/npmx-dev/npmx.dev/blob/553087399dbeac30dd4ec585c710be32c2638124/app/composables/useSettings.ts#L85

However, there are some pages require settings like `searchProvider` and run it on server side, make users who has modified settings get hydration errors, page flashing and even worse cases.

Like if we set `Data source` to `npm` and access [https://npmx.dev/~antfu](https://npmx.dev/~antfu) will trigger page flashing and some unexpected behaviors

Image

### Solution

A easy solution is make all components called `useSettings` to be `client only`, but it's essentially a workaround, I think a better solution is migrating all the settings to cookie (Would it be more reasonable to migrate everything there instead of only migrating necessary ones?)

### Additional context

_No response_

### Logs

```shell-script
inline
  10:27946  error  Redundant role "list" on <ul>  no-redundant-role
  10:31520  error  Redundant role "list" on <ul>  no-redundant-role
  10:43548  error  Redundant role "list" on <ul>  no-redundant-role
  10:47720  error  Redundant role "list" on <ul>  no-redundant-role
  10:52413  error  Redundant role "list" on <ul>  no-redundant-role
  10:59418  error  Redundant role "list" on <ul>  no-redundant-role
  10:63616  error  Redundant role "list" on <ul>  no-redundant-role
  10:68288  error  Redundant role "list" on <ul>  no-redundant-role
  10:71926  error  Redundant role "list" on <ul>  no-redundant-role
  10:75503  error  Redundant role "list" on <ul>  no-redundant-role
  10:79632  error  Redundant role "list" on <ul>  no-redundant-role
  10:83171  error  Redundant role "list" on <ul>  no-redundant-role

✖ 12 problems (12 errors, 0 warnings)
```

Adebesin-Cell:
This is an attempt to fix a sudden flash and weird page display behavior, as seen here

https://github.com/user-attachments/assets/a953237c-4d6f-4244-b77d-9ffb7718ea81

The fix swaps useLocalStorage for useState (SSR-safe), paired with a localStorage sync after mount. During hydration, server and client agree on the same default values. 

  - Replaced useLocalStorage with useState + post-mount localStorage sync in useSettings

Closes #1948

AFTER FIX:

https://github.com/user-attachments/assets/3a35315b-8c17-4f97-96cc-ae9c46ec943c


app/composables/useSettings.ts
```diff
@@ -73,22 +71,86 @@ const DEFAULT_SETTINGS: AppSettings = {
...
+  const settings = useState<AppSettings>(STORAGE_KEY, () => ({ ...DEFAULT_SETTINGS }))
+
+  if (import.meta.client && !syncInitialized) {
+    syncInitialized = true
+
+    // Read localStorage eagerly but apply after mount to prevent hydration
+    // mismatch. During hydration, useState provides server-matching defaults.
+    // After mount, we swap in the user's actual preferences from localStorage.
+    // Uses nuxtApp.hook('app:mounted') instead of onMounted so it works even
+    // when useSettings() is first called from a plugin (no component context).
```
danielroe: I don't think this is the right fix. right now, with onPrehydrate, in many cases we're able to reflect the user's preferences immediately without any flash after mounting.

Adebesin-Cell: Hmm, good point. What approach do you think we could look toward here?

I tried a few iterations but was only able to get satisfactory results using the `useState + localStorage` pattern. Though I can see some flashes (related to client-side sorting, I believe)

The core issue is that `useLocalStorage` from VueUse causes hydration mismatches because the server renders with defaults while the client initializes with stored values immediately. Cos right now, when we load https://npmx.dev/~antfu on page load with `npm registry` as a setting. 

The `onPrehydrate` script handles the visual/DOM side (accent color, background theme, collapsed sections), but the reactive state still needs to reconcile somehow.

Open to suggestions if you have a different pattern in mind, I might be missing something in the Nuxt hydration lifecycle that could help here.

Adebesin-Cell:
Hey @danielroe @serhalp, found some time to look into this again.

I reverted the initial `useState` + post-mount sync approach.


The fix is now scoped to the data-fetching composables (`useOrgPackages`, `useUserPackages`, `useGlobalSearch`). 

A new `useEffectiveSearchProvider` defers the localStorage preference until `app:mounted`, so SSR and initial hydration share the same async-data cache key (`DEFAULT_SETTINGS.searchProvider`, or `?p=` from the URL when present). After mount, the reactive key flips and `useLazyAsyncData` refetches once with the user's preference, basically your second suggestion from the issue thread (load with Algolia on first render, swap to npm on client). `useSettings` itself is unchanged, so pages stay cacheable per-URL.

There's still a bit of delay for the first navigation while the npm provider refetches, worth a separate follow-up to keep the SSR data visible during that swap. ~~Happy to open it as a separate issue~~.

gameroman:
hi @Adebesin-Cell could you resolve the git conflicts please

gameroman (commented):
Should we also add a regression test for this?

Adebesin-Cell:
> Should we also add a regression test for this?

Yeah, #2981 already covers this; it added the hydration e2e test in `test/e2e/hydration.spec.ts`. Don't think we need to merge this one anymore 👍

gameroman:
> > Should we also add a regression test for this?
> 
> Yeah, #2981 already covers this; it added the hydration e2e test in `test/e2e/hydration.spec.ts`. Don't think we need to merge this one anymore 👍

Yeah, looks like it was fixed already 👍

PR #1953, opened 2026-03-05
feat: progressive loading for org packages

### LINKED ISSUE #1946: Org page seems to try to load all packages despite pagination?
### Describe the bug

Try loading https://main.npmx.dev/org/types (make sure you're in infinite scroll mode) and watch your network tab. It first takes about a minute or longer for an initial request and then fetches a gazillion packages:

https://github.com/user-attachments/assets/ca9846d1-f74b-49c3-be12-3331756e05e2

I'm not sure if this a regression or this was already this way, but obviously we shouldn't be trying to load all the data here.

### Additional context

_No response_

### Logs

```shell-script

```

Adebesin-Cell:
## Summary

Fixes #1946 and #2507. Org pages were hanging for large orgs like `@types` (11,000+ packages) and `@sindresorhus` (1,690 packages).

### Root cause

Algolia's `getObjects` API has a 1,000-item limit per request. The existing code sent all package names in a single request, which returned a 400 for orgs with >1,000 packages. This triggered the npm fallback path — one API call per package — causing the page to hang.

### Fix

Batch `getPackagesByName` into chunks of 1,000 with a concurrency limit of 3. All packages are fetched upfront so sorting and filtering operate on the complete dataset.

### What changed

- Extract `getPackagesByNameSlice` as an internal helper for single-batch Algolia `getObjects` calls (max 1,000)
- `getPackagesByName` now batches automatically with a concurrency limit of 3
- Uses nested `for...of` instead of spread to avoid exceeding max function arguments for large result sets
- Added staleness guard in `useOrgPackages` for org name changes during fetch
- No UI changes — `PackageList` handles display as before

### Context

This PR originally included progressive loading and `useVisibleItems` integration. After discussion in the PR, we simplified to just the batching fix — sorting a partial dataset is misleading UX, and the batched Algolia path is fast enough to load everything upfront. Client-side paging is tracked separately in #2529.


ghostdevv (changes requested):
Can we use the new composable James added in #2395?

Adebesin-Cell:
@ghostdevv Updated, extended `useVisibleItems` with an `onExpand` callback so expanding triggers `loadAll()` to fetch remaining packages. Progressive loading is preserved (250 on SSR, rest on demand) while using the composable as you suggested.

.claude/skills/vite-plus
```diff
@@ -0,0 +1 @@
+../../node_modules/vite-plus/skills/vite-plus
```
ghostdevv: can you remove this change

.trae/skills/vite-plus
```diff
@@ -0,0 +1 @@
+../../node_modules/vite-plus/skills/vite-plus
```
ghostdevv: this also

CLAUDE.md
```diff
@@ -0,0 +1,88 @@
+<!--VITE PLUS START-->
```
ghostdevv: this too

i18n/locales/en.json
```diff
@@ -875,7 +875,8 @@
       "failed_to_load": "Failed to load organization packages",
       "no_match": "No packages match \"{query}\"",
       "not_found": "Organization not found",
-      "not_found_message": "The organization \"{'@'}{name}\" does not exist on npm"
+      "not_found_message": "The organization \"{'@'}{name}\" does not exist on npm",
```
ghostdevv: how come there is backticks and `{'@'}`?

ghostdevv (dismissed):
I couldn't find the load more button? Also does the load more load every package, or just another 250?

.claude/skills/vite-plus
Adebesin-Cell: Yeah, those were committed by mistake 🤦

i18n/locales/en.json
Adebesin-Cell: Good catch 👏, removed the quotes.

ghostdevv: what about the `{'@'}`?

Adebesin-Cell: The `@` is vue-i18n’s literal escape syntax. `@` is a reserved character used for interpolation (see [https://vue-i18n.intlify.dev/guide/essentials/syntax.html#special-characters](https://vue-i18n.intlify.dev/guide/essentials/syntax.html#special-characters)), so writing `@types` directly would try to resolve a key called `types` instead of rendering the `@` symbol.

It follows the same pattern used across all our locale files (e.g., `atmosphere (@{handle})`, `Include @types in install`, etc.).

app/composables/npm/useOrgPackages.ts
```diff
@@ -1,9 +1,31 @@
+import type { NpmSearchResponse, NpmSearchResult, PackageMetaResponse } from '#shared/types'
+import { emptySearchResponse, metaToSearchResult } from './search-utils'
+import { mapWithConcurrency } from '#shared/utils/async'
+
+/** Number of packages to fetch metadata for in the initial load */
+const INITIAL_BATCH_SIZE = 50
```
jonchurch: Why only 50? This is a very fast query. Are you limiting the initial data fetch to get a neat page of 50 packages displayed in the UI? That's convenient for the UI but inefficient for the network call budget.

jonchurch (changes requested):
The PR does not currently do what it says in the description.

I am not able to make any pagination happen during testing on the preview, not on scroll nor on changing the filter.

Was this ever working for you in testing @Adebesin-Cell 

checked with https://npmx-u41zkl5iu-npmx.vercel.app/org/sindresorhus

app/composables/npm/useOrgPackages.ts
Adebesin-Cell: Ran some benchmarks against the Algolia getObjects endpoint for @types (11,397 packages):

| Initial batch size | Algolia response time (avg of 3 runs) |
|-------------------|----------------------------------------|
| 50                | ~680ms                                 |
| 100               | ~1,030ms                               |
| 150               | ~1,120ms                               |
| 200               | ~1,065ms                               |
| 250               | ~1,170ms                               |
| 500               | ~1,470ms                               |
| 1,000             | ~2,000ms                               |

This is a single Algolia getObjects POST during SSR. It blocks the HTML response. The registry fetch for the package names list is ~20ms (cached), so Algolia is the bottleneck.

50 to 100 adds ~350ms but doubles the initial content. 100 to 250 adds ~140ms more, but the response is already over a second. The default page size is 25, so 100 gives users 4 pages of content on first paint without needing a second fetch.

I can bump it to 100. It is a good balance between content density and SSR speed. The remaining packages load incrementally (1,000 per batch) as the user scrolls or paginates, or all at once when they filter or sort.

Adebesin-Cell:
> The PR does not currently do what it says in the description.
> 
> I am not able to make any pagination happen during testing on the preview, not on scroll nor on changing the filter.
> 
> Was this ever working for you in testing @Adebesin-Cell
> 
> checked with [npmx-u41zkl5iu-npmx.vercel.app/org/sindresorhus](https://npmx-u41zkl5iu-npmx.vercel.app/org/sindresorhus)



> The PR does not currently do what it says in the description.
> 
> I am not able to make any pagination happen during testing on the preview, not on scroll nor on changing the filter.
> 
> Was this ever working for you in testing @Adebesin-Cell
> 
> checked with [npmx-u41zkl5iu-npmx.vercel.app/org/sindresorhus](https://npmx-u41zkl5iu-npmx.vercel.app/org/sindresorhus)

That’s weird. I did test locally, and loading more consistently triggers when you scroll to the end.

If you switch to table view, you can see it paginates correctly with the total package count showing (screenshot attached).

Screenshot 2026-04-14 at 09 37 30


I’ll look into why the infinite scroll load-more isn’t triggering on the preview deployment and fix it.

jonchurch:
It doesnt make much sense to split this view into pages this way while this view is a sorted list of unsorted data. 

That is, we get a list of all packages, and then the view we present is always sorted. Yet we are not currently sorting the data in the query. So what good is showing me the top downloaded package from a random subset of 50 packages out of 100, 500, 1k+ or whatever?

I think the solution here is to query all data before rendering the page. That will still be significantly faster than the existing behavior, and is a much simpler fix. 

My feedback to you here is to first solve the problem much more simply. It is an extremely low bar to improve the perf here, and the only reason the perf is impacted is because it is hitting the much slower fallback codepath due to the 400 that is hit on the fast path, see https://github.com/npmx-dev/npmx.dev/issues/2507

The actual needed fix is batching the network calls to get around the 1k package limit, not turning this view into a paged view.

jonchurch (changes requested):
[The issue](https://github.com/npmx-dev/npmx.dev/pull/1953#pullrequestreview-4104457738) with not fetching the next "page" on scrolling to the bottom is fixed 👍 

But if you test it now you'll see what I mean in https://github.com/npmx-dev/npmx.dev/pull/1953#issuecomment-4242559834

Lets say I want to see the top downloaded package of sindresorhus org. On first load I will just see whatever comes from the first 50 results. I'll think "hmm, i didnt expect that package to be first? oh well" because as a user I am seeing a sorted list but its unclear to me that its only a subset of 50. Sure it says its showing 50 of 1600, but surely its showing me the top 50 based on my sort choice? (it is not)

But imagine I'm a very Smart user, and I know exactly how the implementatino of this sorted view works! I will simply scroll to the bottom of the list of 50! Currently, it fetches the next 1k, re-sorts the list right? I scroll back up to see what the new top one is. Now I know what the top downloaded package of the first 1050 packages are. Now I want to see the list with all of them sorted! So I scroll down, through 1050 items, to trigger it to fetch the next 1k... I rinse and repeat for however many thousand I need to get through.

Image

My point is that the way this page is presented to the user as pre-sorted makes dynamic pagination here feel really bad as UX

Adebesin-Cell:
> It doesnt make much sense to split this view into pages this way while this view is a sorted list of unsorted data.
> 
> That is, we get a list of all packages, and then the view we present is always sorted. Yet we are not currently sorting the data in the query. So what good is showing me the top downloaded package from a random subset of 50 packages out of 100, 500, 1k+ or whatever?
> 
> I think the solution here is to query all data before rendering the page. That will still be significantly faster than the existing behavior, and is a much simpler fix.
> 
> My feedback to you here is to first solve the problem much more simply. It is an extremely low bar to improve the perf here, and the only reason the perf is impacted is because it is hitting the much slower fallback codepath due to the 400 that is hit on the fast path, see #2507
> 
> The actual needed fix is batching the network calls to get around the 1k package limit, not turning this view into a paged view.

I did approach the issue with batching at first. My commit https://github.com/npmx-dev/npmx.dev/commit/fa6c5908 added Algolia batching to handle the 1k getObjects limit (the root cause of the 400 from #2507), though it was sequential and capped at 1000 packages. The https://github.com/npmx-dev/npmx.dev/commit/6e49401a removed the cap and made batches parallel, which is the actual performance fix. No more falling into the slow npm fallback path that was causing the hang.

The progressive loading layer (first 50, load more on scroll, load all on filter) came later based on https://github.com/npmx-dev/npmx.dev/pull/1953#pullrequestreview-4084288582, suggesting we adopt the useVisibleItems composable from #2395. 

I agree that sorting on a partial dataset is misleading and that the batched Algolia path is fast enough to load everything upfront. I am happy to simplify back to "batch-fetch everything, let PackageList handle display." But since the progressive approach came from review feedback, it would be good to get @ghostdevv's take before I strip it out.

43081j:
ok i've mostly skimmed the diff so let me know if i misunderstood anything.

i think @ghostdevv was suggesting `useVisibleItems` because this PR originally changed the UI to have paging, just without using the same composable the other pages use. In that situation, it would be the right thing to do - consolidate around `useVisibleItems`.

However, going back to the root problem, i'm not sure we need all of this.

As far as I understand, the root problem is that we _send_ too many packages to Algolia in one request so we get an error back.

The solution to that is to **batch the requests, ideally with a concurrency limit** (i.e. so we don't just `Promise.all` the lot, we send a batch-of-batches at a time).

If we do that server-side, the UI receives the full batch as far as i understand and renders it in its entirety (no paging).

unrelated to the Algolia limit, it may be nicer UX to page those client-side one day. But that feels like something which needs a discussion first.

So my two suggested **actions** are:

1. Create a PR which _only_ batches the Algolia owner requests and leaves the UI as it is today
2. Create an issue to discuss adding paging to the author page (clientside or serverside doesn't matter right now and would be part of the discussion)

Adebesin-Cell:
@43081j @jonchurch Thanks for the clear breakdown. That makes sense — Willow's suggestion was the right call given the PR had paging at the time, but the paging itself was solving a symptom rather than the root cause.

I'll simplify this PR down to just the Algolia batching fix (with concurrency limit) and leave the UI as-is. I've opened #2529 to discuss client-side paging for org pages separately as James suggested.

app/composables/npm/useAlgoliaSearch.ts
```diff
@@ -238,11 +236,37 @@ export function useAlgoliaSearch() {
...
+      batches.push(packageNames.slice(i, i + BATCH_SIZE))
+    }
+
+    // Fetch batches with concurrency limit to avoid overwhelming the API
+    const CONCURRENCY = 3
+    const allObjects: NpmSearchResult[] = []
+    for (let i = 0; i < batches.length; i += CONCURRENCY) {
+      const chunk = batches.slice(i, i + CONCURRENCY)
+      const results = await Promise.all(chunk.map(batch => getPackagesByNameSlice(batch)))
+      allObjects.push(...results.flat())
```
43081j: you probably wanna do this in a `for...of` since we could quite easily exceed the max function args and cause it to throw

```ts
for (const result of results) {
  for (const pkg of result) {
    allObjects.push(pkg);
  }
}
```

edit:

actually it can only be 3000 args, right? but still i would do this as we may one day increase the concurrency for example

43081j (approved):

PR #1959, opened 2026-03-06
feat: add binary file warning

### LINKED ISSUE #1945: Package code page shows binary files as a raw string
### Describe the bug

Package code page shows binary files as a raw string. For example, `.woff2` files are shown like this:

https://npmx.dev/package-code/vitepress/v/1.6.4/dist%2Fclient%2Ftheme-default%2Ffonts%2Finter-italic-cyrillic-ext.woff2

Image

I'd expect it to be shown like " --- Binary file --- ". Maybe with a `[show as string]` button. Even further, ideally the images should be shown with a `` tag.

### Additional context

_No response_

### Logs

```shell-script

```

ShroXd:
### 🔗 Linked issue

Resolves #1945 

### 🧭 Context

The binary file is rendered directly, which is not a good user experience. Before supporting different binary file rendering (e.g. image), it would be better to show a warning and allow user to render it anyway.

### 📚 Description

| Scenario | Screenshot |
| :- | :- |
| Light mode | Screenshot From 2026-03-08 09-34-49 |
| Dark mode | Screenshot From 2026-03-08 09-35-16 |

### DIFF
--- a/app/utils/file-types.ts
+++ b/app/utils/file-types.ts
@@ -65,6 +65,7 @@ const BINARY_EXTENSIONS = new Set([
 ])
 
 export function isBinaryFilePath(filePath: string): boolean {
-  const ext = filePath.split('.').pop()?.toLowerCase() ?? ''
+  const dotIndex = filePath.lastIndexOf('.')
+  const ext = dotIndex > -1 ? filePath.slice(dotIndex + 1).toLowerCase() : ''
   return BINARY_EXTENSIONS.has(ext)
 }


43081j:
I'm not sure I would have this "render anyway" option. If the user really wants to look at a binary file, they can "view raw". 

This is how GitHub does it too and it does work pretty well

ShroXd:
> I'm not sure I would have this "render anyway" option. If the user really wants to look at a binary file, they can "view raw".
> 
> This is how GitHub does it too and it does work pretty well

Hi, thanks for the feedback!

That's a good point! I checked the behavior on Github, it does look good. I will update my code.

app/utils/file-types.ts
```diff
@@ -0,0 +1,70 @@
...
+  // Data
+  'bin',
+  'dat',
+  'db',
+  'sqlite',
+  'sqlite3',
+])
+
+export function isBinaryFilePath(filePath: string): boolean {
+  const ext = filePath.split('.').pop()?.toLowerCase() ?? ''
```
43081j: It would be more efficient to use lastIndexOf here and, if it is more than -1, slice from there to get the extension

ShroXd: Nice point! I will refactor here.

43081j:
This looks good to me but I'm wondering if looking at the extension is the most sensible approach. 

Maybe if we're about to load the file anyway, we can load the first few bytes and check for common binary things like a null byte. 

Just a thought, also happy to land this as is. 

Will wait for a second review

TheDevMinerTV:
I don't love this solution. By just checking the path, someone could rename their binary file to something not classified by this list. I'd prefer something that actually checks the MIME type.

Sadly this is kind of tricky in the browser. [`Blob#type`](https://developer.mozilla.org/en-US/docs/Web/API/Blob/type) also just guesses by the file extension. Something like [`file-type`](https://github.com/sindresorhus/file-type) would be more safe but the browser would already have to download the file to classify it, not to mention the mime type database which is also quite huge.

ShroXd:
> I don't love this solution. By just checking the path, someone could rename their binary file to something not classified by this list. I'd prefer something that actually checks the MIME type.
> 
> Sadly this is kind of tricky in the browser. [`Blob#type`](https://developer.mozilla.org/en-US/docs/Web/API/Blob/type) also just guesses by the file extension. Something like [`file-type`](https://github.com/sindresorhus/file-type) would be more safe but the browser would already have to download the file to classify it, not to mention the mime type database which is also quite huge.

Yes.. Only checking the extension is a crude solution, it's easy to bypass the check. My thought is:

- Since this PR is only focusing on UI changes, we can implement better file type checking in a separate PR. Even if someone bypasses the check, it won't introduce any risk to the website.
- For a more robust solution, `file-type` supports reading just the file header to determine its type. Since we may also want to support rendering certain file types (like images) on the code page in the future, it's worth taking some time to carefully design the detection flow.

Issue #1841, opened 2026-03-02
Components missing Storybook documentation


cylewaitforit:
This is a parent issue to track adding stories files for components that are missing them.

- [ ] AboutLogoImg
- [ ] AboutLogoList
- [x] ButtonBase #1270
- [ ] ButtonGroup
- [ ] CodeDirectoryListing
- [ ] CodeFileTree
- [ ] CodeMobileTreeDrawer
- [ ] CodeViewer
- [ ] CompareComparisonGrid
- [ ] CompareFacetCard
- [ ] CompareFacetRow
- [ ] CompareFacetSelector
- [ ] CompareLineChart
- [ ] ComparePackageSelector
- [ ] CompareReplacementSuggestion
- [ ] DiffFileTree
- [ ] DiffHunk
- [ ] DiffLine
- [ ] DiffMobileSidebarDrawer
- [ ] DiffSidebarPanel
- [ ] DiffSkipBlock
- [ ] DiffTable
- [ ] DiffViewerPanel
- [ ] FilterChips
- [ ] FilterPanel
- [ ] GlobalBlogPostFederatedArticles
- [ ] GlobalBlogPostWrapper
- [ ] GlobalBlueskyPostEmbed
- [ ] HeaderAccountMenu
- [ ] HeaderAuthModal
- [ ] HeaderConnectorModal
- [ ] HeaderMobileMenu
- [ ] HeaderOrgsDropdown
- [ ] HeaderPackagesDropdown
- [ ] HeaderSearchBox
- [x] InputBase
- [x] LinkBase #1732 
- [ ] OgImageBlogPost
- [ ] OgImageDefault
- [ ] OgImagePackage
- [ ] OrgMembersPanel
- [ ] OrgOperationsQueue
- [ ] OrgTeamsPanel
- [ ] PackageAccessControls
- [ ] PackageCard
- [ ] PackageChartModal
- [ ] PackageClaimPackageModal
- [ ] PackageCompatiblitiy
- [ ] PackageDependencies
- [ ] PackageDeprecatedTree
- [ ] PackageInstallScripts
- [ ] PackageKeywords
- [ ] PackageLikeCard
- [ ] PackageList
- [ ] PackageListControls
- [ ] PackageListToolbar
- [ ] PackageMaintainers
- [ ] PackageManagerSelect
- [ ] PackageMetricsBadges
- [ ] PackagePlaygrounds
- [ ] PackageReplacement
- [ ] PackageSidebar
[...]

PR #1964, opened 2026-03-06
docs(ui): add ButtonGroup stories to storybook

cylewaitforit:
### 🔗 Linked issue

Part of #1841 

### 🧭 Context

Adds stories for ButtonGroup

[Preview npmx Storybook on Chromatic](https://698b88d5157d89f1f33a6c21-wadelmqcbx.chromatic.com/?path=/docs/components-button-buttongroup--docs)

### 📚 Description

- Created ButtonGroup.stories.ts 
- Moves ButtonBase stories from Base.stories.ts to ButtonBase.stories.ts
- Enables internationalization in ButtonGroup and ButtonBase stories.
- Enables autodocs page for ButtonBase and ButtonGroup


pnpm-lock.yaml
```diff
@@ -4427,6 +4433,13 @@ packages:
   '@storybook/global@5.0.0':
     resolution: {integrity: sha512-FcOqPAXACP0I3oJ/ws6/rrPT9WGhu915Cg8D02a9YxLo0DE9zI+a9A5gRGvmQ09fiWPukqI8ZAEoQEdWUKMQdQ==}
 
+  '@storybook/icons@1.6.0':
```
cylewaitforit: `storybook-i18n` introduces an older version of the `@storybook/icons` library. Opened https://github.com/stevensacks/storybook-i18n/issues/29 so this can be addressed this in a future version of  `storybook-i18n`.

knip.ts
```diff
@@ -65,14 +62,7 @@ const config: KnipConfig = {
...
       entry: ['app/**/*.{ts,vue,css}'],
-      ignoreDependencies: [
-        'docus',
-        'better-sqlite3',
-        '@nuxtjs/mdc',
-        'nuxt!',
-        '@nuxt/ui',
-        'tailwindcss',
-      ],
+      ignoreDependencies: ['docus', 'better-sqlite3', '@nuxtjs/mdc', 'nuxt!'],
```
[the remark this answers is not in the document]
cylewaitforit: These were included in the knip config hints to remove. If knip check passes then coderabbit is wrong.

.storybook/preview-head.html
```diff
@@ -0,0 +1,6 @@
+<style>
+  /* Override docs story canvas background to match npmx theme */
+  .docs-story {
+    background-color: var(--bg, oklch(0.171 0 0)) !important;
+  }
+</style>
```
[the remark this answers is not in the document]
cylewaitforit: The backgrounds in the autodocs pages look correct with this when viewing the light versions of components and removing this makes them not display correctly. Would love to find a way to fix this in the storybook theme overrides instead of needing this preview-head file but at this time have been unsuccessful.

app/components/Button/ButtonBase.stories.ts
```diff
@@ -0,0 +1,116 @@
...
+} satisfies Meta<typeof ButtonBase>
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: {
+    default: 'Button Text',
+  },
+}
```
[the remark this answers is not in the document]
cylewaitforit: Keeping the default story with the default render allow to someone viewing the storybook to change the story to alternative text in place of `Button Text`.  The other stories use the template render in order for the internationalization to work.

ghostdevv:
Could you mark this as ready for review when you are? Then I'll take a look!

knip.ts
cylewaitforit, line 1: This shouldn’t have merged this way. Need to fix. Moving back to draft for a minute.

.storybook/main.ts
```diff
@@ -19,34 +24,6 @@ const config = {
...
-            return undefined
-          }
-        }
-        if (typeof hook === 'function') {
-          docgenPlugin.transform = wrapped as typeof hook
-        } else if (hook) {
-          hook.handler = wrapped as typeof hook.handler
-        }
-      }
-    }
```
JReinhold, line 81: I'm very curious to hear why this is no longer necessary? 🤔 @cylewaitforit

cylewaitforit: Why it's not needed I am not sure, while troubleshooting getting the autodocs to generate from the jsdoc comments for ButtonBase and ButtonGroup I removed it. Builds still worked. 

Since it said it was needed to get builds to not fail, I didn't feel like it needed to be added it back.

cylewaitforit: Thinking about it more, it's probably because these two components don't import their types from other .vue files.  

Maybe that is okay, since I believe @danielroe added that as a known limitation. https://github.com/npmx-dev/npmx.dev/blob/3de681da85de31d8a09f0c2993961b595d824853/CONTRIBUTING.md?plain=1#L997

cylewaitforit: (I am also not married to keeping this removal in this PR. I'm all for deleting code but if it feels out of scope here happy to discuss in a smaller scoped PR.)

JReinhold: ah I think you're right, as long as none of stories and their sub-modules imports types from `.vue` files, I guess this workaround is not needed. If you're okay with that compromise (which it looks like you are based on that comment) then this is probably the rigth way forward.

cylewaitforit:
> Could you mark this as ready for review when you are? Then I'll take a look!

@ghostdevv  Ready when you have time.

cylewaitforit:
Moving this back to draft since it has drifted from main and #2062 and #2125 should probably be merged separately first.

.storybook/main.ts
cylewaitforit: Removed from this PR for now. Can look at it again in the future.

PR #1970, opened 2026-03-07
fix: remove border in package suggestion dropdown items

iiio2:
On the Compare page (/compare), the package search dropdown items had visible borders in light theme but NOT in dark theme, creating an inconsistent look. 

Each dropdown item uses `ButtonBase`, which has border border-border in its base classes. The --border color token resolves to oklch(0.8514 0 0) in light theme (visible against --bg-elevated) and oklch(0.269 0 0) in dark theme (blends with the dark background).

Added `!border-transparent` to the two `ButtonBase` usages inside the dropdown in `PackageSelector.vue` making item borders invisible in both themes while preserving all other button behavior.

Before: 

before

After: 

after

### DIFF
--- a/app/components/OgImage/BlogPost.vue
+++ b/app/components/OgImage/BlogPost.vue
@@ -110,6 +110,8 @@ const formattedAuthorNames = computed(() => {
               v-if="author.avatar"
               :src="author.avatar"
               :alt="author.name"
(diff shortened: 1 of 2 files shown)


iiio2:
Thanks. :blush:

PR #1974, opened 2026-03-07
feat: add bar chart view to compare page

graphieros:
Some users find it easier to read information from tables, others from charts.
This offers the possibility to switch from one view to the other to compare chartable facets.

The charts for downloads history remains independent, and visible with both views.

- bump vue-data-ui to 3.15.11 (bar charts improvements)
- centralise some shared chart utils
- add bar chart component: minimal chart, without CSV download feature since the whole table remains copiable

Desktop:
desktop

Mobile:
mobile

### DIFF
--- a/app/pages/compare.vue
+++ b/app/pages/compare.vue
@@ -283,6 +283,7 @@ useSeoMeta({
         <div v-else-if="packagesData && packagesData.some(p => p !== null)">
           <!-- View tabs -->
           <div
+            v-if="selectedFacets.some(facet => facet.chartable)"
             class="inline-flex items-center gap-1 rounded-md border border-border-subtle bg-bg-subtle p-0.5 mt-4"
             role="tablist"
             :aria-label="$t('compare.packages.section_comparison')"


app/components/Compare/FacetBarChart.vue
```diff
@@ -0,0 +1,267 @@
...
+              fontSize: isMobile.value ? 12 : 18,
+              percentage: { show: false },
+              offsetX: 12,
+              bold: false,
+              color: colors.value.fg,
+              value: {
+                formatter: ({ config }) => {
+                  return config.datapoint.formattedValue
+                },
+              },
```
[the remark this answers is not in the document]
graphieros: config.datapoint exposes any additional attribute passed in the dataset item, so this is fine. There is just a missing TS type def in the library, which will be dealt with in the future.

alexdln:
Visually, I really like everything except the text size. The package name is initially 16px, but since the container is scaled down, it's now ~9px.

I don't know if we can safely increase it, but I think it would be helpful to add a title to the entire line, in the format `<title>{name}: {value}<title>`

@knowler wdyt?

graphieros:
@alexdln I enabled the tooltip, and also increased the font size for names slightly, which should solve the problem

app/components/Compare/FacetBarChart.vue
```diff
@@ -0,0 +1,315 @@
...
+})
+
+const skeletonDataset = computed(() =>
+  props.packages.map((_pkg, i) => {
+    return {
+      name: '_',
+      value: i + 1,
+      color: colors.value.border,
+    }
+  }),
```
alexdln: _Not sure about formatting_

```suggestion
  props.packages.map((_pkg, i) => ({
    name: '_',
    value: i + 1,
    color: colors.value.border,
  })),
```

app/composables/useChartWatermark.ts
```diff
@@ -47,15 +47,41 @@ export function drawSvgPrintLegend(svg: Record<string, any>, colors: WatermarkCo
   return seriesNames.join('')
 }
 
+function injectWatermarkLogo({
```
alexdln: imo generate or render is more suitable, because it does not make any mutations, but simply returns a string

```suggestion
function generateWatermarkLogo({
```

app/pages/compare.vue
```diff
@@ -278,10 +281,73 @@ useSeoMeta({
...
-          <div class="hidden md:block overflow-x-auto">
-            <CompareComparisonGrid :columns="gridColumns" :show-no-dependency="showNoDependency">
-              <CompareFacetRow
+          <!-- View tabs -->
+          <div
+            class="inline-flex items-center gap-1 rounded-md border border-border-subtle bg-bg-subtle p-0.5 mt-4"
+            role="tablist"
+            :aria-label="$t('compare.packages.section_comparison')"
+          >
+            <button
```
alexdln, line 291: Can we reuse ButtonGroup here?

app/pages/compare.vue
```diff
@@ -278,10 +281,73 @@ useSeoMeta({
...
+              id="comparison-tab-table"
+              type="button"
+              role="tab"
+              :aria-selected="comparisonView === 'table' ? 'true' : 'false'"
+              aria-controls="comparison-panel-table"
+              :tabindex="comparisonView === 'table' ? 0 : -1"
```
alexdln: If the current tab is a table, then we disable the "table" tab

```suggestion
              :tabindex="comparisonView === 'table' ? -1 : 0"
```

app/pages/compare.vue
```diff
@@ -278,10 +281,73 @@ useSeoMeta({
...
+              <span>{{ $t('compare.packages.table_view') }}</span>
+            </button>
+
+            <button
+              id="comparison-tab-charts"
+              type="button"
+              role="tab"
+              :aria-selected="comparisonView === 'charts' ? 'true' : 'false'"
+              aria-controls="comparison-panel-charts"
+              :tabindex="comparisonView === 'charts' ? 0 : -1"
```
alexdln:
```suggestion
              :tabindex="comparisonView === 'charts' ? -1 : 0"
```

app/utils/charts.ts
```diff
@@ -595,3 +604,127 @@ export async function copyAltTextForVersionsBarChart({
...
+export function createAltTextForCompareFacetBarChart({
+  dataset,
+  config,
+}: AltCopyArgs<VueUiHorizontalBarDatapoint[], FacetBarChartConfig>) {
+  if (!dataset) return ''
+  const { facet, description, $t } = config
+
+  console.log(config)
+
+  const packages = dataset.map(d => d.name).join(', ')
```
alexdln, line 616:
```suggestion

  const packages = dataset.map(d => d.name).join(', ')
```

graphieros:
@alexdln 

Thank you for your careful review ❤️
I've been known to leave logs 😬

app/pages/compare.vue
alexdln: @graphieros Maybe it's just needed to be added z-index to the active tab and rounded

Image

graphieros: ButtonGroup applies [&>*:not(:first-child)]:border-s-0
It is not adapted for tabs, I reverted to the previous implementation.

alexdln: Ouch, weird...

alexdln:
@graphieros great work, as always 🚀

serhalp:
Something interesting about this is that we could lift the 4 package limit when using the chart view, since it extends vertically rather than horizontally 🤔

graphieros:
> Something interesting about this is that we could lift the 4 package limit when using the chart view, since it extends vertically rather than horizontally 🤔

Future proof horizontal bars !
But there must exist a layout for a table with more packages

PR #1975, opened 2026-03-07
fix: use authors hash for bsky author profile endpoint cache key

userquin:
### 📚 Description

This PR uses a hash from authors for the bsky cache endpoint api to avoid `Error:  [cache] Cache write error. ENAMETOOLONG: name too long, open` errors.

Maybe we can even sort authors by name...


userquin:
No idea about loading JWKs error at CI, on my local `nr build:test` is fine...

userquin:
superseded by #1982

userquin:
we need to find a better approach than using hash

PR #1985, opened 2026-03-07
refactor: use interest invokers for contributor hovercard

knowler:
Refactoring #1596 to use interest invokers, the Popover API, and anchor positioning instead. This is partly an experiment, so don’t let this block you from iterating on your PR @userquin.

Using these web platform APIs means we don’t need to add a whole lot of complexity to pull this off. The only JS we need is to update the `activeContributor` using the `beforetoggle` event for the popover.

This only works in Chromium, so we have a couple of options:

1. Treat the hovercards as a progressive enhancement and work iteratively (polyfill later). If so, this PR is practically ready—I have a few changes I’d like to make though.
2. Try to polyfill the interest invoker behaviour now using the Popover API (well supported). I might end up doing this.

Anchor positioning is only Baseline Newly Available as of December, but it does already have 81% availability in browsers globally.

I personally don’t think it’s worth it to try and create behavioural parity across all of our browser targets here. The content is likely available if the user just clicks through the GitHub links. The only thing that is missing is the governance roles, which might might be able to find a solution (grouping under separate headings would be the most easily accessible, especially from an information architecture perspective).

## TODO

- [ ] Add contextual information for social links in hovercard
- [ ] Re-add exit animation
- [ ] Find a workaround for the double-trigger issue (https://issues.chromium.org/issues/490673948)

### DIFF
--- a/app/components/CallToAction.vue
+++ b/app/components/CallToAction.vue
@@ -42,7 +42,7 @@ function handleCardClick(event: MouseEvent) {
 
 <template>
   <div>
-    <h2 class="text-lg text-fg uppercase tracking-wider mb-6">
+    <h2 id="get-involved" class="text-lg text-fg uppercase tracking-wider mb-6">
       {{ $t('about.get_involved.title') }}
     </h2>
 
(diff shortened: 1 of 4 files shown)


knowler:
I’m going to close this. I think the approach would be great, however, interest invokers themselves are at risk in standards land. We can still glean from them with regard to some of the good research that has be done for them (with regard to accessibility, etc.).

PR #1986, opened 2026-03-07
fix: adjust logo sizing on about page

vladh:
### 🔗 Linked issue

N/A

### 🧭 Context

Currently, the logo grid on the about page has some issues.

First, the hover background is uneven, and in some parts is close to colliding with the logos. In particular, look at the lack of horizontal spacing on the Vercel logo, and the general lack of spacing on the Open Source Pledge logo.

<details>
Screen Shot 2026-03-07 at 22 45 47
</details>
Secondly, some logos are much bigger than others, which makes them look out of place, and also makes the whole grid more irregular. Look at the Nuxt and Vue.js logos, and how big they are relative to the logos above and below them.

<details>

</details>

### 📚 Description

I added more padding to all of the logos.

I replaced the previous per-logo `padding-block` adjustment with a more reliable per-logo `max-height` adjustment. This makes it easier to make the logos look consistent with each other.

I added a negative horizontal margin to the logo container, so that logos are flush with the left and right edges.

I generally adjusted all dimensions to make the logos look better.

cc @alexdln 🙏🏻 

<details>
<summary>Before and after, no hover</summary>

### Before

Screen Shot 2026-03-07 at 22 43 20

### After

Screen Shot 2026-03-07 at 22 43 24
</details>
<details>
<summary>Before and after, with hover background (all hover effects enabled for illustration)</summary>

### Before

Screen Shot 2026-03-07 at 22 45 47

### After

Screen Shot 2026-03-07 at 22 45 52
</details>
<details>
<summary>Before and after on mobile</summary>

### Before

Screen Shot 2026-03-07 at 22 46 53

### After

Screen Shot 2026-03-07 at 22 46 56
</details>


vladh:
Test failures seem unrelated.

alexdln:
They had fixed sizes specifically for consistency in effects. When one element has a hover height of 60px, and the next one has a hover height of 40px, it looks like a bug (and is a bit inconvenient for navigation).

To compensate for this alignment, normalizingIndent was added to the logo settings.

The voidzero group is a special case, since this appearance and such indentations are a strict correspondence to how this group was styled by the authors of the project

Maybe we should make the vertical height bigger for all items (_equal for each, but bigger, f.e. not h-11 but h-15_), but about the other edits - it seems to me that this is trading one problem for another

_P.S. A few tips: create new branches instead of working in main - this will make future support easier for both you and the maintainers. Also, use `details` for long screenshots (click edit on the comment to see how it works)_

vladh:
> When one element has a hover height of 60px, and the next one has a hover height of 40px, it looks like a bug (and is a bit inconvenient for navigation).

Okay! I'll leave this up to your judgement. For me, it's much more important for each logo to look good together with its hover background, since, when hovering, they are both visible at the same time. It's the near-collision between the logo and the edge of the hover background that jumps out as a bug to me personally.

It's true that my change makes the hover background of the different logos differently sized, but in my opinion this is more difficult to notice, because no two hover backgrounds are visible at the same time.

alexdln:
@vladh What do you think about change the effect to scale-120? It will have the same transition as the avatars below, and we'll eliminate the problems with different effects and different sizes for each logo

vladh:
> @vladh What do you think about change the effect to scale-120? It will have the same transition as the avatars below, and we'll eliminate the problems with different effects and different sizes for each logo

Sorry, I didn't quite get this. I'm not sure what you mean by “the effect” that should be changed, and it's also not clear to me what scaling things up would do. If you can describe this more specifically, I'm happy to try it out.

alexdln (approved):
Sorry for the long wait, but I couldn't find a suitable solution that would both preserve your comments and improve the interface to my thoughts. It seems like the current version is a good combination of our ideas. Let me know if you have any thoughts or if it still doesn't seem to work to you

vladh:
@alexdln I think this is a great solution! Thanks for figuring this out. 😊 ❤️

PR #1992, opened 2026-03-08
fix: prevent semver filter placeholder text from being truncated on s…

iiio2:
On iPhone SE 2nd gen (375px width, iOS 14.6), the semver filter placeholder text "Filter by semver (e.g. ^3.0.0)" in the Versions section gets truncated/cut off because the input field doesn't have enough horizontal space. The input uses font-mono at text-xs (12px) and shares its row with an info icon, gap spacing, and padding — leaving insufficient room for the full placeholder on narrow viewports.

Fix: Added max-sm:placeholder:text-2xs class to the filter input in `Versions.vue`. This reduces the placeholder font size from 12px to 11px (text-2xs, defined in uno.config.ts) only on screens below the sm breakpoint. The typed text size remains at 12px, and wider screens are unaffected.

Before: 

before

After: 

after

### DIFF
--- a/app/components/Package/Versions.vue
+++ b/app/components/Package/Versions.vue
@@ -554,7 +554,7 @@ function majorGroupContainsCurrent(group: (typeof otherMajorGroups.value)[0]): b
             :aria-invalid="isInvalidRange ? 'true' : undefined"
             :aria-describedby="isInvalidRange ? 'semver-filter-error' : undefined"
             autocomplete="off"
-            class="flex-1 min-w-0"
+            class="flex-1 min-w-0 max-sm:placeholder:text-2xs"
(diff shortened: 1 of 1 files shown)


knowler:
> maybe we don't need the description in parentheses since we have a tooltip. imo we need a separate label "Filter by Semver" and placeholder like "^3.0.0 or >= 1.4.2"

ya, the description in parentheses isn’t necessary with the tooltip and I think this suggested direction would be great.

- - -

Eventually, it would be great to associate the tooltip content with the `<input>` via `aria-describedby`, but that’s a task for another PR (and would be blocked on changes intend to make to tooltips soon).

ghostdevv:
It sounds like we won't accept this change at the moment, please reopen if I'm wrong :pray:

PR #1995, opened 2026-03-08
refactor: move fund link

### LINKED ISSUE #1835: Make the "fund" link more visible
### Describe the feature

Npm has this big button that allows package maintainers to redirect their users to their funding platform (in the `@babel/core` case it's linking to OpenCollective):

Image

On npmx that's much less visible:

Image

I wonder if there is something we could to to make it more prominent, like adding some accent color to it or making it bigger, maybe moving it in its own section on the right like npm does?

OSS funding is difficult, and "a user is looking at our package on npmx" is a great time for trying to get users to donate. Maybe we should also show the button in the docs page.

### Additional information

### Final checks

parion:
resolves #1835 

### 🧭 Context

For the reasons given in the attached issue, the fund button needed more visibility on package pages. I moved the fund link from the external link list to the top button group. This also enables the `f` key to activate the fund link.

**Please provide any feedback on this design change!** I'm noticing this button bar is getting pretty wide, I was considering scaling it down.

#### Desktop View
image

#### Mobile View
image

### DIFF
--- a/app/pages/package/[[org]]/[name].vue
+++ b/app/pages/package/[[org]]/[name].vue
@@ -729,6 +729,16 @@ onKeyStroke(
   },
 )
 
+onKeyStroke(
+  e => keyboardShortcuts.value && isKeyWithoutModifiers(e, 'f') && !isEditableElement(e.target),
+  e => {
+    if (!fundingUrl.value) return
+    e.preventDefault()
+    navigateTo(fundingUrl.value, { external: true, open: { target: '_blank' } })
+  },
+  { dedupe: true },
+)
+
 const showSkeleton = shallowRef(false)
 </script>
 


RevanthPrk:
Hey @parion,

This looks really cool❤️
But I just felt one thing after seeing your screen shots

Can we change the order of button group? to this
Docs, Code, Compare, Fund

Instead of Fund coming first?

Just my thought and perspective 😇

parion:
image

How's that looking?

RevanthPrk:
@parion This is great now🥳

danielroe:
that button bar is semantically only for things that are 'on npmx' rather than links out

... so I love the highlighting for funding, but I think we need another place for it

Issue #1707, opened 2026-02-27
Anomaly feature should be removed


matthewp:
### Describe the bug

The anomaly feature uses a hard-coded list of packages to adjust: https://github.com/npmx-dev/npmx.dev/pull/1636/changes#diff-c4a16145ea424bfd4f2c1fc0672745fa205c2fb29918b348c0887383b5192e7cR5-R9

This makes npmx an unreliable project when looking at the data and is a vector of abuse. Contributors' other projects look better than competitors by comparison, whether the anomaly is correct or not.

The feature should be removed until/unless a systematic approach is implemented that works the same for all packages.

### Logs

```shell-script

```

danielroe:
anyone can open a PR, and this is 100% in the open so please do call me out if we ever do something underhanded.

you're aware that this _decreases_ vite's downloads?

matthewp:
No one cares about total downloads, only that the chart goes up and to the right.

Self-reporting doesn't fix the problem that some projects (ie those of contributors and enthusiasts) are going to have better looking charts than those who aren't spending the time reporting this (who might not even be aware of npmx). Also no one is reporting anomalies that make the chart look worst. The data is tainted by mistakes already, don't taint it further with subjectivity.

Data is either objective or its not. I'd be happy with a solution here that fixed anomalies in an objective and systematic way.

TheAlexLichter:
> Also no one is reporting anomalies that make the chart look worst

This is literally the only anomaly that exists right now. Bumping Vite's downloads down, so "worsen" them (in your language).

> I'd be happy with a solution here that fixed anomalies in an objective and systematic way.

What kind of systematic way would that be that doesn't detect false positives (think of a surprising popularity spike á la openclaw) for example?

---

On another note: This feature is clearly described and can be toggled easily (see below)

<img width="747" height="560" alt="Image" src="https://github.com/user-attachments/assets/c89edf6c-02c1-42c9-b6bf-f7f10b7f6b21" />

<img width="734" height="654" alt="Image" src="https://github.com/user-attachments/assets/d56d4d43-7d38-4966-a0a6-20567165bbf3" />

matthewp:
> This is literally the only anomaly that exists right now. Bumping Vite's downloads down, so "worsen" them (in your language).

But Vite isn't the only package affected by this anomaly, hence the problem with self-reporting.

sacrosanctic:
> Data is either objective or its not.

[...]

PR #1983, opened 2026-03-07
fix: the weekly data anomaly detection was broken for the Svelte anomalies

samal-rasmussen:
The weekly data anomaly detection was broken for the Svelte anomalies

### 🔗 Linked issue

Follow up to PR: https://github.com/npmx-dev/npmx.dev/pull/1934

### 📚 Description

The previous code was being a bit too naive about weekly range data anomaly detection, and so wasn't including all the affected weeks properly. You actually have to check the weekly range.

PR #1934, opened 2026-03-04
feat: add Svelte CI anomalies to download-anomalies.data.ts

samal-rasmussen:
Add Svelte CI anomalies to download-anomalies.data.ts

PR #1997, opened 2026-03-08
fix: replace manual anomalies with a hampel filter

matthewp:
### 🔗 Linked issue

Previous issue: https://github.com/npmx-dev/npmx.dev/issues/1707

### 🧭 Context

The current implementation of anomaly removal is biased due to the manual nature.

This replaces the implementation with one that uses a hampel filter to automatically remove deviations.

### 📚 Description

The important part here is `applyHampelCorrection`.

It goes over each data point and creates a sliding window of data points, by default **3** on each side.

Of those data points it finds the median, since a spike can't pull it off center.

Then it measures how spread out the neighbors are by calculating the MAD ([see here](https://en.wikipedia.org/wiki/Median_absolute_deviation)).

Then it gives the data point a score and if it's above a threshold then it gets replaced with the median of that window and marked as being an anomaly.

Here are two articles about how hampel filters work:

- https://blogs.sas.com/content/iml/2021/06/01/hampel-filter-robust-outliers.html
- https://www.mathworks.com/help/signal/ref/hampel.html

Here's a screenshot of Vite which still shows its spike removed after this change.

Screenshot from 2026-03-08 17-24-36


danielroe:
this looks very promising! tagging @jycouet who may have some thoughts on this :pray:

jycouet:
Great to have a second look into this.

You probably checked my initial PR, https://github.com/npmx-dev/npmx.dev/pull/1636 I started with hampel implementation 👍

If I remember correctly, the sweet spot was 2.5 or 3 for vite. But this setting would hide "great start" of a lib. (eg: 0 0 0 0 0 0 20000)
I will pull the branch later _(I'm on phone atm)_ as I'm curious to see it, maybe you have a more robust implem' 👍

Another note, when the start or end of the chart is in an anomalie period, any algo can't fix it. A good example is: if you start the chart in the middle of the vite spike. 

Maybe we should add some tests around all this ? (to keep the intent) 

Another note 2: I would love to have more than just vite there today ! 😅

graphieros:
> Another note 2: I would love to have more than just vite there today ! 😅

@jycouet corrections were applied to Svelte in #1934 & #1983

jycouet:
> > Another note 2: I would love to have more than just vite there today ! 😅
> 
> @jycouet corrections were applied to Svelte in #1934 & #1983

That's the drawback of answering on phone
It shows my age 😅

jycouet:
Thanks for this! A few thoughts after pulling the branch:

Reproduction URL that matters:
http://127.0.0.1:3000/package/vite?modal=chart&start=2025-03-09&end=2026-03-07
_(because I suspect there's also a day-of-the-week issue atm, see https://github.com/npmx-dev/npmx.dev/issues/2005 let's not speak about this here)_

### "Great start" test cases:
Packages like `@bramus/specificity` or `@sveltejs/sv-utils` are good real-world examples of the `0 0 0 0 0 → lots of downloads` pattern.
Would be great to add these as test cases (or manual verification) to see how Hampel manages them (or to find good defaults)

### PR suggestion
Could we expose the Hampel filter as its own (separate from manual correction), with tweakable halfWindow / threshold sliders? That way we can compare manual vs Hampel side by side and find the right balance before committing to one approach?
Let me know if you want me to help or if you want to do it?

_// Happy coding_

matthewp:
> Could we expose the Hampel filter as its own (separate from manual correction), with tweakable halfWindow / threshold sliders? That way we can compare manual vs Hampel side by side and find the right balance before committing to one approach?
> Let me know if you want me to help or if you want to do it?

I don't follow, as its own what?

jycouet:
> own

Something like 
image

It's own controls. _(It's more to be able to tests different senarios)_

matthewp:
Yes I can do that.

matthewp:
@jycouet I added sliders and addressed your concerns.

jycouet:
@matthewp I pulled your branch, and see the 2 controls

And with http://127.0.0.1:3000/package/vite?modal=chart&start=2025-03-13&end=2026-03-11 I get:

image

I tried also, removing my local storage to start from "fresh", but I still see the big spike. 
You have the same on your end? I'm missing something?

I didn't look at the code at all, let me know how you want to proceed.

matthewp:
@jycouet That specific date range was affected only. When trying against today it didn't show that. Bumping the default half window to 4 fixes your problem.

jycouet:
@graphieros I think that the merge broke the test

@matthewp I tried with your [fe605db](https://github.com/npmx-dev/npmx.dev/pull/1997/commits/fe605dbaa6cf81ebafcc18ae767ab1255c3ee29a), If I understand your branch, you need to set  Hampel window (4) + Hampel threshold (4) + Apply correction? Right?! And the apply correction is disconnected from the actual behaviorr.

To be able to move forward, I would suggest to: 
1/ start from main
2/ add just these 2 controls Hampel window and Hampel threshold. If it's 0, it's not used, if it's x, it's used by x. Like this, Apply correction stays like main is working.
Like this we are able to play with all values and option to see what to do & understand better.

My thinking is that without manual data, I don't see how to avoid this: 
http://127.0.0.1:3000/package/vite?modal=chart&start=2025-08-15&end=2026-03-13
image

graphieros:
> @graphieros I think that the merge broke the test

Sorry for that, removed the culprit line

matthewp:
@jycouet Removing the biased manual data manipulation is the primary motivation of this pull request. So I'll keep this one until we can resolve whatever issues get us there.

Is "Vite must look perfect at all times" really the primary requirement of this feature? Vs. correcting the unknown (hundreds? thousands?) of packages that likely have anomalies that are not corrected on npmx.dev currently? Currently Vite only looks bad if you enter a date range that no one is going to look at unless they are attempting to play with the controls. 

Tell me what the requirements are and I'll figure out a solution.

jycouet:
@matthewp I understand that it's the motivation, and if we manage to have a better result than manual without manual, I'm all for it. I'm just trying to see how to test the feature! So my idea was to keep everything as is (where we can disable everything), and ADD this feature to be able to see behavior on multiple packages.
If we don't do this, I don't see how we can compare data... We have to play with this branch + live site? I feel that it's less practical. but if you have a way, let me know, I'll follow your advice.

It all started with vite, because it's obvious that there is an issue for a few weeks... But we now have a bit more: https://github.com/npmx-dev/npmx.dev/edit/main/app/utils/download-anomalies.data.ts

> Is "Vite must look perfect at all times" really the primary requirement of this feature? 

I think it's not only vite, and what I like with `pnpm` community, is that anyone can come and improve it by adding a manual data. Or by having a cool algo that could help everyone 🎉

> Currently Vite only looks bad if you enter a date range

Also when we look at the daily view
image

> the default half window to 4 fixes

That's also a "big" decision, what default value to set, and the bigger the more we affect other parts.

> correcting the unknown (hundreds? thousands?) of packages that likely have anomalies that are not corrected on npmx.dev currently?

Manual data is by package right now, but if one day we know something affecting ALL packages, we could also add this one info and all packages will have it 💪

> Tell me what the requirements are and I'll figure out a solution.

I don't know what to answer here, you opend the issue & the PR ^^. I think that we are trying to do better things all together :)
Today I have to say that I like the manual data _(I'm biased, I did it ^^)_
And as I said as first reply here: 
> You probably checked my initial PR, https://github.com/npmx-dev/npmx.dev/pull/1636 I started with hampel implementation. But it didn't cover all situations. I'm happy to see a better implementation.

graphieros:
I think acceptance criteria could be summed up as follows:

1. obvious isolated spikes should be corrected
2. genuine growth ramps should not be flattened
3. low volume packages should not get false positives
4. changing the selected date range should not reintroduce known anomalies
5. daily and aggregated views should behave consistently enough

matthewp:
@jycouet I don't agree that the manual data manipulation is better, it misses an unknown number of packages and introduces bias into the entire site's data set. I didn't review or know about your PR when I submitted this one, to be clear, but I would assume that it did it better than the current state as well.

I think looking at this by wanting specific packages to have nice looking charts is the wrong approach. You are devaluing having a consistent, unbiased method for detecting anomalies. Not having false positives is what is most important, some false negatives will happen, but it will be universally applied.

I think the way to review a change like this is not focusing on specific desired results, but by looking 1) at the documentation on hampel detection and whether it's a good algorithm for this sort of problem. It's used within Matlab which I think is a good indicator that it is. And then 2) is the algorithm implemented correctly, is the code good, etc.

I know that the "everyone chips in with what they know" feels warm and good, and for some things that's true, but when it comes to data it's not the case, everyone chipping in with their individual biases results in a distorted data set. Not maliciously, but biased nevertheless through selection bias, recall bias, survivorship bias, etc.

As for the individual recent changes, I don't know what the correct default values should be, the original values seemed correct to me, I adjusted them based on the view you were looking at, but it wasn't completely clear to me what you were looking at as I do not get the same results locally. So happy to roll that back to previous values.

But first I will reiterate that I am not sure what the requirements are here. I again ask, is it that Vite must never show this spike, regardless of what controls and date ranges are used?

matthewp:
I would like some maintainer feedback please, cc @danielroe do you have views here, thanks.

matthewp:
@graphieros Thank you, which of those criteria do you think are not passing currently? I think it's just (4), and for the date range in @jycouet's last screenshot, only for Vite. Is that correct? Or is there another?

graphieros:
@matthewp it does appear like 4 is the current blocker, from what I see. Is the correction applied at the right stage of the data pipeline ?

I think we need to define a set of packages to test against, like (probably non exhaustive):

- isolated mature-package spike
- great start new package
- low vol package
- cropped range edge anomaly
- daily vs weekly etc

jycouet:
Thank you @matthewp for sharing, I understand better your view.
Not easy to satisfy everyone. One cool thing tho is that you can play with controls to see the view you perfer _(and it will remain)_.

---

I ran some benchmark on: `@bramus/specificity`, `knip`, `remult`, `vite`, `sv` that represent some different styles.
The green line is the hampel one, the other one is the raw value

sorry for the flashbang 🤪

bramus-specificity_BENCHMARK

knip_BENCHMARK

remult_BENCHMARK

sv_BENCHMARK

vite_BENCHMARK

ghostdevv:
what's needing to be done still? minus dealing with the merge conflicts :p

cc: @graphieros @jycouet

graphieros:
> what's needing to be done still? minus dealing with the merge conflicts :p

This PR looks like it is still pending.

jycouet:
Hey @matthewp, thanks for pushing on this, it made me dig deeper Hampel again.

I genuinely think a pure statistical approach is elegant, and in a perfect world I'd love to get there. But as the benchmarks showed, Hampel still struggles with real-world edge cases (great-start packages, boundary spikes on date range changes, low-volume data). These aren't minor => they affect how people read the charts.

As the correction data is open, anyone can see what was corrected and why. Transparency seems like a better answer to the bias concern than removing human judgment entirely.

I think that we can close this PR for now _(or move as draft so that it's not in list of "to look")_.
The conversation was valuable and I'm open to revisiting a hybrid approach down the road.

matthewp:
> As the correction data is open, anyone can see what was corrected and why. Transparency seems like a better answer to the bias concern than removing human judgment entirely.

Transparency in this case would explicitly labeling the correction data as a biased sample. Are you willing to do that?

I'd still like to work on this PR, happy to move into a draft or close, whatever you prefer.

Also open to a separate PR that removes the correction data entirely, so all packages are equally flawed, if you are open to that.

jycouet:
The corrections are primarily based on known npm reporting events (outages, API bugs), not aesthetic preferences for specific packages. I wouldn't call that a biased sample -> it's closer to an errata log.

I was thinking that it's a good thing for everyone, apparently not. I didn't mean to corrupt data, just to make charts more accurate when we know something went wrong on npm's side. When a known anomaly adds 100M+ downloads/week for over a month, it's not exactly a rounding error. Leaving that uncorrected feels like the bigger data integrity issue to me.

Adding a Hampel option alongside the current corrections sounds great to me. Removing "Known anomalies" that I believe aren't destroying data, I'm less sure about, but I'm just an enthusiast who likes coding and learning here, nothing else.

I'm sure the project will lean toward the best, no matter the direction, all good. 
I'll go spend my time and energy on positive stuff. Good luck and happy coding 👍

graphieros:
I suggest we:

- Keep the Hampel implementation and UI controls
- Do not remove manual corrections in this PR.

If Hampel (or another method) proves reliable across all scenarios, we can then later reduce reliance on manual corrections, or keep them only for extreme known events (?), but we should earn that transition **with data**, not assume it upfront.

In the meantime, this PR can be converted to a draft, if this works for you.

matthewp:
@jycouet

> The corrections are primarily based on known npm reporting events (outages, API bugs), not aesthetic preferences for specific packages. I wouldn't call that a biased sample -> it's closer to an errata log.

Known to a subset of people who contribute to this project, that's known as [ascertainment bias](https://catalogofbias.org/biases/ascertainment-bias/). The act of cleaning the data is not neutral, it encodes the cleaner's subjective knowledge into the dataset. Also related is [detection bias](https://catalogofbias.org/biases/detection-bias/).

> I was thinking that it's a good thing for everyone, apparently not. I didn't mean to corrupt data, just to make charts more accurate when we know something went wrong on npm's side. When a known anomaly adds 100M+ downloads/week for over a month, it's not exactly a rounding error. Leaving that uncorrected feels like the bigger data integrity issue to me.

It's not, because the unknown anomalies are not corrected, so when users who visit 2 package pages, one corrected and one not, the uncorrected packages looks less trustworthy, and the user is more likely to favor the package with normal looking charts.

So, unintentionally or not, putting bias into this product leads to the maintainers own packages (of which the maintainers are involved in many high profile projects) having a higher conversion rate than competing packages which have not been corrected. This is the alarming thing and why I originally signaled that the feature should just be removed.

matthewp:
@graphieros 

> If Hampel (or another method) proves reliable across all scenarios, we can then later reduce reliance on manual corrections, or keep them only for extreme known events (?), but we should earn that transition with data, not assume it upfront.

The manual correction is not reliable across all scenarios. One of the reasons I haven't been enthusiastic about coming back to this PR is because there are very high standards for a known anomaly correction mechanism used by MATLAB among other places (Hampel), and there is no standards for the known biased correction mechanism (manual). Can we apply some requirements for manual corrections in order to balance things out better?

graphieros:
> @matthewp 
> 
>> Can we apply some requirements for manual corrections in order to balance things out better?

Sounds fair enough to me.
What are you thinking about ? Documenting sources for manual corrections ?

matthewp:
Yeah something like that, also:

- Official and journalistic sources only (no self-reporting or tweets as reports)
- Applies equally to all packages named in the report. If no packages named, either apply universally or don't apply that event to any packages.

graphieros:
>  Official and journalistic sources only (no self-reporting or tweets as reports)

As decent as this would be for clinical data, it does not seem to be the case for npm downloads. Any manual correction is curated by npmx maintainers. Note that the few corrections applied correct obvious bursts, and the correction can be turned off to view the **raw data**, but we have been through this already. The source of truth remains available which is essential.

This PR should be focused on improving the correction algorithm, and its coexistence with the manual correction is a requirement for now.

matthewp:
@graphieros Happy to table that discussion. I'm a little confused though, you previously [said](https://github.com/npmx-dev/npmx.dev/pull/1997#issuecomment-4061128839)

> I think acceptance criteria could be summed up as follows:
> - obvious isolated spikes should be corrected
> - genuine growth ramps should not be flattened
> - low volume packages should not get false positives
> - changing the selected date range should not reintroduce known anomalies
> - daily and aggregated views should behave consistently enough

Are you now saying this isn't an acceptance criteria for this PR, but that it must also bring back the biased manual corrections?

graphieros:
If the algorithm checks all the boxes of listed requirements, it would basically be redundant with manual corrections. I would applaud such a feat.
 
Until then, manual correction will remain featured.

ghostdevv:
edit: see below message :pray:

ghostdevv:
Right! I asked the other core team members for their thoughts and we've decided that, for now, it is better to close this PR, on the basis that:

- A hampel filter struggles with real-world edge cases (great-start packages, boundary spikes on date range changes, low-volume data) - see above comments for why, [for example these tests from jyc](<https://github.com/npmx-dev/npmx.dev/pull/1997#issuecomment-4061352831>)
- We think that for our default, manual labelling is the best we can do for our users, since we are going to be removing clear issues in the data that arise from NPM weirdness
- We don't agree that the manual labelling is unfair to other package maintainers, and as stated before, the raw data remains accessible and transparent
- No pure statistical approach will be perfect, and we agree that with enough time and research we may find a good trade-off, and maybe to be enabled optionally - but this isn't there yet

While we acknowledge that it's not ideal to have to manually fix the inaccuracies in the data, and that we can't realistically apply this to every package which is a shame, we believe that this is currently the most reliable way to provide a better and more accurate experience without inadvertently introducing false data onto our graphs.

This PR will help inform future improvements to our data cleaning approach, so I think there was value in having this discussion and to dig deeper into hampel filters (and other automated approaches). We just haven't quite found the right solution yet - but thank you for the PR! If you have other ideas in the future, then let us know and we can play around with them to see if it's a good fit.

Issue #2013, opened 2026-03-09
chart weekly display improvement


jycouet:
### Describe the bug

This looks strange

|1 year|2years|
|-|-|
|https://npmx.dev/package/@sveltejs/kit?modal=chart&start=2025-03-10&end=2026-03-08|https://npmx.dev/package/@sveltejs/kit?modal=chart&start=2024-03-10&end=2026-03-08|
|<img width="778" height="808" alt="Image" src="https://github.com/user-attachments/assets/1b71dde4-aae0-44f6-9ccd-c006090c52a9" />|<img width="778" height="803" alt="Image" src="https://github.com/user-attachments/assets/33f10a75-cc67-40b8-b8a6-1a36e88ce474" />
|Default view when we land|Adding 1 more year to the view (going to 2024)|

### Additional context

Looking at it closely, I saw:
- that weekly grouping is happening from start to the end. When the week is not full, the last week displayed is only 1 or 2 days... so it's really small.
- Prediction is available only for month & year

---

Two options to fix this _(that I have already locally! ^^)_

### Option A
Group from End to start, and drop the partial starting week.

That's nice and easy to do. But: 
- Everyday, values are changing in the weekly mode (I love it as it's last rooling week!)
- it's not that consistant with monthy & yearly view that has a prediction and work with set periods

### Option B
Have a prediction for weekly as well!

In addition, this prediction could do a linear estimation based on the last X points _(4 by default)_

<img width="778" height="859" alt="Image" src="https://github.com/user-attachments/assets/fcbc096b-ea48-4011-9714-08200aa490de" />

or (To keep the same package)
<img width="780" height="862" alt="Image" src="https://github.com/user-attachments/assets/c494423d-c66f-4d19-8b5b-dfb4b7818f30" />

It's working pretty well with all view!!! 🚀

jycouet:
Note that npm is using more the option A.

<img width="1932" height="1162" alt="Image" src="https://github.com/user-attachments/assets/f10fd8ff-2347-4252-997f-0aa043a9e36f" />

PR #2014, opened 2026-03-09
fix: weekly chart prediction and data pipeline extraction

jycouet:
### 🔗 Linked issue

Closes: #2013

### 🧭 Context

Weekly chart displayed partial last-week buckets with very low values (1–2 days instead of 7), and prediction/estimation was only available for monthly & yearly granularity.

### 📚 Description

- Extend estimation support to weekly (and daily) granularity
- Add linear regression prediction based on the last N complete data points (configurable, default 4)
- Extract prediction & bucket logic from `TrendsChart.vue` into a new `chart-data-prediction.ts` utility
- Introduce `applyDataPipeline` that orchestrates: prediction → data correction (smoothing/averaging)
- Add a prediction points slider (0–30) in the chart filter UI
- Apply prediction *before* smoothing so partial periods don't skew nearby curves

---

Before

image

After

image

### DIFF
--- a/app/utils/chart-data-prediction.ts
+++ b/app/utils/chart-data-prediction.ts
@@ -96,7 +96,8 @@ export function linearProject(pts: number[]): number | null {
 /**
  * Estimate the full-period value for a partially-complete last bucket.
  *
- * Fallback chain: linear projection → single-point copy → proportional scale-up.
+ * Uses linear projection when enough complete lookback points are available
+ * (`>= predictionPoints`), otherwise falls back to proportional scale-up.
  * Returns the raw last value when the period is already complete or prediction is disabled.
  */
 export function extrapolateLastValue(params: {
@@ -115,9 +116,11 @@ export function extrapolateLastValue(params: {
   if (!(ratio > 0 && ratio < 1) || predictionPoints <= 0) return last
 
   const lookback = series.slice(0, -1).slice(-predictionPoints)
-  const projected = linearProject(lookback)
-  if (projected !== null) return projected
-  if (lookback.length === 1) return lookback[0]!
+
+  if (lookback.length >= predictionPoints) {
+    const projected = linearProject(lookback)
+    if (projected !== null) return projected
+  }
 
   const scaled = last / ratio
   return Number.isFinite(scaled) ? scaled : last
--- a/test/unit/app/utils/chart-data-prediction.spec.ts
+++ b/test/unit/app/utils/chart-data-prediction.spec.ts
@@ -123,12 +123,11 @@ describe('linearProject', () => {
 
 describe('extrapolateLastValue', () => {
   it('returns raw last value when ratio >= 1 (complete bucket)', () => {
-    const lastDateMs = Date.UTC(2025, 1, 1)
     const result = extrapolateLastValue({
       series: [100, 200, 150],
       granularity: 'monthly',
-      lastDateMs,
-      referenceMs: Date.UTC(2025, 2, 1), // bucket fully elapsed
+      lastDateMs: Date.UTC(2025, 1, 1),
+      referenceMs: Date.UTC(2025, 2, 1),
       predictionPoints: 4,
     })
     expect(result).toBe(150)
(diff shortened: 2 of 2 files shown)


graphieros:
Setting the start date to the previous day yields a prediction of 0.
Setting previous start date to a mod != 7 yields predictions that do not seem correct.

Example with vue, keeping the default 14-day prediction window:

image

| day - 1| day - 2| day -3| day -4 | day-5 | day -6 |
|-------|--------|-------|--------|-------|-------|
|image|image|image|image|image|image|

app/utils/chart-data-prediction.ts
```diff
@@ -0,0 +1,157 @@
+import type { ChartTimeGranularity } from '~/types/chart'
+import { applyDataCorrection, type ChartFilterSettings } from '~/utils/chart-data-correction'
+
+export const DEFAULT_PREDICTION_POINTS = 4
+
+// ---------------------------------------------------------------------------
+// Bucket boundaries (UTC)
+// ---------------------------------------------------------------------------
+
```
graphieros, line 9:
```ts
const DAY_MS = 86_400_000

function getUtcDayStart(ts: number): number {
  const d = new Date(ts)
  return Date.UTC(d.getUTCFullYear(), d.getUTCMonth(), d.getUTCDate())
}

// Monday-based week start in UTC
function getWeeklyBucketStartUtc(ts: number): number {
  const dayStart = getUtcDayStart(ts)
  const day = new Date(dayStart).getUTCDay()
  const diffFromMonday = (day + 6) % 7
  return dayStart - diffFromMonday * DAY_MS
}
```

app/utils/chart-data-prediction.ts
```diff
@@ -0,0 +1,157 @@
...
+  const [y, m, day] = d.split('-').map(Number)
+  if (!y || !m || !day) return null
+  return Date.UTC(y, m - 1, day + 1)
+}
+
+/** Start of the bucket containing `ts` (inclusive). */
+export function getBucketStartUtc(ts: number, g: ChartTimeGranularity): number {
+  const d = new Date(ts)
+  if (g === 'yearly') return Date.UTC(d.getUTCFullYear(), 0, 1)
+  if (g === 'monthly') return Date.UTC(d.getUTCFullYear(), d.getUTCMonth(), 1)
```
graphieros, line 43: Now with

```ts
if (g === 'weekly') return getWeeklyBucketStartUtc(ts)
```

app/utils/chart-data-prediction.ts
```diff
@@ -0,0 +1,157 @@
...
+  return Date.UTC(d.getUTCFullYear(), d.getUTCMonth(), d.getUTCDate())
+}
+
+/** End of the bucket containing `ts` (exclusive). */
+export function getBucketEndUtc(ts: number, g: ChartTimeGranularity): number {
+  const d = new Date(ts)
+  if (g === 'yearly') return Date.UTC(d.getUTCFullYear() + 1, 0, 1)
+  if (g === 'monthly') return Date.UTC(d.getUTCFullYear(), d.getUTCMonth() + 1, 1)
+  if (g === 'weekly') return getBucketStartUtc(ts, 'daily') + 7 * 86_400_000
```
graphieros: Replaced with

```ts
if (g === 'weekly') return getWeeklyBucketStartUtc(ts) + 7 * DAY_MS
```

app/utils/chart-data-prediction.ts
```diff
@@ -0,0 +1,157 @@
...
+  granularity: ChartTimeGranularity
+  lastDateMs: number
+  referenceMs: number
+  predictionPoints: number
+}): number {
+  const { series, granularity, lastDateMs, referenceMs, predictionPoints } = params
+  const last = series.at(-1) ?? 0
+
+  // Weekly dates are stored as timestampEnd; bucket start is 6 days earlier
+  const bucketTs = granularity === 'weekly' ? lastDateMs - 6 * 86_400_000 : lastDateMs
```
graphieros:
```ts
const bucketTs = lastDateMs
```

jycouet:
Sorry @graphieros when trying to address `the rabbit`, it destroyed the feature!
- prediction = 0, should mean no prediction, so the raw value
http://127.0.0.1:3000/package/vue?modal=chart&start=2025-03-11&end=2026-03-09
image

Could you check it again? :/

Note: could you add the link to the package with modal & range, it goes faster to find it. Thx

app/utils/chart-data-prediction.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  const bucketTs = lastDateMs
+  const ratio = getCompletionRatio(bucketTs, granularity, referenceMs)
+
+  if (!(ratio > 0 && ratio < 1) || predictionPoints <= 0) return last
+
+  const lookback = series.slice(0, -1).slice(-predictionPoints)
+  const projected = linearProject(lookback)
+  if (projected !== null) return projected
+  if (lookback.length === 1) return lookback[0]!
```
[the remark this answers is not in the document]
graphieros: @jycouet wdyt ?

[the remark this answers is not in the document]
jycouet: `slice(-predictionPoints)` already takes the min naturally. 
If you have 2 points and ask for 4, it just returns the 2 available. No crash, no overflow

To me it's ok. I don't really see why it's flagged!

graphieros:
@jycouet 
can you check the impacts on yearly predictions ?
Take any package, and just switch the granularity to yearly ?

Example with Svelte (open modal, switch to yearly):

| Branch | Prod |
|--------|------|
|image|image|

jycouet:
> can you check the impacts on yearly predictions ? Take any package, and just switch the granularity to yearly ?
> 
> Example with Svelte (open modal, switch to yearly):

Good catch. 
I fixed it by using `scale-up` (like before this PR) when we don't have enough points for prediction. 

That's is sooo cool because: 
- If you set `prediction 0`, it's removing `scale-up` & `linear`, you have the raw raw value _(that was not possible before)_
- It's addressing the `rabbit` question we had earlier! haha

PR #2017, opened 2026-03-10
feat: flag git: and https: dependencies as security concern

### LINKED ISSUE #1084: flag `git:` and `https:` dependencies
They can be manipulated, see https://bsky.app/profile/dsherret.bsky.social/post/3me2bmgc32k2k

ndpvt-web:
## Summary

Adds detection and visual flagging for dependencies that use `git:`, `https:`, or `git+https:` URLs instead of npm registry versions.

These URL-based dependencies bypass npm registry integrity checks and can be manipulated, as noted in the linked discussion. This implementation helps users identify potential security concerns in their dependency trees.

## Changes

- Added `isUrlDependency()` utility function in `server/utils/dependency-analysis.ts` to detect URL-based dependencies
- Added `scanUrlDependencies()` function to scan dependency tree for URL-based deps
- Updated `shared/types/dependency-analysis.ts` with TypeScript types for URL dependency tracking
- Modified `server/utils/dependency-resolver.ts` to include URL dependency scanning in analysis
- Enhanced `app/components/Package/Dependencies.vue` to display warning icons next to URL-based dependencies with tooltips explaining the security concern

## Testing

The implementation follows existing patterns in the codebase and integrates seamlessly with the current dependency analysis system.

Closes #1084

## AI Disclosure

This PR was authored with the assistance of an AI coding assistant (Claude by Anthropic). All changes have been reviewed for correctness and follow the project's existing patterns and conventions.


ndpvt-web:
Thanks for the thorough review! I've addressed all feedback:

1. **Transitive URL dep scanning** - Now iterates over all resolved packages instead of just root, covering dependencies at any depth as required by #1084.
2. **Aligned URL patterns** - `isUrlDependency` now matches all patterns handled by `resolveVersion` (git+http, git+ssh, http, file).
3. **Cached URL dep lookups** - Added computed map in Dependencies.vue to avoid repeated array scans.
4. **Removed orphaned field** - Cleaned up unused `url` field from ResolvedPackage interface.

Updated in the latest commit.

ndpvt-web:
Addressed the additional feedback:

1. **devDependencies scoping** - Now only includes devDependencies for the root package, avoiding misleading warnings for transitive packages.
2. **Parallel scanning** - Replaced sequential loop with `mapWithConcurrency` for consistency with OSV queries.

Updated in the latest commit.

ndpvt-web:
Addressed both issues:

1. **Depth correction** - URL dependencies now correctly record their child depth (root deps -> direct, direct deps -> transitive) instead of using the parent package's depth.
2. **Error handling** - Removed silent error swallowing; scan failures now propagate consistently with OSV error handling patterns.

Updated in the latest commit.

shared/types/dependency-analysis.ts
```diff
@@ -192,6 +205,8 @@ export interface VulnerabilityTreeResult {
   vulnerablePackages: PackageVulnerabilityInfo[]
   /** All deprecated packages in the tree */
   deprecatedPackages: DeprecatedPackageInfo[]
+  /** All dependencies using git: or https: URLs */
+  urlDependencies: UrlDependencyInfo[]
```
serhalp, line 209: The two fields above use the term "package", so it would be nice to have consistency here to avoid giving a false sense of difference:
```suggestion
  /** All packages using unsafe `git:`, `https:` (etc.) URLs in the tree */
  packagesWithUrlVersion: UrlDependencyInfo[]
```

shared/types/dependency-analysis.ts
```diff
@@ -180,6 +180,19 @@ export interface DeprecatedPackageInfo {
...
 }
 
+/**
+ * URL-based dependency info (git:, https:)
+ */
+export interface UrlDependencyInfo {
+  name: string
+  /** The git: or https: URL */
+  url: string
+  /** Depth in dependency tree: root (0), direct (1), transitive (2+) */
```
serhalp, line 190: 🤔 how could root ever happen? this is only a possibility in the `dependencies` array of another package, isn't it?

app/components/Package/Dependencies.vue
```diff
@@ -164,6 +178,19 @@ const numberFormatter = useNumberFormatter()
...
             </LinkBase>
+            <TooltipApp
+              v-if="getUrlDepInfo(dep)"
+              class="shrink-0 text-orange-700 dark:text-orange-500"
+              :text="getUrlDepInfo(dep)!.url"
+            >
+              <button
+                type="button"
+                class="p-2 -m-2"
+                :aria-label="`git/https dependency: ${getUrlDepInfo(dep)!.url}`"
```
serhalp, line 189: This hardcoded English string needs to be internationalized. Check out guidance in CONTRIBUTING.md if you aren't familiar!

server/utils/dependency-analysis.ts
```diff
@@ -255,6 +256,70 @@ function getSeverityLevel(vuln: OsvVulnerability): OsvSeverityLevel {
...
+/**
+ * Check if a dependency URL is a git: or https: URL that should be flagged.
+ */
+function isUrlDependency(url: string): boolean {
+  return (
+    url.startsWith('git:') ||
+    url.startsWith('git+') ||
+    url.startsWith('http:') ||
+    url.startsWith('https:') ||
+    url.startsWith('file:')
```
serhalp, line 268: Per https://docs.npmjs.com/cli/v11/configuring-npm/package-json#git-urls-as-dependencies, these are also valid and insecure:
- `github:foo/bar`
- `gist:foo`
- `bitbucket:foo/bar`
- `gitlab:foo/bar`
- `foo/bar` (this means GitHub)

🤔 We can't just assume any protocol is insecure, as `npm:` is fine.

Maybe something like this works?
```suggestion
    // starts with any `foo:` except `npm:`
    url.match(/^(?!npm:)[^:]+:/) ||
    // foo/bar
    url.match(/^[^/]+/[^/]+$/)
```

serhalp: oooh wait a sec, I think we can just do...
```suggestion
    !semver.valid(url) && !url.startsWith('npm:')
```

👀 although we'd have to think through ["loose" mode and coercion](https://npmx.dev/package/semver) to be sure we're using the same logic as package managers do.

server/utils/dependency-analysis.ts
```diff
@@ -289,6 +354,14 @@ export const analyzeDependencyTree = defineCachedFunction(
         return depthOrder[a.depth] - depthOrder[b.depth]
       })
 
+    // Scan for git: and https: URL dependencies in all packages
+    const urlDepResults = await mapWithConcurrency(
+      packages,
+      pkg => scanUrlDependencies(pkg.name, pkg.version, pkg.depth, pkg.path),
+      OSV_DETAIL_CONCURRENCY,
```
serhalp, line 361: We should create a separate constant for this, as it's unrelated to fetching OSV details

server/utils/dependency-analysis.ts
```diff
@@ -255,6 +256,70 @@ function getSeverityLevel(vuln: OsvVulnerability): OsvSeverityLevel {
...
+ * Returns a map of package names to their URL dependencies.
+ */
+async function scanUrlDependencies(
+  name: string,
+  version: string,
+  depth: DependencyDepth,
+  path: string[],
+): Promise<UrlDependencyInfo[]> {
+  try {
+    const packument = await fetchNpmPackage(name)
```
serhalp, line 283: This will be very slow, since it's re-fetching the packument for every package in the tree. [This data was already constructed to begin with by fetching all these packuments](https://github.com/npmx-dev/npmx.dev/blob/d9a3178f8553e866474afca2aa113f45f1167d27/server/utils/dependency-resolver.ts#L142) and the `path` that "leads" to a given package is returned, so it seems it should be possible to derive what we need directly, no? 🤔

serhalp: If it helps simplify things, I think it would even be fine to just [attach a new field for this in `resolveDependencyTree` directly](https://github.com/npmx-dev/npmx.dev/blob/d9a3178f8553e866474afca2aa113f45f1167d27/server/utils/dependency-resolver.ts#L176).

server/utils/dependency-analysis.ts
```diff
@@ -255,6 +256,70 @@ function getSeverityLevel(vuln: OsvVulnerability): OsvSeverityLevel {
   return 'unknown'
 }
 
+/**
+ * Check if a dependency URL is a git: or https: URL that should be flagged.
+ */
+function isUrlDependency(url: string): boolean {
```
serhalp, line 262: I think this function is a good candidate for great good unit test coverage. Let's do some table tests covering all the supported formats (see my comment below).

serhalp (commented):
🙌🏼 Thanks for the PR!

server/utils/dependency-analysis.ts
serhalp: FYI I found this live example of a package with a dependency pointing at `github:`: https://npmx-r89mhfl1x-npmx.vercel.app/package/react-native-gauge

serhalp:
@ndpvt-web hey there! are you interested in continuing to work on this PR?

PR #2021, opened 2026-03-10
fix(ui): allow dismissing mobile search input

### LINKED ISSUE #1881: Close mobile search input on click outside
### Describe the bug

There is no way to close this search input on a subpage.
Image

### Additional context

_No response_

### Logs

```shell-script

```

epifaniofrancisco:
### 🔗 Linked issue

Resolves #1881

### 🧭 Context

On mobile, when the search input was expanded on a subpage, there was no way to close it. Users were stuck with the search bar taking up the full header.

### 📚 Description

Added three ways to dismiss the mobile search:
- Click outside the header
- Press Escape key
- Click the X close button

Also fixed the search input height to match the header buttons for visual consistency.

### DIFF
--- a/app/components/AppHeader.vue
+++ b/app/components/AppHeader.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { onClickOutside } from '@vueuse/core'
 import { LinkBase } from '#components'
 import type { NavigationConfig, NavigationConfigWithGroups } from '~/types'
 import { isEditableElement } from '~/utils/input'
@@ -143,6 +144,7 @@ const route = useRoute()
 const isMobile = useIsMobile()
 const isSearchExpandedManually = shallowRef(false)
 const searchBoxRef = useTemplateRef('searchBoxRef')
+const headerRef = useTemplateRef('headerRef')
 
 // On search page, always show search expanded on mobile
 const isOnHomePage = computed(() => route.name === 'index')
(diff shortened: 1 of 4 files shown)


app/components/AppHeader.vue
```diff
@@ -200,10 +210,22 @@ onKeyStroke(
   },
   { dedupe: true },
 )
+
+onKeyStroke(
+  e =>
+    isKeyWithoutModifiers(e, 'Escape') &&
+    isMobile.value &&
+    isSearchExpanded.value &&
+    !isOnSearchPage.value,
```
serhalp, line 219: maybe this could be extracted, since we duplicate this complex condition in a few places? it would also be more expressive by self-documenting with a name

app/components/AppHeader.vue
```diff
@@ -301,15 +323,25 @@ onKeyStroke(
...
 
+      <!-- Mobile: Close search button (collapses search) -->
+      <ButtonBase
+        v-if="isSearchExpanded && !isOnSearchPage"
+        type="button"
+        class="sm:hidden flex-shrink-0"
+        :aria-label="$t('nav.close_search')"
+        @click="collapseMobileSearch"
+        classicon="i-lucide:x"
+      />
```
knowler, line 343: I think we should just collapse the input if it doesn’t have a value or if it’s not focused. We shouldn’t need an additional button for this. And if I recall, we added a button to clear the input which would effectively serve the same purpose if the search did have a value.

ghostdevv:
Hey, thanks for the PR! There ended up being a smaller one that seems to also solve it, so we'll close this for now :pray:

PR #2025, opened 2026-03-10
feat: add version history page

### LINKED ISSUE #1846: Linkable package page to show list of versions
### Describe the feature

In the Babel repo, we currently link to https://www.npmjs.com/package/@babel/types?activeTab=versions, in https://github.com/babel/babel/blob/cd22e3a8faac6e83ca99389f78046d0e2947bce5/.github/workflows/issue-triage.yml#L102

It does not seem like there is something similar on npmx.dev. I would like to point our users to somewhere where they can easily see if a version is available or not, without having to expand a bunch of small menus to do it.

### Additional information

### Final checks

ShroXd:
### 🔗 Linked issue

Resolves #1846 

### 🧭 Context

Like the original issues mentioned, it would be useful to have a full version history page with filtering functionality.

@WilcoSp is currently working on the changelog page feature in #1233. We've [discussed](https://discord.com/channels/1464542801676206113/1481064687255687271) this and agreed that both pages will be implemented and shipped separately, with the interaction between them designed based on real-world usage.

### 📚 Description

Since it's common for a package to have hundreds of versions, some package like typescript even have 3,700 versions. To provide good performance and user experience, I did some optimization and trade off during developing.

#### 1. Two phase data fetching

Phase 1 (on page load) calls `getVersions` to fetch only a lightweight summary — version strings, dist-tags, and publish times. This is enough to immediately render the "Current Tags" section and all group headers without waiting for full metadata.
Phase 2 (lazy, on first group expand) calls `fetchAllPackageVersions` to retrieve full metadata (deprecated status, provenance). The result is cached in `fullVersionMap` as a `shallowRef` — once loaded it's reused across all subsequent group expansions, never re-fetched.

_The tradeoff_: a short one-time loading delay on first expand, in exchange for a significantly faster initial page load.

#### 2. Virtual scrolling + SSR fallback

Since `WindowVirtualizer` is client-only, it's wrapped in `<ClientOnly>`, with a `#fallback` slot providing a static SSR substitute — just the first 20 group headers, no interactivity.

The SSR fallback iterates `versionGroups` directly, a flat list of group objects, each containing an array of versions
The client-side `virtualizer` consumes `flatItems` — a flattened array that interleaves group headers and their expanded version rows into a single indexed list.

This follows existing pattern in [package list page](https://github.com/npmx-dev/npmx.dev/blob/a170292a266c1d89fd150281e4a863154b944ce0/app/components/Package/List.vue#L156-L201).

#### 3. Debounced version filter

Since the filtering work is done entirely on the frontend, it can put significant performance pressure on packages with a large number of versions, such as typescript. And to support virtual scrolling, version groups are flattened into a `flatItems` array — a single indexed list of group headers and version rows that `WindowVirtualizer` can render efficiently. Therefore, a debounce on the filter input is necessary to avoid triggering expensive recomputation on every keystroke.

Some further optimization options include chunked `flatItems` construction with `requestIdleCallback` and moving the filtering work into a Web Worker. However, both require non-trivial design and implementation to handle potential race conditions. We also plan to add changelog support to this page, which will introduce additional matching overhead. For now, the current implementation is "good enough" — once the full feature set is complete, we can do thorough performance profiling to find the best optimization approach.

#### 4. Disable prefetching for package page of each version

Per-version links in the history list have prefetching disabled, as expanding a group could generate a lot of simultaneous prefetch requests and unnecessarily strain the network and significantly impact page performance. Although this makes package page navigation slightly slower, considering most user behavior on this page is browsing and filtering, the tradeoff is acceptable.

#### 5. Some interesting cases

- [A package with 3,700 versions](https://npmx-ez2ae9l7z-npmx.vercel.app/package/typescript/versions)
- [A package with 19,498 versions](https://npmx-ez2ae9l7z-npmx.vercel.app/package/@mahdiarjangi/phetch-cli/versions)
- [A package with single major version](https://npmx-ez2ae9l7z-npmx.vercel.app/package/@vue/compiler-dom/versions)
- [A package with a lot of current tags](https://npmx-ez2ae9l7z-npmx.vercel.app/package/astro/versions)

### 📸 Screenshots

| Scenario | Screenshot |
| :- | :- |
| Light mode | version-history-page-light |
| Dark mode | version-history-page-dark |
| Expand version group | version-history-expand-group |
| Filter | version-history-filtering |


app/pages/package/[[org]]/[name]/versions.vue
```diff
@@ -0,0 +1,724 @@
...
+  name: 'package-versions',
+})
+
+const route = useRoute()
+const router = useRouter()
+
+const packageName = computed(() => {
+  const { org, name } = route.params as { org?: string; name: string }
+  return org ? `${org}/${name}` : (name as string)
+})
```
Comment (shuuji3) [inline app/pages/package/[[org]]/[name]/versions.vue]: Since we enables the experimental typed pages (ref. [Experimental Features · Nuxt Advanced v4](https://nuxt.com/docs/4.x/guide/going-further/experimental-features#typedpages)), you can get the fully typed route with this:

```suggestion
const route = useRoute('package-version')
const router = useRouter()

const packageName = computed(() => {
  const { org, name } = route.params
  return org ? `${org}/${name}` : name
})
```

shuuji3 (changes requested):

app/pages/package/[[org]]/[name]/versions.vue
Comment (ShroXd) [reply app/pages/package/[[org]]/[name]/versions.vue]: > Since we enables the experimental typed pages (ref. [Experimental Features · Nuxt Advanced v4](https://nuxt.com/docs/4.x/guide/going-further/experimental-features#typedpages)), you can get the fully typed route with this:

Thanks for mentioning that! I will fix this before sending the request.

Because this PR is still WIP and I'm tweaking the UX details, I'm using `as` and other shortcuts to cheat the type system for now — saves tokens for both AI and my brain. 😉

Comment (shuuji3) [reply app/pages/package/[[org]]/[name]/versions.vue]: I see, that makes sense. 👍🏻

app/pages/package/[[org]]/[name]/versions.vue
```diff
@@ -0,0 +1,539 @@
...
+                  <button
+                    v-if="item.type === 'header'"
+                    type="button"
+                    class="flex items-center gap-3 px-4 py-2.5 w-full text-start hover:bg-bg-subtle transition-colors"
+                    :class="index < flatItems.length - 1 ? 'border-b border-border' : ''"
+                    :aria-expanded="expandedGroups.has(item.groupKey)"
+                    :aria-label="
+                      expandedGroups.has(item.groupKey)
+                        ? $t('package.versions.collapse', { tag: item.label })
+                        : $t('package.versions.expand', { tag: item.label })
```
Comment (knowler) [inline app/pages/package/[[org]]/[name]/versions.vue]: Don’t worry about updating these labels. Using `aria-expanded` is already sufficient for AT users to know what state the button is in and consequently what state it will change to if they activate it. Doing both can be confusing (e.g. screen reader might read `Collapse 4.x versions, button expanded` or `Expand 4.x versions, button collapsed`).

This goes for anywhere else in the PR that includes the words “expand” or “collapse” inside of the `aria-label`.

app/components/Package/Versions.vue
```diff
@@ -521,15 +529,27 @@ function majorGroupContainsCurrent(group: (typeof otherMajorGroups.value)[0]): b
...
-      </ButtonBase>
+      <div class="flex items-center gap-3">
+        <LinkBase
+          :to="versionsPageRoute"
+          variant="button-secondary"
+          class="text-fg-subtle hover:text-fg transition-colors min-w-6 min-h-6 p-1 rounded"
+          :title="$t('package.versions.view_all_versions')"
+          classicon="i-lucide:history"
+          data-testid="view-all-versions-link"
+        >
```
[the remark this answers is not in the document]
ShroXd: Vue3 handles this correctly because v-if/v-else always renders exactly one root element at runtime. The `$attrs` would be bind to the real root element. This is confirmed via the passed tests, no fix is needed.

app/pages/package/[[org]]/[name].vue
```diff
@@ -207,6 +207,8 @@ const { diff: sizeDiff } = useInstallSizeDiff(packageName, resolvedVersion, pkg,
 //    → Preserve the server-rendered DOM, don't flash to skeleton.
 const nuxtApp = useNuxtApp()
 const route = useRoute()
+// Gates template rendering only — data fetches intentionally still run.
+const isVersionsRoute = computed(() => route.name === 'package-versions')
```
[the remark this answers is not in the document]
Comment (ShroXd) [reply app/pages/package/[[org]]/[name].vue:216]: Because `immediate` is set once at mount — skipped requests won't re-fire on navigation, leaving data permanently missing. Added a code comment to clarify this.

app/pages/package/[[org]]/[name]/versions.vue
Comment (ShroXd) [reply app/pages/package/[[org]]/[name]/versions.vue]: Thanks for pointing it out!

test/nuxt/components/PackageVersions.spec.ts
```diff
@@ -72,10 +72,15 @@ describe('PackageVersions', () => {
...
+      // Find version links (exclude anchor links, external links, and action buttons)
       const versionLinks = component
         .findAll('a')
-        .filter(a => !a.attributes('href')?.startsWith('#') && a.attributes('target') !== '_blank')
+        .filter(
+          a =>
+            !a.attributes('href')?.startsWith('#') &&
+            a.attributes('target') !== '_blank' &&
+            !a.attributes('data-testid')?.includes('view-all-versions'),
+        )
```
graphieros: This whole findAll + filter is repeated throughout the test, and might be dried out in a function

ShroXd: > This whole findAll + filter is repeated throughout the test, and might be dried out in a function

Thanks for pointing it out! It was on my reminder before, but I forgot to check the reminder. 😅

graphieros (approved):
LGTM 
Thank you :)

btea:
image

Perhaps, like npmjs.org, we could also display the download count for each version.

ShroXd:
@btea Thanks for mentioning that — download counts would be helpful for users to find the best version. That said, I'd prefer to implement this in a separate PR.

The download data requires an additional network request, and there are a few details worth thinking through carefully: the data only covers the past 7 days, truly inactive versions will show no data, etc. There are also UX decisions around how to display and sort by download counts. We also plan to support changelog viewing on this page.

Shipping a _good enough but not perfect_ version page first will give us a good baseline for future iteration.

btea:
That sounds amazing. I'm looking forward to this feature. 👍

shuuji3 (approved):
Great new feature in npmx 🙌 (packed with so many implementation details!)

I agree with all the points about the trade-offs. A Web Worker might be too much for now. I feel the current version is fast enough, and the loading indicator also helps.

PR #2028, opened 2026-03-10
feat(ui): added word wrap to code viewer

### LINKED ISSUE #2027: add word wrap to code viewer
### Describe the feature

currently the code viewer does not have word wrap would be cool to have word wrap

Image

### Additional information

### Final checks

stephansama:
### 🔗 Linked issue

resolves #2027

### 🧭 Context

scrolling horizontally is annoying when trying to just read code in my opinion.

### 📚 Description

I just added a small button in the toolbar to toggle word wrap and a class to enable and disable word wrap

I did use gemini to help with this PR however i implemented a similar feature on my personal [blog](https://blog.stephansama.info/) on code blocks. just used the ai to quickly get up to speed


whitep4nth3r:
Hi and thank you for this change! 

In my opinion, and for accessibility reasons (for example for people who don't use a mouse) I believe that it's always better to wrap code regardless just using CSS. 

For that reason I don't think we need a toggle and the wrapping can just be implemented using CSS.

whitep4nth3r:
Perhaps we should even have a max line length in the code viewer like in many IDEs.

stephansama:
> Hi and thank you for this change!
> 
> In my opinion, and for accessibility reasons (for example for people who don't use a mouse) I believe that it's always better to wrap code regardless just using CSS.
> 
> For that reason I don't think we need a toggle and the wrapping can just be implemented using CSS.

i was trying to get it to work with css only however i was having a little issue. like it would mostly work but not all the way. i can take another crack at it again when i get out of work. 🫡

nosthrillz:
Hi @whitep4nth3r 

I created this PR in response to the same issue, avoiding a toggle
https://github.com/npmx-dev/npmx.dev/pull/2070

Hope I'm not stepping on toes, it's my first OSS PR ever 😅

gameroman:
> In my opinion, and for accessibility reasons (for example for people who don't use a mouse) I believe that it's always better to wrap code regardless just using CSS.

I think having a toggle/setting + shortcut would be better because sometimes you do not want the code to be wrapped

stephansama:
> > In my opinion, and for accessibility reasons (for example for people who don't use a mouse) I believe that it's always better to wrap code regardless just using CSS.
> 
> I think having a toggle/setting + shortcut would be better because sometimes you do not want the code to be wrapped

i think the toggle is better for user experience as i prefer word wrap however regular people that i meet hate word wrap. giving the choice to the people i think is the correct move

PR #2039, opened 2026-03-12
fix(ui): remove install package command version for latest

bluwy:
### 🔗 Linked issue

n/a

### 🧭 Context

#1969 made a change that caused the version to show up in the install command, even if the package link does not have a version, e.g. https://main.npmx.dev/package/vite

image

The version shouldn't be showed in this case

### 📚 Description

This PR returns it to like before, http://localhost:3000/package/vite

image


alex-key:
Since you still could click on the latest version semver and url would be changed (and `requestedVersion` added) I would improve logic to omit this prop in case resolvedVersion is the latest. But it's up to you

bluwy:
Done. Personally I'd show the tag too, if it's `beta` for example so it's `pkg@beta` and not `pkg@1.2.3`, but I'd like to get some form of this fix out first so it doesn't show a pinned version install.

PR #2042, opened 2026-03-12
fix(ui): remove redundant calendar icons from date inputs

### LINKED ISSUE #2038: Consider removing additional calendar icon from date inputs
Within TrendsChart.vue there's two date inputs which get given a calendar icon. The way this is done caused issues on mobile chrome due to a browser bug (worked around in #2037), though it causes wider issues that aren't as easy to address. Chromium on desktop and Firefox on desktop and mobile both have already have calendar icons as part of their date input widgets, which leads to a weird double calendar icon situation.

Desktop chromium example:

Image

The chromium one can be hidden but is the only way to access the picker for a mouse user, and the firefox one can't be hiddden.

On Firefox Android this can actually lead to the input value still being clipped which is not desirable.

marlonwq:
### 🧭 Context
Resolves #2038

### 📚 Description
This PR removes the manual calendar icons, preventing duplication.

### 🛠️ Changes
- Removed redundant `<span>` icons with class `i-lucide:calendar`.
- Adjusted `InputBase` padding from `ps-7` to `ps-2` to align text correctly after icon removal.

### 🧪 Testing
- Verified on Chrome (Desktop): Only the native icon remains, picker works as expected.
image
image

### DIFF
--- a/app/components/Package/TrendsChart.vue
+++ b/app/components/Package/TrendsChart.vue
@@ -1618,7 +1618,7 @@ watch(selectedMetric, value => {
                 v-model="startDate"
                 type="date"
                 :max="DATE_INPUT_MAX"
-                class="w-full min-w-0 bg-transparent ps-2"
+                class="w-full min-w-0 bg-transparent"
                 size="medium"
               />
             </div>


app/components/Package/TrendsChart.vue
```diff
@@ -1633,16 +1629,12 @@ watch(selectedMetric, value => {
...
-                class="absolute inset-is-2 i-lucide:calendar w-4 h-4 text-fg-subtle shrink-0 pointer-events-none"
-                aria-hidden="true"
-              />
               <InputBase
                 id="endDate"
                 v-model="endDate"
                 type="date"
                 :max="DATE_INPUT_MAX"
-                class="w-full min-w-0 bg-transparent ps-7"
+                class="w-full min-w-0 bg-transparent ps-2"
```
alexdln: I think we can just leave default component paddings now (2.5 in this case)

```suggestion
                class="w-full min-w-0 bg-transparent"
```

app/components/Package/TrendsChart.vue
```diff
@@ -1613,16 +1613,12 @@ watch(selectedMetric, value => {
...
-                class="absolute inset-is-2 i-lucide:calendar w-4 h-4 text-fg-subtle shrink-0 pointer-events-none"
-                aria-hidden="true"
-              />
               <InputBase
                 id="startDate"
                 v-model="startDate"
                 type="date"
                 :max="DATE_INPUT_MAX"
-                class="w-full min-w-0 bg-transparent ps-7"
+                class="w-full min-w-0 bg-transparent ps-2"
```
alexdln:
```suggestion
                class="w-full min-w-0 bg-transparent"
```

alexdln (approved):

alexdln:
@shuuji3 sorry, minor changes request 🙃

shuuji3:
No problem! I thought I canceled the queue by approving it 😄

graphieros:
> The mobile view still hide the day, but I think this is a separate issue 😅

FIxed in https://github.com/npmx-dev/npmx.dev/pull/2037/changes

marlonwq:
Wow, thanks for the quick reply!

I've implemented the suggestions 🙂

graphieros:
I'm glad this one is in :)

lukewarlow:
@graphieros that's not correct btw, firefox android does still have a bug but it seems related to the w-full and min-w-0 on the input. Chrome android had a separate bug that has been fixed by that patch.

Issue #1826, opened 2026-03-02
highlight license changes


dominikg:
### Describe the feature

Sometimes packages change license. They should do it in a major release and highlight it in their changelogs but thats not always the case.

Violating a license can have legal consequences so adding a banner that mentions the old and new license can help draw attention to it

### Additional information

- [ ] Would you be willing to help implement this feature?

### Final checks

- [x] Read the [contribution guide](https://github.com/npmx-dev/npmx.dev/blob/main/CONTRIBUTING.md).
- [x] Check existing [issues](https://github.com/npmx-dev/npmx.dev/issues).

aisiklar:
hi, i sent a pull request to fix this
https://github.com/npmx-dev/npmx.dev/pull/2188

as this is literally my first PR to an oss codebase, i hope, all is ok, i have not done something wrong in this PR.
Thank you

PR #2048, opened 2026-03-12
fix: use package license for skills

### LINKED ISSUE #2043: License detection fails for nitro skills
### Describe the bug

`nitro` package has the skill (https://main.npmx.dev/package/nitro) and it is licensed as MIT (there's `LICENSE` file and `license: "MIT"` in `package.json` properly), but the dialog shows "No license specified" warning.

screenshot of skill info dialog with additional warning

### Additional context

_No response_

### Logs

```shell-script

```

maxchang3:
### 🔗 Linked issue

resolves #2043

### 🧭 Context

Currently, the logic for retrieving a skill’s license only checks the frontmatter.

As a result, when a skill does not specify a license there, the system displays a misleading "No license specified" warning, even if the parent package already defines a valid license.

### 📚 Description

This PR improves license handling in `skills.ts` by retrieving and passing the package license as a fallback when a skill does not specify its own license.

---

_(A related question: if a skill's license differs from the package's license, should we explicitly surface this (or even treat it as a warning)?)_

### DIFF
--- a/server/utils/skills.ts
+++ b/server/utils/skills.ts
@@ -162,9 +162,12 @@ export async function fetchSkillContent(
 /**
  * Validate skill frontmatter and return warnings.
  */
-export function validateSkill(frontmatter: SkillFrontmatter): SkillWarning[] {
+export function validateSkill(
+  frontmatter: SkillFrontmatter,
+  packageLicense?: string,
+): SkillWarning[] {
   const warnings: SkillWarning[] = []
-  if (!frontmatter.license) {
+  if (!frontmatter.license && !packageLicense) {
     warnings.push({ type: 'warning', message: 'No license specified' })
   }
   if (!frontmatter.compatibility) {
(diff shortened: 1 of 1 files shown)


marlonwq:
Great addition with the fallback logic!

Regarding your question: I definitely think we should surface it when a skill's license differs from the package's license.

As discussed in #1826, license transparency is a major pain point and surfacing these mismatches prevents 'legal traps' for users who assume the entire package follows the root license. A subtle UI warning or just a 'license override' badge would be a huge DX win.

ghostdevv:
@marlonwq suggestion makes sense! will mark this as draft while it's being worked on :pray: @maxchang3 let me know if you'd like any help, or don't have time to take it on right now :pray:

maxchang3:
@ghostdevv Hi, since the warning messages are still hardcoded (see #2049), would it make more sense to add a new warning after that fix lands? Or we can do this in a follow-up PR, since this one only affects the internal warning generation logic.

I'm also happy to adapt this to this PR.

PR #2052, opened 2026-03-12
fix: align weekly chart buckets from end to match npm downloads

### LINKED ISSUE #2041: Weekly chart: align buckets from end to match npm trends
Follow-up of #2013 / #2014.

Weekly chart last value doesn't match the table (npm rolling week). Example: [@sveltejs/kit vs svelte](https://main.npmx.dev/compare?packages=@sveltejs/kit,svelte) — table shows 1.7M but chart shows ~1.4M.

Fix: build weekly buckets from end date backwards (drop partial first bucket) so the last point matches [npm trends](https://npmtrends.com/@sveltejs/kit-vs-svelte).

PR incoming soon™

jycouet:
Closes #2041

## Problem

Weekly buckets were aligned from the **start** of the date range, so the last bucket often covered a different 7-day window than the npm rolling week.

## Solution

- **Build weekly buckets from the end date backwards** so the last bucket aligns exactly with the npm "last week" window
- **Extract bucketing logic** into `app/utils/chart-data-buckets.ts` with `fillPartialBucket` for the (now partial) first bucket
- **Extract date helpers** into `app/utils/date.ts` (`parseIsoDate`, `toIsoDate`, `addDays`, etc.)
- **Handle single-point series** edge case
- Add comprehensive tests for both buckets and date utilities

## Verification

On `/compare?packages=svelte,@sveltejs/kit`, table and chart now show identical values (3.3M / 1.4M).

### DIFF
--- a/app/composables/useCharts.ts
+++ b/app/composables/useCharts.ts
@@ -56,8 +56,8 @@ function splitIsoRangeIntoChunksInclusive(
     const actualEnd = cursorEnd.getTime() < finalEnd.getTime() ? cursorEnd : finalEnd
 
     chunks.push({
-      startIso: formatIsoDateOnly(cursorStart),
-      endIso: formatIsoDateOnly(actualEnd),
+      startIso: toIsoDate(cursorStart),
+      endIso: toIsoDate(actualEnd),
     })
 
     cursorStart = addDays(actualEnd, 1)


jycouet:
> This is great !
> 
> Just one thing, the Known Anomalies checkbox is now togglable when there are no recorded corrections. I think this might be a regression (?).

Haaa, yes I did this change. It's actually bugging me that sometime I can click and sometime not. We have the tooltip on top to know if it will have wrong data, so having this conditional doesn't feel correct.

It's a bit like "disabling smooting if if's the same value"

What do you think ?

graphieros:
> What do you think ?

As a user, if I can click but it does not change anything, it might feel as a bug.
I don't think we can rely on users reading the tooltip ;)

I wonder if the tooltip, however, could also say: 'No recorded anomalies' ?

jycouet:
On a flat chart _(a lot of small libraries with almost no download)_ all sliders have also no affect on the trend. 

I have a half strong opinion on this 😜
But you decide, no worries, I can remove this change 👍
I'm waiting for you final word 👀

graphieros:
Yeah, you're right about the sliders...
Can we compromise, just adding 'No recorded anomalies' in the tooltip ?

(And I prefer that we decide together ;)

jycouet:
> Yeah, you're right about the sliders...
> Can we compromise, just adding 'No recorded anomalies' in the tooltip ?
> 
> (And I prefer that we decide together ;)

) <- closing your open parentesis ;)

Me too, I would like to decide together, just wanted to say that we could remove it. I would not have hard feelings, you will still be a great OSS friend 😉 _BTW we have to meet one day !🧡_

Your suggestion just before looks perfect to me. I'll update that after coffee ☺️

graphieros:
> Me too, I would like to decide together, just wanted to say that we could remove it. I would not have hard feelings, you will still be a great OSS friend 😉 _BTW we have to meet one day !🧡_

Yes, you must let know the next time you're in Paris ^^

graphieros:
I'm sorry @jycouet I edited your previous message instead of answering it.
You are right, even I don't read tooltips apparently!

PR #2056, opened 2026-03-13
fix: correct code link for alias

### LINKED ISSUE #2044: Incorrect links for aliased links in code view
In the code tab, we have dynamic links within the code preview. They allow you to quickly navigate to other packages.

However, if the project uses aliases in dist as well - they are also considered packages, and the links lead to non-existent pages

https://npmx.dev/package-code/nuxt/v/4.3.1/dist%2Fcomponents%2Fruntime%2Fclient-component.js

eryue0220:
### 🔗 Linked issue

resolve #2044

### 🧭 Context

support alias path like `~/app/components/Button.vue`, `@/app/components/Button.vue`  or `#/app/components/Button.vue`


gameroman (commented):
Also when package uses something like this https://main.npmx.dev/package-code/empathic/v/2.0.0/find.mjs#L3 would be nice to correctly resolve that too

eryue0220:
> Also when package uses something like this https://main.npmx.dev/package-code/empathic/v/2.0.0/find.mjs#L3 would be nice to correctly resolve that too

Updated.

gameroman (approved):
There are some linter warnings but this looks good 👍

test/fixtures/jsdelivr-cdn/empathic@2.0.0/find.mjs
```diff
@@ -0,0 +1,73 @@
...
+import * as walk from 'empathic/walk'
+/**
+ * Find an item by name, walking parent directories until found.
+ *
+ * @param name The item name to find.
+ * @returns The absolute path to the item, if found.
+ */
+export function up(name, options) {
+  const start = (options && options.cwd) || ''
+  for (const dir of walk.up(start, options)) {
```
ghostdevv: we should ignore the lint errors in here

server/utils/import-resolver.ts
```diff
@@ -206,9 +506,26 @@ export function createImportResolver(
...
+      exportsMap,
+      currentFile,
+      files,
+    )
+    const resolved =
+      relativeResolved != null
+        ? relativeResolved
+        : internalResolved != null
+          ? internalResolved
+          : selfResolved
```
ghostdevv:
```suggestion
    const resolved = relativeResolved ?? internalResolved ?? selfResolved
```

ghostdevv (changes requested):
I'm in two minds, one that we shouldn't try and resolve special aliases as they will only work for special packages that are shipping code intended to be processed by something like vite, and the other that we should because it won't hurt - I think the latter is ok as we can always change it later.

Can we also support `$` as an alias prefix?

ghostdevv (approved):

gameroman (approved):

PR #2060, opened 2026-03-13
refactor: move duplicated repositoryUrl in composable

MatteoGabriele:
### 🧭 Context
Spotted repeated logic regarding the repository URL

### 📚 Description
I've moved the logic regarding repositoryUrl to a separate composable to reduce duplicated code and simplify component code.

### DIFF
--- a/app/components/BuildEnvironment.vue
+++ b/app/components/BuildEnvironment.vue
@@ -1,6 +1,4 @@
 <script setup lang="ts">
-import type { BuildInfo } from '#shared/types'
-
 const { footer = false, buildInfo: buildInfoProp } = defineProps<{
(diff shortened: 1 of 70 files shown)


app/composables/useRepositoryUrl.ts
```diff
@@ -0,0 +1,41 @@
+import { joinURL } from 'ufo'
+
+function normalizeGitUrl(url: string): string {
```
alexdln: Looks like you can reuse normalizeGitUrl from shared/utils/git-provider (OgImage/Package used it)

MatteoGabriele: oh that's even better ❤️

PR #2063, opened 2026-03-13
refactor: remove linter warnings

MatteoGabriele:
### 📚 Description
Removes all linter warnings

### DIFF
--- a/app/composables/npm/useUserPackages.ts
+++ b/app/composables/npm/useUserPackages.ts
@@ -47,7 +47,7 @@ export function useUserPackages(username: MaybeRefOrGetter<string>) {
 
   const asyncData = useLazyAsyncData(
     () => `user-packages:${searchProviderValue.value}:${toValue(username)}`,
-    async (_, { signal }) => {
+    async (_nuxtApp, { signal }) => {
       const user = toValue(username)
       if (!user) {
         return emptySearchResponse()


server/api/atproto/bluesky-author-profiles.get.ts
```diff
@@ -39,7 +39,7 @@ export default defineCachedEventHandler(
       return { authors: [] }
     }
 
-    const handles = authors.map(a => a.blueskyHandle).filter(v => v != null)
+    const handles = authors.map(a => a.blueskyHandle).filter(handle => handle != null)
```
alexdln, line 42: Check only that not equal to null... What if empty string... Maybe:

```suggestion
    const handles = authors.map(a => a.blueskyHandle).filter(Boolean)
```

lunaria/components.ts
```diff
@@ -425,7 +423,7 @@ function SvgLocaleSummary(
       file.localizations.find(localization => localization.lang === lang)?.status === 'missing',
   )
   const outdatedFiles = status.filter(file => {
-    const localization = file.localizations.find(localization => localization.lang === lang)
+    const localization = file.localizations.find(l => l.lang === lang)
```
alexdln: Let's use full names, maybe localizationItem in this case

config/i18n.ts
```diff
@@ -406,9 +406,8 @@ export const datetimeFormats = Object.values(currentLocales).reduce((acc, data)
 }, {} as DateTimeFormats)
 
 export const numberFormats = Object.values(currentLocales).reduce((acc, data) => {
-  const numberFormats = data.numberFormats
-  if (numberFormats) {
-    acc[data.code] = { ...numberFormats }
+  if (data.numberFormats) {
+    acc[data.code] = { ...data.numberFormats }
```
alexdln, line 410: Why you changed cloning to Object.assign everywhere, but not here? 🙃

app/composables/useMarkdown.ts
```diff
@@ -88,20 +88,20 @@ function parseMarkdown({ text, packageName, plain }: UseMarkdownOptions): string
   html = html.replace(/~~(.+?)~~/g, '<del>$1</del>')
 
   // Links: [text](url) - only allow https, mailto
-  html = html.replace(/\[([^\]]+)\]\(([^)]+)\)/g, (_match, text, url) => {
+  html = html.replace(/\[([^\]]+)\]\(([^)]+)\)/g, (_match, matchedText, url) => {
```
alexdln: Minor:
```suggestion
  // Links: [matchedText](url) - only allow https, mailto
  html = html.replace(/\[([^\]]+)\]\(([^)]+)\)/g, (_match, matchedText, url) => {
```

Or I will prefer to call it textGroup and urlGroup

app/composables/npm/useUserPackages.ts
```diff
@@ -47,7 +47,7 @@ export function useUserPackages(username: MaybeRefOrGetter<string>) {
 
   const asyncData = useLazyAsyncData(
     () => `user-packages:${searchProviderValue.value}:${toValue(username)}`,
-    async ({ $npmRegistry }, { signal }) => {
+    async (_, { signal }) => {
```
alexdln: We don't have character limits 😅

```suggestion
    async (_nuxtApp, { signal }) => {
```

config/i18n.ts
MatteoGabriele: I've only took care of the one the linter was complaining about

app/composables/npm/useUserPackages.ts
MatteoGabriele: `_` is usually used when the key is not in use, and it is widely utilized in the project at the moment, so it maintains consistency.

config/i18n.ts
alexdln: Can you please share the failed rule? I'd be interested to read how it defines what needs to be changed

server/api/atproto/bluesky-author-profiles.get.ts
MatteoGabriele: TypeScript won't understand the filter and still warn about null

app/composables/npm/useUserPackages.ts
alexdln: A variable starting with "\_" is a js-devs convention that it's unused or private, and linters follow this as well. Simply using "\_" is like a mini-version of both inserting an argument and skipping a rule

\_nuxtApp is also used frequently. And I'm more in favor of making it immediately obvious for future contributors what's under this unused variable

server/api/atproto/bluesky-author-profiles.get.ts
alexdln: Oh, sure, typescript... sorry

config/i18n.ts
MatteoGabriele:
```
help: If in-place mutation is acceptable, use `Object.assign` or direct property assignment instead of spreading
note: `Object.assign` mutates the first argument. Disable this rule if copy-on-write behavior is required.
```

alexdln: Oh, since you're spreading object subkey - it's fine here (not just `{ ...numberFormats }` but `{ ...data.numberFormats }`).

A bit weird, but got it, thanks

PR #2070, opened 2026-03-13
fix: wrap code viewer lines by computing width

### LINKED ISSUE #2027: add word wrap to code viewer

nosthrillz:
### 🔗 Linked issue

Attempts to resolve [2027](https://github.com/npmx-dev/npmx.dev/issues/2027)

### 🧭 Context

The current approach has overflow-x-auto, but wrapping might be better.
It's unclear whether that's preferred, so making the PR to allow a preview.

### 📚 Description

We actually can compute available screen width, so we can avoid overflow, and instead set the width and use flex-wrap combined with pre-wrap to get the desired effect.

### Potential points of contention / concerns
I didn't visually test the highlighting, I'm yet unsure what the implications would be :)

### DIFF
--- a/app/utils/chart-data-buckets.ts
+++ b/app/utils/chart-data-buckets.ts
@@ -34,17 +34,13 @@ export function buildWeeklyEvolution(
   if (sorted.length === 0) return []
 
   const rangeStartDate = parseIsoDate(rangeStartIso)
+  const rangeEndDate = parseIsoDate(rangeEndIso)
 
-  // Align from last day with actual data (npm has 1-2 day delay, today is incomplete)
-  const lastNonZero = sorted.findLast(d => d.value > 0)
-  const pickerEnd = parseIsoDate(rangeEndIso)
-  const effectiveEnd = lastNonZero ? parseIsoDate(lastNonZero.day) : pickerEnd
-  const rangeEndDate = effectiveEnd.getTime() < pickerEnd.getTime() ? effectiveEnd : pickerEnd
-
-  // Group into 7-day buckets from END backwards
   const buckets = new Map<number, number>()
+
   for (const item of sorted) {
-    const offset = Math.floor((rangeEndDate.getTime() - parseIsoDate(item.day).getTime()) / DAY_MS)
+    const itemDate = parseIsoDate(item.day)
+    const offset = Math.floor((rangeEndDate.getTime() - itemDate.getTime()) / DAY_MS)
     if (offset < 0) continue
     const idx = Math.floor(offset / 7)
     buckets.set(idx, (buckets.get(idx) ?? 0) + item.value)
--- a/app/components/Code/Viewer.vue
+++ b/app/components/Code/Viewer.vue
@@ -135,57 +135,57 @@ watch(
   flex: 1;
   min-width: 0;
   max-width: calc(100% - var(--line-numbers-width));
+}
 
-  &:deep(pre) {
-    margin: 0;
-    padding: 0;
-    background: transparent !important;
-    overflow: visible;
-    max-width: 100%;
-  }
+.code-content:deep(pre) {
+  margin: 0;
+  padding: 0;
+  background: transparent !important;
+  overflow: visible;
+  max-width: 100%;
+}
 
-  &:deep(code) {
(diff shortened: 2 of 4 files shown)


gameroman:
A toggle would be nice and also line numbers should adapt


Example comparison

https://npmx.dev/package-code/grammy/v/1.41.1/out%2Fbot.d.ts

https://npmx-axzqrskfc-npmx.vercel.app/package-code/grammy/v/1.41.1/out%2Fbot.d.ts

nosthrillz:
Switching to Draft because the line numbers need to be recomputed. As attempted in `https://github.com/npmx-dev/npmx.dev/pull/2028/changes` as well

WilcoSp:
the wrapping is very readable and atm I also only see the issue of the line numbers & the highlighting included with it.

a file that may be good to check would be in nuxt the /dist/index.mjs file because it has large imports

nosthrillz:
After spending hours on this, I've hit a roadblock. This is a non-trivial issue, even though it seemed like "bite-sized" initially.

Been getting massive performance hits on large files due to recomputing line heights for 10^3 .. 10^5 nodes if you count in the highlighting.

What I tried:
- measuring lines with JS (getComputedStyle / offsetHeight). But this blocks the main thread for large files like the [nuxt dist](https://npmx.dev/package-code/nuxt/v/4.4.2/dist%2Findex.mjs) example. (There's a reason GH hides it, eh?)
- v-html line numbers + multiplier spacers (still measuring). Can lead to as bad as 120s hang INP score for the nuxt file.
- chunks of HTML + infinite scroll. Aside from the hydration mismatches that I didn't continue investigating, the UX is a bit unfortunate as it no longer allows you to easily jump to a certain line, or simply scroll ahead multiple pages at a time.
- Non-blocking expansion in a window-pane mode. Worked out, but still 16s INP because we're parsing 10^4+ DOM nodes
- CSS counters + truncation at 1000 lines. Kinda like GitHub. The problem here is that going to line 1001 on load will block again, until the full page is shown. Could potentially make it a hybrid approach.

The jist of it is that, to wrap the 8500 Shiki-highlighted lines, the browser needs a lot of time counting line heights, and reconciling, because we need the window resize, so it's not a one-off

Going forward:
1. GitHub-style show chunk buttons. Show lines X-Y, with "Load lines above" / "Load lines below" buttons at edges and a "Show all" nuclear button. At least the users would be aware that clicking load all will be consuming. Problem is that we would have to give up on anchor tags to lines
2. Infinite scroll (both directions, chunked). Works, but not ideal UX :)
3. Actuall virtual scroller like twitter. There's [virtual-scroller](https://github.com/catamphetamine/virtual-scroller) solving for exactly that. We'd show a virtual window pane for the nodes. It would be smooth AF. The big problem is, that it's impossible to "jump to line" or "find on page". Not to mention it would have to be CSR.


IMHO I think the GitHub style option fits the UX the closest.

Please advise

app/utils/chart-data-buckets.ts
```diff
@@ -34,17 +34,13 @@ export function buildWeeklyEvolution(
...
-  const effectiveEnd = lastNonZero ? parseIsoDate(lastNonZero.day) : pickerEnd
-  const rangeEndDate = effectiveEnd.getTime() < pickerEnd.getTime() ? effectiveEnd : pickerEnd
-
-  // Group into 7-day buckets from END backwards
   const buckets = new Map<number, number>()
+
   for (const item of sorted) {
-    const offset = Math.floor((rangeEndDate.getTime() - parseIsoDate(item.day).getTime()) / DAY_MS)
+    const itemDate = parseIsoDate(item.day)
+    const offset = Math.floor((rangeEndDate.getTime() - itemDate.getTime()) / DAY_MS)
```
jycouet, line 43: Are these changes intended in this PR?

nosthrillz: They are not, a mistake on my part for merging latest instead of rebasing :) 

I will have to remove them. Don't think this PR will end up being merged to be honest, given the above conversation

PR #2075, opened 2026-03-14
feat: add CSP and some security headers to HTML pages

serhalp:
### 🔗 Linked issue

N/A but discussed on Discord here: https://discord.com/channels/1464542801676206113/1465055739293991033/1481201825242681394

### 🧭 Context

We're missing a few basic security protections such as a Content Security Policy on HTML responses, `X-Frame-Options`, etc.

You can see the result of a basic scan here: https://developer.mozilla.org/en-US/observatory/analyze?host=main.npmx.dev.

### 📚 Description

Add a CSP to the HTML layout and a few security headers (`X-Content-Type-Options`, `X-Frame-Options`, `Referrer-Policy`) on server responses.

Adding the CSP to the HTML layout directly with a `<meta>` tag is _way_ simpler than trying to configure a response header conditionally only on html responses while excluding API responses, static assets, special internal paths, and whatever else, and then setting the CSP on pre-rendered pages also by either configuring static headers for those or using the `<meta>` approach only for those 😓.

The CSP `img-src` directive imports `TRUSTED_IMAGE_DOMAINS` from the image proxy module so the two stay in sync automatically, and similarly for supported git hosts.

> [!WARNING]
> It's unlikely I found all the hosts we need to allowlist. Please help me find any missing ones!

This PR also adds an e2e test that should fail on future PRs if a request is blocked by a CSP violation in the app's main happy path.


serhalp:
found CSP errors on the blog page:
- img-src needs to allow https://cdn.bsky.app/img/ and https://api.star-history.com/
- connect-src needs to allow https://public.api.bsky.app/xrpc/

serhalp:
I _think_ I've exercised every single existing npmx feature and I didn't trigger any CSP errors 👍🏼. Merging.

Note that atm apparently the npm cli connector thing doesn't work on Vercel preview URLs (unrelated to this PR, it's a CORS config issue), so I'll double-check that on main.npmx.dev right after merging.

PR #2078, opened 2026-03-14
feat(i18n): use rangeFormat for pagination

### LINKED ISSUE #1921: Search results with inconsistent pagination number format
### Describe the bug

The pagination number format in the search page is inconsistent.
Example `1-79306 of 79,306` instead of `1-79,306 of 79,306`.

Additionally, I would like to discuss the idea that when we display all results, it should resemble something like 
- Showing all 79,306 results
- All 79,306 results
- 1-79,306

### Additional context

Image

kytta:
### 🔗 Linked issue

Resolves #1921

### 🧭 Context

This change number formatting for the page ranges in pagination from ad-hoc `{start}-{end}` to proper range formatting via [Intl.NumberFormat.prototype.formatRange()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/formatRange).

This ensures that the locale-specific range formatting is always used, for example:

- Bengali: `২৫,৪৮৬ এর মধ্যে ৭৫১–১,০০০` (uses Bengali numbers)
- Japanese: `25,486 件中 751～1,000` (uses tilde)
- Hindi: `2,83,319 में से 2,83,251–2,83,319` (uses Indian base system)

### 📚 Description

The linked issue #1921 talks about number formatting not being consistent. It's true that, currently, only the total number of results is formatted properly, but not start and end. My first instinct was to just wrap them in `$n`, but then I remembered about the `formatRange` function.

Since there is no shorthand for range formatting in vue-i18n, I have used the numberFormatter composable to call it myself in the template code. Moving from `{start}-{end}` to just `{range}` required updating all i18n strings, of course. I have also noticed and fixed the invalid string in Turkish.

I have not used any AI tooling for this contribution.

### DIFF
--- a/i18n/locales/tr-TR.json
+++ b/i18n/locales/tr-TR.json
@@ -882,7 +882,7 @@
       "paginated": "Sayfalı",
       "items_per_page": "Sayfa başına öğe",
       "per_page": "/ sayfa",
-      "showing": "Gösteriliyor",
+      "showing": "{range} / {total} gösteriliyor",
       "previous": "Önceki",
       "next": "Sonraki",
       "nav_label": "Sayfalama navigasyonu"


userquin (approved):
LGTM

Thx ❤️

userquin:
https://caniuse.com/?search=formatRange

PR #2086, opened 2026-03-15
fix: fetch resolved version for install size

gameroman:
Fixes package install size resolution by fetching resolved version for install size

Example: https://npmx.dev/package/fd-package-json/v/%5E2.0.0

Before:

image

After:

image


serhalp (approved):
oooooh was that why? thanks for fixing!

👀 do we have the same bug on the compare page?

PR #2090, opened 2026-03-15
fix(i18n): add missing pt-BR translations

offeringofpie:
### 📚 Description

Added some missing pt-BR translations.


shuuji3 (approved):
Thanks for the update! :pray:

PR #2101, opened 2026-03-16
fix: added scroll preservation opt in to the compare page

### LINKED ISSUE #2098: Changing Input in Compare Package shouldn't reset scroll
### Describe the bug

Currently changing all of the inputs in the [Compare Package page](https://npmx.dev/compare?packages=svelte,react-npm-yu251) resets the scroll to the top. Would be nice if it keeps its scroll to not disorient the users.

### Additional context

_No response_

### Logs

```shell-script

```

mihaizaurus:
### 🔗 Linked issue

resolves #2098 

### 🧭 Context

Fixes the compare page jumping back to the top when compare state is updated on the page.

On /compare, both package selection and facet selection sync state into the route query, being treated like normal navigations by the router scroll behavior, which caused the page to reset. This change makes scroll preservation opt-in at the page level and enables it for the compare page, so query-only updates keep the current scroll position.

Do note that this does not change normal page-to-page navigation scroll behavior, hash scrolling, or back/forward saved-position restoration.

I basically changed two files and added one test:
- Added `preserveScrollOnQuery: true` to the compare page meta
- Updated app/router.options.ts to return false from scrollBehavior for same-path, same-hash navigations on pages that opt in
- Added router-level unit coverage for:
    - saved position restoration
    - scroll preservation on compare-style query updates
    - hash anchor scrolling
    - normal navigation scrolling to top

I wanted to implement this in a clean way that fixes both affected compare controls with one router-level change:
- package updates via packages=...
- facet updates via facets=...

I wanted to avoid duplicating special-case URL update logic in multiple compare components/composables, and thus keep the behavior scoped to pages that explicitly opt in.

https://github.com/user-attachments/assets/03b7d66c-0c72-4538-acf8-7dfc6dbcdf32

### DIFF
--- a/app/router.options.ts
+++ b/app/router.options.ts
@@ -1,7 +1,7 @@
 import type { RouterConfig } from 'nuxt/schema'
 
 export default {
-  scrollBehavior(to, _from, savedPosition) {
+  scrollBehavior(to, from, savedPosition) {
     // If the browser has a saved position (e.g. back/forward navigation), restore it
 
     if (savedPosition) {
@@ -11,8 +11,8 @@ export default {
     // Preserve the current viewport for query-only updates on pages that opt in,
     // such as compare where controls sync state to the URL in-place.
     if (
-      to.path === _from.path &&
-      to.hash === _from.hash &&
+      to.path === from.path &&
+      to.hash === from.hash &&
       to.meta.preserveScrollOnQuery === true
     ) {
       return false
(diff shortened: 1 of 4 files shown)


graphieros (approved):
Works like a charm
LGTM

app/router.options.ts
serhalp: nit: `_` prefix is for unused vars

```suggestion
  scrollBehavior(to, from, savedPosition) {
```

serhalp (approved):

PR #2105, opened 2026-03-16
fix: tag sorting on package versions page

### LINKED ISSUE #2102: Confusing UI for "latest" tag in versions history
### Describe the bug

I'm looking at https://main.npmx.dev/package/@babel/core/versions

Image

It really makes it look like 8.0.0-rc.2 is marked as "latest". At first I thought I made a publishing mistake and accidentally published that RC as latest, rather than 7.29.0.

### Additional context

Ref https://github.com/npmx-dev/npmx.dev/pull/2025 

Looking at the screenshot in that PR, I'd expect 7.29.0 to be marked as "latest" and shown in big:

cc @ShroXd (and thanks for that PR :) )

### Logs

```shell-script

```

ShroXd:
### 🔗 Linked issue

Resolves #2102 

### 🧭 Context

A implementation bug caused the confusing UI.

### 📚 Description

Fix unstable ordering of the tagged versions array. Tags are now sorted by semantic priority, with the exception of latest which remains pinned at the top.

| Scenario | Screenshot |
| :- | :- |
| After | QQ20260316-231723 |
| Before | version list page before |

### DIFF
--- a/app/pages/package/[[org]]/[name]/versions.vue
+++ b/app/pages/package/[[org]]/[name]/versions.vue
@@ -5,10 +5,11 @@ import { compare, validRange } from 'semver'
 import {
   buildVersionToTagsMap,
   buildTaggedVersionRows,
+  compareTagRows,
+  compareVersionGroupKeys,
   filterVersions,
   getVersionGroupKey,
   getVersionGroupLabel,
-  getTagPriority,
 } from '~/utils/versions'
 import { fetchAllPackageVersions } from '~/utils/npm/api'
 
@@ -71,14 +72,7 @@ const latestTagRow = computed(() => tagRows.value.find(r => r.tags.includes('lat
 const otherTagRows = computed(() =>
   tagRows.value
     .filter(r => !r.tags.includes('latest'))
-    .sort((rowA, rowB) => {
-      const priorityA = Math.min(...rowA.tags.map(getTagPriority))
-      const priorityB = Math.min(...rowB.tags.map(getTagPriority))
-      if (priorityA !== priorityB) return priorityA - priorityB
-      const timeA = versionTimes.value[rowA.version] ?? ''
-      const timeB = versionTimes.value[rowB.version] ?? ''
-      return timeB.localeCompare(timeA)
-    }),
+    .sort((rowA, rowB) => compareTagRows(rowA, rowB, versionTimes.value)),
 )
 
 function getVersionTime(version: string): string | undefined {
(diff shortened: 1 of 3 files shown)


app/utils/versions.ts
```diff
@@ -51,6 +51,37 @@ export function getPrereleaseChannel(version: string): string {
   return match ? match[1]!.toLowerCase() : ''
 }
 
+/**
+ * Priority order for well-known dist-tags.
+ * Lower number = higher priority in display order.
+ * Unknown tags fall back to Infinity and are sorted by publish date descending.
```
nicolo-ribaudo, line 57: I was reading through the code and wondering how it's possible that "latest" is not a well-known tag. It might be worth having a comment here explaining that it's handled manually elsewhere :)

ShroXd: > I was reading through the code and wondering how it's possible that "latest" is not a well-known tag. It might be worth having a comment here explaining that it's handled manually elsewhere :)

Nice point, thanks for bringing it up! I reconsidered `TAG_PRIORITY` — it's definitely weird to exclude `latest` from it. Although we filter it out in the UI, it doesn't make sense for a general constant to be affected by specific UI logic.
So I've added `latest: 0` and shifted the other priorities by 1.

app/pages/package/[[org]]/[name]/versions.vue
```diff
@@ -64,7 +65,21 @@ async function ensureFullDataLoaded() {
...
+const otherTagRows = computed(() =>
+  tagRows.value
+    .filter(r => !r.tags.includes('latest'))
+    .sort((a, b) => {
+      const pa = getTagPriority(a.primaryTag)
+      const pb = getTagPriority(b.primaryTag)
+      if (pa !== pb) return pa - pb
+      const ta = versionTimes.value[a.version] ?? ''
+      const tb = versionTimes.value[b.version] ?? ''
+      return tb.localeCompare(ta)
```
Comment (graphieros) [inline app/pages/package/[[org]]/[name]/versions.vue]: nit: I would recommend to use readable vars

Comment (ShroXd) [reply app/pages/package/[[org]]/[name]/versions.vue]: > nit: I would recommend to use readable vars

Nice point! I will refactor this.

mbtools:
The version page is great! And thanks for this fix, too

Maybe move the sort logic to `sortTagsLatestFirst` in `versions.ts` and add some tests to `test/unit/app/utils/versions.spec.ts` 🙏 

Just a bit feedback for your next PRs 😉 

1. It's not obvious that you can enter semver ranges to filter

image

image

It's nicely done on the package page. I suggest using the same for the version page:

image

2, The dates for the dist-tags are missing the timestamp on hover

image

Nice work,
Marc

ShroXd:
@mbtools Thanks for pointing it out!
I refactored my code and added unit tests. And I will finish the remaining UX part in a new PR to make this PR focus on the fix.

mbtools:
Nice! LGTM 👍

PR #2108, opened 2026-03-16
fix(ui): enlarge chart skeleton title block

### LINKED ISSUE #2106: Improve skeleton loaders in compare page bar charts
### Describe the bug

The current skeleton loaders for the bar charts in the compare page do not have sufficient height for the title part of the loader, leading to CLS when loading datapoints.

Image

Kemalau:
## Why
The compare page bar-chart skeleton was shorter than the rendered chart title/subtitle block, which caused layout shift when datapoints loaded.

## What changed
- increased the title/subtitle skeleton placeholders in `FacetBarChart`
- increased the bar row placeholder height/spacing to better match the loaded chart
- added a focused component test for the loading skeleton structure

## Verification
- `pnpm exec oxlint app/components/Compare/FacetBarChart.vue test/nuxt/components/compare/FacetBarChart.spec.ts`
- `pnpm exec oxfmt --check app/components/Compare/FacetBarChart.vue test/nuxt/components/compare/FacetBarChart.spec.ts`
- `pnpm test test/nuxt/components/compare/FacetBarChart.spec.ts` *(blocked in this environment: Playwright browser launch fails because `libatk-1.0.so.0` is missing on the host)*

Closes #2106.

### DIFF
--- a/app/components/Compare/FacetBarChart.vue
+++ b/app/components/Compare/FacetBarChart.vue
@@ -283,23 +283,27 @@ const config = computed<VueUiHorizontalBarConfig>(() => {
       </VueUiHorizontalBar>
 
       <template #fallback>
-        <div class="flex flex-col gap-2 justify-center items-center mb-2">
-          <SkeletonInline class="h-4 w-16" />
-          <SkeletonInline class="h-4 w-28" />
-        </div>
-        <div class="flex flex-col gap-1">
-          <SkeletonInline class="h-7 w-full" v-for="pkg in packages" :key="pkg" />
+        <div class="flex flex-col gap-3" data-test="facet-bar-chart-skeleton">
+          <div class="flex flex-col items-center gap-2 mb-3">
+            <SkeletonInline class="h-5 w-36 max-w-full" />
+            <SkeletonInline class="h-4 w-52 max-w-full" />
+          </div>
+          <div class="flex flex-col gap-2">
+            <SkeletonInline class="h-8 w-full" v-for="pkg in packages" :key="pkg" />
+          </div>
         </div>
       </template>
     </ClientOnly>
 
     <template v-else>
-      <div class="flex flex-col gap-2 justify-center items-center mb-2">
-        <SkeletonInline class="h-4 w-16" />
-        <SkeletonInline class="h-4 w-28" />
-      </div>
-      <div class="flex flex-col gap-1">
-        <SkeletonInline class="h-7 w-full" v-for="pkg in packages" :key="pkg" />
+      <div class="flex flex-col gap-3" data-test="facet-bar-chart-skeleton">
+        <div class="flex flex-col items-center gap-2 mb-3">
+          <SkeletonInline class="h-5 w-36 max-w-full" />
+          <SkeletonInline class="h-4 w-52 max-w-full" />
+        </div>
+        <div class="flex flex-col gap-2">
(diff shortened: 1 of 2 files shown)


graphieros (changes requested):
I think the height of the skeleton bars for the chart section should be more dynamic.
There is still a size difference on desktop:

Image

But more importantly, a fixed height for skeleton bars will not match on mobile.

test/nuxt/components/compare/FacetBarChart.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+        label: 'Bundle size',
+        description: 'Compare install footprint',
+        facetLoading: true,
+      },
+    })
+
+    const skeleton = component.find('[data-test="facet-bar-chart-skeleton"]')
+    expect(skeleton.exists()).toBe(true)
+
+    const lines = skeleton.findAllComponents({ name: 'SkeletonInline' })
```
shaundai, line 20: checks for individual heights of placeholder items is not very useful.  if we changed the height of the title to 8px for example, we'd have to update all of the unit tests for this.  let's instead focus on the input instead and making sure that it exists. (eg: 
```
testPackages = ['react', 'vue']

props: { ..... etc etc
...packages: testPackages
const lines = skeleton.findAllComponents({ name: 'SkeletonInline' })
expect(lines).toHaveLength(2 + testPackages.length)

```

test/nuxt/components/compare/FacetBarChart.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+    expect(lines).toHaveLength(4)
+    expect(lines[0]?.attributes('class')).toContain('h-5')
+    expect(lines[1]?.attributes('class')).toContain('h-4')
```
shaundai, line 23: not a blocker for this PR, but this spec file could benefit from more comprehensive unit tests.  success state, empty state, testing dynamic updates like the `watch` on `props.packages` etc

app/components/Compare/FacetBarChart.vue
```diff
@@ -283,23 +283,27 @@ const config = computed<VueUiHorizontalBarConfig>(() => {
...
     </ClientOnly>
 
     <template v-else>
-      <div class="flex flex-col gap-2 justify-center items-center mb-2">
-        <SkeletonInline class="h-4 w-16" />
-        <SkeletonInline class="h-4 w-28" />
-      </div>
-      <div class="flex flex-col gap-1">
-        <SkeletonInline class="h-7 w-full" v-for="pkg in packages" :key="pkg" />
+      <div class="flex flex-col gap-3" data-test="facet-bar-chart-skeleton">
```
shaundai, line 299: we're repeating the same element twice. could we extract this into its own separate component and reuse in multiple places?

shaundai (changes requested):

test/nuxt/components/compare/FacetBarChart.spec.ts
graphieros: This is indeed not really in the scope of this issue, but can be part of an iteration pertaining to test coverage improvements.

app/components/Compare/FacetBarChart.vue
graphieros: Repeating just twice, especially when only localised to this file only, is ok.

ghostdevv:
I think this has already been fixed, but thank you for your contributions!

PR #2121, opened 2026-03-17
feat: opt-in repeated pulse animation on download graph

trueberryless:
### 🧭 Context

As initially suggested on Discord, it would be nice to bring pack the looping pulse animation of the weekly download graph. Because it is not accessibly, we leave it disabled by default, and allow users to opt-in with a new setting. 

### 📚 Description

Adapted the `WeeklyDownloadGraph.vue` component (pinging @graphieros as they are the export of their component ❤️ ). Additionally introduced a new setting, new text strings in `en.json` etc.

Also pinging @knowler to review this change for accessibility please 🙏

### DIFF
--- a/app/components/Package/SelectionView.vue
+++ b/app/components/Package/SelectionView.vue
@@ -3,8 +3,7 @@ defineProps<{
   viewMode?: ViewMode
 }>()
 
-const { selectedPackages, clearSelectedPackages, selectedPackagesParam, closeSelectionView } =
-  usePackageSelection()
+const { selectedPackages, clearSelectedPackages, selectedPackagesParam } = usePackageSelection()
 
 const { data, pending } = useAsyncData(
   async () => {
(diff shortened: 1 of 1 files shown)


graphieros:
I think you should also update the settings state when using the easter egg

trueberryless:
> I think you should also update the settings state when using the easter egg

I don't quite fully understand what the easter egg is, could you explain what would need updating there, please? 🥺

graphieros:
> > I think you should also update the settings state when using the easter egg
> 
> I don't quite fully understand what the easter egg is, could you explain what would need updating there, please? 🥺

Oh of course, sorry.
WeeklyDownloadStats.vue listens to keyboard events and toggles the animation when the sequence of arrows 'up right left up left right' is detected.

So if a user happens to use the easter egg, the state in the settings can be out of sync.

trueberryless:
Is this what you would also expect or do you think it should always activate when hitting the combination?

graphieros:
> Is this what you would also expect or do you think it should always activate when hitting the combination?

I'm sorry I edited by mistake your comment.

More like toggling through the easter egg changes the settings state too, so it is in sync and remains. Wdyt ?

graphieros:
> I’m curious if instead of creating a specific settings for this, if maybe we should create one called “Whimsical mode” or “Whimsy mode” which we could group future features under, so that we have a bit of a better contract with the user. In the description, we could still list this feature as an example, but also should note that some features might be distracting.

I agree, the setting could be more broad indeed

i18n/locales/en.json
```diff
@@ -139,7 +139,7 @@
     "hide_platform_packages": "Hide platform-specific packages in search",
     "hide_platform_packages_description": "Hide native binary packages like {'@'}esbuild/linux-x64 from results",
     "enable_graph_pulse_loop": "Enable looping of pulse effect on mini graph",
-    "enable_graph_pulse_loop_description": "Activate the repeated animation of the pulse effect on the weekly download graph",
+    "enable_graph_pulse_loop_description": "Activate the repeated animation of the pulse effect on the weekly download graph. This can be distracting for some users.",
```
graphieros: Slight reformulation:

Activate a continuous pulse animation on the weekly download graph. This animation may be distracting for some users.

PR #2138, opened 2026-03-18
fix: normalize metrics language across possible values

### LINKED ISSUE #2131: English Localization for ESM and Types support should be consistent across possible values
### Describe the bug

Currently, the tooltip for ESM says `No ESM Support` if there is no ESM but `ESM Supported` if there is ESM. CJS Says `CJS Supported`. We should normalize the first one to `ESM Supported` so it's consistent with the others.

Additionally, the Types support possible values are `Types included`, `Types available via {package}`, and `No TypeScript types`. We should normalize the last one to the others and say `No types`.

Totally chill if we want to normalize another direction but figured I'd just note this in an issue and PR it really quick.

### Additional context

_No response_

### Logs

```shell-script

```

bnb:
### 🔗 Linked issue

resolves #2131

### 🧭 Context

Tiny change to normalize a few strings just to bring the structure of what's being communicated in-line so it's consistent across the board.

Normalizes two strings into the same formats as their peers.

### 📚 Description

Currently, the tooltip for ESM says `No ESM Support` if there is no ESM but `ESM Supported` if there is ESM. CJS Says `CJS Supported`. This PR normalizes the first to be `ESM Supported` so it's consistent with the others.

The tooltops for Types are `Types included`, `Types available via {package}`, and `No TypeScript types`. This PR normalizes the last to be the same as the other two and changes it to `No types`.

### DIFF
--- a/i18n/locales/en.json
+++ b/i18n/locales/en.json
@@ -524,11 +524,11 @@
     "metrics": {
       "esm": "ES Modules supported",
       "cjs": "CommonJS supported",
-      "no_esm": "No ES Modules support",
+      "no_esm": "ES Modules unsupported",
       "types_label": "Types",
       "types_included": "Types included",
       "types_available": "Types available via {package}",
-      "no_types": "No TypeScript types"
+      "no_types": "No types"
     },
     "license": {
       "view_spdx": "View license text on SPDX",


bnb:
definitely didn't open this against my own fork first; if you saw that no you didn't 🤫

PR #2134, opened 2026-03-18
feat: use i18n for hardcoded strings in components and pages

Sigmabrogz:
Closes #2128

Replaced hardcoded strings with i18n translations across the following files:
- `app/pages/blog/index.vue`
- `app/pages/diff/[[org]]/[packageName]/v/[versionRange].vue`
- `app/components/diff/ViewerPanel.vue`
- `app/components/Package/SkillsModal.vue`

Added missing keys to `en.json`.

PR #2140, opened 2026-03-18
fix(i18n): add missing translation keys

### LINKED ISSUE #2128: Add missed translations to terms
We have hardcoded text in some places. We need to convert them to i18n terms and use them as translations. Some of those places:
- https://github.com/npmx-dev/npmx.dev/blob/4c4fe63f17a700e1db45c7537207d11101355e48/app/pages/blog/index.vue
- https://github.com/npmx-dev/npmx.dev/blob/adb39592880ff29d588b010c839868c8d3b8b704/app/pages/diff/%5B%5Borg%5D%5D/%5BpackageName%5D/v/%5BversionRange%5D.vue
- https://github.com/npmx-dev/npmx.dev/blob/adb39592880ff29d588b010c839868c8d3b8b704/app/components/diff/ViewerPanel.vue
- https://github.com/npmx-dev/npmx.dev/blob/adb39592880ff29d588b010c839868c8d3b8b704/app/components/Package/SkillsModal.vue

nadaniels:
### 🔗 Linked issue

Resolves #2128
Related PR: #2134 

### 📚 Description
This PR addresses the linked issue by adding missing translation keys and removing hardcoded text, including 22 new keys across 7 files.

Additional Notes:
- Removes the fallback excerpt text in the [blog list](https://github.com/npmx-dev/npmx.dev/compare/main...nadaniels:npmx.dev:fix-2128-missing-translation-keys?expand=1#diff-5385ef7432b85023c447818e699f06a16aa5f9fcbcb1bfbf586cadccafbdbb80L40), since the blog `description` is enforced by the schema.
- Purposefully does not add a key to `skills-npm` in the [SkillsModal](https://github.com/npmx-dev/npmx.dev/blob/adb39592880ff29d588b010c839868c8d3b8b704/app/components/Package/SkillsModal.vue) as mentioned in the linked issue, since it's an official package name. A key for `skills CLI` was added, following the existing pattern for `npm CLI`.

I'm not sure how best to proceed with the other PR that was opened while this was in progress. Happy to combine efforts if that's the best way to move forward, given additional guidance. Thanks!

### DIFF
--- a/app/pages/diff/[[org]]/[packageName]/v/[versionRange].vue
+++ b/app/pages/diff/[[org]]/[packageName]/v/[versionRange].vue
@@ -134,10 +134,13 @@ useSeoMeta({
 
     <!-- Error: invalid route -->
     <div v-if="!versionRange" class="container py-20 text-center">
-      <p class="text-fg-muted mb-4">
-        {{ $t('compare.version_invalid_url') }}
-        <code class="font-mono text-sm">/diff/package/v/from...to</code>
-      </p>
+      <i18n-t keypath="compare.version_invalid_url_format.hint" tag="p" class="text-fg-muted mb-4">
+        <code class="font-mono text-sm"
+          >/diff/{{ packageName }}/v/{{
+            $t('compare.version_invalid_url_format.from_version')
+          }}...{{ $t('compare.version_invalid_url_format.to_version') }}</code
+        >
+      </i18n-t>
       <NuxtLink :to="packageRoute(packageName)" class="btn">{{
         $t('compare.version_back_to_package')
       }}</NuxtLink>
--- a/i18n/locales/en.json
+++ b/i18n/locales/en.json
@@ -1190,7 +1190,11 @@
     "comparing_versions_label": "Comparing versions...",
     "version_back_to_package": "Back to package",
     "version_error_message": "Failed to compare versions.",
-    "version_invalid_url": "Invalid comparison URL. Use format:",
+    "version_invalid_url_format": {
+      "hint": "Invalid comparison URL. Use format: {0}",
+      "from_version": "from",
+      "to_version": "to"
+    },
     "version_selector_title": "Compare with version",
     "summary": "Summary",
     "deps_count": "{count} dep | {count} deps",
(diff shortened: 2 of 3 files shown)


app/pages/diff/[[org]]/[packageName]/v/[versionRange].vue
```diff
@@ -135,21 +135,26 @@ useSeoMeta({
     <!-- Error: invalid route -->
     <div v-if="!versionRange" class="container py-20 text-center">
       <p class="text-fg-muted mb-4">
-        Invalid comparison URL. Use format: /diff/package/v/from...to
+        {{ $t('compare.version_invalid_url') }}
+        <code class="font-mono text-sm">/diff/package/v/from...to</code>
```
Comment (nadaniels) [inline app/pages/diff/[[org]]/[packageName]/v/[versionRange].vue]: I wasn't sure what the best way to handle this would be, and if we should attempt to add translations for `package`/`from`/`to`. Seems like if we did, we'd want additional brackets or something to separate them out from the "hard-coded" parts of the url (e.g. `/diff/[packageName]/v/[from]...[to]`). Not totally sure what a i18n key for that would look like.

Comment (MatteoGabriele) [reply app/pages/diff/[[org]]/[packageName]/v/[versionRange].vue]: I do think we need to translate the "from" and "to" strings. 
something like `$t('compare.version_diff_format', { packageName })` and the string would be: 

en.json: 
```json
{
  "version_invalid_format": {
    "hint": "Invalid comparison URL. Use format: {0}",
    "format": "/diff/{packageName}/v/from...to"
  },
}
```

then you use the [i18n-t ](https://vue-i18n.intlify.dev/guide/advanced/component#basic-usage)component with something like this

```html
<i18n-t keypath="version_invalid_format.hint" tag="p" for="format">
    <code>{{ $t("version_invalid_format.format") }}</code>
</i18n-t>
```

I hope this is what you were looking for

Comment (nadaniels) [reply app/pages/diff/[[org]]/[packageName]/v/[versionRange].vue]: Thanks for that! I added the i18n component, and split the format details into additional keys for `from` and `to`, while leaving the rest hard-coded (with the `packageName` interpolated). Hopefully that reduces the potential for introducing errors to the full string in the translations.

Comment (MatteoGabriele) [reply app/pages/diff/[[org]]/[packageName]/v/[versionRange].vue]: Thanks! I do find the HTML a bit harder to read, but I think it would probably reduce the number of typos when translating the example.

PR #2152, opened 2026-03-19
feat(docs): add custom badge generator

trueberryless:
### 🧭 Context

In the last npmx meetup, we were discussing the idea of adding a generator inside the docs for custom badges to simplify the process of getting the Markdown content.

So I played around with the idea a little bit and landed on this implementation, which I find pretty useful.

### 📚 Description

I moved the custom badges docs to their dedicated page because they are like 50% of the features page and I think since there are many subheadings, it makes sense to be able to see them in the ToC (therefore the dedicated page).

And then I created a little `BadgeGenerator.vue` which let's the user input the package name and the type of info they wanna have in the badge with a little preview.

Feel free to suggest reverts or further ideas for this improvement.

I also played around a bit with updating the package in the ## Examples section dynamically to what the user currently has typed in the input, but I couldn't figure out a clean way to create the code block like this with a component, which sucks a little bit. Therefore, I dropped this, but if you know how we could do that, feel free to guide me to a possible solution.

Here is a deployment preview of the right page: https://docsnpmxdev-git-fork-trueberryless-feature-docs-cus-74ef86-npmx.vercel.app/guide/badges

### DIFF
--- a/app/components/Package/ExternalLinks.vue
+++ b/app/components/Package/ExternalLinks.vue
@@ -23,15 +23,6 @@ const homepageUrl = computed(() => {
   return homepage
 })
 
-const fundingUrl = computed(() => {
-  let funding = displayVersion.value?.funding
-  if (Array.isArray(funding)) funding = funding[0]
-
-  if (!funding) return null
-
-  return typeof funding === 'string' ? funding : funding.url
-})
-
 const PROVIDER_ICONS: Record<string, IconClass> = {
   github: 'i-simple-icons:github',
   gitlab: 'i-simple-icons:gitlab',
@@ -95,11 +86,6 @@ const repoProviderIcon = computed((): IconClass => {
     <li v-if="jsrInfo?.exists && jsrInfo.url">
       <LinkBase :to="jsrInfo.url" :title="$t('badges.jsr.title')" classicon="i-simple-icons:jsr">
         {{ $t('package.links.jsr') }}
-      </LinkBase>
-    </li>
-    <li v-if="fundingUrl">
-      <LinkBase :to="fundingUrl" classicon="i-lucide:heart">
-        {{ $t('package.links.fund') }}
       </LinkBase>
     </li>
   </ul>
--- a/app/components/Package/Header.vue
+++ b/app/components/Package/Header.vue
@@ -241,6 +241,15 @@ const likeAction = async () => {
     isLikeActionPending.value = false
   }
 }
+
+const fundingUrl = computed(() => {
+  let funding = props.displayVersion?.funding
+  if (Array.isArray(funding)) funding = funding[0]
+
+  if (!funding) return null
+
+  return typeof funding === 'string' ? funding : funding.url
+})
 </script>
 
 <template>
@@ -312,6 +321,15 @@ const likeAction = async () => {
             </span>
           </ButtonBase>
         </TooltipApp>
+
+        <LinkBase
+          variant="button-secondary"
+          v-if="fundingUrl"
+          :to="fundingUrl"
+          classicon="i-lucide:handshake text-accent"
+        >
+          <span class="max-sm:sr-only">{{ $t('package.links.fund') }}</span>
+        </LinkBase>
       </div>
     </div>
   </header>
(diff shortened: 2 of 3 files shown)


graphieros:
What do you think of having the option to create a custom badge ? Any label, any value

trueberryless:
> What do you think of having the option to create a custom badge ? Any label, any value

Since the existing parameters allow for that, I would say we could create a second component for the parameter section, which allows to set all the parameters. But I didn't wanted to overload the start of the page with such an overwhelming component. Awesome idea, gonna work on it 👍

jycouet:
> > What do you think of having the option to create a custom badge ? Any label, any value
> 
> Since the existing parameters allow for that, I would say we could create a second component for the parameter section, which allows to set all the parameters. But I didn't wanted to overload the start of the page with such an overwhelming component. Awesome idea, gonna work on it 👍

You can say that you want 2 contributions it's ok 😅

---

Awesome work @trueberryless _IRL meetups are everything ^^_

trueberryless:
> You can say that you want 2 contributions it's ok 😅

Actually, I want to do the addition of the second component in this PR as well, to ship this feature as one single unit. That way it feels cohesive.

trueberryless:
I'm gonna work on the "any value" part tomorrow, as this is actually not currently supported by the API, so I'll have to add it 👍

docs/app/components/BadgeGenerator.vue
```diff
@@ -5,41 +5,36 @@ const isValid = ref(true)
...
-    copied.value = false
-  }, 2000)
+  try {
+    await navigator.clipboard.writeText(markdown)
+    copied.value = true
+    setTimeout(() => {
+      copied.value = false
+    }, 2000)
+  } catch {
+    console.error('Failed to copy to clipboard')
```
graphieros: This should probably be forwarded to the user explicitely, maybe as an error state in the copy button

docs/app/components/BadgeGeneratorParameters.vue
```diff
@@ -77,14 +77,25 @@ watch([pkg, type, queryParams], () => {
...
-    copied.value = false
-  }, 2000)
+  try {
+    await navigator.clipboard.writeText(markdown)
+    copied.value = true
+    setTimeout(() => {
+      copied.value = false
+    }, 2000)
+  } catch {
+    console.error('Failed to copy to clipboard')
```
graphieros: The copy to clipboard could probably be a utility (but this refactoring could be dealt with in a separate issue to address all instances)

graphieros:
In the list, I would recommend to position the colour badges at the start of the lines, so they are all nicely aligned. Wdyt ?

docs/app/components/BadgeGeneratorParameters.vue
ghostdevv: I think there is a use clipboard or smth?

gameroman: https://vueuse.org/core/useClipboard/

trueberryless: Interesting, this didn't work for me, but that might have been a problem with dependencies not installing correctly. I'm gonna try again 👍

trueberryless:
> In the list, I would recommend to position the colour badges at the start of the lines, so they are all nicely aligned. Wdyt ?

You mean these badges here at the end to the front?

image

I'm not totally sure if this would make it cleaner. First, some have two colors depending on the value, like vulnerabilities or types. Second, this would make the actual type not aligned, and the boldness of it wouldnt be as noticeable. I personally think that the colour badges should be a subtle addition at the end.

trueberryless:
Alright, I think I got every suggestion!

Feel free to suggest more adjustments, like ordering of the sidebar (maybe badges should be right below features), refactoring to some utilities or such ❤️

ghostdevv:
After some sleep I'll figure out why I broke knip and we can merge this!

ghostdevv:
This reminds me actually that @yuyinws made a [badge builder website](https://npmx-badge.vercel.app/) - if there is anything missing from this impl we could add it in a follow up :eyes:

trueberryless:
We could add a link to it for more complex generations in another PR 🙌

PR #2155, opened 2026-03-19
refactor: remove npms

trueberryless:
### 🧭 Context

As [discussed on Discord](https://discord.com/channels/1464542801676206113/1464926498027995170/1484282327189160037), we decided to remove all npms traces since the data is 4 years outdated.

### 📚 Description

https://npms.io uses and provides outdated data, like e.g. this Astro package, which is 4 years old:

image

The only place where npms is really used is [custom badges](https://docs.npmx.dev/guide/features#custom-badges), which provide SVGs via an API mainly for markdown in Readmes. We decided that no data is better than wrong data, therefore deleting all traces.

### DIFF
--- a/docs/shared/utils/badges.ts
+++ b/docs/shared/utils/badges.ts
@@ -15,10 +15,6 @@ export const BADGE_TYPES = Object.freeze([
   'types',
   'maintainers',
   'deprecated',
-  'quality',
-  'popularity',
-  'maintenance',
-  'score',
   'name',
 ] as const)
 


43081j:
i think the e2e badge tests need updating 👀

trueberryless:
Hmm, I discovered a lot more places where "popularity", "quality", "maintainence" and "score" are used, I'll push a commit that removes those as well, but they might not be neccessary to remove 🤔

trueberryless:
Okay, I might have missed many use cases of the four scores throughout the project. I am not totally sure if they all use npms, but they seems to show up everywhere. I'm not sure if we should maybe just remove them from custom badges, if the other occurances even use npms or what exactly to do here 🤔 

Also I think a lot of i18n files would still need to be updated (remove the translations), but I didn't do it for now, because I wanted to get confirmation, whether I am doing the right thing (we are speaking of removing about 1000 lines 😅).

We can always revert the last commit, but I'd like to get a little guidance here...

43081j:
if you can fix the type errors, looks good to me 👍 

edit: ah and some conflicts with main

trueberryless:
Autofix seems to have run some i18n script which adds in empty translations. Feel free to commit here, I have to run 👋

43081j:
no clue what autofix is doing, why it adds all these empty keys...

but if ci passes, whatever. ill believe whatever i18n lint we run is doing its job 😅

trueberryless:
maybe this fixes it, let's see: https://github.com/npmx-dev/npmx.dev/commit/82d00cd764d2bc5877ca77b718bef6010876087f

trueberryless:
I hope we don't remove any feature that was valueable to many people: Just the be clear, I did not exactly investigate _what_ I have removed here, I just searched for all occurances of the four percentage stats (quality, popularity, maintenance and overall score) 🪭 

But I really agree, that the dependency of npms shouldn't exist and if we find a better alternative, we can implement the features again 👍

ghostdevv (approved):
I can't find anything else either, so LGTM xD

PR #2162, opened 2026-03-20
fix: add legacy support for clipboard copy functionality

### LINKED ISSUE #2151: "Copy as Markdown" button on README doesn't copy to clipboard in Safari
### Describe the bug

**Root cause:**

**Steps to reproduce:**

3. Paste — nothing was copied

**Suggested fix:**

rzzf:
### 🔗 Linked issue

resolves #2151


rzzf:
This doesn't seem to have any effect.

PR #2171, opened 2026-03-20
feat(ui): make version colors consistent

MatteoGabriele:
### 🧭 Context

The "latest" tag on the VersionSelector dropdown does not match the colors used for versions in the Package Version sidebar.

Screenshot 2026-03-20 at 15 39 38

### 📚 Description
Creates the `badge-accent` shortcut to synchronize accent colors for VersionSelector and Package Version matching.

Screenshot 2026-03-20 at 15 40 15


RYGRIT:
Should we use capital letters?

ghostdevv:
> Should we use capital letters?

I think it's more consistent to leave it as is, but feel free to open an issue? or you can make a PR? generally I'll try to merge units of work to keep stuff moving, so it's better as a follow up if you see what I mean

PR #2176, opened 2026-03-21
feat: add semver range filter and element title support

ShroXd:
### 🔗 Linked issue

n/a

### 🧭 Context

Following work of #2105 , was mentioned in the PR review comment.
cc: @mbtools

### 📚 Description

Add description for semver range filtering and validator. The normal error message would make the page shake during input, so an inline icon is used to indicate the error state instead.

https://github.com/user-attachments/assets/5b41bc59-92d0-4636-a326-dbc3a6e77f75

Also added titles to elements.

| Element | Screenshot |
| :- | :- |
| Version | QQ20260321-170411 |
| Time | QQ20260321-170358 |


mbtools:
Works well. Try `^6.0.0 || ^7.0.0 || ^8.0.0-0` for `vite` package. The input box should be a bit wider and the placeholder "Filter by semver". 

After filtering, this icons keeps spinning. Maybe replace with a fixed "funnel" icon indicating that the list was filtered.

image


My final wish:

I'm testing with `verdaccio` package which has lots of versions. A small improvement would be to mark the majors as deprecated, for example, `5.x`, `4.x`, and below, in case all versions underneath it are deprecated. 

image

Similarly, if a tag points to a deprecated version, we could add a "deprecated" badge in the tag list:

image

image

ShroXd:
Hi @mbtools, thanks for the feedback! I have updated my PR.

Because of the design of fetching, the information about whether a version is deprecated can only be known after phase 2 is finished. Therefore I changed the behavior of the network request, making phase 2 fire automatically after phase 1 is finished.

This puts additional rendering pressure on the main thread. The performance profiling shows a significant number of `runIfDirty` calls, which is caused by the long reactive chain of `fullVersionMap` and derived data. Some potential optimizations include batch updating of `fullVersionMap` or even using markRaw to skip the reactive process for `fullVersionMap`. However, the page is still fast, so I would avoid early optimization and address it if we detect any performance issues on this page later.

mbtools:
I tried on my phone and it's super fast even with over a thousand versions in a package like @nx/js. 

LGTM 👍 

PS wishlist: two toggle buttons to hide deprecated and pre-release versions. I would hide them by default since these versions are only relevant for maintainers or early adopters ie a fraction of the users.

ShroXd:
@mbtools Happy to hear you like it!

The idea of a toggle button to hide some versions is great! I would move it to a separate PR because although the implementation is simple, some business logic needs to be discussed, like:

1. Should we allow users to configure it in the settings page?
2. The scope of hiding versions. Some packages like [astro](https://main.npmx.dev/package/astro/versions) have a lot of tagged versions.
3. Although some versions are deprecated, they are still used by many projects, like Vue 2. So the default value should be considered carefully.

I can write an issue later. Once we gather enough thoughts from the community, I can finish the remaining things.

ghostdevv (approved):

PR #2182, opened 2026-03-21
fix(ui): make readme markdown copy work in safari

### LINKED ISSUE #2151: "Copy as Markdown" button on README doesn't copy to clipboard in Safari

MathurAditya724:
### 🔗 Linked issue

Fixes #2151

### 🧭 Context

The \"Copy as Markdown\" button in the README section wasn’t working in Safari.
It could fail silently because we were waiting for an async fetch before writing to clipboard.

### 📚 Description

This PR makes the README copy flow more reliable in Safari:

- enabled VueUse clipboard `legacy` mode so it can fall back to `execCommand('copy')` when needed
- reuse already-fetched markdown if it exists
- only fetch markdown on click when it’s missing

Result: \"Copy as Markdown\" now works consistently, including Safari.


app/pages/package/[[org]]/[name].vue
```diff
@@ -108,9 +109,12 @@ function prefetchReadmeMarkdown() {
 }
 
 async function copyReadmeHandler() {
-  await fetchReadmeMarkdown()
+  let markdown = readmeMarkdownData.value?.markdown
+  if (!markdown) {
+    await fetchReadmeMarkdown()
+    markdown = readmeMarkdownData.value?.markdown
+  }
```
Comment (ghostdevv) [inline app/pages/package/[[org]]/[name].vue]: I tried this out in a webkit based browser and it didn't work for me, but that kinda seems like it makes sense since there is still an async action happening. There is this blog post that seems promising, but I haven't managed to get it to work so far - will try again later

https://wolfgangrittner.dev/how-to-use-clipboard-api-in-safari/

ghostdevv (changes requested):

MathurAditya724:
I re-tested this on WebKit/Safari with the latest changes and it now works reliably, including the first click after navigation.  I added a loading state + markdown prefetch, and kept the clipboard legacy fallback.  Here’s a quick demo -

https://github.com/user-attachments/assets/c0afbc0a-d864-4bef-b191-4c9903a5b7d6

MatteoGabriele:
Thanks a lot for the contribution, but I will close this PR as a fix has already been merged. 🙏

PR #2184, opened 2026-03-21
feat: upgrade to vue-router v5 to migrate away from unplugin-vue-router

trueberryless:
### 🧭 Context

Renovate's Dependency Dashboard shows that the `unplugin-vue-router` is depricated and should be migrated according to https://router.vuejs.org/guide/migration/v4-to-v5.html#From-unplugin-vue-router:

image

PR link: https://github.com/npmx-dev/npmx.dev/issues/2

### 📚 Description

This PR updates all occurances of `vue-router` to at least v5 and removes most `unplugin-vue-router` packages (for some reason, there are still some occurances in the lockfile 🤔)


ghostdevv:
I think we should do this work in #2047 as I think the Nuxt upgrade is also required? WDYT? @trueberryless

trueberryless:
> I think we should do this work in #2047 as I think the Nuxt upgrade is also required? WDYT? @trueberryless 

Yeah, I didn't see that one, but I don't have write access to those Renovate branches anyway.

I get the same type errors tho as in that upgrade. So this might wanna wait for the other PR

ghostdevv:
I'll close this for now then, but we can reopen this one if you have an idea of how to fix it but my guess currently is that it may've been an accidental breaking change in Nuxt that needs addressing, and as such we need Daniel.

PR #2185, opened 2026-03-21
feat: package quality score

### LINKED ISSUE #48: display package quality score and health indicators
maintenance activity, type coverage, docs quality - similar to jsr

Flo0806:
### 🔗 Linked issue

Resolves: #48

### 🧭 Context

With this npmx has a quality and health check inside the detail sidebar.

### 📚 Description

This checks can looks like:
- Has a substantial README
- Has a description
- Recently updated
- TypeScript types
- Has a license
- Has a repository link
- Has publish provenance
- Supports ES modules
- No known vulnerabilities
- Not deprecated

Maximum possible points are 14.

### Tests
Unit tests for point calculations

image

### DIFF
--- a/app/components/Package/QualityScore.vue
+++ b/app/components/Package/QualityScore.vue
@@ -3,10 +3,32 @@ const props = defineProps<{
   score: PackageScore
 }>()
 
+const categoryLabels: Record<ScoreCategory, string> = {
+  documentation: $t('package.score.categories.documentation'),
+  maintenance: $t('package.score.categories.maintenance'),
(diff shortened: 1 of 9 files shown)


43081j:
thanks for taking a look at this, but i think we really need to spend some time discussing what the right metrics are, how we want to score things (if we want to "score" them at all).

i see there's an open issue but we still haven't decided any of this. e.g. do we want a score? do we want just a list of health checks? what should be weighted in what way? etc etc.

i think we need to move back to the issue and have what'll probably be a lengthy discussion before we can change any code.

PR #2186, opened 2026-03-21
fix: use `displayVersion` instead of `pkg`

### LINKED ISSUE #2163: License is always shown from latest version
### Describe the bug

Take a look at glob:
https://npmx.dev/package/glob

Now at v9 of glob:
https://npmx.dev/package/glob/v/9.3.5

both show blueoak license but the latter is actually ISC.

### Additional context

_No response_

### Logs

```shell-script
!
```

Flo0806:
### 🔗 Linked issue

Fixes: #2163

### 🧭 Context

If you switch to older/other versions in a package which have a different version the version doesn't change

### 📚 Description

Problem is the loaded packages (latest/current version) was used to show license informations. Now we use the `displayVersion` for that and we now always see the current license.

### DIFF
--- a/app/components/OgImage/Package.vue
+++ b/app/components/OgImage/Package.vue
@@ -158,7 +158,11 @@ try {
         </span>
 
         <!-- License (if any) -->
-        <span v-if="displayVersion?.license || pkg?.license" class="flex items-center gap-2" data-testid="license">
+        <span
+          v-if="displayVersion?.license || pkg?.license"
+          class="flex items-center gap-2"
+          data-testid="license"
+        >
           <svg
             viewBox="0 0 32 32"
             :fill="primaryColor"
--- a/app/components/OgImage/Package.vue
+++ b/app/components/OgImage/Package.vue
@@ -158,7 +158,7 @@ try {
         </span>
 
         <!-- License (if any) -->
-        <span v-if="pkg?.license" class="flex items-center gap-2" data-testid="license">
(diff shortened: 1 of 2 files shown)


ghostdevv (changes requested):
Ideally we fix this where we're loading the packument from the registry, since presumably we're either getting `license` from the top level of the packument which is just referring to latest. We should be getting license, and most other data, from the packument version instead if that makes sense?

Flo0806:
> Ideally we fix this where we're loading the packument from the registry, since presumably we're either getting `license` from the top level of the packument which is just referring to latest. We should be getting license, and most other data, from the packument version instead if that makes sense?

displayVersion resolves always the current license, doesn't it? From this it works as well for me with this fix. Did I overlook anything?

RYGRIT:
Perhaps she meant to directly handle the `license` attribute in the pkg object, so we don't have to do the same thing in every rendering place.

ghostdevv:
> Perhaps she meant to directly handle the `license` attribute in the pkg object, so we don't have to do the same thing in every rendering place.

apologies for the delay, but yea! like ideally when we're actually getting this data from the registry we'd use the correct field. It could be more complicated than it seems though, I haven't had the time to actually poke around to see how, so let me know if it's not possible with our current setup. But then this would also hopefully fix stuff like the og images using the wrong license that was also reported in the original issue.

Flo0806:
I'll have a look 🙏

Issue #2153, opened 2026-03-19
charts improvements (a11y among others)


graphieros:
This issue lists all charts a11y improvements to be done.
It is not exhaustive, and can be completed iteratively, and include other types of chart improvements or implementations.

## Improve series recognition for users with achromatopsia

- [x] bar charts - add background patterns

## Improve chart navigability for keyboard users
- [x] update vue-data-ui components with keyboard navigation in the chart area & legends + customisable visual hints

## Improve chart data access
- [x] update vue-data-ui components with hidden data tables & hints
- [ ] access to visual data tables when possible

## various chart options
- [x] quadrant chart on the compare page

graphieros:
@knowler, I'm pasting here the second point you made in your comment on #2193:

> Provide a visual of the data table since that can be valuable.

I think there are two different cases, on which I would love to have your thoughts:

## 1. large charts (downloads in the modal and the compare page, versions modal, bar charts in the compare page)
  a. we could consider having a toggle (or tabs) to access a visual data table when the chart is constrained by a modal or by a small grid (like the bar charts in the compare page). Would the built-in sr data table be necessary in this case, since the data table would have to be accessed through an extra step ?
  b. in the compare page, we could display the data table under the chart itself, without a toggle. In this case, would the built-in sr data table be redundant ? If so, an update would be necessary in vue-data-ui to add a flag to allow disabling this table from the configuration.
  
## 2. small charts (downloads sparkline)
displaying a data table for the weekly downloads sparkline chart of the package page (with 52 rows) is obviously not desirable, leading to a similar solution as 1a (adding a button to show a data table in a modal ?).

PR #2193, opened 2026-03-22
feat: charts a11y improvements

graphieros:
This PR is part of #2153 

- Bump vue-data-ui to 3.16.0 ([release notes](https://github.com/graphieros/vue-data-ui/releases/tag/v3.16.0))
  - keyboard navigation in the chart areas to cycle through datapoints when charts are in focus
  - sr hidden data tables & hints

- Add translations for hints & sr sections
- Update focus styles

Updated charts:
- sparkline (weekly downloads on package page)
- downloads chart in modal (package page)
- versions chart in modal (package page=
- downloads charts (compare page)
- bar charts (compare page)

### DIFF
--- a/app/components/Package/TrendsChart.vue
+++ b/app/components/Package/TrendsChart.vue
@@ -1390,7 +1390,9 @@ const chartConfig = computed<VueUiXyConfig>(() => {
     theme: isDarkMode.value ? 'dark' : ('' as VueDataUiTheme),
     a11y: {
       translations: {
-        keyboardNavigation: $t('package.trends.chart_assistive_text.keyboard_navigation_horizontal'),
+        keyboardNavigation: $t(
+          'package.trends.chart_assistive_text.keyboard_navigation_horizontal',
+        ),
         tableAvailable: $t('package.trends.chart_assistive_text.table_available'),
         tableCaption: $t('package.trends.chart_assistive_text.table_caption'),
       },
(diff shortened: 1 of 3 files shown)


graphieros:
> Anyway it looks good and is already a great improvement, thanks 🤍

Thank you @alexdln :)

This is part of gradual a11y improvements, and I know it can be even better.
I would suggest to ship this for a start, at least we have kbd navigation, then I can continue to improve (and learn notably from @knowler).

i18n/locales/en.json
```diff
@@ -482,6 +482,12 @@
       "maintainer_template": "{avatar} {char126}{name}"
     },
     "trends": {
+      "chart_sr": {
```
ghostdevv: what does `sr` mean?

ghostdevv (changes requested):

i18n/locales/en.json
graphieros: screen reader
I should probably prefer full names

ghostdevv (approved):

knowler (approved):
This is a great iterative improvement.

- - -

Eventually:

1. We’ve already discussed this in the builder Discord, but we could add some relevant semantics to the data point exploration thing (e.g. `role=slider` or something). This likely will provide some feedback for us when the data point changes (perhaps without a live region).
    - Right now even though there is a screen reader description for how to use the image controls, there’s no feedback for them. A (debounced) live region could help with this as another iterative step without adding all of the extra requirements for the semantic change.
2. Provide a visual of the data table since that can be valuable.

PR #2203, opened 2026-03-22
fix: show copy buttons on touch devices and fix aria-label mismatch

### LINKED ISSUE #1771: Copy buttons for commands are missed on mobile
### Describe the bug

On Package page each command has a "Copy" button next to it. It is visible on hover state for the desktop version. But it is missed on mobile devices since there is no hover state.

**Steps to reproduce:**

- Use mobile device or emulator in your browser
- Go to https://npmx.dev/package/nuxt#get-started
- In section "GET STARTED" check no button is visible to copy commands
- Still copied if you click on command, but it's not intuitive

### Additional context

Image

Image

thegreatalxx:
### 🔗 Linked issue
resolves #1771

### 🧭 Context
Two related issues with `CopyToClipboardButton`: (1) copy buttons were hidden on touch devices because they relied on a CSS hover state that never fires on mobile, and (2) the button's `aria-label` could differ from its visible text, causing an accessibility label/content mismatch for screen readers.

### 📚 Description
- Added `@media (hover: none)` CSS block to make copy buttons visible and interactable on touch-only devices, where hover is unavailable.
- Fixed the `aria-label` computation so it defaults to `copyText` (the same string shown to sighted users) rather than a separate value, eliminating the WCAG label-content-name mismatch.

A regression test (`test/nuxt/components/CopyToClipboardButton.spec.ts`) verifies that:
- `aria-label` matches the visible `copyText` when no explicit override is given
- An explicit `ariaLabelCopy` prop takes precedence
- The copied state correctly reflects `copiedText` in both label and visible text

### DIFF
--- a/app/components/CopyToClipboardButton.vue
+++ b/app/components/CopyToClipboardButton.vue
@@ -89,17 +89,9 @@ function handleClick() {
 }
 
 @media (hover: none) {
-  /* On touch devices show the button statically (no hover possible) */
+  /* On touch devices, hide the button since hover is not available */
   .copyButton {
-    clip: auto;
-    clip-path: none;
-    height: auto;
-    overflow: visible;
-    width: auto;
-    opacity: 1;
-    translate: none;
-    pointer-events: auto;
-    transition: none;
+    display: none;
   }
 }
 </style>


knowler:
Hey there, there already is an open PR (https://github.com/npmx-dev/npmx.dev/pull/998) to resolve this issue, even though it wasn’t linked (since it came before the issue was filed). We’re actively working on that one—just waiting for a dependency upgrade, so I will be closing this one. Thank you anyway!

PR #2228, opened 2026-03-23
feat: highlight active item in package docs sidebar

iiio2:
Clicking a sidebar TOC item on the docs page navigates to the section, but the sidebar doesn't visually indicate which item is currently selected. Now we can see which item is selected. 

Before: 

before-docs-one

After: 

after-docs-one


ghostdevv (changes requested):
Hmm I wonder if we should use an [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver) instead so that it highlights the top most visible one? That would mean that it wouldn't always show the one you have in your hash though, but when you load the page it would which I think works?

howwohmm (commented):
nice improvement — the before/after screenshots make the value clear.

one thing: updateActiveTocLink only runs on hashchange and toc data changes. if the user scrolls past a heading without clicking the TOC, the active indicator won't update. an IntersectionObserver on the heading elements would make it feel polished — though that could be a follow-up PR to keep this one focused.

the onTocClick + nextTick pattern to handle click-before-hashchange is a nice touch.

PR #2230, opened 2026-03-23
fix: remove conflicting inline tailwind classes from package docs

iiio2:
The TOC section count (e.g. (102)) appears bold in production but normal in development due to conflicting Tailwind utility classes between inline HTML and the CSS rule.

Cause: The <a> tag had inline font-semibold competing with .toc-content > ul > li > a's font-medium, and Tailwind's CSS ordering differs between dev and production builds.

Fix: Removed inline utility classes from the <a> tag in render.ts, letting the existing CSS rule handle styling so font-normal on the (102) span always wins.

Now in development: 

dev-formal

In production: 

in-production

Now in Production:

https://npmxdev-git-fork-iiio2-fix-remove-conflicting-inlin-3f1bac-npmx.vercel.app/package-docs/vue/v/3.5.30


ghostdevv:
I tried to fix the removal of the `hover` style by adding it back, but then it doesn't make the title stand out since it's no longer semi bold. Then I realised that it seems that in production it's working correctly? I'll close this for now, but please re-open if I'm wrong? There is caching of the generated docs which makes it harder to figure out sometimes :melting_face: 

image

PR #2231, opened 2026-03-23
[WIP]chore: normalize import order across codebase

RYGRIT:
### 🧭 Context

+ Organize imports per CONTRIBUTING.md: type → external → internal → relative.
+ Fixes 96 files across app/, server/, test/, and other directories.


RYGRIT:
> are we doing this programmatically?

Yeah, Claude has written an automatic conversion script for me, but I haven't submitted it yet. I don't know if there's a better solution for me to explore further, or do you have any good solutions?

ghostdevv:
> > are we doing this programmatically?
> 
> Yeah, Claude has written an automatic conversion script for me, but I haven't submitted it yet. I don't know if there's a better solution for me to explore further, or do you have any good solutions?

Unless we can do this in the oxlint/oxfmt, I'd lean towards not doing it at all really. There will be different preferences so there would likely need some discussion too. Plus Nuxt has auto imports for a lot of stuff, for better or worse, which reduces the need somewhat.

If you'd like this then perhaps we could start with an issue instead?

RYGRIT:
Makes perfect sense. I completely agree that doing this via a one-off script isn't sustainable, and enforcing it through oxlint/oxfmt is the right approach.

Since there are also Nuxt auto-imports to consider and different preferences among the team, I'll close this PR for now. I'll open an issue as you suggested so we can discuss whether/how to implement an automated import sorting rule in our setup. Thanks for the quick feedback!

PR #2236, opened 2026-03-23
feat: welcome first contributors when PR opens

trueberryless:
### 🧭 Context

I noticed we already have a welcoming messages for new contributors, however this message is not sent until the PR is merged. This way, maintainers are unknowingly during PR review that this is the first PR of the person. Therefore, I created a second action which writes and immediate message to the PR as soon as it gets opened.

### 📚 Description

This PR adds a new workflow which uses https://github.com/zephyrproject-rtos/action-first-interaction as this fork is more performant and flexible (see [example issue](https://github.com/actions/first-interaction/issues/365)) than the official action. It gives the new contributor confidence and tells them what all those bot messages mean, so they are not overwhelmed.

### DIFF
--- a/.github/workflows/welcome-open.yml
+++ b/.github/workflows/welcome-open.yml
@@ -29,5 +29,3 @@ jobs:
 
             3. One or more of our maintainers will take a look and may ask you to make changes.
                We try to be responsive, but don’t worry if this takes a few days.
-
-            Please feel free to ask any questions you may have!


trueberryless:
> Maybe we could do both messages? The opentelling them to ask any questions they might have and/or join the discord and then the close message telling them congratulations and you can go to here to get a contributor role

Awesome idea, so many welcomes, I'm almost overwhelmed.

trueberryless:
I also adapted the new action to use [a fork](https://github.com/zephyrproject-rtos/action-first-interaction) of the first-interaction action, because the original is not as performant (because it's not using the `/search` API) and would fail when no `issue_message` is passed. See more [in this issue](https://github.com/actions/first-interaction/issues/365).

trueberryless:
> I _think_ they could be part of the same .yml file and then we could use the `if` on the job to detect the event? Not 100% though

I don't like putting them in one action because it adds a lot of unnecessary overhead that can cleanly be avoided because we have two different GitHub events (`opened` vs `closed`). So why not keeping them separated?

ghostdevv:
> > I _think_ they could be part of the same .yml file and then we could use the `if` on the job to detect the event? Not 100% though
> 
> I don't like putting them in one action because it adds a lot of unnecessary overhead that can cleanly be avoided because we have two different GitHub events (`opened` vs `closed`). So why not keeping them separated?

I lean towards one just to reduce the file count in the workflows dir - I thought that if we're using the `if` thing then it won't cost us anything?

trueberryless:
> I lean towards one just to reduce the file count in the workflows dir - I thought that if we're using the `if` thing then it won't cost us anything?

I wouldn't consider myself an expert in how GitHub Actions optimize such conditional steps or jobs. I was just thinking about the maintenance of one bigger workflow vs two smaller ones. But I guess more files in the workflow folder also doesn't make maintenance easier, so I would be _okay_ to combine the two (it's just more prune to errors as always with GitHub Actions, let the trail and error begin 😅)

ghostdevv:
> > I lean towards one just to reduce the file count in the workflows dir - I thought that if we're using the `if` thing then it won't cost us anything?
> 
> I wouldn't consider myself an expert in how GitHub Actions optimize such conditional steps or jobs. I was just thinking about the maintenance of one bigger workflow vs two smaller ones. But I guess more files in the workflow folder also doesn't make maintenance easier, so I would be _okay_ to combine the two (it's just more prune to errors as always with GitHub Actions, let the trail and error begin 😅)

Yeaaa GitHub actions is :sparkles: fun :sparkles: - oki let's just leave it as two and if we need to we can merge them together later

ghostdevv:
> I also adapted the new action to use [a fork](https://github.com/zephyrproject-rtos/action-first-interaction?rgh-link-date=2026-03-23T16%3A46%3A06Z) of the first-interaction action, because the original is not as performant (because it's not using the `/search` API) and would fail when no `issue_message` is passed. See more [in this issue](https://github.com/actions/first-interaction/issues/365).

where'd you find the fork? could we just use actions script like the close one?

trueberryless:
We have been using the fork action for quite some time now in Starlight ([ref](https://github.com/withastro/starlight/blob/51e60beb9344000e9f1d3fb76b838458fe5db798/.github/workflows/welcome-bot.yml#L17)) and it runs [flawlessly](https://github.com/withastro/starlight/actions/workflows/welcome-bot.yml) (apart from yesterdays changes, where I tried to use the official action, but it's too restrictive).

I advise to use the fork one instead of a new script as it already works. You can also read many benefits of [in this PR from HiDeoo](https://github.com/withastro/starlight/pull/1139).

.github/workflows/welcome-open.yml
```diff
@@ -25,7 +25,9 @@ jobs:
                If they spot any issues you will see some error messages on this PR.
                Don’t hesitate to ask any questions if you’re not sure what these mean!
```
trueberryless, line 26: I'm not sure if we should repeat the "ask any questions" in line 33 as it already says so here...

ghostdevv: I swear I can read

ghostdevv:
> We have been using the fork action for quite some time now in Starlight ([ref](https://github.com/withastro/starlight/blob/51e60beb9344000e9f1d3fb76b838458fe5db798/.github/workflows/welcome-bot.yml#L17)) and it runs [flawlessly](https://github.com/withastro/starlight/actions/workflows/welcome-bot.yml) (apart from yesterdays changes, where I tried to use the official action, but it's too restrictive).
> 
> I advise to use the fork one instead of a new script as it already works. You can also read many benefits of [in this PR from HiDeoo](https://github.com/withastro/starlight/pull/1139).

ah sweet, let's do it

trueberryless:
> fix: buy glasses 

😂😂😂 I just can't anymore 🤣🤣🤣

ghostdevv:
> > fix: buy glasses
> 
> 😂😂😂 I just can't anymore 🤣🤣🤣

this is why they invented squash merging :relieved: :joy:

PR #2239, opened 2026-03-23
test: collect coverage from untested files too

AriPerkkio:
### 🔗 Linked issue

(looks like everyone leaves this empty)

### 🧭 Context

Currently code coverage is not collected from untested files. You'll need to define `coverage.include` to instruct Vitest which files are considered as source files.

https://vitest.dev/guide/coverage.html#including-and-excluding-files-from-coverage-report

### 📚 Description

Defines source directories in `coverage.includes` with extensions of `ts` and `vue`.

This drops code coverage:

```diff
  ------------------------------------------------------------|---------|----------|---------|---------|
  File                                                        | % Stmts | % Branch | % Funcs | % Lines |
  ------------------------------------------------------------|---------|----------|---------|---------|
- All files                                                   |   62.43 |    50.36 |   58.68 |   63.53 |
+ All files                                                   |   48.50 |    42.03 |   46.77 |   48.81 |

```


ghostdevv (approved):

ghostdevv (changes requested):
I re-added the `exclude` as I think it should still be there? WDYT?

Also the CI has been failing, any ideas? If not I'll try and take a look soon 🙏 I'll mark as draft while we work on that, and then we can make it ready again when it's working!

AriPerkkio:
> I re-added the `exclude` as I think it should still be there? WDYT?

`*.json` does not match the `include` glob's extensions (`.ts`, `.vue`) so it's not needed at all. Does `app/`, `cli/`, `server/` or `shared/` directories contain `node_modules`? 🤔 

Also `node_modules` is always excluded so it doesn't need to be added. https://github.com/vitest-dev/vitest/blob/1f2d318493363855b66a22caaf7c1c10579029d5/packages/vitest/src/node/config/resolveConfig.ts#L497-L523

> Also the CI has been failing, any ideas? If not I'll try and take a look soon 🙏 I'll mark as draft while we work on that, and then we can make it ready again when it's working!

No idea really. Tests should not fail based on collected coverage.

ghostdevv:
> > I re-added the `exclude` as I think it should still be there? WDYT?
> 
> `*.json` does not match the `include` glob's extensions (`.ts`, `.vue`) so it's not needed at all. Does `app/`, `cli/`, `server/` or `shared/` directories contain `node_modules`? 🤔
> 
> Also `node_modules` is always excluded so it doesn't need to be added. [vitest-dev/vitest@`1f2d318`/packages/vitest/src/node/config/resolveConfig.ts#L497-L523](https://github.com/vitest-dev/vitest/blob/1f2d318493363855b66a22caaf7c1c10579029d5/packages/vitest/src/node/config/resolveConfig.ts#L497-L523)

yeaaa you right, I kinda hoped it'd fix CI though :laughing: 

> > Also the CI has been failing, any ideas? If not I'll try and take a look soon 🙏 I'll mark as draft while we work on that, and then we can make it ready again when it's working!
> 
> No idea really. Tests should not fail based on collected coverage.

:melting_face:

ghostdevv (approved):
seems to be working now, ty!

PR #2245, opened 2026-03-23
feat: add timeline tab to package page

### LINKED ISSUE #2244: Add a timeline tab to packages
### Describe the feature

This should basically list the versions published along with any significant changes that came with them.

Initially we should show:

- CJS/ESM
- Trust/provenance
- Dependency count
- Install size
- License

### Additional information

### Final checks

43081j:
### 🔗 Linked issue

#2244

### 🧭 Context

A timeline will help show how a package has changed over time in good and bad ways.

### 📚 Description

This is basically a timeline which shows the published versions of a package.

Alongside each version, it shows significant changes.

Positive changes flagged:

- ESM
- Dependency count decrease
- Install size decrease
- Trusted publishing enabled
- Provenance enabled

Negatives flagged:

- CJS
- Dependency count increase
- Install size increase
- Trusted publishing disabled
- Provenance disabled

Neutrals flagged:

- License changes


MatteoGabriele:
Super cool feature! I took a quick look at it, and there are a couple of UI/UX aspects that aren't very clear

#### What happens when you click on each version?
I know the "diff" link appears and lets you check differences between versions, but it’s not obvious that clicking the version itself opens that feature. Maybe we can improve discoverability and the link's purpose?

#### Re-rendering
Every time I click a version, it does all the requests again (cached),  but it seems to be causing a lot of repainting

43081j:
> Every time I click a version, it does all the requests again (cached), but it seems to be causing a lot of repainting

That's because its a new page load really. isn't this how the code/docs/etc tabs work too?

if not, whatever "make it work" config those have, we need the same here. any help with that would be nice

server/api/registry/timeline/[...pkg].get.ts
```diff
@@ -0,0 +1,103 @@
...
+    if (!pkgParam) {
+      throw createError({ statusCode: 404, message: 'Package name is required' })
+    }
+
+    let packageName: string
+    try {
+      packageName = decodeURIComponent(pkgParam)
+    } catch {
+      throw createError({ statusCode: 400, message: 'Invalid package name encoding' })
+    }
```
Comment (ghostdevv) [inline server/api/registry/timeline/[...pkg].get.ts:43]: we don't try/catch this anywhere else from what I can tell, should we be?

server/api/registry/timeline/[...pkg].get.ts
```diff
@@ -0,0 +1,103 @@
...
+            time: packument.time[v]!,
+            license: typeof license === 'string' ? license : undefined,
+            type: typeof version.type === 'string' ? version.type : undefined,
+            hasTypes: hasBuiltInTypes(version) || undefined,
+            hasTrustedPublisher: version._npmUser?.trustedPublisher ? true : undefined,
+            hasProvenance: version.dist?.attestations ? true : undefined,
+            tags: tagsByVersion.get(v) ?? [],
+          }
+        })
+        .sort((a, b) => new Date(b.time).getTime() - new Date(a.time).getTime())
```
Comment (ghostdevv) [inline server/api/registry/timeline/[...pkg].get.ts]:
```suggestion
        .sort((a, b) => Date.parse(b.time) - Date.parse(a.time))
```

server/api/registry/timeline/[...pkg].get.ts
```diff
@@ -0,0 +1,103 @@
...
+          if (license && typeof license === 'object' && 'type' in license) {
+            license = (license as { type: string }).type
+          }
+
+          return {
+            version: v,
+            time: packument.time[v]!,
+            license: typeof license === 'string' ? license : undefined,
+            type: typeof version.type === 'string' ? version.type : undefined,
+            hasTypes: hasBuiltInTypes(version) || undefined,
```
Comment (ghostdevv) [inline server/api/registry/timeline/[...pkg].get.ts:74]:
```suggestion
            hasTypes: hasBuiltInTypes(version),
```

could we just do that?

ghostdevv (changes requested):

MatteoGabriele:
@43081j, I didn't realize it was actually linking me to the package version.
What if the version link took users directly to the package page? I worry that simply linking to the specific version while staying in the timeline may not clearly show what happened.

MatteoGabriele:
I was also wondering if it might be a good idea to merge this page's functionalities with the existing Version history feature. They seem similar enough to be combined into a single page. 🤔

howwohmm (commented):
this is excellent work — the semver-ordered predecessor comparison is a particularly smart choice to avoid false diffs from interleaved backport releases. a few things from reading the full diff:

1. **N+1 install-size fetches with no concurrency limit.** the watcher fires `fetchSize()` for all 25 versions simultaneously on page load. consider a concurrency limiter (e.g. `p-limit(5)`) to avoid overwhelming the browser's connection pool on slow networks.

2. **`sizesLoading` gates `loadMore()` too aggressively.** if any size fetch is in-flight, the user can't paginate. since sizes are best-effort annotations, decouple them — guard `loadMore` only on `loadingMore` itself, not on `sizesLoading`.

3. **page-boundary sub-event pop-in.** `versionSubEvents` computes predecessors only from loaded entries. the last version on each page initially shows no sub-events, then retroactively gains them when the next page loads (bringing its semver predecessor into scope). one fix: fetch `offset - 1` from the server to ensure the boundary predecessor is always available.

4. **`useAsyncData` key collision.** the key `` `timeline:${packageName.value}` `` is evaluated once at setup. if the component instance handles navigation between packages (possible via address bar), the cache key won't distinguish them. using a getter function (Nuxt 3.8+) or including a version hash would be safer.

the `cloneDeep` on sub-event comparison data and the paginated server endpoint design are both well done.

43081j:
> What if the version link took users directly to the package page? I worry that simply linking to the specific version while staying in the timeline may not clearly show what happened

agreed. it was this originally, i have changed it back.

> I was also wondering if it might be a good idea to merge this page's functionalities with the existing Version history feature. They seem similar enough to be combined into a single page.

I'm also not sure. they have different purposes. it just so happens the only timeline entries right now are version specific, but there's nothing stopping us adding timeline events which are not associated with a version.

server/api/registry/timeline/[...pkg].get.ts
Comment (43081j) [reply server/api/registry/timeline/[...pkg].get.ts:74]: these `undefined` fallbacks are so the resulting JSON doesn't contain the keys unnecessarily. its basically an optimisation so we don't send a bunch of defaulted keys over the wire

ghostdevv (approved):

app/pages/package-timeline/[[org]]/[packageName].vue
```diff
@@ -0,0 +1,417 @@
...
+  try {
+    const data = await fetchTimeline(timelineEntries.value.length)
+    timelineEntries.value = [...timelineEntries.value, ...data.versions]
+    totalVersions.value = data.total
+  } catch {
+    loadError.value = true
+  } finally {
+    loadingMore.value = false
+  }
+}
```
Comment (MatteoGabriele) [inline app/pages/package-timeline/[[org]]/[packageName].vue:99]: We never stop fetching the timeline, so it keeps repeating itself as long as I keep clicking the button.

https://github.com/user-attachments/assets/f647ccbb-b963-4302-9633-4667b59ea085

Comment (serhalp) [reply app/pages/package-timeline/[[org]]/[packageName].vue:99]: 👀 Is this still an issue?

app/pages/package-timeline/[[org]]/[packageName].vue
```diff
@@ -0,0 +1,417 @@
...
+      }
+    }
+
+    // License changes
+    const currentLicense = current.license ?? 'Unknown'
+    const previousLicense = previous.license ?? 'Unknown'
+    if (currentLicense !== previousLicense) {
+      events.push({
+        key: 'license',
+        positive: false,
```
Comment (alexdln) [inline app/pages/package-timeline/[[org]]/[packageName].vue]: It might be positive. For example when it changed from no license to MIT

app/pages/package-timeline/[[org]]/[packageName].vue
```diff
@@ -0,0 +1,417 @@
...
+            v-if="versionSubEvents.has(entry.version)"
+            class="relative border-s border-border/50 ms-3 mt-2"
+          >
+            <li
+              v-for="ev in versionSubEvents.get(entry.version)"
+              :key="ev.key"
+              class="mb-2 ms-4 relative last:mb-0"
+            >
+              <span
+                class="absolute -start-[calc(1rem+0.375rem)] top-0.5 flex items-center justify-center w-3 h-3 rounded-full border"
```
Comment (alexdln) [inline app/pages/package-timeline/[[org]]/[packageName].vue]:
```suggestion
                class="absolute -start-[1.375rem] top-0.5 flex items-center justify-center w-3 h-3 rounded-full border"
```

server/api/registry/timeline/[...pkg].get.ts
```diff
@@ -0,0 +1,103 @@
...
+            hasTrustedPublisher: version._npmUser?.trustedPublisher ? true : undefined,
+            hasProvenance: version.dist?.attestations ? true : undefined,
+            tags: tagsByVersion.get(v) ?? [],
+          }
+        })
+        .sort((a, b) => Date.parse(b.time) - Date.parse(a.time))
+
+      return {
+        versions: allVersions.slice(offset, offset + limit),
```
Comment (alexdln) [inline server/api/registry/timeline/[...pkg].get.ts:83]: You are processing data on each request. Might be better loading and sorting in a separate cached method and here just call this [cached] function and slice the result of it (maybe even no need to cache this route)

app/pages/package-timeline/[[org]]/[packageName].vue
Comment (43081j) [reply app/pages/package-timeline/[[org]]/[packageName].vue:99]: not anymore 👍

alexdln:
@43081j Is it possible to add a loader for the sizes? It took about 30 seconds for me (nuxt), and I thought there was nothing special in the package (and would have just gone to another page to look further keeping sure that nothing changed in terms of size)

43081j:
im going to try refactor some things and will come back to that once i have 👍

43081j:
ok calling it a night for now but basically i've reworked the size calculation to happen in a paged way just like the timeline itself does.

so for each page of timeline (25 items), we send do 2 requests:

- timeline itself (which is SSR'd as far as i understand, nuxt stuff)
- size calculations (always lazily loaded)

load more then triggers these two requests again.

in cases where we fail to compute the size, or for whatever reason we can't figure the dependency tree out, we skip those versions in the size response. so notices dont show with "changed by 100%!".

also adds a css loading bar at the top that shows when these requests are in flight

server/api/registry/timeline/[...pkg].get.ts
Comment (43081j) [reply server/api/registry/timeline/[...pkg].get.ts:83]: you are right but im going to do this in a follow up to avoid this PR dragging on, so i can split it up more in future with incremental PRs

gameroman:
Why does the timeline tab require the `/v/:version` part?

app/pages/package-timeline/[[org]]/[packageName].vue
```diff
@@ -0,0 +1,444 @@
...
+      const sizeRatio =
+        previousSize.totalSize > 0
+          ? (currentSize.totalSize - previousSize.totalSize) / previousSize.totalSize
+          : 0
+      const depDiff = currentSize.dependencyCount - previousSize.dependencyCount
+
+      const sizeIncreased = sizeRatio > SIZE_INCREASE_THRESHOLD
+      const sizeDecreased = sizeRatio < -SIZE_INCREASE_THRESHOLD
+      const depsIncreased = depDiff > DEP_INCREASE_THRESHOLD
+      const depsDecreased = depDiff < -DEP_INCREASE_THRESHOLD
```
Comment (gameroman) [inline app/pages/package-timeline/[[org]]/[packageName].vue:186]:
```suggestion
      const sizeIncreased = sizeRatio >= SIZE_INCREASE_THRESHOLD
      const sizeDecreased = sizeRatio <= -SIZE_INCREASE_THRESHOLD
      const depsIncreased = depDiff >= DEP_INCREASE_THRESHOLD
      const depsDecreased = depDiff <= -DEP_INCREASE_THRESHOLD
```

gameroman:
Another feature I think would be nice to have is to show on timeline if that specific version is deprecated

gameroman:
Also if size increase/decrease is `>= 3x`, show size increased/decreased `3x` instead of `by 200%`

gameroman:
This feature looks very nice overall

app/pages/package-timeline/[[org]]/[packageName].vue
Comment (43081j) [reply app/pages/package-timeline/[[org]]/[packageName].vue:186]: these were on purpose as exclusive thresholds, i.e. so they have to be _exceeded_ to match.

if there's consensus that they should be inclusive, i can change them though.

Comment (gameroman) [reply app/pages/package-timeline/[[org]]/[packageName].vue:186]: I'm fine with either, I just think that threshold for reduced `depsDecreased` could be smaller,
always nice to celebrate dependency reductions :)

Could it be relative to total dependencies or something?

Comment (43081j) [reply app/pages/package-timeline/[[org]]/[packageName].vue:186]: the problem then becomes more difficult since large packages need to drop a huge amount to meet it.

maybe lets go with this for now and figure it out in a follow up so we don't drag it out. i'd like to iterate on this

serhalp:
Maybe the timeline "dots" could be better leveraged visually? I find it hard to parse the timeline at a glance when there are multiple majors and pre-releases interleaved.

Here's an awful sketch of what I'd suggest:

Screenshot 2026-04-10 at 11 13 11

Basically: make the dot for every release under the `latest` tag the same colour as the current latest, but lighter; repeat for any other tag, like `beta`; make the tag badges pop more; and maybe call out the first release of a major somehow.

43081j:
makes sense.

i'll take a look

43081j:
im going to leave it as it is for now and tackle that in another PR.

so this is ready to go i think!

alexdln (approved):

alexdln:
_I think we'll also need og image for this page_

gameroman (approved):
LGTM, can iterate later

ghostdevv (approved):
works on my machine

PR #2261, opened 2026-03-24
chore: improve externals configuration in chromatic

JReinhold:
### 🧭 Context

To cut down on Chromatic snapshot usage, we're trying to be more aggressive about not cache busting as often. Currently changes to `nuxt.config.ts` and `app/assets/main.css` accounts for over 5000 snapshots a week, which are most likely not necessary.

If we see that the cache is now busted to rarely, we can continue iterating on this.

### DIFF
--- a/chromatic.config.json
+++ b/chromatic.config.json
@@ -3,5 +3,5 @@
   "onlyChanged": true,
   "autoAcceptChanges": "main",
   "exitZeroOnChanges": false,
-  "externals": ["public/**", "app/assets/**", ".storybook/**", "nuxt.config.ts", "uno.config.ts"]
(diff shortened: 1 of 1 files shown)


ghostdevv:
(sorry, was just seeing why that welcome action failed)

ghostdevv:
> Thanks @JReinhold !
> 
> I had to move the theme out of `uno.config.ts` in #2252. So depending on if this merges before or after I will probably add `uno.theme.ts` to this array soon. I think we would want to bust cache on changes to either file but I imagine that neither file would really change all that often though.

I just merged your PR with main and added the change there!

PR #2266, opened 2026-03-24
fix(a11y): use native checkboxes for compare facet selector

PhyPhy07:
### 🔗 Linked issue

No Issue Number , This PR was constructed to address a TODO comment in FacetSelector.vue

### 🧭 Context

On /compare, users choose which metrics appear in the table using the facet strip at the top. That strip was implemented as buttons with aria-pressed, which behaves like a custom toggle rather than a standard control. Screen readers and keyboard users generally get clearer, more predictable behavior from real checkboxes (with an associated label). The component already called this out in TODOs. This change implements those controls as native checkboxes while leaving selection logic, URL syncing, and rules unchanged (e.g. at least one facet must stay on, and “coming soon” options stay off).

### High Level Summary of Changes

Replaced facet row ButtonBase chips with a <label> + checkbox per metric, including disabled state when a facet is “coming soon” or when it’s the only selected facet (so the “≥ one metric” rule still holds).

Preserved existing category All / None actions and styling intent (chip-like borders/backgrounds, focus styles).
Updated FacetSelector.spec.ts to assert on checkboxes (data-facet-id, checked/disabled, change → toggleFacet) and removed an unused test helper.
Added aria-label on each category’s facet role="group" so the subgroup is named in the accessibility tree.

### 📚 Description

The facet strip on the compare page was using ButtonBase with aria-pressed to let users toggle metrics on and off. The visual design implied a checkbox group, but the semantics didn't back that up — screen readers and other assistive tech treat push-buttons differently from form controls, so the experience wasn't matching the mental model of "pick which rows to show."

- Swapped those out for a real <label> + <input type="checkbox"> per metric. The underlying behavior is all the same: 
(ex. useFacetSelection is untouched, selections still sync to the facets query param, etc...)

Tests are updated to reflect the new markup — data-facet-id, checked/disabled state, change handling — so any regressions in selection behavior will surface.

Better semantics, same behavior.

### Screenshot: Before/After UI Change:
Current Comparison UI:
Current UI
After Selection Boxes Implemented:
SelectionBoxes

### DIFF
--- a/app/components/Compare/FacetSelector.vue
+++ b/app/components/Compare/FacetSelector.vue
@@ -23,6 +23,20 @@ function onFacetChange(facet: FacetInfoWithLabels) {
   toggleFacet(facet.id)
 }
 
+/** Visual variant for each facet chip (border/background/cursor). */
+function facetChipVariantClass(facet: FacetInfoWithLabels): string {
+  if (facet.comingSoon) {
+    return 'text-fg-subtle/50 bg-bg-subtle border-border-subtle cursor-not-allowed'
+  }
+  if (isFacetCheckboxDisabled(facet)) {
+    return 'text-fg-muted bg-bg-muted border-border opacity-90 cursor-not-allowed'
+  }
+  if (isFacetSelected(facet.id)) {
+    return 'text-fg-muted bg-bg-muted border-border cursor-pointer'
+  }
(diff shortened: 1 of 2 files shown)


knowler:
Hello @PhyPhy07, thank you for contributing. Unfortunately, we already have a PR in progress (#1536) for this one and so I will be closing this one.

PR #2273, opened 2026-03-25
feat:  compare download charts with sparklines

### LINKED ISSUE #2270: line charts (multiple series) - propose a 'split' view option to display series in individual sparklines

graphieros:
Resolves #2270 

This implements an additional view for the download charts of the compare page, to address a11y issues related to color vision deficiency, and display each series on a separate sparkline chart. @Adebesin-Cell also co-authored this PR with:
- Persisting compare view (table or bars) and chart view (combined line or sparklines) in URL
- tab component

A toggle allows to switch from the 'Combined view' to the 'Split view'.

| Dark mode | Light mode |
|------------|------------|
|dark mode|light mode|

Hovering a sparkline chart syncs the selected index on all charts to ease comparison:

image

Long package names are ellipsed to avoid any Y offset that a wrap would cause.

Other:
- colours for sparkline use either predefined package colours or default to vue-data-ui's internal palette. If the number of compared series turns out to be higher than the number of palette colours, the assigned colours will loop the palette (same as the combined line chart does internally).
- bump vue-data-ui to 3.16.5 (dashed segments on sparklines)
- show estimation segments on the sparklines of the compare page:

image

Notes:
 - there is no printing feature yet for the split view, this could be part of a future iteration.

### DIFF
--- a/app/components/Tab/Panel.vue
+++ b/app/components/Tab/Panel.vue
@@ -11,7 +11,7 @@ const getTabId = inject<(value: string) => string>('tabs-tab-id')
 const getPanelId = inject<(value: string) => string>('tabs-panel-id')
 
 if (!selected || !getTabId || !getPanelId) {
-  throw new Error('TabItem must be used inside a TabRoot component')
+  throw new Error('TabPanel must be used inside a TabRoot component')
 }
 
 const isSelected = computed(() => selected.value === props.value)


app/components/Chart/SplitSparkline.vue
```diff
@@ -0,0 +1,234 @@
...
+
+import('vue-data-ui/style.css')
+
+const props = defineProps<{
+  dataset?: Array<
+    VueUiXyDatasetItem & {
+      color?: string
+      series: number[]
+      dashIndices?: number[]
+    }
```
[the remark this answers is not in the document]
graphieros: This will be addressed in a future iteration, and requires an update in vue-data-ui for VueUiSparkline to support dashed indices

app/components/Package/TrendsChart.vue
```diff
@@ -1630,6 +1633,51 @@ watch(selectedMetric, value => {
...
+      role="tablist"
+      :aria-label="$t('package.trends.chart_view_toggle')"
+    >
+      <button
+        id="combined-chart-layout-tab"
+        type="button"
+        role="tab"
+        :aria-selected="isSparklineLayout ? 'false' : 'true'"
+        aria-controls="combined-chart-layout-panel"
+        :tabindex="isSparklineLayout ? 0 : -1"
```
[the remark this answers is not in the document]
graphieros: This is the desired outcome.

app/components/Package/TrendsChart.vue
```diff
@@ -1630,6 +1633,51 @@ watch(selectedMetric, value => {
     id="trends-chart"
     :aria-busy="activeMetricState.pending ? 'true' : 'false'"
   >
+    <div
+      v-if="isMultiPackageMode"
+      class="inline-flex items-center gap-1 rounded-md border border-border-subtle bg-bg-subtle p-0.5 mt-4 mb-8"
+      role="tablist"
+      :aria-label="$t('package.trends.chart_view_toggle')"
+    >
+      <button
```
Adebesin-Cell: Just a little nitpick here, can we use `ButtonBase` instead of plain button here to keep buttons consistent across the app?

app/components/Package/TrendsChart.vue
```diff
@@ -1630,6 +1633,51 @@ watch(selectedMetric, value => {
...
+            ? 'border-transparent text-fg-subtle hover:text-fg'
+            : 'bg-bg border-border shadow-sm text-fg'
+        "
+        @click="isSparklineLayout = false"
+      >
+        <span class="i-lucide:chart-line size-[1em]" aria-hidden="true" />
+        <span>{{ $t('package.trends.chart_view_combined') }}</span>
+      </button>
+
+      <button
```
Adebesin-Cell: Same here.

graphieros: ButtonBase does not exactly fit the bill out of the box here. 
We might a proper tab components to replace these blocks, but there are only 2 for now.
Feel free to be my co-author on this one :)
(for this and | or the URL thing)

RYGRIT:
I found two a11y-related bugs:

1. The Reset button cannot be selected using the Tab key.
image

2. I'm not sure how to best describe this issue. For instance, initially in React, my currently selected date is 2026-02-11. When I press Tab to switch to Solid.js, it correctly displays 2026-02-11—so far, so good. However, when I press Shift-Tab to switch back to React, it displays the data corresponding to 2026-03-25 instead.

https://github.com/user-attachments/assets/be9756ad-7a3a-40b8-8a5d-616cdf49af66

graphieros:
> 1. The Reset button cannot be selected using the Tab key.

@RYGRIT  The reset button is focusable. The problem is that the css var for the accent color does nothing when the app loads the default accent color. This requires a separate issue, and is not related specifically to this page.

image

> 2. I'm not sure how to best describe this issue...

Well spotted. This is an issue with the library which does not keep the state from the shared index when using synced instances (I'll try to update it ASAP, but it should not be blocking, since navigation inside one syncs all)

RYGRIT:
> The reset button is focusable. The problem is that the css var for the accent color does nothing when the app loads the default accent color. This requires a separate issue, and is not related specifically to this page.

Yes, as you mentioned, I discovered that the issue regarding the accent color actually lies not with the "Reset" button, but rather with the "Open Options" button (I tested this again in an Incognito window). I believe this was simply a misjudgment on my part, stemming from the focus order: after the focus lands on the chart, pressing the Tab key shifts the focus to the "Reset" button; I had previously—and mistakenly—assumed that the "Reset" button was the next element in the focus sequence following the control buttons located to the right of the thumbnail.

graphieros:
@RYGRIT 
vue-data-ui was updated to fix the problem you mentioned for the state loss of the selected index when tabing between chart instances

Quick link to test: https://npmx-8zj6uev7m-npmx.vercel.app/compare?packages=vue,svelte,react,astro

RYGRIT:
It seems the problem persists while I am working on the first chart.

https://github.com/user-attachments/assets/ffc56c9e-1a5b-48f0-8ba6-6feaa760f0e3

graphieros:
> It seems the problem persists while I am working on the first chart.

I don't see the problem, what is it exactly ? If you tab away from the component, the state will be reset. That's normal, like with the mouse, mouseout resets the state, for the last value to always be shown when not interacting with the charts. But when tabing inside to cycle through chart instances, the state is maintained:

https://github.com/user-attachments/assets/8d253951-38bb-4b19-982a-4822af73b7d4

RYGRIT:
I'm not sure why the behavior I'm seeing here is different from yours.🤦‍♂️When I press Shift + Tab to return to the previous chart, the state resets.


https://github.com/user-attachments/assets/2733ed95-e09b-436a-b421-b0bb4ad1a3ba

graphieros:
> I'm not sure why the behavior I'm seeing here is different from yours.🤦‍♂️When I press Shift + Tab to return to the previous chart, the state resets.

The difference is that you did not cycle through datapoints after landing on the second instance.
I'll try to find why, but that's kind of a niche edge case

RYGRIT:
> I'll try to find why, but that's kind of a niche edge case

I think I've figured it out: the current behavior is that after performing operations on an active chart, I can navigate forward through a full cycle; however, attempting to navigate backward causes the state to reset. If, on the other hand, I perform a left or right shift operation *before* attempting to navigate backward, the backward navigation then preserves the current state.

graphieros:
> I think I've figured it out: the current behavior is that after performing operations on an active chart, I can navigate forward through a full cycle; however, attempting to navigate backward causes the state to reset. If, on the other hand, I perform a left or right shift operation _before_ attempting to navigate backward, the backward navigation then preserves the current state.

Awesome, thanks to your diagnosis I found a fix in the library, will update soon

graphieros:
@RYGRIT 

I think we are good ?
https://npmx-pz79qtyck-npmx.vercel.app/compare?packages=vue,svelte,react,astro

RYGRIT:
> If you tab away from the component, the state will be reset.

If we need to meet this requirement, it currently appears that we are unable to do so.

graphieros:
> > If you tab away from the component, the state will be reset.
> 
> If we need to meet this requirement, it currently appears that we are unable to do so.

Nah, now that's a feature for kbd users ;)
Can't really reset on blur, because we would lose state when tabing through instances. I think it's fine.

But I can add a keydown.esc to reset maybe ?

RYGRIT:
Perfect explanation, that design trade-off definitely makes sense. Great work on the fixes. 👍

> But I can add a keydown.esc to reset maybe ?

That's a good idea

graphieros:
> Perfect explanation, that design trade-off definitely makes sense. Great work on the fixes. 👍

Ok, I added a keydown.esc that resets (won't work if the mouse is inside, which is normal).

Adebesin-Cell:
Hey @graphieros! Taking you up on the co-author offer 😄

I've pushed a commit that addresses the tab buttons and URL persistence:

**Branch:** [`Adebesin-Cell/npmx.dev@2270-line-charts-multiple-series---propose-a-split-view-option-to-display-series-in-individual-sparklines`](https://github.com/Adebesin-Cell/npmx.dev/tree/2270-line-charts-multiple-series---propose-a-split-view-option-to-display-series-in-individual-sparklines)

You can cherry-pick the commit or I can open a PR into your branch, whatever works best for you!

graphieros:
> You can cherry-pick the commit or I can open a PR into your branch, whatever works best for you!

Awesome:)
I think it would be best that you open a PR into this branch.

🙏 can you please add some tests for the tab component ?

Adebesin-Cell:
> > You can cherry-pick the commit or I can open a PR into your branch, whatever works best for you!
> 
> Awesome:) I think it would be best that you open a PR into this branch.
> 
> 🙏 can you please add some tests for the tab component ?

You can review here: https://github.com/npmx-dev/npmx.dev/pull/2282

graphieros:
@Adebesin-Cell  🙏 it seems we have failing checks post merge, can you please take a look ?

Adebesin-Cell:
> @Adebesin-Cell 🙏 it seems we have failing checks post merge, can you please take a look ?

For sure!

Adebesin-Cell:
All previous errors are resolved now 🚀

graphieros:
> All previous errors are resolved now 🚀

These are great improvements, thanks :)

app/components/Chart/SplitSparkline.vue
```diff
@@ -0,0 +1,261 @@
...
+          bottom: 0,
+        },
+      },
+    }
+  })
+})
+</script>
+
+<template>
+  <div>
```
MatteoGabriele: Can we remove this extra div?

app/components/Chart/SplitSparkline.vue
```diff
@@ -0,0 +1,261 @@
...
+})
+
+watch(
+  () => colorMode.value,
+  value => {
+    resolvedMode.value = value === 'dark' ? 'dark' : 'light'
+  },
+  { flush: 'sync' },
+)
+
```
MatteoGabriele, line 49: Will this work as well? Not sure if you actually need the component to be mounted, or you just need to trigger it at the beginning

```suggestion
onMounted(() => {
  rootEl.value = document.documentElement
})

watch(
  () => colorMode.value,
  value => {
    resolvedMode.value = value === 'dark' ? 'dark' : 'light'
  },
  { flush: 'sync', immediate: true },
)

```

app/components/Tab/Item.vue
```diff
@@ -0,0 +1,76 @@
...
+const getTabId = inject<(value: string) => string>('tabs-tab-id')!
+const getPanelId = inject<(value: string) => string>('tabs-panel-id')!
+
+const isSelected = computed(() => selected.value === props.value)
+const resolvedTabId = computed(() => props.tabId ?? getTabId(props.value))
+const resolvedPanelId = computed(() => getPanelId(props.value))
+
+function select() {
+  selected.value = props.value
+}
```
MatteoGabriele: This is a minor detail, but I'm not a big fan of non-null assertions. Since this component can only exist within a TabRoot context, we could throw an error if no injected properties are present.

We still need to make the select function an arrow function, or it would be hoisted, and TypeScript wouldn't understand that we're ok with it. In this case, you can either use an arrow function or keep the same function but add `selected!.value = props.value`. This would be fine here because we're sure it failed before otherwise.

```suggestion
const selected = inject<WritableComputedRef<string>>('tabs-selected')
const getTabId = inject<(value: string) => string>('tabs-tab-id')
const getPanelId = inject<(value: string) => string>('tabs-panel-id')

if (!selected || !getTabId || !getPanelId) {
  throw new Error('TabItem must be used inside a TabRoot component')
}

const isSelected = computed(() => selected.value === props.value)
const resolvedTabId = computed(() => props.tabId ?? getTabId(props.value))
const resolvedPanelId = computed(() => getPanelId(props.value))

const select = () =>{
  selected.value = props.value
}
```

app/components/Tab/Panel.vue
```diff
@@ -0,0 +1,29 @@
...
+defineOptions({ name: 'TabPanel' })
+
+const props = defineProps<{
+  value: string
+  panelId?: string
+}>()
+
+const selected = inject<WritableComputedRef<string>>('tabs-selected')!
+const getTabId = inject<(value: string) => string>('tabs-tab-id')!
+const getPanelId = inject<(value: string) => string>('tabs-panel-id')!
```
MatteoGabriele: potentially the same thing here. (see previous comment)

app/components/Tab/Root.vue
```diff
@@ -0,0 +1,46 @@
...
+
+const internalValue = shallowRef(props.defaultValue ?? '')
+
+const selectedValue = computed({
+  get: () => props.modelValue ?? internalValue.value,
+  set: (v: string) => {
+    internalValue.value = v
+    emit('update:modelValue', v)
+  },
+})
```
MatteoGabriele: You can simplify this by using the defineModel macro. 
I don't think you need a defaultValue, since the modelValue can have a default value.

```suggestion
const props = withDefaults(
  defineProps<{
    idPrefix?: string
  }>(),
  {
    idPrefix: 'tab',
  },
)

const selectedValue = defineModel<string>()
```

PR #2274, opened 2026-03-25
fix: do not strip package name from description

gameroman:
### 🧭 Context

Removes stripping of package name from description as it has no real benefit and only brings confusion 

### 📚 Description

| Package | `del` | `sockjs` |
| ------- | ----- | -------- |
| Before | {4E38E7BC-36F7-4264-877C-F041D7D645A8} | {D3751241-DDEB-4D57-B6CB-8ABE9FED3ABA} |
| After | {CC03655E-C81F-4E05-AB90-3B894F29BB7C} | {02A30574-037E-4CDF-A305-60B48D0AB85E} |


gameroman:
Yes, that makes sense 👍

PR #2276, opened 2026-03-25
fix(ui): version selector older groups and versions history link

### LINKED ISSUE #2247: All versions of a package dropdown
### Describe the bug

Image

this button doesn't do anything

### Additional context

https://npmx.dev/package/npkill

tinsever:
### 🔗 Linked issue

Resolves #2247

### 🧭 Context

The package version dropdown could not meaningfully expand when the visible "group" was a single tagged release but older versions existed without tags: there was nothing to expand, yet users still needed a way to load and show those older rows. The "View all versions" link in the selector also pointed at the package overview instead of the dedicated version history page.

A follow-up issue also existed in the first-load expansion path: clicking a tagged row with nested versions could unintentionally reveal every hidden untagged group, even when that row was only supposed to expand its own nested versions.

### 📚 Description

**Version selector**  
- After loading the full version list, tagged groups can act as a toggle for "extra" groups (older releases without dist-tags): `showAllGroups` and `visibleVersionGroups` hide untagged groups until the user expands the controlling tagged row, and collapse them again on second toggle.  
- `toggleGroup` is refactored so loading runs once, nested multi-version groups still expand/collapse, and the expand control stays correct for loading, nested, and “show more groups” cases.  
- Fixed the first-load expansion behavior so `showAllGroups` is only enabled when the reloaded row actually controls additional hidden groups. Expanding a tagged row with nested versions now opens only that row instead of also revealing unrelated untagged groups.  
- Keyboard handling (listbox) uses `isGroupOpen` / `canToggleGroup` so Enter matches the new semantics; ArrowLeft can collapse the additional-groups state.  
- Expand/collapse buttons use `aria-expanded` / labels derived from `isGroupOpen`.  
- When `distTags`, `versions`, or `currentVersion` change, `showAllGroups` resets so stale UI does not stick around.

**Routing**  
- Added `packageVersionsRoute(packageName)` in `app/utils/router.ts` for the package-versions route.  
- `Versions.vue` uses that helper instead of duplicating org/name parsing.  
- The selector footer link now uses `packageVersionsRoute` so “view all versions” goes to version history, not the main package route.

**Tests**  
- New Vitest case in `test/nuxt/components/VersionSelector.spec.ts`: single tagged release + fetch returns an older version; expand shows `0.9`, collapse hides it again.  
- Added a regression case in `test/nuxt/components/VersionSelector.spec.ts` covering a tagged row with nested same-major versions plus an unrelated older group; expanding that tagged row no longer reveals the unrelated hidden group.


tinsever:
Sorry this took a few iterations, it worked on my machine! 🤣

PR #2290, opened 2026-03-26
chore: add i18n redirect

danielroe:
### 📚 Description

this adds a redirect to our new i18n dashboard.

it won't do anything until we release to main, so I won't move the i18n.npmx.dev subdomain until then

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -100,7 +100,7 @@
           "value": "i18n.npmx.dev"
         }
       ],
-      "destination": "https://npmx.dev/translation-status"
+      "destination": "https://main.npmx.dev/translation-status"
     }
   ]
 }


userquin:
Not sure if we should use rewrites instead redirects, check https://vercel.com/kb/guide/can-i-redirect-from-a-subdomain-to-a-subpath

Vercel redirects/rewrites don't mention this case.

vercel.json
```diff
@@ -91,6 +91,16 @@
...
+    },
+    {
+      "source": "/(.*)",
+      "has": [
+        {
+          "type": "host",
+          "value": "i18n.npmx.dev"
+        }
+      ],
+      "destination": "https://npmx.dev/translation-status"
```
cylewaitforit: question:

Since it seems like that translation-status page would only be updated as often as releases go out, would it be more beneficial for it to point to the version from the main branch since that would be update more frequently?

```suggestion
      "destination": "https://main.npmx.dev/translation-status"
```

PR #2296, opened 2026-03-27
fix: add spacing between tab toggle and comparison table

iiio2:
<table>
<tr>
 <td>Before</td>
 <td>After</td>
<tr>
 <td>before</td>
 <td>after</td>
</table>


iiio2:
Thanks.

Issue #1760, opened 2026-02-28
Use pointer cursor for interactive elements


alexdln:
## General
We have two main positions regarding the cursor:
* the historical context of the W3C (_use pointer cursor for links only_) and the system cursor on devices;
* the established user experience and the standard supported by most sites (_including Google and Wikipedia_).

Last time we gave the default cursor a chance, and at the moment it is written as the rule in our contributing guide.

## Standards

To begin with, what we consistently agree on is that we need a clear and understandable standard and unification of rules. Also it must comply with a11y standards.

Regarding the a11y standards, we were unable to find clear wording for all standards, and in fact, this is clearly stated only in the W3C ([spec](https://www.w3.org/TR/css-ui-4/#valdef-cursor-pointer)) in its original version*

In terms of usage, there is also no absolute standard - most sites prefer pointer, OS almost always use default, some libraries use default

\* W3C recently updated their standard and specified that pointer can be used for interactive elements ([Video](https://www.youtube.com/watch?v=BdNFBgnzbIY), [working group discussion](https://github.com/w3c/csswg-drafts/issues/1936#issuecomment-420722704)), but apparently as a result of wrong changes this line was removed ([commit](https://github.com/w3c/csswg-drafts/commit/1a57a4025c8778dab45ef8d5fbd482e4d6c4ebce#r177842392))

## Internal talks

Even though we want to follow the standard (and given that the working group has decided that using pointer is acceptable) - we also take the established user experience as a key focus. And it almost unequivocally indicates that we're more accustomed to buttons on a website with a pointer.

This is supported by our personal experience as users and as the target group of our website. In terms of accessibility, we must also take into account existing practices, as changing them could make daily experience worse for some users.

## Result

Based on all this, **we're returning the pointer cursor** for **all** interactive elements (_without any distinction or definition of what a link is and what a button is_). This needs to be updated in the contributing guide and updated throughout our site.

graphieros:
Note on how to enable cursor pointer on chart components.

### Add `useCursorPointer` attr in charts configurations

```ts
const chartConfig = computed<VueUiXyConfig>(() => {
  return {
    //...
    chart: {
      //...
      userOptions: {
        //...
        useCursorPointer: true
      }
    }
  }
})
```
List of components on which to apply the change in the  chart config's `userOptions` object:
- TrendsChart.vue
- TimelineChart.vue
- VersionDistribution.vue
- FacetBarChart.vue
- FacetScatterChart.vue

[...]

PR #2303, opened 2026-03-28
style: use pointer cursor for download button

iiio2:
Here, 

pointer

### DIFF
--- a/app/components/Package/DownloadButton.vue
+++ b/app/components/Package/DownloadButton.vue
@@ -55,7 +55,7 @@ async function downloadPackage() {
       type="button"
       @click="downloadPackage"
       :disabled="loading"
-      class="border-border-subtle bg-bg-subtle! text-xs text-fg-muted hover:enabled:(text-fg border-border-hover)"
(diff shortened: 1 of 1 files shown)


serhalp:
🤔 Have you seen our guidelines for this? https://github.com/npmx-dev/npmx.dev/blob/7688cd77ae88cdb33bd97367a0afc1f875e8a768/CONTRIBUTING.md?plain=1#L423-L428

@knowler what's your stance on a "Download" button? In this case it isn't strictly a link to a file (you can't open it in a new tab, etc.), so I'm thinking it's more of an "action" and no pointer is desirable?

knowler:
@serhalp Based on our guidelines, that logic makes sense to me. I think a pointer could also give people the false impression that they could drag the download link to a bookmarks bar, however, that doesn’t work.

We actually have an open issue that is seeking make cursors the default everywhere (#1760), so maybe we might be revisiting this in the near future.

ghostdevv:
I think what I'm taking from the discussion is that we should close this for now, and revisit it later based on the result from the `cursor-pointer` issue. Please reopen if I'm wrong!

Issue #2150, opened 2026-03-19
Pages missing Storybook stories


cylewaitforit:
This is a parent issue to track adding stories files for pages that are missing them.

- [x] about #2149 
- [x] accessibility #2301
- [x] brand #2468  
- [x] blog #2679
- [ ] compare
- [ ] diff
- [x] error #2601 
- [ ] index
- [x] leaderboard #2789 
- [ ] org
- [ ] package
- [ ] package-changelog 
- [ ] package-code
- [ ] package-docs
- [ ] package-timeline
- [x] pds #2514 
- [x] privacy #2319 
- [ ] profile
- [x] recharging #2496 
- [ ] search
- [x] settings #2545 
- [x] translation-status #2559 
- [ ] username
- [ ] username-orgs

PR #2319, opened 2026-03-29
docs(ui): add stories for Privacy page

cylewaitforit:
### 🔗 Linked issue

#2150 

### 🧭 Context

This would also enable a storybook mock page, storybook a11y checks, and chromatic visual regression tests for this page as documented by the storybook stories.

### 📚 Description

Adds stories for Privacy page.

Creates shared `pagesDecorator` to reduce duplication in stories.

https://698b88d5157d89f1f33a6c21-hffnzxcven.chromatic.com/?path=/story/pages-privacy--default


package.json
```diff
@@ -9,6 +9,9 @@
     "url": "https://roe.dev"
   },
   "type": "module",
+  "imports": {
+    "#storybook/*": "./.storybook/*"
+  },
```
ghostdevv: I'm not sure if the `#` has some special convention in nuxt, as some aliases are currently defined in the vite config and typed in the nuxt config - but I _think_ there are other aliases that are provided by nuxt? All that to say, we should probably use the vite config for this alias (if possible) for consistency - and do you know if using `#` is ok here, or if there is a more idiomatic prefix to use?

cylewaitforit: `#` is the [node convention for subpath imports](https://nodejs.org/api/packages.html#subpath-imports:~:text=Entries%20in%20the%20%22imports%22%20field%20must%20always%20start%20with%20%23%20to%20ensure%20they%20are%20disambiguated%20from%20external%20package%20specifiers).  It is [supported by storybook](https://storybook.js.org/docs/writing-stories/mocking-data-and-modules/mocking-modules#configuration) which is why I used it here.  I didn't know nuxt had used  `#` but it might be fighting against the current then because most other tooling is [working to support node subpath imports better.](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#subpath-imports-starting-with-)

cylewaitforit: If preferred in this PR, I could update the `import { pageDecorator } from '#storybook/decorators'` lines to just use the full path resolutions for now and avoid the subpath simplification.

cylewaitforit: Thinking more on this and felt that the subpath import changes felt like a stretch to be in the scope of these changes. So I've removed them to simplify this change set and will consider introducing it it it's own PR where the pros and cons can be weighed on their own.

PR #2331, opened 2026-03-30
fix: relax devtools-specific security headers in dev

RYGRIT:
### 🧭 Context

Nuxt DevTools relies on a same-origin iframe and dev-only runtime connections, which were being blocked by our CSP and frame restrictions during development.

Instead of disabling `security-headers` entirely in dev, this change keeps the existing security headers in place and only relaxes the DevTools-specific parts needed for the local dev runtime.

### 📚 Description

This PR updates `security-headers` so that:

- DevTools-only CSP requirements are allowed in dev
- the DevTools route uses `SAMEORIGIN` instead of `DENY`
- DevTools-specific relaxations are not applied when DevTools is explicitly disabled

### DIFF
--- a/modules/security-headers.ts
+++ b/modules/security-headers.ts
@@ -24,10 +24,10 @@ export default defineNuxtModule({
     const devtools = nuxt.options.devtools
 
     const isDevtoolsRuntime =
-      nuxt.options.dev
-      && devtools !== false
-      && (devtools == null || typeof devtools !== 'object' || devtools.enabled !== false)
-      && !process.env.TEST
+      nuxt.options.dev &&
+      devtools !== false &&
+      (devtools == null || typeof devtools !== 'object' || devtools.enabled !== false) &&
+      !process.env.TEST
 
     // These assets are embedded directly on blog pages and should not affect image-proxy trust.
     const cspOnlyImgOrigins = ['https://api.star-history.com', 'https://cdn.bsky.app']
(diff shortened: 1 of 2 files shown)


ghostdevv (changes requested):
Could we instead add the necessary nuxt permissions in dev?

RYGRIT:
@ghostdevv I’m not fully sure what you mean by “nuxt permissions” here. If you mean a built-in Nuxt/DevTools setting, I couldn’t find one that seems to apply here.

From what I can tell, the issue is caused by our security headers rather than a Nuxt option: DevTools uses a same-origin iframe under `/__nuxt_devtools__/client/`, so instead of skipping `security-headers` entirely in dev, I think the better fix is to keep the headers and only relax the dev-only bits it needs, like allowing `'self'` in `frame-src`, using `SAMEORIGIN` for the DevTools route instead of DENY, and adding any required dev connect-src entries if needed.

ghostdevv:
> @ghostdevv I’m not fully sure what you mean by “nuxt permissions” here. If you mean a built-in Nuxt/DevTools setting, I couldn’t find one that seems to apply here.
> 
> From what I can tell, the issue is caused by our security headers rather than a Nuxt option: DevTools uses a same-origin iframe under `/__nuxt_devtools__/client/`, so instead of skipping `security-headers` entirely in dev, I think the better fix is to keep the headers and only relax the dev-only bits it needs, like allowing `'self'` in `frame-src`, using `SAMEORIGIN` for the DevTools route instead of DENY, and adding any required dev connect-src entries if needed.

What I meant was can we expand the permissions in dev to support the devtools iframe? I'm weary against relaxing the policy in dev as it may make it harder for us to catch issues with the security policy as we have one less testing ground before production - if that makes sense?

RYGRIT:
Yep, that makes sense. I kept the headers in dev and only relaxed the bits DevTools needs (`frame-src 'self'`, `SAMEORIGIN` for `/__nuxt_devtools__/`, and the dev-only websocket source), so we still exercise the CSP in dev instead of bypassing it entirely.

PR #2346, opened 2026-03-31
feat: add kawai logo in npmx.dev homepage

alfonsusac:
### 🧭 Context

April fools is coming up, something fun needs to be done. I present to you a kawaii variant of the npmx home page. 

### 📚 Description

- `?kawai=true` is detected in index.vue using window.location.search
- Hydration issue is solved using onPrerender()
- Kawai logo is rendered using `` stored in /public/extra
- On load, it will fade in. On hover, it will scale-105
- Non-persistent (unlike Bsky's implementation) and can be removed by removing query.

Discussion:
https://discord.com/channels/1464542801676206113/1486838820556116051/1488491068285522021

### DIFF
--- a/test/nuxt/a11y.spec.ts
+++ b/test/nuxt/a11y.spec.ts
@@ -141,6 +141,7 @@ import {
   BlueskyPostEmbed,
   BuildEnvironment,
   ButtonBase,
+  LandingLogo,
   LinkBase,
   CallToAction,
   ChartPatternSlot,
@@ -326,6 +327,14 @@ describe('component accessibility audits', () => {
       const component = await mountSuspended(AppHeader, {
         props: { showConnector: false },
       })
+      const results = await runAxe(component)
+      expect(results.violations).toEqual([])
+    })
+  })
+
+  describe('LandingLogo', () => {
+    it('should have no accessibility violations', async () => {
+      const component = await mountSuspended(LandingLogo)
       const results = await runAxe(component)
       expect(results.violations).toEqual([])
     })


app/pages/index.vue
```diff
@@ -24,6 +24,9 @@ defineOgImageComponent('Default', {
   title: 'npmx',
   description: 'a fast, modern browser for the **npm registry**',
 })
+
+const route = useRoute()
+const isKawaii = computed(() => route.query.kawai === 'true')
```
WilcoSp: Maybe an idea to only need to have `?kawai` instead of `?kawai=true`.
you can achieve this with 
```js
 'kawai' in route.query
```


further looking very good

alexdln: Shouldn't it be?

```suggestion
const isKawaii = computed(() => route.query.kawaii === 'true')
```

(kawai**i**, second "i")

app/pages/index.vue
```diff
@@ -36,8 +39,16 @@ defineOgImageComponent('Default', {
...
+          <img
+            v-if="isKawaii === true"
+            width="400"
+            class="pb-8 motion-safe:animate-fade-in motion-safe:hover:scale-105 motion-safe:transition"
+            src="/extra/npmx-cute.svg"
+            alt="npmx kawaii logo"
+          />
+          <AppLogo v-if="isKawaii !== true" class="w-42 h-auto sm:w-58 md:w-70" />
           <span
+            v-if="isKawaii !== true"
```
alexdln: Hydration error :(

Can you please check how it works for example in color-picker with onprehydrate and try to repeat?

https://github.com/npmx-dev/npmx.dev/blob/d86c9c448237c0dd71d780e668aa514825972ddb/app/components/Settings/AccentColorPicker.vue#L6

alexdln: Also technically, it's possible to solve the problem at the server level. Call api route as src instead (`img src="api/home-logo"`), check there referrer query and return either image A or image B depends on it

But I'm not sure it will work as fast as native approach

alfonsusac:
Ok I made some changes: 
- fixed the hydration issue (no longer showing hydration warning)
- improve entry animation (scale-in)
- improve responsiveness (make logo smaller in phone)

Questions:
1. coderabbit is suggesting me creating i18n token for the alt text, i can do that, but can i directly write the en.json entry in the same PR? can i also write the id-ID.json entry while im at it too? 👀

app/pages/index.vue
alfonsusac: Fixed!

app/pages/index.vue
```diff
@@ -24,6 +24,28 @@ defineOgImageComponent('Default', {
   title: 'npmx',
   description: 'a fast, modern browser for the **npm registry**',
 })
+
+// const route = useRoute()
+// const isKawaii = computed(() => route.query.kawaii === 'true')
+onPrehydrate(el => {
+  const isKawaii = window.location.search.includes('kawaii=true')
```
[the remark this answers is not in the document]
alfonsusac: Fixed in latest commit! Thanks!

app/pages/index.vue
```diff
@@ -25,8 +25,27 @@ defineOgImageComponent('Default', {
...
-const isKawaii = computed(() => route.query.kawaii === 'true')
+// const route = useRoute()
+// const isKawaii = computed(() => route.query.kawaii === 'true')
+onPrehydrate(el => {
+  const isKawaii = window.location.search.includes('kawaii=true') 
+  const normal = el.querySelector('#npmx-index-h1-logo-normal') as HTMLElement
+  const kawaii = el.querySelector('#npmx-index-h1-logo-kawaii') as HTMLElement
+  const env = el.querySelector('#npmx-index-h1-logo-env') as HTMLElement
+  const tagline = el.querySelector('#npmx-index-tagline') as HTMLElement
+  if (!normal || !kawaii || !env || !tagline) return
```
alexdln: _Strange but I can't add suggestions..._

```ts
  const isKawaii = new URLSearchParams(window.location.search).has('kawaii') // as mentioned in other comment
  if (!isKawaii) return; // dom operations are expensive and there is no need for them if we don't have param
  const normal = el.querySelector<HTMLElement>('#npmx-index-h1-logo-normal') // You have condition to check that every element is present so there shouldn't be types issues
  const kawaii = el.querySelector<HTMLElement>('#npmx-index-h1-logo-kawaii')
  const envLogo = el.querySelector<HTMLElement>('#npmx-index-h1-logo-env')
```

app/pages/index.vue
```diff
@@ -24,6 +24,29 @@ defineOgImageComponent('Default', {
...
+  if (isKawaii) {
+    normal.style.display = 'none'
+    kawaii.style.display = 'block'
+    env.style.display = 'none'
+    tagline.style.display = 'none'
+  } else {
+    normal.style.display = 'block'
+    kawaii.style.display = 'none'
+    env.style.display = 'block'
+    tagline.style.display = 'block'
```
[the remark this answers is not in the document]
alfonsusac: Good catch. Thank you! Fixed in latest commit!

app/pages/index.vue
alfonsusac: Fixed in latest commit!

app/pages/index.vue
```diff
@@ -36,8 +39,16 @@ defineOgImageComponent('Default', {
           dir="ltr"
           class="relative flex items-center justify-center gap-2 header-logo font-mono text-5xl sm:text-7xl md:text-8xl font-medium tracking-tight mb-6 motion-safe:animate-fade-in motion-safe:animate-fill-both"
         >
-          <AppLogo class="w-42 h-auto sm:w-58 md:w-70" />
+          <img
+            v-if="isKawaii === true"
+            width="400"
+            class="pb-8 motion-safe:animate-fade-in motion-safe:hover:scale-105 motion-safe:transition"
+            src="/extra/npmx-cute.svg"
+            alt="npmx kawaii logo"
```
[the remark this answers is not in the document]
alfonsusac: Should I write the en.json value for the new localization token? Or should I do it in separate PR?

app/pages/index.vue
```diff
@@ -24,6 +24,29 @@ defineOgImageComponent('Default', {
...
+    normalLogo.style.display = 'none'
+    kawaiiLogo.style.display = 'block'
+    logoEnv.style.display = 'none'
+    logoTagline.style.display = 'none'
+  } else {
+    normalLogo.style.display = 'block'
+    kawaiiLogo.style.display = 'none'
+    logoEnv.style.display = 'block'
+    logoTagline.style.display = 'block'
+  }
```
alexdln: This logic probably won't work because we have condition above now...

app/pages/index.vue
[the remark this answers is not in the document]
alfonsusac: Got it! Fixed in latest commit! Thanks!

app/pages/index.vue
alfonsusac: AAA this is embarassing 😅. Fixed in latest commit. Thanks!

PR #2355, opened 2026-04-01
feat: add logo to OAuth metadata

nicolo-ribaudo:
### 🧭 Context

When I go in my PDS's account management and look at the connected apps, npmx.dev has no logo. For comparison, PDSls has one:
image

(npmx is missing in the screenshot because right now I'm getting error 401 when trying to log in)

The difference seems to be that https://pdsls.dev/oauth-client-metadata.json links to a logo while https://npmx.dev/oauth-client-metadata.json does not.

### 📚 Description

This PR adds a logo_uri set to npmx.dev/logo.svg. I'm not 100% sure that SVG is a supported format and I don't know how to test it, I guess if we merge and it does not work then we switch to .ico like PDSls? :sweat_smile:

### DIFF
--- a/server/utils/atproto/oauth.ts
+++ b/server/utils/atproto/oauth.ts
@@ -41,11 +41,11 @@ export function getOauthClientMetadata(pkAlg: string | undefined = undefined): O
     ? webUriSchema.parse(`${clientUri}/.well-known/jwks.json`)
     : undefined
 
-  // If anything changes here, please make sure to also update /shared/schemas/oauth.ts to match
   return {
     client_name: 'npmx.dev',
     client_id,
     client_uri: clientUri,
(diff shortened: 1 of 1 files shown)


server/utils/atproto/oauth.ts
```diff
@@ -41,11 +41,11 @@ export function getOauthClientMetadata(pkAlg: string | undefined = undefined): O
     ? webUriSchema.parse(`${clientUri}/.well-known/jwks.json`)
     : undefined
 
-  // If anything changes here, please make sure to also update /shared/schemas/oauth.ts to match
```
nicolo-ribaudo, line 44: This file does not exist anymore

ghostdevv (approved):

PR #2357, opened 2026-04-01
fix: command for `create-*` packages in Deno

### LINKED ISSUE #2337: Deno command for `create-` packages is wrong/non-functional
### Describe the bug

Using `create-example` as an example, while others show `npm create example` or `bun create example`, Deno shows `deno run example` which yields `error: Module not found "file:///Users/username/example".`

The correct command(s) for Deno are:

Deno >2.7 (what should be shown imo):
`deno create npm:example`

Deno <2.7:
`deno init --npm example`

I do intend to submit a PR to fix.

### Additional context

Deno 2.7 added the `deno create` command as an alias to the `deno init` command. However the current command shown in npmx is wrong for both Deno 2.7 and all previous versions before Deno 2.7

knotbin:
### 🔗 Linked issue

resolves #2337

### 🧭 Context

For example package `create-example`, Current Deno command `deno run example` is incorrect and errors out. 

This PR fixes that and changes it to the correct `deno create` command.

### 📚 Description

This is a simple change that just changes the Deno command for create to the correct `deno create npm:example` command.


knotbin:
@ghostdevv So sorry, I should have scrolled farther when I got the CodeRabbit review, I assumed the first comment was its full review. 😅

app/utils/install-command.ts
```diff
@@ -137,6 +137,9 @@ export function getExecuteCommandParts(options: ExecuteCommandOptions): string[]
   if (options.isCreatePackage) {
     const shortName = getCreateShortName(options.packageName)
     if (shortName !== options.packageName) {
+      if (options.packageManager === 'deno') {
+        return ['deno', 'create', `npm:${shortName}`]
+      }
```
ghostdevv: Could we instead set the shortName to `npm:`? How are we handling the other cases where we need to prefix with `npm:` for deno?

ghostdevv (changes requested):

ghostdevv:
> @ghostdevv So sorry, I should have scrolled farther when I got the CodeRabbit review, I assumed the first comment was its full review. 😅

no worries!

app/utils/install-command.ts
knotbin: I added a helper function `getCreatePackageSpecifier` to handle this logic, similar to the `getPackageSpecifier` function that handles other cases of the `npm:` prefix. 

This logic is purposefully separate from the other `npm` prefix logic because JSR is not yet handled (out of this PR's scope) for `create-*` packages and if checks for matching JSR packages of `create-*` packages are added in the future (as they are for the other logic), it would be very different from the JSR logic in `getPackageSpecifier` because:
- `jsr:` specifier isn't needed for create command, unlike install command
- JSR "create" or "template" packages are required to have a `./create` export

app/utils/install-command.ts
```diff
@@ -125,6 +125,24 @@ export interface ExecuteCommandOptions extends InstallCommandOptions {
...
+  if (shortName === packageName) {
+    return null
+  }
+  if (packageManager === 'deno') {
+    // npm compatibility: npm:package
+    return `npm:${shortName}`
+  }
+
+  return shortName
+}
```
ghostdevv, line 144:
```suggestion
function getCreatePackageSpecifier(options: ExecuteCommandOptions): string | null {
  if (!options.isCreatePackage) {
    return null
  }

  const shortName = getCreateShortName(options.packageName)
  if (shortName === options.packageName) {
    return null
  }

  if (options.packageManager === 'deno') {
    // npm compatibility: npm:package-name
    return `npm:${shortName}`
  }

  return shortName
}
```

small nit pick 😅

ghostdevv: I can't make changes to your branch directly unfortunately :thinking: Could you also merge with main? Then we can merge! @knotbin

knotbin: Done, thank you!

PR #2365, opened 2026-04-03
refactor(cli)!: use mkdtempDisposable and require node >=24.4

trivikr:
### 🔗 Linked issue

N/A

### 🧭 Context

The connector’s `packageInit()` flow created a temporary directory with `mkdtemp()` and then manually removed it in a `finally` block. Since this codebase now runs on Node.js 24.x, we can use the built-in disposable temp directory API instead of managing that cleanup ourselves.

### 📚 Description

This PR updates the CLI package init flow to use `fs.promises.mkdtempDisposable()` with `await using` in `packageInit()`.


ghostdevv:
I updated the engines field, I think we can get away with this since we're in pre-release :p

PR #2373, opened 2026-04-03
fix: skip emoji conversion inside code blocks

### LINKED ISSUE #2364: `:1234:` is converted to emoji when found w/in an IPv6 address
### Describe the bug

npmx:

Image

---

GitHub:

Image

---

https://npmx.dev/package/ipaddr.js has example IPv6 addresses in code blocks that include the string `:1234:`, which npmx's slack emoji turns into 🔢. This shouldn't happen.

### Additional context

- https://npmx.dev/package/ipaddr.js
- https://github.com/whitequark/ipaddr.js

### Logs

```shell-script

```

fxderico:
### 🔗 Linked issue
Fixes #2364

### 🧭 Context
Emoji shortcodes like `:1234:` were being converted even inside code blocks in package READMEs.

### 📚 Description
`convertToEmoji` was running against the full HTML string, so `:1234:` in IPv6 addresses inside `<code>` and `<pre>` tags got turned into 🔢. Updated the function to match code/pre blocks first and skip them, only converting shortcodes in the rest of the content. Removed the now-unused `emojisKeysRegex` variable.

### DIFF
--- a/shared/utils/emoji.ts
+++ b/shared/utils/emoji.ts
@@ -1905,15 +1905,16 @@ const emojis = {
   'wales': '🏴󠁧󠁢󠁷󠁬󠁳󠁿',
 }
 
-const emojisKeysRegex = new RegExp(
-  Object.keys(emojis)
-    .map(key => `:${key}:`)
-    .join('|'),
-  'g',
-)
-export function convertToEmoji(text: string): string {
-  return text.replace(emojisKeysRegex, match => {
-    const key = match.slice(1, -1) as keyof typeof emojis
-    return emojis[key] || match
-  })
+export function convertToEmoji(html: string): string {
+  return html.replace(
+    /(<code[\s>][\s\S]*?<\/code>|<pre[\s>][\s\S]*?<\/pre>)|(:[\w+-]+:)/gi,
+    (match, codeBlock: string | undefined, shortcode: string | undefined) => {
+      if (codeBlock) return codeBlock
+      if (shortcode) {
+        const key = shortcode.slice(1, -1) as keyof typeof emojis
+        return emojis[key] || shortcode
+      }
+      return match
+    },
+  )
 }


shared/utils/emoji.ts
```diff
@@ -1905,15 +1905,16 @@ const emojis = {
...
-    .map(key => `:${key}:`)
-    .join('|'),
-  'g',
-)
-export function convertToEmoji(text: string): string {
-  return text.replace(emojisKeysRegex, match => {
-    const key = match.slice(1, -1) as keyof typeof emojis
-    return emojis[key] || match
-  })
+export function convertToEmoji(html: string): string {
```
ghostdevv, line 1908: this previously took in `text`, but now `html` - is this definitely the case everywhere? and if not I guess it doesn't matter?

ghostdevv (changes requested):

gameroman:
Hi @ffedex thanks for the PR

Could you please add some tests to verify the behavior and to prevent future regressions

shared/utils/emoji.ts
gameroman: Looking at the only usage of `convertToEmoji`

https://github.com/npmx-dev/npmx.dev/blob/a89588c923bacfe4c2dadcd4b8282cdd83e15231/server/utils/readme.ts#L796

It previously also took `html` actually, but it was not obvious, so I think this is correct

shared/utils/emoji.ts
```diff
@@ -1905,15 +1905,16 @@ const emojis = {
...
+  return html.replace(
+    /(<code[\s>][\s\S]*?<\/code>|<pre[\s>][\s\S]*?<\/pre>)|(:[\w+-]+:)/gi,
+    (match, codeBlock: string | undefined, shortcode: string | undefined) => {
+      if (codeBlock) return codeBlock
+      if (shortcode) {
+        const key = shortcode.slice(1, -1) as keyof typeof emojis
+        return emojis[key] || shortcode
```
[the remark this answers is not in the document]
gameroman: The coderabbit's suggestion is correct

PR #2379, opened 2026-04-04
fix: bridge SSR payload mismatch when client search provider differs from server default

### LINKED ISSUE #2380: SSR hydration mismatch and "no results" flash when searchProvider is set to npm
### Describe the bug

When the user's `searchProvider` setting is `npm` (instead of the default `algolia`), several pages experience SSR hydration errors and broken behavior.

## Affected routes

- `/search?q=vue` — hydration error + brief flash of correct results followed by a "no results" screen
- `/settings` — hydration error
- Likely also `/~[username]` and `/org/[orgname]` routes (any route using search composables)

## Root cause (issue is created post MR creation 😅 )

The server always fetches search data using the default provider (`algolia`), producing an SSR cache key like `search:algolia:vue`. When the client hydrates with `searchProvider = 'npm'`, it looks for `search:npm:vue` in the payload — which doesn't exist. This cache key mismatch causes `useLazyAsyncData` to miss the SSR cache, triggering a client-side refetch and hydration errors.

### Logs

```shell-script
## Steps to reproduce

1. Go to `/settings`, change the search data source to **npm**
2. Navigate directly (via URL bar) to `/search?q=vue`
3. **Observe:** hydration errors in console; search results flash then disappear, replaced by "no results"
4. Navigate directly to `/settings`
5. **Observe:** hydration error in console
```

gusa4grr:
### 🔗 Linked issue

Resolves #2380 

### 🧭 Context

When the user's `searchProvider` value is `npm`, there will be hydration errors and more. Routes affected are `search?q=vue` or `/settings` page, but it may be more, like `/~[username]` route

Moreover, on `search?q=vue`, the user is additionally presented with "no results screen" after flashing actual search results 😅 

In this MR I:
* I introduced the Search SSR Bridge between both searchProviders, as I found that the results are of the same structure and assumed I could "reuse" them during hydration
* added vue org to the e2e fixture, so it mimics now the same scenario as on the prod app, as before in tests, there were no "org" checks in search results
* aligned the "interaction" spec with arrow-up/down behaviour to accommodate new expected behaviour 

In order to see this behaviour yourself, you need to:
1. Go to settings, change your provider to `npm` from algolia
2. write in url bar manually `/search?q=vue`. Observe hydration + no results
3. write in url bar manually `/settings`. Observe hydration error

More details in "details" section below 😉 

### Images attached:

search page:

image

settings page:

image

- Centralize `?p` query param merge into `useSearchProvider()` getter
- Remove duplicated `searchProviderValue` computed from all consumers
- Add `bridgeSearchSSRPayload()` to copy SSR cache to client's provider key during hydration
- Destructure asyncData before spreading to avoid shadowing custom `data` ref
- Update e2e tests for org suggestion cards in keyboard navigation

### 📚 Description

During my adventures with the [implementing server-synced user preferences](https://github.com/npmx-dev/npmx.dev/pull/1189) (please check that one too if you fancy 🙂 ), I stumbled upon the e2e tests failing with hydration errors, which were:
* landing on /search?q=vue
* settings page
Also, the `interactions` spec and `url-compatibility` spec have each failed tests 🤯 

So I went ahead and started digging into what "have I done" so it broke.
after bunch of hours of little to no success I (finally) decided to check how does the `main` branch behave and just went ahead to npmx.dev prod app and I found that with `npm` as searchProvider, th things are actually not working as they should.

This became visible during my implementation, as preferences are now server-synced, which just uncovered those issues. And the combination of the fact that in the "test" environment, searchProvider` is defaulted to `npm`.

I continued digging and found a misalignment in the searchProvider value during hydration. Basically the fetch was always happening with algolia, and then when NPM "results" were tried to be hydrated on client - it sees nothing due to cache key mismatch between providers.

so I dig more and more and I come up with the solution you see in this MR.

I'm happy to discuss further or implement a better approach.

NOTE: When navigating via the website to the search page, all is OK. like for instance if you're on the `/settings` and write in npmx search bar vue => it will render OK. The secret here is the `?p=npm` query parameter, which is programmatically added, thereby enabling the search page to find results.

### DIFF
--- a/app/composables/useGlobalSearch.ts
+++ b/app/composables/useGlobalSearch.ts
@@ -1,4 +1,3 @@
-import { normalizeSearchParam } from '#shared/utils/url'
 import { debounce } from 'perfect-debounce'
 
 // Pages that have their own local filter using ?q
@@ -9,11 +8,6 @@ const SEARCH_DEBOUNCE_MS = 100
 export function useGlobalSearch(place: 'header' | 'content' = 'content') {
   const { settings } = useSettings()
   const { searchProvider } = useSearchProvider()
-  const searchProviderValue = computed(() => {
-    const p = normalizeSearchParam(route.query.p)
-    if (p === 'npm' || searchProvider.value === 'npm') return 'npm'
-    return 'algolia'
-  })
 
   const router = useRouter()
   const route = useRoute()
@@ -111,7 +105,7 @@ export function useGlobalSearch(place: 'header' | 'content' = 'content') {
   return {
     model: searchQueryValue,
     committedModel: committedSearchQuery,
-    provider: searchProviderValue,
+    provider: searchProvider,
     startSearch: flushUpdateUrlQuery,
   }
 }
--- a/app/components/SearchProviderToggle.client.vue
+++ b/app/components/SearchProviderToggle.client.vue
@@ -2,11 +2,6 @@
 const route = useRoute()
 const router = useRouter()
 const { searchProvider } = useSearchProvider()
-const searchProviderValue = computed(() => {
-  const p = normalizeSearchParam(route.query.p)
-  if (p === 'npm' || searchProvider.value === 'npm') return 'npm'
-  return 'algolia'
-})
 
 const isOpen = shallowRef(false)
 const toggleRef = useTemplateRef('toggleRef')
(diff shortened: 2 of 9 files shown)


gusa4grr:
Fixes #2380

app/composables/useGlobalSearch.ts
```diff
@@ -68,7 +62,7 @@ export function useGlobalSearch(place: 'header' | 'content' = 'content') {
       })
       return
     }
-    router.push({
+    void router.push({
```
ghostdevv: are these voids necessary?

ghostdevv (changes requested):

app/composables/useGlobalSearch.ts
gusa4grr: oh, my bad, I just "fixed" the errors in my IDE, but apparently that's my local `oxc.typeAware: true` config, which marks this as error.

I'll revert the changes

gusa4grr: done @ghostdevv 👍🏻

PR #2383, opened 2026-04-05
fix: resolve semantic versioning issue

### LINKED ISSUE #2342: Navigating to a nonexistent package version does not error
### Describe the bug

https://npmx.dev/package/axios/v/150.150.150
You can do on whatever semantic version you want to...

### Additional context

Any semantic version

### Logs

```shell-script

```

ElshadHu:
### 🔗 Linked issue

Fixes #2342
### 🧭 Context

Requesting a non-existent package version (e.g. `/package/axios/v/150.150.150`) returned HTTP 200 instead of 404. The page was silently showing content as if the version existed.

Added version validation in `useResolvedVersion` and unit tests to cover the fix.

### 📚 Description

The `npm.antfu.dev` API (fast-npm-meta) echoes back any version string without validating it against the registry. For non-existent versions it returns a 200 response but omits `publishedAt`. Because the fetch succeeded and returned a truthy version string, the existing 404 guards in the page never triggered.

The fix adds a check in `useResolvedVersion`: when an exact version is requested (starts with a digit) and the response has no `publishedAt`, a second request is made to the `/versions/` endpoint on the same API to confirm the version actually exists. If it does not, the composable returns `undefined`, which the existing 404 guards in [name].vue already handle correctly - both on the server (throwing a 404) and on the client (showing the error page).

Dist-tags (`latest`, `next`) and semver ranges start with a letter and skip this check since they resolve to a different version string by design.

**Files changed:**
- useResolvedVersion.ts - added registry validation when `publishedAt` is absent
- use-resolved-version.spec.ts - added unit tests covering valid versions, non-existent versions, old versions without `publishedAt`, dist-tags, and scoped packages

### DIFF
--- a/app/composables/npm/useResolvedVersion.ts
+++ b/app/composables/npm/useResolvedVersion.ts
@@ -1,4 +1,5 @@
 import type { PackageVersionsInfo, ResolvedPackageVersion } from 'fast-npm-meta'
+import semver from 'semver'
 
 export function useResolvedVersion(
   packageName: MaybeRefOrGetter<string>,
@@ -18,7 +19,7 @@ export function useResolvedVersion(
       // error (no publishedAt, no validation). When publishedAt is missing for
       // an exact version request, cross-check the versions list to confirm the
       // version actually exists in the registry.
-      if (version && /^\d/.test(version) && !data.publishedAt) {
+      if (version && semver.valid(version) && !data.publishedAt) {
         const versionsData = await $fetch<PackageVersionsInfo>(
           `https://npm.antfu.dev/versions/${name}`,
         )
--- a/test/nuxt/composables/use-resolved-version.spec.ts
+++ b/test/nuxt/composables/use-resolved-version.spec.ts
@@ -118,7 +118,7 @@ describe('useResolvedVersion', () => {
   })
 
   it('does not cross-check dist-tags against the versions list', async () => {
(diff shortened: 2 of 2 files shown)


app/composables/npm/useResolvedVersion.ts
```diff
@@ -13,6 +14,20 @@ export function useResolvedVersion(
         ? `https://npm.antfu.dev/${name}@${version}`
         : `https://npm.antfu.dev/${name}`
       const data = await $fetch<ResolvedPackageVersion>(url)
+
+      // The fast-npm-meta API echoes back non-existent exact versions without
+      // error (no publishedAt, no validation). When publishedAt is missing for
+      // an exact version request, cross-check the versions list to confirm the
+      // version actually exists in the registry.
```
ghostdevv, line 21: I'd almost rather fix this upstream, [want to make a PR](https://github.com/antfu/fast-npm-meta/)?

ElshadHu: Sure, I can make a PR.

ElshadHu: I made a [PR](https://github.com/antfu/fast-npm-meta/pull/37) in upstream, if there is any issue, please let me know

ghostdevv:
I'll close this in anticipation of https://github.com/antfu/fast-npm-meta/pull/37 being merged - if that isn't then we can revisit this :pray:

PR #2384, opened 2026-04-05
perf: parallelize jsDelivr README fallback probes

trivikr:
### 🔗 Linked issue

N/A

### 🧭 Context

README fallback probing in `server/utils/readme-loaders.ts` was strictly sequential. On cache misses, packages with missing or nonstandard README filenames could pay for several failed jsDelivr requests before the loader found the right file or gave up, which made worst-case README render time noticeably slower than necessary.

### 📚 Description

This changes the jsDelivr README fallback path to reduce that long-tail latency while keeping the fallback behavior conservative.

* Build a prioritized candidate list that tries the npm registry’s `readmeFilename` first when it exists.
* Probe jsDelivr candidates in small parallel batches instead of one-by-one.
* Preserve existing fallback semantics for missing, nonstandard, and truncated npm READMEs.
* Add unit coverage for batched probing and candidate prioritization in `test/unit/server/utils/readme-loaders.spec.ts`.


server/utils/readme-loaders.ts
```diff
@@ -35,15 +44,27 @@ export async function fetchReadmeFromJsdelivr(
...
+    const responses = await Promise.all(
+      batch.map(async filename => {
+        try {
+          const url = `https://cdn.jsdelivr.net/npm/${packageName}${versionSuffix}/${filename}`
+          const response = await fetch(url)
+          if (!response.ok) {
+            return null
+          }
+
+          return await response.text()
```
ghostdevv: should we move this to the end so we only read the body of the response that works?

ghostdevv (changes requested):

server/utils/readme-loaders.ts
trivikr: Done in https://github.com/npmx-dev/npmx.dev/pull/2384/changes/fd7d050a1a8633bfc138f497aec0765f963603d3

trivikr:
Also updated PR to fetch provided readme before starting the fallback batch in https://github.com/npmx-dev/npmx.dev/pull/2384/changes/72bb0b216fcb61b2f8950a33f6e6c96ebd4901e6

server/utils/readme-loaders.ts
```diff
@@ -35,15 +42,27 @@ export async function fetchReadmeFromJsdelivr(
...
+        } catch {
+          return null
+        }
+      }),
+    )
+
+    const matchedReadme = responses.find((response): response is Response => response !== null)
+    if (matchedReadme) {
+      return await matchedReadme.text()
     }
```
ghostdevv:
```suggestion
    for (const response of responses) {
      const text = await response?.text()
      if (text) return text
    }
```

I guess it's possible that the readme response is empty, so I wonder if you should do something like this 🤔

ghostdevv (changes requested):

server/utils/readme-loaders.ts
trivikr: Yes. We can also skip whitespace only readme bodies by checking for `text?.trim()`
```suggestion
    for (const response of responses) {
      const text = await response?.text()
      if (text?.trim()) {
        return text
      }
    }
```

server/utils/readme-loaders.ts
```diff
@@ -35,15 +42,29 @@ export async function fetchReadmeFromJsdelivr(
...
+          return null
+        }
+      }),
+    )
+
+    for (const response of responses) {
+      const text = await response?.text()
+      if (text?.trim()) {
+        return text
       }
```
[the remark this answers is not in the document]
ghostdevv: :melting_face: what do you think? @trivikr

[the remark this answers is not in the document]
trivikr: Done in https://github.com/npmx-dev/npmx.dev/pull/2384/changes/82266864a89ffdf083de4871ab2ba4192d8055a5

ghostdevv (approved):

Issue #2340, opened 2026-03-31
semver range seems to not calculate correctly for comparison operators


denis-sokolov:
`^7.5 <7.6` matches 7.6, this seems unintentional.

(Same for `~7.5` matching 7.6)

<img width="410" height="390" alt="Image" src="https://github.com/user-attachments/assets/e144f0b8-8d70-48a1-a5cc-0c88bb774616" /> <img width="422" height="211" alt="Image" src="https://github.com/user-attachments/assets/8f675ade-d216-4fb6-b16f-a119c46399f4" />

(Example taken from [the prisma package page](https://npmx.dev/package/prisma))

joshuaisaact:
Hey so I fixed this in the above PR. It turns out we were always adding the latest version at the end of the filter array. 

It might be that was intentional, but I think I agree with you - it doesn't really make much sense, from a UX perspective. 

If I filter by semver I should only see the semvers I want to filter by, I shouldn't always see latest too.

denis-sokolov:
Thank you!

Yes, I was fully confused by the feature and I thought I can not trust the results given what it was showing. 

If we want to always show the latest, I would visualize it. E.g. move the latest version above the search field.

PR #2389, opened 2026-04-05
fix(ui): respect semver filter when showing latest tag

joshuaisaact:
The 'latest' tag was unconditionally added to visible results even when it didn't match the active semver filter. 

This caused versions like 7.6.0 to appear when filtering by ranges like '^7.5 <7.6'.

Remove the special case handling for the 'latest' tag and let it flow naturally through the filtering logic. 

This makes the behaviour predictable and fixes the bug where non-matching versions were displayed.

### 🔗 Linked issue

Fixes issue #2340

### 🧭 Context

We gave 'latest' releases special handling in the semver filter, which meant they always appeared in certain circumstances. 

### 📚 Description

Fixes the linked bug. Also simplifies the code - not sure why we gave latest versions special treatment.

### DIFF
--- a/app/components/Package/Versions.vue
+++ b/app/components/Package/Versions.vue
@@ -210,7 +210,18 @@ const visibleTagRows = computed(() => {
           getTagVersions(row.tag).some(v => filteredVersionSet.value.has(v.version)),
       )
     : rowsMaybeFilteredForDeprecation
-  return rows.slice(0, MAX_VISIBLE_TAGS)
+  const first = rows.slice(0, MAX_VISIBLE_TAGS)
+
+  // When no filter is active, ensure 'latest' is always shown (even if not fully loaded)
+  if (!isFilterActive.value) {
+    const latestTagRow = rows.find(row => row.tag === 'latest')
+    if (latestTagRow && !first.includes(latestTagRow)) {
+      first.pop()
+      first.push(latestTagRow)
+    }
+  }
+
+  return first
 })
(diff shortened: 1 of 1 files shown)


app/components/Package/Versions.vue
```diff
@@ -210,14 +210,7 @@ const visibleTagRows = computed(() => {
...
       )
     : rowsMaybeFilteredForDeprecation
-  const first = rows.slice(0, MAX_VISIBLE_TAGS)
-  const latestTagRow = rows.find(row => row.tag === 'latest')
-  // Ensure 'latest' tag is always included (at the end) if not already present
-  if (latestTagRow && !first.includes(latestTagRow)) {
-    first.pop()
-    first.push(latestTagRow)
-  }
-  return first
```
serhalp, line 220: This behaviour was ensuring that the `latest` tag is always included even if we haven't fully loaded all versions for the package under the hood. This is important when _no semver filter has been typed at all_.

This code isn't specific to the semver filter. So I believe we need to keep this behaviour when `!isFilterActive.value`. Make sense?

joshuaisaact: Ohhhhh right. Yeah, that makes a ton of sense, my bad. I'll update it

joshuaisaact: Ok resolving this comment @serhalp  - I've updated it so the old logic is retained when there are no filters.

PR #2393, opened 2026-04-05
fix(ui): preserve root dir state when sibling dir is expanded

### LINKED ISSUE #2363: Source tree-view expands 2 directories instead of 1
### Describe the bug

- Got to https://npmx.dev/package-code/ky/v/1.14.3/distribution%2Ftypes%2Fcommon.d.ts
- Collapse `core/`
- Try expanding `types/`

Expected: `types/` is expanded

Actual: both `types/` and `core/` are expanded

https://github.com/user-attachments/assets/b27cb3f9-9356-44e3-8156-7533a5629d42

### Additional context

_No response_

### Logs

```shell-script

```

trivikr:
### 🔗 Linked issue

resolves https://github.com/npmx-dev/npmx.dev/issues/2363

### 🧭 Context

The package code sidebar was re-expanding a collapsed sibling directory when another directory in the same tree was opened. On `ky@1.14.3`, collapsing `core/` and then expanding `types/` caused both directories to appear expanded again.

This change fixes the tree expansion logic so only the intended directory opens, and adds coverage for the regression.

### 📚 Description

The root cause was that the code tree auto-expanded the current file path from recursive `CodeFileTree` instances, not just from the root tree. That meant a subtree remount could re-apply ancestor expansion and reopen a directory the user had just collapsed.

This PR:
- limits auto-expansion of the current file path to the root file tree instance
- keeps file tree expansion state updates explicit by replacing the stored `Set` after mutations
- adds a Nuxt regression test covering the sibling-directory case from the issue

### DIFF
--- a/app/composables/useFileTreeState.ts
+++ b/app/composables/useFileTreeState.ts
@@ -7,7 +7,6 @@ export function useFileTreeState(baseUrl: string) {
     } else {
       expanded.value.add(path)
     }
-    expanded.value = new Set(expanded.value)
   }
 
   function isExpanded(path: string) {
@@ -22,7 +21,6 @@ export function useFileTreeState(baseUrl: string) {
       prefix = prefix ? `${prefix}/${part}` : part
       expanded.value.add(prefix)
     }
-    expanded.value = new Set(expanded.value)
   }
 
   return {
--- a/app/composables/useFileTreeState.ts
+++ b/app/composables/useFileTreeState.ts
@@ -7,6 +7,7 @@ export function useFileTreeState(baseUrl: string) {
     } else {
       expanded.value.add(path)
     }
+    expanded.value = new Set(expanded.value)
   }
 
   function isExpanded(path: string) {
@@ -21,6 +22,7 @@ export function useFileTreeState(baseUrl: string) {
       prefix = prefix ? `${prefix}/${part}` : part
       expanded.value.add(prefix)
     }
+    expanded.value = new Set(expanded.value)
   }
 
   return {
--- a/test/nuxt/components/CodeFileTree.spec.ts
+++ b/test/nuxt/components/CodeFileTree.spec.ts
@@ -1,0 +1,89 @@
+import { mountSuspended } from '@nuxt/test-utils/runtime'
+import { describe, expect, it, vi } from 'vitest'
+import CodeFileTree from '~/components/Code/FileTree.vue'
+
+const mockTree: PackageFileTree[] = [
+  {
+    name: 'distribution',
+    type: 'directory',
+    path: 'distribution',
+    children: [
+      {
+        name: 'core',
+        type: 'directory',
+        path: 'distribution/core',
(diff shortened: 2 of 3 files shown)


test/nuxt/components/CodeFileTree.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+          version: '1.14.3',
+          filePath: '',
+        },
+      },
+    },
+  })
+}
+
+describe('CodeFileTree', () => {
+  it('keeps a collapsed sibling directory closed when another sibling expands', async () => {
```
jhroemer: I'm personally a little wary of tests like this, because the space of "this thing should not happen" is incredibly large. There's nothing wrong with the expectation, but maybe it can just be part of a normal collapse/expand test, instead of being a test specifically for this bug. Wdyt?

johnrommeloctaviano (approved):

app/composables/useFileTreeState.ts
```diff
@@ -7,6 +7,7 @@ export function useFileTreeState(baseUrl: string) {
     } else {
       expanded.value.add(path)
     }
+    expanded.value = new Set(expanded.value)
```
ghostdevv: what does this do?

ghostdevv (changes requested):
Could you take a look at the other review comments too?

app/composables/useFileTreeState.ts
serhalp: I think `expanded.value.clear()` might be clearer here (pun intended)

test/nuxt/components/CodeFileTree.spec.ts
trivikr: Renamed in https://github.com/npmx-dev/npmx.dev/pull/2393/changes/05dcb211cb1015ad2ce07ac02597932a8d4f55d0

test/nuxt/components/CodeFileTree.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+    expect(typesButton).toBeDefined()
+    await typesButton!.trigger('click')
+
+    await vi.waitFor(() => {
+      expect(wrapper.text()).toContain('common.d.ts')
+      expect(wrapper.text()).not.toContain('constants.d.ts')
+    })
+
+    wrapper.unmount()
+  })
```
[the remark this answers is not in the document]
trivikr: Moved in https://github.com/npmx-dev/npmx.dev/pull/2393/changes/f13968d59b32474bd7b33d27352b7b8e1ed0d7d9

app/composables/useFileTreeState.ts
trivikr: Removed in https://github.com/npmx-dev/npmx.dev/pull/2393/changes/35073d6548f1e9c66ec40f8041b99b5e146b2208

ghostdevv (approved):

PR #2415, opened 2026-04-07
fix(cli): enable web auth for package:init

### LINKED ISSUE #2414: Claiming a package is not possible with passkeys-only 2FA
### Describe the bug

Hi, 

When claiming a package, it is not possible to do so if your npm account only has passkeys for 2FA in npmjs.com. 

This relates to the removal of TOTP based 2FA for new accounts: [Github Blog](https://github.blog/changelog/2025-10-10-strengthening-npm-security-important-changes-to-authentication-and-token-management/ )

My understanding is that this bug affects all accounts created after the removal of TOTP based 2FA, where the user has set up 2FA with passkeys. There is no way to set up TOTP anymore, which npmx.dev expects for the "Claim package" operation.

# Reproducing the issue

## Prerequisites for reproducing the issue:

The following steps assumes you have set up Two-Factor Authentication in npmjs.com with passkeys as your only 2FA method. Here's a screenshot of my Account Settings -> Modify 2FA screen in npmjs.com. Using this account I discovered the bug:

Image

## Steps to reproduce:

1. Connect to npm CLI.
2. Search for a package that does not already exist and click "Claim package". The "Claim Package Name" dialog should appear.
3.  Click the "Claim Package Name" button.

## Expected behavior:

Web auth should be used, so that npmjs.com opens where you can authenticate with your passkey. 

## Actual behavior: 

Image

The "Local Connector" dialog in npmx.dev appears, showing: 

**Enter OTP to continue** - with a input field for OTP code and a "Retry with OTP" button.

It also displays: **---OR--- Retry with web auth.**

Retrying web auth does not work, it loops you back to the "Local Connector" dialog.

In this case, the user has no OTP set up and therefore no OTP code to enter, so it is a very confusing message to get. 

--> There is now no way for the user to continue, so they cannot claim the package.

### Root cause

When "Retry with web auth" is clicked, frontend sends POST /execute to the CLI backend with body:
```json
{
    "interactive":true,
    "openUrls":false
}
```

"Interactive" means using web auth (as opposed to OTP).

In **cli/src/server.ts** at app.post('/execute'), executeOperation() is called, which creates a `execOptions: ExecNpmOptions` object containing this `'interactive': true` flag, then passes it to the `"package:init"` operation at line 812.

However, there's a special case for package:init that says it explicitly does not support interactive mode even though that was what requested by the frontend:

```ts
case 'owner:add':
    result = await ownerAdd(params.user, params.pkg, execOptions)
    break
case 'owner:rm':
    result = await ownerRemove(params.user, params.pkg, execOptions)
    break
case 'package:init':
    // package:init has its own special execution path (temp dir + publish)
    // and does not support interactive mode
    result = await packageInit(params.name, params.author, options.otp)
    break
``` 

Notice how the package:init is passed only options.otp and not execOptions - which contains interactive mode properties.

# Fix (PR incoming)

I don't presume to know for what reason there is a special case in place here, but it seems to me that sending the execOptions containing interactive mode properties would make web auth work and thereby allowing 2FA passkeys to be used.

I have successfully reproduced and fixed the issue in my local env. Will hit you with a PR soon.

# Other notes

Aside from this bug, this process is confusing and not very user friendly. It makes multiple assumptions on what the user should understand, such as: OTP from where, npm or npmx.dev? What does OTP mean in this case - is it a authenticator app? Code sent to your email? Am I supposed to wait from a email here?

Also, I'm fairly certain this is also broken for new accounts who has not set up 2FA passkeys - though that takes users down a different path than what I described above. 

### Additional context

_No response_

### Logs

```shell-script

```

ehs5:
### 🔗 Linked issue

Fixes #2414

### 🧭 Context

**package:init** (Claim Package) never attempted web auth (AKA interactive mode) when requested by frontend. A special case had been put in place to cause this behaviour in **cli/src/server.ts** - but this completely blocked users who only have set up 2FA with passkeys from claiming a package. The only other option to proceed was OTP, which is not available to set up as a 2FA method at npmjs.com anymore.

I suspect npm users that have not set up any 2FA also could not complete the "Claim package" operation - though I haven't verified this entirely.

### 📚 Description

The root cause of the issue was in **cli/src/server.ts**. In executeOperation(), there was a special case where interactive mode was not executed when operation was of type "package:init". 

I rewrote this so that in case "package:init", packageInit() now takes execOptions as param instead of just otp code, making it run similarly to other operations. The execOptions object contains necessary params to complete the web auth procedure.

To make this work, packageInit() in **cli/src/npm-client.ts** now runs execNpm() similarly to the other operations, whereas before it had its own, mostly duplicated, execution of execFileAsync(). To achieve this, an optional cwd param had to be added to the ExecNpmOptions type. This is so that the tempDir path (where package.json is temporarily stored when claiming package) can be passed down to execFileAsync() for OTP and pty.spawn() for interactive, allowing for running the npm commands in the temp folder.

The code has been successfully run from my local env allowing my account, that does not have OTP as 2FA, to claim a package.

### DIFF
--- a/cli/src/npm-client.ts
+++ b/cli/src/npm-client.ts
@@ -157,6 +157,8 @@ const AUTH_URL_TIMEOUT_MS = 90_000
 export interface ExecNpmOptions {
   otp?: string
   silent?: boolean
+  /** Working directory for the npm command. */
+  cwd?: string
   /** When true, use PTY-based interactive execution instead of execFile. */
   interactive?: boolean
   /** When true, npm opens auth URLs in the user's browser.
@@ -214,6 +216,7 @@ async function execNpmInteractive(
     name: 'xterm-256color',
     cols: 120,
     rows: 30,
+    cwd: options.cwd,
     env,
   })
 
@@ -337,6 +340,7 @@ async function execNpm(args: string[], options: ExecNpmOptions = {}): Promise<Np
     const { command, args: processArgs } = resolveNpmProcessCommand(npmArgs)
     const { stdout, stderr } = await execFileAsync(command, processArgs, {
       timeout: 60000,
+      cwd: options.cwd,
       env: createNpmEnv(),
     })
 
@@ -559,12 +563,12 @@ export async function listUserPackages(user: string): Promise<NpmExecResult> {
  * Creates a minimal package.json in a temp directory and publishes it.
  * @param name Package name to claim
  * @param author npm username of the publisher (for author field)
- * @param otp Optional OTP for 2FA
+ * @param options Execution options (otp, interactive, etc.)
  */
 export async function packageInit(
   name: string,
   author?: string,
-  otp?: string,
+  options?: ExecNpmOptions,
 ): Promise<NpmExecResult> {
   validatePackageName(name)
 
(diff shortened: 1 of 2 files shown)


ghostdevv:
Do you know why this special case was added originally?

ehs5:
Not entirely, but based  on the comment in server.ts:

```ts
// package:init has its own special execution path (temp dir + publish)
// and does not support interactive mode

```

In npm-client.ts, the code needs to run "npm publish" from the temp folder that's created for the package.json to be published. It might be it wasn't immediately obvious on how to do that, or a reluctance to change the  function that runs npm commands, execNpm(), since the other operations called execNpm() where the command would be run in the current directory. I got around that by passing the temp directory as an optional cwd parameter to execNpm().

serhalp (approved):
LGTM. Thank you for fixing and testing!

I tried to write some test coverage for this, but all I could come up with was really gross and wouldn't be catching much, so I think we can just live without tests here.

ehs5:
Thank you! This was actually my first open source contribution to any project ever. Really nice to be able to contribute to such a cool and visible project and be met with nothing but positivity from you guys. 🙏🏻

PR #2425, opened 2026-04-08
fix(i18n): update Czech translations

VentyCZ:
### 📚 Description

Updating Czech translations. Also, adding two new translation strings for proper aria label support on collapsible sections.


ghostdevv (approved):

ghostdevv:
ty, good catch!! could you split the translations and key addition up next time though :eyes:

VentyCZ:
> ty, good catch!! could you split the translations and key addition up next time though 👀

Yea, will do :) Just wanted to translate it while I'm at it. Will probably go hunt more missing / non-translatable strings in the future.

PR #2432, opened 2026-04-09
fix: pass query params to OG image ISR function on Vercel

### LINKED ISSUE #2431: Compare page OG image renders empty state on production
The OG image for the compare page always renders the empty state on production, even when packages are provided via query params.

**URL:** https://npmx.dev/compare?packages=zustand,redux&view=charts

**Expected:** OG image shows package comparison with download bars (zustand vs redux)
**Actual:** OG image shows generic "Compare npm packages side-by-side" empty state

## Evidence

- Works correctly on local dev server (`localhost:3001`)
- On Vercel production, the `nuxt-og-image-options` script tag in the page HTML contains the correct serialized props (`packages: ["zustand", "redux"]`)
- But the rendered OG image (both PNG and HTML endpoints) shows the empty state; packages never reach the component
- Even cache-busted fresh renders (`x-vercel-cache: MISS`) produce the same empty image
- The etag is identical across all URL variations (with/without `_query`, with/without packages)

## Reproduction

```bash
# Works locally (dev):
curl http://localhost:3001/__og-image__/image/compare/og.png?packages=zustand,redux -o local.png

# Broken on production:
curl 'https://npmx.dev/__og-image__/image/compare/og.html?packages=zustand,redux' | grep -i 'zustand'
# Returns nothing — empty state
```

## Introduced in

PR #2277 (feat: add OG image for compare pages) — merged Apr 6

Adebesin-Cell:
## Summary

- Adds `passQuery: true` to the `/__og-image__/**` ISR route rule
- Vercel's ISR `passQuery` defaults to `false`, which strips all query parameters before passing the request to the serverless function
- This caused the compare page OG image to always render the empty state because the function never received `packages=zustand,redux` or the `_query` parameter that `nuxt-og-image` relies on

## Root cause

Vercel's [prerender config](https://vercel.com/docs/build-output-api/v3/primitives#prerender-configuration-file) has a `passQuery` option that defaults to `false`. Without it, the ISR function receives the request with query params stripped — so the OG image handler always fetched `/compare` (no packages) instead of `/compare?packages=zustand,redux`.

## Test plan

- [ ] Deploy preview and verify OG image at `/compare?packages=zustand,redux`
- [ ] Check `/__og-image__/image/compare/og.html?packages=zustand,redux&_query=%7B%22packages%22:%22zustand,redux%22%7D` shows package names
- [ ] Verify other OG images (package pages, homepage) still work

Closes #2431


nuxt.config.ts
```diff
@@ -128,7 +128,13 @@ export default defineNuxtConfig({
...
     '/:pkg/.well-known/skills/**': { isr: 3600 },
     '/:scope/:pkg/.well-known/skills/**': { isr: 3600 },
-    '/__og-image__/**': getISRConfig(3600),
+    '/__og-image__/**': {
+      isr: {
+        expiration: 3600,
+        passQuery: true,
+        allowQuery: ['packages', '_query'],
+      },
+    },
```
alexdln, line 138: Can you please specify that these parameters are only allowed for /__og-image__/image/compare/og.html?

alexdln:
Overall, it's a standard solution. The only thing I doubt is that there can be any number of these parameters and we could exceed the limits (_incl. with different params order_). On the other hand, we also have an infinite number of packages and it's fine there.

I think we can try it and if anything happens we'll roll it back and make a universal image

p.s. can we add sorting [here](https://github.com/npmx-dev/npmx.dev/blob/main/app/pages/compare.vue#L143) so that at least we don't store different images with the same content depending on the order?

nuxt.config.ts
Adebesin-Cell: I think we can add to allow query for only the comparison page?

Adebesin-Cell:
> Overall, it's a standard solution. The only thing I doubt is that there can be any number of these parameters and we could exceed the limits (_incl. with different params order_). On the other hand, we also have an infinite number of packages and it's fine there.
> 
> I think we can try it and if anything happens we'll roll it back and make a universal image
> 
> p.s. can we add sorting [here](https://github.com/npmx-dev/npmx.dev/blob/main/app/pages/compare.vue?rgh-link-date=2026-04-09T11%3A32%3A29Z#L143) so that at least we don't store different images with the same content depending on the order?

Good point!

app/pages/compare.vue
```diff
@@ -140,7 +140,7 @@ async function exportComparisonDataAsMarkdown() {
 }
 
 defineOgImageComponent('Compare', {
-  packages: () => packages.value,
+  packages: () => packages.value.toSorted((a, b) => a.localeCompare(b)),
```
alexdln, line 143: _we already used toSorted, so should be fine_

alexdln (approved):

alexdln:
Thanks for the fix 🤍

PR #2435, opened 2026-04-09
ci: add e18e dependency diff workflow

serhalp:
### 🔗 Linked issue

N/A

### 🧭 Context

See https://github.com/e18e/action-dependency-diff.

### 📚 Description

This configures the e18e/action-dependency-diff workflow. I've tried to tune it to avoid being noisy, but we may have to iterate a bit on it once merged.

Note that the action doesn't have sophisticated monorepo handling, so it won't report the base app + cli + docs separately, since they share a lockfile.

### DIFF
--- a/.github/workflows/dependency-diff.yml
+++ b/.github/workflows/dependency-diff.yml
@@ -18,7 +18,6 @@ concurrency:
 
 permissions:
   contents: read
-  pull-requests: write
 
 jobs:
   dependency-diff:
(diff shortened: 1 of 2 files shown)


.github/workflows/dependency-diff.yml
```diff
@@ -0,0 +1,39 @@
+name: dependency-diff
+
+on:
+  pull_request:
```
43081j, line 4: note that it'll only succeed on PRs made by people with permission to `write` the repo iirc

that can be solved by setting `pull_request_target` but people are usually scared of that because of the security implications. a very well written workflow will be fine but it still increases risk i guess that we one day let something slip in

lets discuss on discord 😄

43081j (approved):

PR #2469, opened 2026-04-11
fix: retry replacement suggestions after transient fetch failures

trivikr:
### 🔗 Linked issue

N/A

### 🧭 Context

The compare page caches replacement lookups per package in `useCompareReplacements`. That cache treated transient fetch failures the same as a real “no replacement” response by storing `null` for both cases. Once that happened, the package was considered already checked and would never be fetched again for the rest of the session.

In practice, a temporary API failure, offline blip, or server 500 could permanently suppress replacement suggestions until the app was reloaded.

### 📚 Description

This PR changes compare-page replacement caching so only successful fetch results are persisted.

Successful `null` responses are still cached, which preserves the existing optimization for packages that truly have no replacement suggestion. Thrown fetch errors are no longer written into the cache, which allows the package to be retried on a later watch cycle instead of being suppressed for the rest of the session.

### DIFF
--- a/app/composables/useCompareReplacements.ts
+++ b/app/composables/useCompareReplacements.ts
@@ -42,15 +42,18 @@ export function useCompareReplacements(packageNames: MaybeRefOrGetter<string[]>)
         namesToCheck.map(async name => {
           try {
             const replacement = await $fetch<ModuleReplacement | null>(`/api/replacements/${name}`)
-            return { name, replacement }
+            return { name, replacement, failed: false as const }
           } catch {
-            return { name, replacement: null }
+            return { name, failed: true as const }
           }
         }),
       )
 
       const newReplacements = new Map(replacements.value)
-      for (const { name, replacement } of results) {
+      for (const result of results) {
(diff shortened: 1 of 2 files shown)


app/composables/useCompareReplacements.ts
```diff
@@ -42,15 +42,18 @@ export function useCompareReplacements(packageNames: MaybeRefOrGetter<string[]>)
...
         }),
       )
 
       const newReplacements = new Map(replacements.value)
-      for (const { name, replacement } of results) {
+      for (const result of results) {
+        if (result.failed) continue
+
+        const { name, replacement } = result
         newReplacements.set(name, replacement)
```
ghostdevv:
```suggestion
        if (result.failed) continue
        newReplacements.set(result.name, result.replacement)
```

ghostdevv (dismissed):
cc: @gameroman

gameroman (approved):

serhalp (approved):
LGTM. Good catch, thanks for fixing!

Issue #776, opened 2026-02-02
[discussion][ux] add guidelines for a consistent ui/ux


dominikg:
Currently lots of features get added quickly where many of them look coherent in themselves but joined together they create an experience that is not the same.

As a starting point some of my thoughts when going through the UI

- spacing vs border vs (background-)color
- vertical and horizontal alignment, also on text baseline
- icon positioning (prefix or postfix, currently its a mix without a recognizable pattern)
- tooltips on numeric fields ( eg explain what xx / yy means on dependencies)
- link to internal routes where it makes sense (dependencies in ui and package.json instead only npm-graph)
- no hidden interactive ui (that only becomes visible on hover like the copy button on npm install box)
- prefer full words over abbreviations when space is available (DEPS -> Dependencies, VULNS -> Vulnerabilities)
- no surprises/consistency with user inputs. (eg typing on the homepage search box adds the first char as search query to history)
- when to use animations/transitions
- semantic html and a11y requirements

graphieros:
We are probably missing Base components to standardize the UI

garthdw:
Another one to add is where accent color should be used.

e.g. Some links on hover use accent color others don't.

jellydeck:
related #1071
common base button & link components

PR #2474, opened 2026-04-11
feat: add package name copy button with package manager dropdown icon on search view

### LINKED ISSUE #2341: [FEAT] Impl copy button with package manager dropdown icon's on search view
### Describe the feature

Implementing a copy button directly on the search tiles, here for example; https://npmx.dev/search?q=typescript+native 

saving two more clicks instead of having to click on the package then copy.

Image

### Additional information

### Final checks

akadotsh:
### 🔗 Linked issue
Closes https://github.com/npmx-dev/npmx.dev/issues/2341

### 🧭 Context
Adds a copy-to-clipboard feature for package names in the search results view.

### DIFF
--- a/app/components/Package/Card.vue
+++ b/app/components/Package/Card.vue
@@ -40,6 +40,12 @@ const pkgDescription = useMarkdown(() => ({
 }))
 
 const numberFormatter = useNumberFormatter()
+const packageName = computed(() => props.result.package?.name ?? '')
+
+const { copied: copiedPkgName, copy: copyPkgName } = useClipboard({
+  source: packageName,
+  copiedDuring: 2000,
+})
 </script>
 
 <template>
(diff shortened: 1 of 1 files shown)


ghostdevv (changes requested):
thanks for the PR! I think the copy button should be always visible to the right of the package name, perhaps just the icon - there was an issue or comment about this for the ones we have already but I can't find it (other than #776) - what do you think?

app/components/Package/Card.vue
```diff
@@ -49,14 +55,23 @@ const numberFormatter = useNumberFormatter()
         :is="headingLevel ?? 'h3'"
         class="font-mono text-sm sm:text-base font-medium text-fg group-hover:text-fg transition-colors duration-200 min-w-0 break-all"
       >
-        <NuxtLink
-          :to="packageRoute(result.package.name)"
-          :prefetch-on="prefetch ? 'visibility' : 'interaction'"
-          class="decoration-none after:content-[''] after:absolute after:inset-0"
```
alexdln, line 55: The logic that the entire card should be clickable seems not to work now. I think this is a more important feature on the search page than copying a name. 
The copy button on the side will probably fix this (but it will create an extra interaction)

Could you please share, what user case would this might solve? I feel like user know the name they're looking for in the search already and then open the needed package and interact there

PR #2476, opened 2026-04-11
feat: enable turning off code ligatures

Kiwow:
Closes #2423

Adds a new setting that controls whether ligatures are used in code. The setting sets a data attribute on the root, which is picked up by a selector in global styles.

I opted to have changes in the setting call the `toggleCodeLigatures` function rather than using `v-model`. This is needed because we need to run the side effect of updating the data attribute, which happens inside a `watch` is registered in `useCodeLigatures`. I still don't really like this, because the attribute update is dependent on `useCodeLigatures` being called, even though you can access and update the setting itself on `settings`. Other settings work like this as well though (e.g. the data attribute for `keyboardShortcuts` gets updated only if `useKeyboardShortcuts` is called), so for the time being, I've kept it this way. I'd be interested in revisiting this in the future, perhaps updating the `useSettings` composable to handle side effects on changes of individual settings. Please let me know if that makes sense to you.

I've provided translations for the new setting's label in English and Czech (my native language).

### DIFF
--- a/app/composables/useSettings.ts
+++ b/app/composables/useSettings.ts
@@ -257,7 +257,7 @@ export function useCodeContainer() {
   }
 }
 
-export function useCodeLigatures() {
+export const useCodeLigatures = createSharedComposable(function useCodeLigatures() {
   const { settings } = useSettings()
 
   const codeLigatures = computed(() => settings.value.codeLigatures)
@@ -285,4 +285,4 @@ export function useCodeLigatures() {
     codeLigatures,
     toggleCodeLigatures,
   }
-}
+})
--- a/app/composables/useSettings.ts
+++ b/app/composables/useSettings.ts
@@ -39,6 +39,8 @@ export interface AppSettings {
     autoOpenURL: boolean
   }
   codeContainerFull: boolean
+  /** Enable/disable ligatures in code */
+  codeLigatures: boolean
   sidebar: {
     collapsed: string[]
   }
@@ -65,6 +67,7 @@ const DEFAULT_SETTINGS: AppSettings = {
     autoOpenURL: false,
   },
   codeContainerFull: false,
+  codeLigatures: true,
   sidebar: {
     collapsed: [],
   },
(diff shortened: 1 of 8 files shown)


Kiwow:
I don't totally understand `createSharedComposable`, but after reading a bit of the [`effectScope` RFC](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0041-reactivity-effect-scope.md), I *think* it's a good thing to use here?

~~I'm happy to add it if anyone reviewing this thinks it makes sense to use it.~~

EDIT: After thinking about it a bit more, I think it makes sense to use it here, so I've done so

app/composables/useSettings.ts
```diff
@@ -253,3 +256,33 @@ export function useCodeContainer() {
     toggleCodeContainer,
   }
 }
+
+export const useCodeLigatures = createSharedComposable(function useCodeLigatures() {
```
ghostdevv, line 260:
```suggestion
export const useCodeLigatures = createSharedComposable(() => {
```

ghostdevv (approved):

app/composables/useSettings.ts
Kiwow: I saw this in the [`useKeyboardShortcuts` composable](https://github.com/npmx-dev/npmx.dev/blob/81179de6a7ee4dc3b180696a5fa0aee49b3d783a/app/composables/useSettings.ts#L116) and it made sense to me as a way to get nicer stack traces in logs. I think it's nice to have a named function for that reason.

To be fair, in prod builds, it won't make a difference, so I'm up for either solution. Your version is definitely more readable.

<details>
<summary>To compare, here's the outputs of `console.trace` (running in dev) when called at the start of the composable and the start of the watcher function, respectively:</summary>

Named function expression:

Two stack traces. Both contain 'useCodeLigatures' as part of the trace, as well as Vue-specific functions


Anonymous arrow function:

Two stack traces. Outside of Vue-specific functions, the traces only contain references to anonymous functions
</details>

ghostdevv: damn, ok we can keep it how you have it :eyes:

ghostdevv:
Will wait for @serhalp's review

ghostdevv (approved):

serhalp (approved):
thank you!

PR #2478, opened 2026-04-11
feat: add a badge type for package likes

### LINKED ISSUE #2412: like counter badge
### Describe the feature

It would be great if I could add a "like" counter badge using the official npmx badges instead of parsing `https://npmx.dev/api/social/likes/[pkg]`

### Additional information

### Final checks

t128n:
### 🔗 Linked issue

Resolves #2412

### 🧭 Context

In #2412, it was requested to have a badge variant that shows the likes a specific package has on npmx.dev.

This PR introduces a `likes`-type badge.

### 📚 Description

This PR adds a badge type called `likes` and integrates the logic of the `https://npmx.dev/api/social/likes/[pkg]` endpoint into it.

### DIFF
--- a/server/api/registry/badge/[type]/[...pkg].get.ts
+++ b/server/api/registry/badge/[type]/[...pkg].get.ts
@@ -496,6 +496,13 @@ const badgeStrategies = {
       color: isDeprecated ? COLORS.red : COLORS.green,
     }
   },
+
+  'likes': async (pkgData: globalThis.Packument) => {
+    const likesUtil = new PackageLikesUtils()
+    const { totalLikes } = await likesUtil.getLikes(pkgData.name)
+
+    return { label: 'likes', value: String(totalLikes ?? 10), color: COLORS.blue }
+  },
 }
 
 const BadgeTypeSchema = v.picklist(Object.keys(badgeStrategies) as [string, ...string[]])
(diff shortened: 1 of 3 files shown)


server/api/registry/badge/[type]/[...pkg].get.ts
```diff
@@ -496,6 +496,13 @@ const badgeStrategies = {
       color: isDeprecated ? COLORS.red : COLORS.green,
     }
   },
+
+  'likes': async (pkgData: globalThis.Packument) => {
+    const likesUtil = new PackageLikesUtils()
+    const { totalLikes } = await likesUtil.getLikes(pkgData.name)
+
+    return { label: 'likes', value: String(totalLikes ?? 0), color: COLORS.blue }
```
Comment (ghostdevv) [inline server/api/registry/badge/[type]/[...pkg].get.ts]: Maybe the default colour should be red? 👀 We could use the same colour that we do on the frontend

ghostdevv (changes requested):

server/api/registry/badge/[type]/[...pkg].get.ts
Comment (t128n) [reply server/api/registry/badge/[type]/[...pkg].get.ts]: Good idea - originally went with blue because of the likes itself are on stored via ATProtocol but red feels nice as well.

Went with the badge-default red

https://github.com/npmx-dev/npmx.dev/blob/5cb6bc6f66efec3da98aa9ab96323da208a7235e/server/api/registry/badge/%5Btype%5D/%5B...pkg%5D.get.ts#L36

instead of `text-red-500` (`oklch(63.7% 0.237 25.331)` / `#fb2c36`) which is used for the heart in the `LikeCard`

image

Comment (ghostdevv) [reply server/api/registry/badge/[type]/[...pkg].get.ts]: The only reason I suggested the frontend one was an attempt at distinguishing from other badges that use red to indicate a failure - but I think we could change this in a follow up if necessary

ghostdevv (approved):

PR #2482, opened 2026-04-12
fix(i18n): consistent casing for common.view_on en keys

ghostdevv:
### 📚 Description

I noticed the view on keys mostly started with a capital letter (`View on ...`) but for these two it was lowercase (`view on ...`)


alexdln (approved):

alexdln:
Good catch
_We need somehow check in prs that the case is always specified as normal, and then lowercase and uppercase are done through styles_

PR #2485, opened 2026-04-12
feat(i18n): add Chinese translations

WishMelz:
### 📚 Description
update Simplified Chinese and Traditional Chinese translations


serhalp (approved):
Thank you!

PR #2486, opened 2026-04-12
fix: improve anchors parsing in readmes

alexdln:
### 🧭 Context

Some packages (like nuxt) add anchors to headers themselves to control them in other services. This, coupled with the fact that we wrapped all header content in a link, caused hydration errors and wrong rendering

Added processing for a few common cases
* if it's a title-anchor, then the user's link is ignored and ours is added;
* if there was a link inside the title, our anchor will stand separately after;
* if the link is empty, then we will show only content text

_The class from the other file does not exist, but it caused warning with several thousand lines description and was terribly interfering_

### DIFF
--- a/server/utils/readme.ts
+++ b/server/utils/readme.ts
@@ -558,8 +558,8 @@ export async function renderReadmeHtml(
     return `<h${semanticLevel} id="${id}" data-level="${depth}"${preservedAttrs}><a href="#${id}">${displayHtml}</a></h${semanticLevel}>\n`
   }
 
+  const anchorTokenRegex = /^<a(\s.+)?\/?>$/
   renderer.heading = function ({ tokens, depth }: Tokens.Heading) {
-    const anchorTokenRegex = /^<a(\s.+)?\/?>$/
     const isAnchorHeading =
       anchorTokenRegex.test(tokens[0]?.raw ?? '') && tokens[tokens.length - 1]?.raw === '</a>'
 
--- a/server/utils/readme.ts
+++ b/server/utils/readme.ts
@@ -549,11 +549,23 @@ export async function renderReadmeHtml(
       toc.push({ text: plainText, id, depth })
     }
 
+    // The browser doesn't support anchors within anchors and automatically extracts them from each other,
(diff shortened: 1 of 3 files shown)


server/utils/readme.ts
```diff
@@ -549,11 +549,23 @@ export async function renderReadmeHtml(
       toc.push({ text: plainText, id, depth })
     }
 
+    // The browser doesn't support anchors within anchors and automatically extracts them from each other,
+    // causing a hydration error. To prevent this from happening in such cases, we use the anchor separately
+    if (htmlAnchorRe.test(displayHtml)) {
+      return `<h${semanticLevel} id="${id}" data-level="${depth}"${preservedAttrs}>${displayHtml}<a href="#${id}"></a></h${semanticLevel}>\n`
```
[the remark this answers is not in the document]
alexdln: It's for different type of blocks

serhalp, line 555: Could we just always do it this way rather than conditionally? 🤔

server/utils/readme.ts
```diff
@@ -549,11 +549,23 @@ export async function renderReadmeHtml(
...
+    }
+
     return `<h${semanticLevel} id="${id}" data-level="${depth}"${preservedAttrs}><a href="#${id}">${displayHtml}</a></h${semanticLevel}>\n`
   }
 
   renderer.heading = function ({ tokens, depth }: Tokens.Heading) {
-    const displayHtml = this.parser.parseInline(tokens)
+    const anchorTokenRegex = /^<a(\s.+)?\/?>$/
```
serhalp: perf nit: we could define this constant regex outside this `renderer.heading` so it doesn't get recompiled on every heading call.

serhalp (commented):
would it be possible to add some test coverage for this? it's pretty complex parsing logic!

server/utils/readme.ts
alexdln: Probably more UX question - is it appropriate for us to only be able to click the anchor by hovering over that side area, not the entire header? I didn't want to break the current UX, so I just added a condition for this rare, specific case with anchor inside heading

serhalp (approved):
🚀

PR #2199, opened 2026-03-22
fix(ui): reduce badge width overestimation in fallback text measurement

MathurAditya724:
### 🔗 Linked issue

Closes #2187

### 🧭 Context

Badge widths looked too wide in production for long values (for example https://main.npmx.dev/api/registry/badge/engines/vitest), while local dev looked tighter.

The root cause is that when exact canvas measurement is unavailable, the fallback used a fixed per-character width, which overestimates strings containing many narrow characters (spaces, separators, punctuation).

### 📚 Description

This updates the badge text-width fallback logic to be less aggressive and more realistic, replace the fixed-width fallback with a small character-class estimator

PR #2487, opened 2026-04-12
feat: improve badge width estimation

### LINKED ISSUE #1601: Measuring text for the purpose of rendering badges fails on Vercel
### Describe the bug

On local machine, `/api/registry/badge/version/react-pdf?label=wwwwwwwwwwww` or `iiiiiiiiiiii` most likely will render correctly:

Image

Image

On Vercel, measuring text with canvas is crashing, resulting in inaccurate fallback measurements:

Image

Image

### Additional context

_No response_

### Logs

```shell-script
No logs, which is exactly why it's so hard to debug for me personally
```

t128n:
### 🔗 Linked issue

Resolves #1601

### 🧭 Context

In production, measuring text with canvas is crashing, resulting in inaccurate fallback measurements:

overflowing badge

Introduced a character lookup table of "manually" measured character widths for more
accurate lookups.

### 📚 Description

Tweaked the `estimateTextWidth` function to use a on-character level lookup table for more accurate results when the Canvas API from `@napi-rs/canvas` is unavailable.
Lookup table derived from running

```ts
// ...
const ctx = createCanvas(1, 1).getContext('2d')
const chars = ' !"#$%&\'()*+,-./:;<=>?@[\\]^_`{|}~0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz'
ctx.font = BADGE_FONT_SHORTHAND
const entries = [...chars].map(ch => `'${ch === "'" ? "\\'" : ch === '\\' ? '\\\\' : ch}': ${Math.ceil(ctx.measureText(ch).width)}`)
console.log('default: {\n  ' + entries.join(', ') + '\n}')
```

locally, where fonts and the API are available.

Using a generous fallback for characters that aren't in the map, such as emojis.

#### Overflow / Long Labels

| Case | Prod | PR |
|------|------|----|
| Long label |  |  |
| Long value |  |  |
| Long package name |  |  |

#### Shields.io Style (`?style=shieldsio`)

| Case | Prod | PR |
|------|------|----|
| Normal |  |  |
| Long label |  |  |
| Emoji label |  |  |

#### Emojis

| Case | Prod | PR |
|------|------|----|
| Emoji in label |  |  |
| Emoji in value |  |  |
| Multiple emojis |  |  |
| Emoji-only label |  |  |

#### CJK Characters

| Case | Prod | PR |
|------|------|----|
| Chinese label |  |  |
| Japanese label |  |  |
| Korean label |  |  |
| Mixed CJK + ASCII |  |  |

#### Special & Punctuation Characters

| Case | Prod | PR |
|------|------|----|
| Symbols |  |  |
| Narrow chars (`iiiii`) |  |  |
| Wide chars (`WWWWW`) |  |  |
| Mixed widths |  |  |

#### Badge Types Smoke Test

| Type | Prod | PR |
|------|------|----|
| `version` |  |  |
| `license` |  |  |
| `downloads` |  |  |
| `size` |  |  |
| `types` |  |  |
| `deprecated` |  |  |


ghostdevv (approved):
this doesn't break on the last test case from #2199 so we can yolo this :p

ghostdevv (changes requested):
actually, rq why is it "when canvas not available"? @t128n

t128n:
> actually, rq why is it "when canvas not available"? @t128n 

Text is usually measured using createCanvasContext. On Vercel, createCanvasContext can’t be created (returns null, likely due to missing native node.js addons in the serveless runtime), so it falls back to estimateTextWidth. estimateTextWidth makes a best guess for characters’ widths and has edge cases with wider characters like `w`.
That's the whole context.

So Canvas not available was referring to the createCanvasContext (or what its name is - currently on mobile and it's hard to look outside the PRs boundary to find the imports).

ghostdevv:
> > actually, rq why is it "when canvas not available"? @t128n
> 
> Text is usually measured using createCanvasContext. On Vercel, createCanvasContext can’t be created (returns null, likely due to missing native node.js addons in the serveless runtime), so it falls back to estimateTextWidth. estimateTextWidth makes a best guess for characters’ widths and has edge cases with wider characters like `w`. That's the whole context.
> 
> So Canvas not available was referring to the createCanvasContext (or what its name is - currently on mobile and it's hard to look outside the PRs boundary to find the imports).

ah but that's always been true right? like that fact hasn't changed in this PR?

also that's why we have the napi-rs canvas? or no?

t128n:
Yeah, my assumptions were flawed - I thought it would be about the whole canvas not being available, but I made some further tests real quick.

So, `@napi-rs/canvas` & `canvasContext` generally are available, but if I'm not missing anything major, it's producing 0 widths because Geist as a font is not available on the Vercel serverless function (my assumption based on the testing).

```ts
function measureTextWidth(text: string, font: string): number | null {
  const context = getCanvasContext()

  if (context) {
    context.font = font

    const measuredWidth = context.measureText(text).width

    if (Number.isFinite(measuredWidth) && measuredWidth > 0) { // Returns false as font not available and nothing can be measured
      return Math.ceil(measuredWidth)
    }
  }

  return null // yields null for canvas context and then falls back to estimateTextWidth 
}
```

So it's rather: `use pre-measured widths for a lookup table` rather than `use table for badges when canvas not available`

wojtekmaj:
> ah but that's always been true right? like that fact hasn't changed in this PR?

Yea, it is my flawed implementation that I wasn't able to debug because I didn't know what errors were happening on Vercel, and it is not reproducible locally at all.

ghostdevv:
I think we can create an issue to try and load the font and see what the perf is like on that, if we can get it working that is. If there are Vercel issues then we can help with debugging that this time around as we should now have access to logs

serhalp:
@ghostdevv what's the status of this PR? I see you enabled auto-merge but you still have changes requested 🤔🤔🤔

t128n:
> @ghostdevv what's the status of this PR? I see you enabled auto-merge but you still have changes requested 🤔🤔🤔

There was #2587 created to fix the actual bug (Geist font not available, and therefore the endpoint falling back to the measuring algorithm) instead of working around it by improving the text-measurement algorithm

ghostdevv (approved):
~~actually, rq why is it "when canvas not available"? @t128n~~ sorry this is what was in the box when I went to approve - GitHub fully functioning software! xD I think this should be good and we can do the follow up later...

PR #2488, opened 2026-04-12
fix: encode username in constructed gravatar URL

tomaioo:
## Summary

Security: Unencoded username in API route construction

## Problem

**Severity**: `Medium` | **File**: `app/components/User/Avatar.vue:L30`

The username is inserted directly into `/api/gravatar/${props.username}`. A crafted username containing path separators or reserved URL characters can change the effective request path and potentially hit unintended endpoints.

## Solution

Wrap the username with `encodeURIComponent(props.username)` when constructing the URL path.

## Changes

- `app/components/User/Avatar.vue` (modified)

### DIFF
--- a/app/components/User/Avatar.vue
+++ b/app/components/User/Avatar.vue
@@ -31,7 +31,7 @@ const textClass = computed(() => {
   }
 })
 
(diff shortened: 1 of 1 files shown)


serhalp (approved):
Good catch, thank you!

PR #2472, opened 2026-04-11
feat: replace quadrant with scatter with selectable axes

graphieros:
Resolves #2449 

This replaces the controversial quadrant chart with a scatter plot, where 2 facets can be compared.
No weights, no debates.

- 2 selects to choose facets.
- a same facet can be chosen for both axes (diagonal sorting)
- usual actions in the chart menu (exports, alt copy, annotator)
- bump vue-data-ui to 3.17.13 ([release notes](https://github.com/graphieros/vue-data-ui/releases/tag/v3.17.13))
- also resolves #2106 

| Dark mode| Light mode|
|-----------|------------|
|image|image|

- Hovered datapoint (or accessed via keyboard):
image

PR #2493, opened 2026-04-12
fix: improve compare scatter chart legend

graphieros:
Follow up to #2472 

This improves the legend of the scatter chart on the compare page, by positioning the items on the side, below the select inputs.
On mobile, the previous behavior is maintained.

image

One downside: the legend is not part of the png export, however since the chart bears labels above datapoints, we can live with that.

### DIFF
--- a/app/components/Compare/FacetScatterChart.vue
+++ b/app/components/Compare/FacetScatterChart.vue
@@ -370,7 +370,11 @@ onMounted(async () => {
 
             <!-- Custom legend -->
             <template #legend="{ legend }" v-if="readyTeleport">
-              <div id="compare-scatter-legend-inner"></div>
+              <div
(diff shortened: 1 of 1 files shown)


alexdln (approved):
It looks great. And those effects of displaying the exact values ​​when hovering over an element are just amazing!

graphieros:
> It looks great. And those effects of displaying the exact values ​​when hovering over an element are just amazing!

Thank you ^^

Want I would like eventually is to also have the same datapoint highlighting when hovering | focusing legend items, but this requires a non-trivial update of the lib (adding exposed `onEnter` & `onLeave` in the legend scoped slot, which... would need to be applied to all components with legends). It's now in my roadmap (https://github.com/graphieros/vue-data-ui/issues/326) ;)

PR #2494, opened 2026-04-12
fix(i18n): consistent common.view_on ua and ru keys

romansp:
### 📚 Description

Similar to #2482 updated `common.view_on` keys for ua and ru for consistency.


ghostdevv (approved):

ghostdevv:
ty!

PR #2496, opened 2026-04-12
docs(ui): add stories for Recharging page

cylewaitforit:
### 🔗 Linked issue

#2150 

### 🧭 Context

### 📚 Description

Adds stories for Recharging page.

Adds [MSW](https://mswjs.io/) and [MSW storybook addon](https://github.com/mswjs/msw-storybook-addon?tab=readme-ov-file) for mocking API calls on pages in Storybook. (Can be broken into separate PR if this feels too large.)


cylewaitforit:
@coderabbitai review

pnpm-workspace.yaml
```diff
@@ -41,10 +41,14 @@ savePrefix: ''
 shellEmulator: true
 
 catalogs:
+  msw:
+    'msw': '^2.13.2'
```
ghostdevv, line 45: how come this has it's own catalog?

.storybook/.public/mockServiceWorker.js
```diff
@@ -0,0 +1,335 @@
+/* eslint-disable */
```
ghostdevv, line 1: is this file generated or vendored?

.storybook/handlers.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export const repoStatsHandler = http.get('/api/repo-stats', () => {
+  return HttpResponse.json({
+    contributors: 123,
+    commits: 1234,
+    pullRequests: 1234,
+  })
+})
+
+export const contributorsHandler = http.get('/api/contributors', () => {
```
ghostdevv, line 11: Does this work without mocking?

ghostdevv (changes requested):

pnpm-workspace.yaml
cylewaitforit: Fair question. Since MSW has other uses and isn’t specifically tied to the Storybook dependency update cycle I figured it would be better out of the Storybook catalog.

Maybe that’s a premature move. Since right now I was only adding the MSW to the `.storybook/public/`.

If there is preference that it’s in its moved into the Storybook catalog, moved to be part of the default catalog, or not defined in catalog at all I’d be fine with whatever. I’ve only used catalogs for a short time. So there’s probably best practices I haven’t fully internalized yet.

.storybook/.public/mockServiceWorker.js
cylewaitforit: It’s initialized with `npx msw init`.

https://mswjs.io/docs/cli/init

.storybook/handlers.ts
cylewaitforit: You can see in the team section on the about page in Storybook currently that without mocking it doesn’t get the api response. 

https://storybook.npmx.dev/?path=/story/pages-about--default

These two pages aren’t all that heavy on API content so they fallback pretty gracefully when it’s not available. 

I think having this API mocking layer in place will be even more important for some of the data heavy pages.

pnpm-workspace.yaml
ghostdevv: we can leave as is for now! if we want later we can always merge it into storybook/general one :pray: I only asked as it's currently on it's own

ghostdevv (approved):

PR #2505, opened 2026-04-13
fix: source profile data from npm

### LINKED ISSUE #2504: `~user` profile pages show packages by repo owner, not npm maintainer
# The Problem

When I visit a user profile I expect to see a list of the packages that npm user is a maintainer of.

However, when using the (default) Algolia search provider the profile page lists any package whose git repository is owned by a github/gitlab/bitbucket user or org with that same name. 

**It is not actually showing packages for which said npm user is a maintainer!**

That's the root cause of #1882, #1888, and the source of confusion in #2057

[Here is a PoC](https://npmx.dev/~patak) showing a spoofed package on @patak-cat's npm user profile, it's trivial to spoof packages and get them onto a user's profile page (sorry @patak-cat!)

Image

This also results in anemic download stats on user profile pages. 
* [~jdalton](https://npmx.dev/~jdalton) with 11k weekly downloads (should be ~1 Billion/week)
* my boy [~ulisesgascon](https://npmx.dev/~ulisesgascon) w/ 794 weekly downloads (should be ~3 Billion/week)
* me with 286 weekly [~jonchurch](https://npmx.dev/~jonchurch) (should be 150 million/week)
## The Cause

The user profile page uses Algolia [npm-search](https://github.com/algolia/npm-search) to filter to `owner.name:username` . Algolia's `owner.name` field  is documented in their indexer as being the ["GitHub owner or npm owner"](https://github.com/algolia/npm-search/blob/4736244247b63a6344e517bf0a79411155a5e51a/README.md?plain=1#L111-L116) 
  
The indexer's [`getOwner`](https://github.com/algolia/npm-search/blob/4736244247b63a6344e517bf0a79411155a5e51a/src/formatPkg.ts#L316-L358) checks `repository.user` before anything npm related:

```ts
if (repository?.user) {
  if (repository.host === 'github.com')    return { name: repository.user, ... }
  if (repository.host === 'gitlab.com')    return { name: repository.user, ... }
  if (repository.host === 'bitbucket.org') return { name: repository.user, ... }
}
if (lastPublisher) return lastPublisher
return author || null
```

When you see correct data on a user's profile page, it's either because they happen to have a repo field w/ the same github username as their npm username, or there is no repo field AND said user was the last person to publish the package.

The caller's comment states the intent explicitly:
[`// always favor the repository owner`](https://github.com/algolia/npm-search/blob/4736244247b63a6344e517bf0a79411155a5e51a/src/formatPkg.ts#L135).

## The Fix

There are two different approaches IMO
### option 1: code reuse

Ironically, the org page works as I'd expect the user page to. It pulls its information from npm and doesnt have the above "bug".

So the first fix would be to just reuse the org logic, which fetches from npm first and then enriches w/ algolia metadata. 
~Opened PR #2505 to do so~ (superseded by #2513)

That's the one we can land today to at least fix the problem of user pages sourcing its package list from a non npm identity.

### option 2: fix upstream
Edit: The PR https://github.com/npmx-dev/npmx.dev/pull/2513

I _live_ for one line fixes though and we have two we can pursue here for perhaps the "correct" fix if the intent from npmx is to use algolia over the npm api for this. That requires fixing upstream, then tweaking the query used in this codebase.

Idk how likely a fix is to land (can someone cc whatever contact yall have on the npm-search index?) but we could have `owners.name` be facetable so we can filter on it. 

I'll let algolia folks tell me if that would be expensive in query time and or dollars for their index (likely need to reindex too? idk how algolia works). But the PR to do that is here: https://github.com/algolia/npm-search/pull/1357

If that's upstreamed, its another one line change to the filter on the user profile page would fix thing:

```diff
-filters: `owner.name:${username}`
+filters: `owners.name:${username}`
```
[here](https://github.com/npmx-dev/npmx.dev/blob/f5bb3ed2f77fba0485a7c5dc732343b1dd400e11/app/composables/npm/useAlgoliaSearch.ts#L188) and [here](https://github.com/npmx-dev/npmx.dev/blob/f5bb3ed2f77fba0485a7c5dc732343b1dd400e11/app/composables/npm/useAlgoliaSearch.ts#L188) in `useAlgoliaSearch.ts`

jonchurch:
fixes #2504 

This PR reuses the org code to power the `useUserPackages.ts` hook.

Why? well the endpoint to get packages for orgs also works for users, by design. So we can reuse the API w/o any loss of accuracy. 

~It's a dirty lil hack rn, just reuse that same hook. I spent enough time root causing this bug, so this is the short term fix that I prefer and prefer to have reviewed.~ Seeing the red checks made me come back and try to do this slightly cleaner lol. Now im jsut duplicatng the logic of the org hook whereas before I was using the org hook itself. It's messy.

There is an existing bug btw on the Org pages, which is present in this PR as well. The org logic is limited to fetching 1k packages at once from npm, but there is no limit or pagination currently. See https://www.npmx.dev/~sindresorhus , on an empty cache that makes requests for every single one of his 1690 packages to fetch metadata as a fallback after the 400. Issue for that is #2507

## Before 📸 
Screenshot 2026-04-13 at 5 40 06 PM

## After 📸 
Screenshot 2026-04-13 at 5 39 58 PM

### DIFF
--- a/app/composables/npm/useUserPackages.ts
+++ b/app/composables/npm/useUserPackages.ts
@@ -1,22 +1,9 @@
-/** Default page size for incremental loading (npm registry path) */
-const PAGE_SIZE = 50 as const
-
-/** npm search API practical limit for maintainer queries */
-const MAX_RESULTS = 250
-
 /**
- * Fetch packages for a given npm user/maintainer.
+ * Fetch all packages for a given npm user.
  *
- * The composable handles all loading strategy internally based on the active
- * search provider. Consumers get a uniform interface regardless of provider:
- *
- * - **Algolia**: Fetches all packages at once via `owner.name` filter (fast).
- * - **npm**: Incrementally paginates through `maintainer:` search results.
- *
- * @example
- * ```ts
- * const { data, status, hasMore, isLoadingMore, loadMore } = useUserPackages(username)
- * ```
+ * Mirrors {@link useOrgPackages} — both use the same npm registry endpoint
+ * (`/-/org/<name>/package`) which accepts usernames and org names alike.
+ * The only difference: unknown users return an empty list instead of a 404.
  */
 export function useUserPackages(username: MaybeRefOrGetter<string>) {
   const route = useRoute()
@@ -26,24 +13,7 @@ export function useUserPackages(username: MaybeRefOrGetter<string>) {
     if (p === 'npm' || searchProvider.value === 'npm') return 'npm'
     return 'algolia'
   })
-  // this is only used in npm path, but we need to extract it when the composable runs
-  const { $npmRegistry } = useNuxtApp()
-  const { searchByOwner } = useAlgoliaSearch()
-
-  // --- Incremental loading state (npm path) ---
-  const currentPage = shallowRef(1)
-
-  /** Tracks which provider actually served the current data (may differ from
-   *  searchProvider when Algolia returns empty and we fall through to npm) */
-  const activeProvider = shallowRef<'npm' | 'algolia'>(searchProviderValue.value)
-
-  const cache = shallowRef<{
-    username: string
-    objects: NpmSearchResult[]
-    total: number
-  } | null>(null)
-
-  const isLoadingMore = shallowRef(false)
+  const { getPackagesByName } = useAlgoliaSearch()
 
   const asyncData = useLazyAsyncData(
     () => `user-packages:${searchProviderValue.value}:${toValue(username)}`,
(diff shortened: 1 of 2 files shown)


jonchurch (commented):
There's actually a scroll bug right now on the preview that Im investigating (I ripped out pagination bc the org approach does not use it, but left in the scroll based updates for paging without thinking about it)

https://npmxdev-git-fork-jonchurch-fix-user-profile-package-93cb4c-npmx.vercel.app/~qwerzl

Edit: fixed it in https://github.com/npmx-dev/npmx.dev/pull/2505/changes/5069b1325e2ec4680655014f17b369e3c41f5e3b

jonchurch:
Closing this as https://github.com/algolia/npm-search/pull/1357 was merged so the correct fix will not look like this anymore!

PR #2512, opened 2026-04-14
fix: compare scatter chart scale labels overlapping on first load

graphieros:
Follow up to #2472 

The charts of the compare page (bars and scatter chart) are loaded inside a hidden tab (v-show).

The scatter chart component internally autosizes on mounted to adapt the chart area and avoid overlaps between the Y-axis label and its scale labels. However, since the component mounts in a hidden tab, the auto-sizing is done on empty dimensions, leading to overlaps on the first load.

This fixes the issue by triggering a re-render of the scatter component when the charts tab is activated, enabling the autosizing to function in proper conditions.

**Edit: changing the tab display condition using `v-if` instead of `v-show` solves the issue, without coupling the chart component to the tab state through props.**

Test:
- load the compare page with the table tab active
- switch to the charts tab: there should be no label overlapping on Y

### DIFF
--- a/app/components/Compare/FacetScatterChart.vue
+++ b/app/components/Compare/FacetScatterChart.vue
@@ -11,6 +11,7 @@ import('vue-data-ui/style.css')
 const props = defineProps<{
   packagesData: ReadonlyArray<PackageComparisonData | null>
   packages: string[]
+  isActiveTab: boolean
 }>()
 
 const colorMode = useColorMode()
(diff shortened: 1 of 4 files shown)


43081j:
hmm if its costly to switch to `v-if`, could it be better to just use an intersection observer in `onMount` instead?

instead of passing active tab state around, check if self (the chart) is intersecting (i.e. is visible) and re-render if it is.

its a bit more effort but this PR currently couples the chart and the outer component more than it should be imo

graphieros:
> hmm if its costly to switch to `v-if`, could it be better to just use an intersection observer in `onMount` instead?
> 
> instead of passing active tab state around, check if self (the chart) is intersecting (i.e. is visible) and re-render if it is.
> 
> its a bit more effort but this PR currently couples the chart and the outer component more than it should be imo

The intersection observer would trigger re-renders every time the chart comes into view, which is overkill. But I agree about the coupling.
I'll see if I can just change the tabs component to use v-if, hopefully it won't cause any regression^^

43081j:
only from "not intersecting" to "intersecting", i.e. keep a flag that changes when we see `0`

but yeah i think `v-if` is the nicest solution if it doesn't cause poor UX

graphieros:
@43081j 

Turns out just changing `v-show` to `v-if `does the job nicely.
Thank you for keeping me honest, I tend to overcomplicate sometimes :)

43081j (approved):

PR #2520, opened 2026-04-14
feat: prefill package manager from query param

### LINKED ISSUE #2515: Let other websites and tools define the default package manager
### Describe the feature

pnpm v11 will start using npmx.dev for `pnpm docs` and `pnpm home` commads (see [post](https://bsky.app/profile/pnpm.io/post/3mjhihoyjwc2w)). @zkochan had [the idea](https://discord.com/channels/1464542801676206113/1466869459590451448/1493608143769440507) to have a URL parameter that tell the user came from a given package manager, so npmx automatically show the right pm cli commands (for example, for get started code blocks).

This would also be useful for docs that already default to a different package manager (like https://vite.dev, that defaults to pnpm). So they can define their preferred default when linking to pnpmx.

For usage from the pnpm cli or pnpm docs, we could use `?pm=pnpm` for this. This should override the user choice (hopefully it is usually the same).

From docs (like vite), we could think of having a `?dpm=pnpm` (default package manager). But in the vite docs, we also let the user chose their preferred pm, and we could change the links to match the user selection, though. So in that case, the `dpm` isn't really needed. We could start with `?pm` only and see if there are real issues reported with it before adding more complexity.

### Additional information

### Final checks

alexdln:
### 🔗 Linked issue

Resolves #2515

### 🧭 Context

Third-party package managers or individual services may find it useful to link to us so that a specific package manager is immediately configured.

Added support for selecting and saving a package manager based on the query param (`?pm=pnpm`)

### DIFF
--- a/app/composables/useSelectedPackageManager.ts
+++ b/app/composables/useSelectedPackageManager.ts
@@ -12,6 +12,10 @@ export const useSelectedPackageManager = createSharedComposable(
 
     // Sync to data-pm attribute on the client
     if (import.meta.client) {
+      const queryPM = new URLSearchParams(window.location.search).get('pm')
+      if (queryPM && packageManagers.some(pm => pm.id === queryPM)) {
+        pm.value = queryPM as PackageManagerId
+      }
       // Watch for changes and update the attribute
       watch(
         pm,
--- a/app/utils/prehydrate.ts
+++ b/app/utils/prehydrate.ts
@@ -42,7 +42,7 @@ export function initPreferencesOnPrehydrate() {
     let pm = 'npm'
 
     // Support package manager preference in query string (for example, ?pm=pnpm)
-    const queryPM = new URLSearchParams(document.location.search).get('pm')
+    const queryPM = new URLSearchParams(window.location.search).get('pm')
     if (queryPM && validPMs.has(queryPM)) {
       pm = queryPM
       localStorage.setItem('npmx-pm', pm)
--- a/app/utils/prehydrate.ts
+++ b/app/utils/prehydrate.ts
@@ -39,20 +39,28 @@ export function initPreferencesOnPrehydrate() {
       document.documentElement.dataset.bgTheme = preferredBackgroundTheme
     }
 
-    // Read and apply package manager preference
-    const storedPM = localStorage.getItem('npmx-pm')
-    // Parse the stored value (it's stored as a JSON string by useLocalStorage)
     let pm = 'npm'
-    if (storedPM) {
-      try {
-        const parsed = JSON.parse(storedPM)
-        if (validPMs.has(parsed)) {
-          pm = parsed
-        }
-      } catch {
-        // If parsing fails, check if it's a plain string (legacy format)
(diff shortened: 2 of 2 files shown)


mbtools:
Nice feature! Thanks for implementing @alexdln. 

Can we also do `?pm=none` and hide the complete "Get Started" section? 🙏 

PS: The "Download" button in "Get Started" is misplaced. It would be much better in the link section under the description.

alexdln:
> Can we also do ?pm=none and hide the complete "Get Started" section? 🙏

@mbtools I think this could be part of a larger interface customization project (section availability, order, customizations). For now, I don't think it's a good issue for it and to hide it only through a parameter (how to show it again later, how it should work like for users, how to avoid cls, etc.).

> PS: The "Download" button in "Get Started" is misplaced. It would be much better in the link section under the description.

Hmm, I'm not sure. This is the section that describes how to install the package. From that perspective, I think the option to download the package as `tgz` pretty makes sense

---

I think you can create issues and we'll try to collect the thoughts of other participants before making the right decision

patak-cat:
I agree with @alexdln, hiding the getting started section is orthogonal to what PM you are using. This PM could later control other sections, too.

@alexdln it seems that when you have ?pm=yarn, then that is remembered. I think we should only store in local storage a user selection, not one that was set with a URL. This was at least surprising to me. User of vite docs may end up with their default preference overriden. There are some interesting dynamics if the preferred pm is diff in different websites and one wants to override the other but let's hope that isn't the case. I think for now, the ?pm={pm} url should only set the PM temporarily and not store it as a preferrence.

ghostdevv:
> I think for now, the ?pm={pm} url should only set the PM temporarily and not store it as a preferrence.

Or we could store it if you haven't set your own preference, and if you have it's only temporary

patak-cat:
> Or we could store it if you haven't set your own preference, and if you have it's only temporary

If someone visit a link from vite, it may get pnpm stored if it is the first time visiting npmx. That feels strange to me. It would be more natural that the preference is only store when the user interact with the dropdown (or later on in settings).

alexdln:
@patak-cat I thought about it. But let's say a user visit a package with the pnpm parameter. And from there, they navigate to types. It would be wrong if they suddenly had npm there...

patak-cat:
> @patak-cat I thought about it. But let's say a user visit a package with the pnpm parameter. And from there, they navigate to types. It would be wrong if they suddenly had npm there...

Good point, should the `pm` query param be sticky when doing internal navigation? We can merge it as is and later improve it if you think changing the preference won't be an issue for users.

patak-cat (approved):

patak-cat:
Let's try it out, and we can iterate on it as we start to get this feature used by others.

alexdln:
There is an option to distribute this parameter to all links (but this is a heavy load), we can also make an additional entry in sessionStorage (but now, with a permanently saved session, it is approximately the same as localStorage...)

So yeah, let's give it a try for now. In general, I think if a service clearly recommends some pm and users clearly come to us with that pm in params, it's likely that they're comfortable with that pm

PR #2524, opened 2026-04-14
fix(docs): allow hyphens in fenced codeblock language (#2437)

### LINKED ISSUE #2437: docs feature codeblock hyphenated lang regex
**Handle hyphenated fenced languages in `@example` blocks.**

These regexes only consume `\w`, so a valid fence like ```` ```glimmer-ts ```` is parsed as language `glimmer` and leaves `-ts` at the start of the code body. That breaks example rendering for some Shiki languages now loaded in `server/utils/shiki.ts`, instead of falling back cleanly.

_Originally posted by @coderabbitai[bot] in https://github.com/npmx-dev/npmx.dev/pull/2382#discussion_r3042830927_

MukundaKatta:
## Summary

Fixes #2437.

The fenced-codeblock language regexes in `server/utils/docs/render.ts` and `server/utils/docs/text.ts` matched `\w`, which doesn't include `-`. For Shiki languages with hyphens (`glimmer-ts`, `vue-html`, `objective-c`, etc.), the `-suffix` was dropped from the captured language and leaked into the top of the code body — the code block rendered with broken highlighting and a stray `-ts` at the start.

Swaps `\w` for `[\w-]` in the three affected regexes:

- `render.ts` — the two regexes that extract `lang` and strip the fence from `example.doc`
- `text.ts` — the fenced-block extraction in `renderMarkdown`

## Test plan

- [x] Added regression test in `test/unit/server/utils/docs/text.spec.ts` for `glimmer-ts` covering both sides of the bug: the `-ts` suffix no longer leaks into the body, and highlighting kicks in.
- [ ] CI (lint + existing test suite)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/server/utils/docs/render.ts
+++ b/server/utils/docs/render.ts
@@ -191,9 +191,9 @@ async function renderJsDocTags(tags: JsDocTag[], symbolLookup: SymbolLookup): Pr
     : null
   const examplePromises = examples.map(async example => {
     if (!example.doc) return ''
-    const langMatch = example.doc.match(/```(\w+)?/)
+    const langMatch = example.doc.match(/```([\w-]+)?/)
     const lang = langMatch?.[1] || 'typescript'
-    const code = example.doc.replace(/```\w*\n?/g, '').trim()
+    const code = example.doc.replace(/```[\w-]*\n?/g, '').trim()
     return highlightCodeBlock(code, lang)
   })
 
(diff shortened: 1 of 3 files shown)


MatteoGabriele:
Sorry, but we do not accept contributions that are entirely AI-generated. Please review the project rules of conduct before you start contributing.

PR #2525, opened 2026-04-14
fix(i18n): update language switch immediately without refresh

BittuBarnwal7479:
### 🔗 Linked issue

#2526 

### 🧭 Context

When switching language from Settings, the first change didn’t fully apply unless the page was refreshed. 
This PR fixes both so the UI updates immediately and stays consistent.

### 📚 Description

I updated the language-change flow in Settings so locale updates are applied through the same path every time, including the first switch. That removes the partial-translation state and the refresh workaround.

I tested this manually by:

switching locales multiple times from the Settings page
checking header icons immediately after each switch
confirming shortcuts still navigate correctly

Before: 

https://github.com/user-attachments/assets/ee0adf00-24e6-44bb-9bd9-bee58a122e27

After:

https://github.com/user-attachments/assets/99ba50f4-ec11-44fb-8dd2-289fb48327e7


romansp (changes requested):
Thanks for looking into that as this was bugging me too.

Let's keep this PR minimal though and fix language switch issue only.

Icons and shortcuts in `AppHeader` seem to be this way by design right now. I don't know the reason behind choosing `,` and `s` might actually be better. Anyways, I suggest to revert this part. Let's open a new issue if you think that's a problem and we address it there instead.

BittuBarnwal7479:
> Thanks for looking into that as this was bugging me too.
> 
> Let's keep this PR minimal though and fix language switch issue only.
> 
> Icons and shortcuts in `AppHeader` seem to be this way by design right now. I don't know the reason behind choosing `,` and `s` might actually be better. Anyways, I suggest to revert this part. Let's open a new issue if you think that's a problem and we address it there instead.

Thanks, makes sense. I’ll keep this PR minimal and revert the AppHeader icon changes so it only fixes the language switch issue.

BittuBarnwal7479:
@romansp please review.

romansp (approved):

romansp:
This looks good to me. Thanks @BittuBarnwal7479.

To whoever will be merging I suggest use squash strategy to avoid revert commits appearing on `main`.

app/pages/settings.vue
```diff
@@ -244,8 +249,8 @@ const setLocale: typeof setNuxti18nLocale = newLocale => {
                 <SelectField
                   id="language-select"
                   :items="locales.map(loc => ({ label: loc.name ?? '', value: loc.code }))"
-                  v-model="currentLocale"
-                  @update:modelValue="setLocale($event as typeof currentLocale)"
+                  :modelValue="currentLocale"
```
graphieros: currentLocale is evaluated as a locale string, not a string, hence the failing CI type test

graphieros (changes requested):
Thank you for your PR :)
Please address the failing types test

app/pages/settings.vue
```diff
@@ -8,13 +10,21 @@ const keyboardShortcutsEnabled = useKeyboardShortcuts()
 const { toggleCodeLigatures } = useCodeLigatures()
 
 // Create a computed property to handle locale binding properly
```
graphieros: This comment can be removed (or rephrased to propertly reflect the content)

app/pages/settings.vue
BittuBarnwal7479: Done

graphieros (approved):

PR #2545, opened 2026-04-16
docs(ui): add stories for Settings page

cylewaitforit:
### 🔗 Linked issue

#2150 

### 🧭 Context

### 📚 Description

     
Adds stories for Settings page.

Removes conditional logic for '@nuxtjs/color-mode' when using storybook and moves `__NUXT_COLOR_MODE__` mocking to the preview-head file. This both reduces the workarounds for storybook in the nuxt config as well as prevents the storybook from crashing when a theme was selected on the Settings page in the storybook.

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -15,7 +15,8 @@ export default defineNuxtConfig({
     '@vite-pwa/nuxt',
     '@vueuse/nuxt',
     '@nuxtjs/i18n',
-    ...(isStorybook ? [] : ['@nuxt/fonts', '@nuxtjs/color-mode']),
+    '@nuxtjs/color-mode',
+    ...(isStorybook ? [] : ['@nuxt/fonts']),
   ],
 
   $test: {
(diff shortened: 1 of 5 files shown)


nuxt.config.ts
```diff
@@ -15,7 +15,8 @@ export default defineNuxtConfig({
     '@vite-pwa/nuxt',
     '@vueuse/nuxt',
     '@nuxtjs/i18n',
-    ...(isStorybook ? [] : ['@nuxt/fonts', '@nuxtjs/color-mode']),
+    '@nuxtjs/color-mode',
+    ...(isStorybook ? [] : ['@nuxt/fonts']),
```
ghostdevv, line 19: Image

ghostdevv (approved):

nuxt.config.ts
cylewaitforit: We also have to load those in the preview-head file at the moment. So we have them,  just not from `@nuxt/fonts`.

Would be great to not have any special carve outs for storybook in the nuxt config eventually.

ghostdevv: 100%

PR #2546, opened 2026-04-16
fix: use fast-npm-meta in timeline

43081j:
### 🔗 Linked issue

N/A

### 🧭 Context

Timeline is currently awfully slow for packages with a large amount of versions.

### 📚 Description

We were fetching the whole packument which is _pretty big_.

So this switches to using fast-npm-meta which is basically just
trimmed down metadata.


graphieros (approved):
:)

ghostdevv (approved):

PR #2549, opened 2026-04-16
feat(ui): add Ctrl+N/P navigation in command palette

jcalixte:
Hi, I propose the shortcuts in the Command Palette `ctrl+n/p` for list navigation that can be a good fit for keyboard-heavy users for list navigation without affecting anyone else.

Disclaimer : I've primarly used Claude Code to generate this Pull Request.

## Summary

- Adds Emacs `Ctrl+N` (next) and `Ctrl+P` (previous) as alternative navigation shortcuts in the command palette, alongside the existing `ArrowDown`/`ArrowUp` keys
- Updates the `command_palette.instructions` i18n string to mention `Ctrl+N/P`

## Implementation

The two existing `ArrowDown`/`ArrowUp` conditions in `handleGlobalKeydown` were extended with an `||` clause — minimal change with no logic duplication. I reused the existing `getCommandElements()`, `focusCommand()`, and `focusInput()` helpers.


jhroemer:
@jcalixte I agree that the arrow-keys are awkward, but tab works well for me at least. Any reason this doesn't work for you? 

I'd personally be a bit hesitant about adding something emacs specific (I'm not an emacs user, but I understand `N/P` is emacs specific?), or I would then at least question if we should also, or instead, support vim-like `j` or `k` shortcuts. There are also browser extensions that allow some of these things (I use Vimium myself), and there might be a line to draw between what's supported by default and what's user specific. 

To be clear, not saying to do one or the other, just questioning whether adding emacs specific modifiers as a default is too opinionated.

jcalixte:
Hi @jhroemer, thanks for taking time for answering me!

I completely forgot that the tab / shift+tab was a possibility as I'm learning new ways to develop others than in VS Code, I may have been in a rabbit hole these past months 🫣

I agree with you that this feature may be too opinionated.

If you agree, I'll close this PR then.

Thank you!

PR #2554, opened 2026-04-16
fix: resolve readme copy functionality in Safari

### LINKED ISSUE #2151: "Copy as Markdown" button on README doesn't copy to clipboard in Safari

MatteoGabriele:
### 🔗 Linked issue

resolves #2151

### 🧭 Context

The copy Readme button doesn't work in Safari

### 📚 Description

Unfortunately, for security reasons, Safari doesn't allow the Clipboard API to be used with async operations, so it silently fails without copying anything.
I've created a custom useClipboardAsync that addresses the security issues and handles the async call.

### DIFF
--- a/.lighthouserc.cjs
+++ b/.lighthouserc.cjs
@@ -25,7 +25,7 @@ module.exports = {
   ci: {
     collect: {
       startServerCommand: 'pnpm preview',
-      startServerReadyPattern: 'Listening',
+      startServerReadyPattern: '(Listening|Local|:3000)',
       url: [
         'http://localhost:3000/',
         'http://localhost:3000/search?q=nuxt',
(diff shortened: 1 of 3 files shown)


knowler (approved):
Works! I remember running into this one time and using a similar solution.

graphieros (approved):
Beautiful 🌿

ghostdevv:
Ty for solving this! Did you play around with useClipboardItems from vueuse? I tried to get it to work but I couldn't 🫠

MatteoGabriele:
> Ty for solving this! Did you play around with useClipboardItems from vueuse? I tried to get it to work but I couldn't 🫠

I did, but I would crash on me, so I went the custom way. At least for the asynchronous operations. Perhaps opening an issue on vueuse could be an option as well.

PR #2556, opened 2026-04-17
fix: round chart numbers in `applyDataCorrection` for consistent rounding

ulrichstark:
### 🔗 Linked issue

There's not.
This is a pretty straightforward and tiny fix.

### 🧭 Context

The weekly download charts shows numbers with fraction when hovering over it and smoothing is enabled:

image

Here you can see the exact number of the screenshot coming in to the code location I'm changing with this PR:

image

### 📚 Description

Download numbers are now rounded up before showing them to the user.
I can also do the change in `applyDataCorrection` instead if that's preferred.

### DIFF
--- a/app/components/Chart/SplitSparkline.vue
+++ b/app/components/Chart/SplitSparkline.vue
@@ -31,9 +31,7 @@ const props = defineProps<{
 
 const { locale } = useI18n()
 const colorMode = useColorMode()
-const numberFormatter = useNumberFormatter({
-  maximumFractionDigits: 0,
-})
+const numberFormatter = useNumberFormatter()
 const resolvedMode = shallowRef<'light' | 'dark'>('light')
 const rootEl = shallowRef<HTMLElement | null>(null)
 const palette = getPalette('')
--- a/app/components/Package/TrendsChart.vue
+++ b/app/components/Package/TrendsChart.vue
@@ -1540,7 +1540,7 @@ const chartConfig = computed<VueUiXyConfig>(() => {
           const rows = items
             .map((d: Record<string, any>) => {
               const label = String(d?.name ?? '').trim()
-              const raw = Math.round(Number(d?.value ?? 0))
+              const raw = Number(d?.value ?? 0)
               const v = compactNumberFormatter.value.format(Number.isFinite(raw) ? raw : 0)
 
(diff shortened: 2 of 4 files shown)


graphieros:
@ulrichstark 

Thank you for your PR :)

This issue was already addressed in #2519 (sparkline) and #2550 (other charts).
However, I think the approach of rounding numbers directly @`applyDataCorrection` is a better one, if you want to rework your solution, I would be happy to merge it.

ulrichstark:
Hey @graphieros, thanks for the prompt review! Just reverted my first commit and made the change in `applyDataCorrection` by adding rounding (up) as last step. I hope this is what you meant. Enjoy your weekend!

graphieros (changes requested):
@ulrichstark 

Perfect.
Prior changes applied higher up need to be reverted / adapted, now that the rounding is rooted:

- TrendsChart.vue : https://github.com/npmx-dev/npmx.dev/pull/2550/changes : the rounding can be removed
- SplitSparkline.vue : the `maximumFractionDigits` param can be removed from the composable call
- WeeklyDownloadStats.vue: the `maximumFractionDigits` param can be removed from the composable call

graphieros (approved):
LGTM 🌿

ulrichstark:
> LGTM 🌿

Thanks for assisting on this change!

Does this mean the original issue way already fixed but not yet released to production? This would also explain that I failed to reproduce it locally, but could on production and therefore could only proof it using the devtools on the production site.

graphieros:
> Does this mean the original issue way already fixed but not yet released to production? This would also explain that I failed to reproduce it locally, but could on production and therefore could only proof it using the devtools on the production site.

Yes, it will be in production with the next release
In the meantime, as a contributor, I recommend you use https://main.npmx.dev/ as your daily driver :)

ulrichstark:
> > Does this mean the original issue way already fixed but not yet released to production? This would also explain that I failed to reproduce it locally, but could on production and therefore could only proof it using the devtools on the production site.
> 
> Yes, it will be in production with the next release In the meantime, as a contributor, I recommend you use https://main.npmx.dev/ as your daily driver :)

Ahhh perfect. Thank you very much once again!!

PR #2559, opened 2026-04-17
docs(ui): add stories for Translation Status page

cylewaitforit:
### 🔗 Linked issue

#2150 

### 🧭 Context

### 📚 Description

     
 Adds stories for Translation Status page. 
 
Moved `i18nStatusHandler` into its own file in a new handlers directory since the amount of mock data in that handler had grown considerably to support mocking this page.


app/pages/translation-status.stories.ts
```diff
@@ -0,0 +1,30 @@
...
+} satisfies Meta<typeof TranslationStatus>
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+/** `/lunaria/status.json` is intercepted by MSW. Showing a variety of completion level translation statuses for a subset of locales. */
+export const Default: Story = {}
+
+/** No API response — the fetch never succeeds so `fetchStatus` stays as `'pending'`. Shows skeleton blocks in the locale list and skeleton inlines in body text. */
+export const WithoutTranslationData: Story = {
```
cylewaitforit, line 24: When the translation data is unavailable or loading, the page uses ` new Date().toISOString()`. That will mean that the visual regression snapshot for this test will constantly have a different date on it. 

While we could probably mock `Date()` to make that less flaky for the visual regression tests, I wonder if the page should show something else when the data isn't present. Open to other's thoughts.

ghostdevv: For now we can mock the date, would you like to create an issue to discuss the change in? (or just jump straight to a PR if you have a clear idea, and think it won't require large discussion) - If you want to that is! :sweat_smile:

cylewaitforit: Opened #2589

cylewaitforit: I'll move this back to draft while that is in review.

cylewaitforit: Merged those changes back in.

ghostdevv (approved):

PR #2564, opened 2026-04-18
chore: increase operation per run in stale workflow

MatteoGabriele:
### 🧭 Context
Running 100 operations per run was likely too conservative considering the size of npmx, and we weren't able to process all the items we had.
Currently, we have 210 issues, 97 PRs, and comments to review, so we need to expand our quota.

### 📚 Description
Increase the operation-per-run to 500 and see if this is enough. We have 5000 API requests per hour, and this workflow runs once at 2 am, so it's still really safe.


graphieros (approved):
The syntax looks ok to me !

PR #2566, opened 2026-04-18
chore: remove stale bot

MatteoGabriele:
### 🧭 Context
We are removing it because it is considered too noisy.

### 📚 Description
Removes the workflow


MatteoGabriele:
I've also manually deleted all comments from the bot that I've found so far.

patak-cat (approved):

PR #2580, opened 2026-04-18
chore: add workflow for stale prs and issue type bug

MatteoGabriele:
### 🧭 Context
We need an automated system to manage stale issues and PRs to keep the project clean. ✨ 
We have decided to focus only on "Bug" issues, as they typically have a shorter lifespan, along with all PRs. These will be tagged and then automatically closed.

### 📚 Description
These changes add the built-in GitHub stale action to automate the entire stale-tagging process and auto-close.
Issues of type "Bug" and PRs will be tagged after 30 days of inactivity and then automatically closed 7 days later.


43081j (approved):

graphieros (approved):

patak-cat:
Leaving trace here of what I said in the chat, too, that my personal preference is to only do this for PRs (I would probably have 3 months instead of 1), and in the case of bugs only add the stale label but avoid the comment and auto-close.

But I'm not doing a lot in the repo side, so I'll let you decide what works best for you. Thanks for modifying the previous version @MatteoGabriele! It is great that features are left untouched 🙏🏼

MatteoGabriele:
@patak-cat We're not going to leave any comments at any stage: this is just a stale tag after 1 month, then it closes after another 7 days, and it's always possible to reopen, obviously. 
All that's needed to unstale is a comment or actual work being done. A healthy PR or an urgent bug cannot be left untouched for 37 days and still be considered relevant, at least for the time being. Later on, perhaps we could increase the days if we think we're slowing down.

PR #2586, opened 2026-04-19
perf: prevent duplicates in file tree sprite

ghostdevv:
### 🔗 Linked issue

n/a

### 🧭 Context

#1347

### 📚 Description

I noticed that there is duplicate icons in the file tree sprite, which doesn't seem to be intended - but I may be wrong. If I'm correct then this would shrink it currently from 576K to 260K :tada:


serhalp (approved):
LGTM, good find

It would be nice to add some perf benchmarks to CI eventually!

PR #2601, opened 2026-04-20
docs(ui): add stories for Error page

cylewaitforit:
### 🔗 Linked issue

#2150 

### 🧭 Context

### 📚 Description

     
Adds stories for Error page.

### DIFF
--- a/app/error.stories.ts
+++ b/app/error.stories.ts
@@ -1,0 +1,79 @@
+import Error from './error.vue'
+import type { Meta, StoryObj } from '@storybook-vue/nuxt'
+import type { NuxtError } from '#app'
+
+const meta = {
+  title: 'pages/error',
+  component: Error,
+  parameters: {
+    layout: 'fullscreen',
+  },
+} satisfies Meta<typeof Error>
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+/** Package, org, or page not found. */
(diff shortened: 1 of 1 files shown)


app/error.stories.ts
```diff
@@ -0,0 +1,79 @@
...
+  },
+}
+
+/** Error with a detailed message to test text wrapping and layout. */
+export const LongMessage: Story = {
+  args: {
+    error: {
+      status: 500,
+      message:
+        'An unexpected error occurred while processing your request. Our team has been notified and is working to resolve the issue. Please try again in a few moments.',
```
ghostdevv:
```suggestion
        'Internal server error. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aenean tristique ex ac nisi dapibus maximus. Curabitur feugiat lorem eros, sed eleifend purus facilisis at.',
```

since it's mocked should be use lipsum here or something - maybe like this xD just to potentially avoid confusion as this message specifically isn't actually used anywhere

ghostdevv (changes requested):

app/error.stories.ts
cylewaitforit: Yeah that is better. Updated to use the suggestion instead.

ghostdevv (approved):

PR #2616, opened 2026-04-23
feat(ui): add keytrace-style identity visualizer (poc)

### LINKED ISSUE #2583: Adopt keytrace
### Describe the feature

As we connect with more services (from npmjs.com, atproto, to github.com, mastodon, etc) we need a way to link these services identities into a single User entity. [Keytrace](https://keytrace.dev) allows us to do exactly that:
> One identity, many proofs.
> Keytrace links your atproto handle to the accounts you already own — GitHub, DNS, npm, Mastodon, PGP — with signed, portable proofs. You be you, everywhere.

Adopting Keytrace will simplify [surfacing of Project/User data](https://github.com/npmx-dev/npmx.dev/issues/2568) in npmx.

To start with, we could implement a profile for each domain that shows the atproto avatar / banner / display name / description and a list of linked accounts from keytrace. Basically implementing a keytrace visualizer in npmx. So for npmx.dev/zeu.dev we should see her linked accounts.

Second step is to allow to link accounts directly from npmx. We can do that in another issue, too. Creating this issue to kick start the process of adopting keytrace and have a place to discuss about it in GitHub.

### Additional information

### Final checks

BittuBarnwal7479:
### 🔗 Linked issue

resolves #2583

---

### 🧭 Context

This PR started as a UI POC. It is now updated to integrate real Keytrace claim data on /profile/{domain}.

---

### 📚 Description

What changed?
- Integrated real Keytrace claim fetching via @keytrace/claims
- Profile header now renders real identity data (name/avatar; banner support in UI)
- Linked accounts now come from real claims (not synthetic account generation)
- Reverify flow is wired end-to-end to real claims
- Added safer proofMethod mapping with validation (no unsafe cast)
- Fallback behavior is now explicit:
  - real data when available
  - no-claims state when user has no claims
  - service-unavailable state when Keytrace cannot be reached
  - removed misleading mock profile fallback

 Verification
- Tested real handle lookup (example: zeu.dev) returns verified claims
- Confirmed profile page renders linked accounts from live Keytrace data
- Confirmed no-claims and unavailable states show empty/non-misleading account output
- Reverify updates status/last-checked using real claim matching

Notes
- This PR remains read-only UI (no linking flow creation yet)
- Account-link creation/verification authoring flow can be follow-up work


app/components/ProfileHeader.vue
```diff
@@ -0,0 +1,71 @@
...
+          @error="onAvatarError"
+        />
+        <span v-else class="font-mono text-fg-subtle" aria-hidden="true">{{ initials }}</span>
+      </div>
+
+      <div class="min-w-0">
+        <h2 class="font-mono text-xl sm:text-2xl font-medium min-w-0 break-words">
+          {{ profile.name || 'Unknown Profile' }}
+        </h2>
+        <p class="mt-2 text-fg-muted min-w-0 break-words">{{ profile.description }}</p>
```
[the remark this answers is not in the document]
BittuBarnwal7479: Check again.

app/components/LinkedAccounts.vue
```diff
@@ -0,0 +1,51 @@
...
+      <SkeletonBlock v-for="index in 3" :key="index" class="h-20 rounded-md" />
+    </div>
+
+    <p v-else-if="!accounts.length" class="mt-4 text-fg-muted">No linked accounts</p>
+
+    <ul v-else class="mt-4 space-y-2">
+      <li v-for="account in accounts" :key="`${account.platform}-${account.username}`">
+        <AccountItem :account="account" />
+      </li>
+    </ul>
```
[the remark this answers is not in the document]
BittuBarnwal7479: Check again.

BittuBarnwal7479:
@patak-cat please review.

app/components/Compare/FacetSelector.vue
```diff
@@ -101,30 +101,31 @@ function deselectAllFacet(category: string) {
         :aria-labelledby="`facet-category-label-${category}`"
         data-facet-category-facets
       >
-        <ButtonBase
+        <button
```
patak-cat: Would you revert this change and do it in a separate PR if you see value with it? Let's keep this PR only about adding the read-only keytrace claims

app/pages/profile/[identity]/index.vue
```diff
@@ -232,20 +238,38 @@ defineOgImage(
...
+      <div class="flex flex-col gap-2">
+        <h2
+          class="font-mono text-2xl sm:text-3xl font-medium min-w-0 break-words"
+          :title="$t('profile.likes')"
+          dir="ltr"
+        >
+          {{ $t('profile.likes') }}
+          <span v-if="likes">({{ likes.records?.length ?? 0 }})</span>
+        </h2>
+        <p class="text-fg-muted text-sm">{{ $t('profile.public_interests_description') }}</p>
```
Comment (patak-cat) [inline app/pages/profile/[identity]/index.vue]: Let's remove this explanation, too. Please propose it in a separate PR so it is easier to merge each of these.

patak-cat:
Thanks @BittuBarnwal7479! This is a great first step:
image

The avatar in the profile above it isn't loading for me:
image

I think we should remove this box and integrate it with the main profile info above it. But I'll let other folks that are working on design check things like spacing and overall UI.

BittuBarnwal7479:
> Thanks @BittuBarnwal7479! This is a great first step: image
> 
> The avatar in the profile above it isn't loading for me: image
> 
> I think we should remove this box and integrate it with the main profile info above it. But I'll let other folks that are working on design check things like spacing and overall UI.

Hii @patak-cat thanks for point out this.
For some claim types (for example DNS identities like zeu.dev), Keytrace does not return [avatarUrl] so the avatar was empty and appeared to disappear.

patak-cat:
> For some claim types (for example DNS identities like zeu.dev), Keytrace does not return [avatarUrl] so the avatar was empty and appeared to disappear.

This is fine. The one that isn't showing is the one in the Profile Header above the Linked Accounts. The image for zeu.dev links to [this svg from dicebear](https://api.dicebear.com/9.x/initials/svg?seed=zeu.dev) and it is broken when it is used in the site. I don't know what is the idea for this section:
image

I think you can remove it.

BittuBarnwal7479:
> > For some claim types (for example DNS identities like zeu.dev), Keytrace does not return [avatarUrl] so the avatar was empty and appeared to disappear.
> 
> This is fine. The one that isn't showing is the one in the Profile Header above the Linked Accounts. The image for zeu.dev links to [this svg from dicebear](https://api.dicebear.com/9.x/initials/svg?seed=zeu.dev) and it is broken when it is used in the site. I don't know what is the idea for this section: image
> 
> I think you can remove it.

Hi @patak-cat , thanks for the review.
I’ve now addressed:

- Removed the Identity profile header
- Kept real Keytrace avatars when present and only use fallback when missing.
- Fixed avatar rendering for identities where Keytrace does not return avatarUrl (for example DNS claims) by using deterministic fallback avatars.
- Routed both profile and account avatars through our internal image proxy, which resolves the CSP blocking issue from direct Dicebear URLs.

Tobbe:
Thanks for working on this @BittuBarnwal7479 

I just tried it at https://npmx-rcact1v5f-npmx.vercel.app/profile/tobbe.dev and this is what I see
image

Keytrace itself shows this

image

1. Would it be possible to make npmx also say "bsky" instead of just "other"?
2. I think it'd be nice with some explanation of what the different badges mean. I initially had "stale" on both my linked account, and didn't know what that meant or what to do about it. I eventually just tried refreshing them at keytrace, and that changed them to the current "verified" state.

BittuBarnwal7479:
> Thanks for working on this @BittuBarnwal7479
> 
> I just tried it at https://npmx-rcact1v5f-npmx.vercel.app/profile/tobbe.dev and this is what I see image
> 
> Keytrace itself shows this
> 
> image
> 1. Would it be possible to make npmx also say "bsky" instead of just "other"?
> 2. I think it'd be nice with some explanation of what the different badges mean. I initially had "stale" on both my linked account, and didn't know what that meant or what to do about it. I eventually just tried refreshing them at keytrace, and that changed them to the current "verified" state.

Thanks for review!
I'll update the labels to properly show bsky instead of other, and i agree that adding clearer explanations for the badge states would improve the UX.

I also noticed the merge conflicts and will resolve them.

I’m currently in the middle of my final B.Tech end exams, so updates might be a bit slower for the next few days, but I’ll continue working on the requested changes as soon as possible.

BittuBarnwal7479:
@Tobbe pls review.

Tobbe:
@BittuBarnwal7479 Great that it shows Bluesky now!

image

But it looks like the text inside the badges is slightly misaligned. Do you agree?

BittuBarnwal7479:
Hi @Tobbe , thanks for taking a look!

I checked the badge alignment in DevTools and it appears centered on my side. I've attached a screenshot for reference.

UI rendering can vary across browsers and systems sometime.

image

Tobbe:
Try zooming out to 90% and you might see the alignment issue.

Instead of setting vertical padding, try adjusting the line height. That should work more reliably across different zoom levels. Try for example `leading-[1.15]` (or some other multiplier)

BittuBarnwal7479:
> Try zooming out to 90% and you might see the alignment issue.
> 
> Instead of setting vertical padding, try adjusting the line height. That should work more reliably across different zoom levels. Try for example `leading-[1.15]` (or some other multiplier)

View in 100%
image

View in 90%
image

OS: window (Chrome)

BittuBarnwal7479:
> Try zooming out to 90% and you might see the alignment issue.
> 
> Instead of setting vertical padding, try adjusting the line height. That should work more reliably across different zoom levels. Try for example `leading-[1.15]` (or some other multiplier)


We can add more padding across top-bottom (1.15).

image

Tobbe:
> We can add more padding across top-bottom (1.15).

I never got anything using padding to work properly on my OS (MacOS) and my specific screen resolution/scaling. 
Tweaking the line height was more reliable for me. Maybe it makes the browser to use some other calculations that better handle sub-pixel rendering/rounding

BittuBarnwal7479:
> > We can add more padding across top-bottom (1.15).
> 
> I never got anything using padding to work properly on my OS (MacOS) and my specific screen resolution/scaling. Tweaking the line height was more reliable for me. Maybe it makes the browser to use some other calculations that better handle sub-pixel rendering/rounding

Thanks for the suggestion!

I updated the badge styles to use `leading-[1.15]` as suggested.
Could you take another look and let me know if it looks better on your setup?

Tobbe:
> I updated the badge styles to use `leading-[1.15]` as suggested.
> Could you take another look and let me know if it looks better on your setup?

It didn't make a difference. Is `leading-[1.15]` enough to make a difference on your setup?
You also shouldn't need to set vertical padding at all if the line height is tall enough.

Do your best. I won't block the PR on this. I hate messing with CSS, so don't want to force it on you either.

BittuBarnwal7479:
> > I updated the badge styles to use `leading-[1.15]` as suggested.
> > Could you take another look and let me know if it looks better on your setup?
> 
> It didn't make a difference. Is `leading-[1.15]` enough to make a difference on your setup? You also shouldn't need to set vertical padding at all if the line height is tall enough.
> 
> Do your best. I won't block the PR on this. I hate messing with CSS, so don't want to force it on you either.

Thanks for the suggestion. I updated the badges to use `leading-5` so the line height matches the badge height (`h-5`).

I checked it on my side, including macOS (Using chrome extension :-)), and the alignment looks good now. Let me know if you still see any issues.

image

Tobbe:
Amazing work! That's perfect 🤩

BittuBarnwal7479:
@patak-cat  if everything looks good now, feel free to merge. All requested changes have been addressed and checks are green. Thanks! 🙌

trueberryless:
Could you please remove all the translations except English? We like to keep feature PRs minimalistic and let the native-speaker update their translation files. Adding all those translations here would result in all translations being marked as complete, which messes with our Lunaria workflow.

gameroman (changes requested):
I agree with @trueberryless's review, we should only add translations for languages we know

BittuBarnwal7479:
> Could you please remove all the translations except English? We like to keep feature PRs minimalistic and let the native-speaker update their translation files. Adding all those translations here would result in all translations being marked as complete, which messes with our Lunaria workflow.

Sure thanks for review.

BittuBarnwal7479:
@trueberryless I've addressed the feedback and removed the translation changes. Could you please review again?

trueberryless:
most locale files still contain unwanted changes, please address again!

server/utils/keytrace.ts
```diff
@@ -0,0 +1,93 @@
+import type { ClaimVerificationResult } from '@keytrace/claims'
```
graphieros, line 1: 🙏 Can you please add unit tests for this utility ?

graphieros (changes requested):

BittuBarnwal7479:
> most locale files still contain unwanted changes, please address again!

@trueberryless Thanks for catching that. Some unintended locale file changes were introduced by an automated tool during my workflow.
image

server/utils/keytrace.ts
BittuBarnwal7479: Sure

graphieros:
> @trueberryless Thanks for catching that. Some unintended locale file changes were introduced by an automated tool during my workflow.

I think you need to remove the keys from the schema

BittuBarnwal7479:
> > @trueberryless Thanks for catching that. Some unintended locale file changes were introduced by an automated tool during my workflow.
> 
> I think you need to remove the keys from the schema

But this can create another issue on en translation file.

graphieros:
> > I think you need to remove the keys from the schema
> 
> But this can create another issue on en translation file.

Unused translation keys must be removed from the schema.
Perhaps it is easier to revert all changes made to locale files and schema, and then just add what you need.

BittuBarnwal7479:
> > > I think you need to remove the keys from the schema
> > 
> > 
> > But this can create another issue on en translation file.
> 
> Unused translation keys must be removed from the schema. Perhaps it is easier to revert all changes made to locale files and schema, and then just add what you need.

@graphieros Thanks for the feedback. I wanted to explain the issue I'm running into.

Initially, I removed the translation changes from the non-English locale files and kept the PR focused on `en.json` and the schema. However, the CI then reported that `common.error` is an unused translation key. If I keep the key, `i18n:report` fails. If I remove it, the schema is regenerated accordingly, but the automated i18n tooling also updates additional locale files, which reintroduces changes outside the scope of this PR.

I've been trying to keep the PR aligned with the requirement of only modifying `en.json` and the schema, but the automated generation and validation steps seem to be pulling in extra locale changes. I'm investigating the correct workflow to resolve this cleanly and will push another update once I have it sorted out.

Thanks for your patience.

BittuBarnwal7479:
After investigating, I think we should remove the unused `common.error` key across all locales since it's flagged by the i18n tooling and causes the report to fail. I'll keep the new Keytrace strings in `en.json` only and include just this required cleanup.
Thanks!!

BittuBarnwal7479:
> After investigating, I think we should remove the unused `common.error` key across all locales since it's flagged by the i18n tooling and causes the report to fail. I'll keep the new Keytrace strings in `en.json` only and include just this required cleanup. Thanks!!

@trueberryless pls suggest changes.

Tobbe:
@BittuBarnwal7479 I feel sorry for you this is taking so long to get merged. Please hang in there! You're doing great work

Tobbe:
> After investigating, I think we should remove the unused `common.error` key across all locales since it's flagged by the i18n tooling and causes the report to fail. I'll keep the new Keytrace strings in `en.json` only and include just this required cleanup. Thanks!!

FWIW, I agree. You removed the only usage of `common.error` here 
image
https://github.com/npmx-dev/npmx.dev/pull/2616/changes#diff-29ac6f4151ba0eecfe277dda22bb157cc2f04cf6e3a85f6db96b82f2719c4073L248

So it only makes sense to also remove it from the translations schema and all translation files

BittuBarnwal7479:
> > After investigating, I think we should remove the unused `common.error` key across all locales since it's flagged by the i18n tooling and causes the report to fail. I'll keep the new Keytrace strings in `en.json` only and include just this required cleanup. Thanks!!
> 
> FWIW, I agree. You removed the only usage of `common.error` here image https://github.com/npmx-dev/npmx.dev/pull/2616/changes#diff-29ac6f4151ba0eecfe277dda22bb157cc2f04cf6e3a85f6db96b82f2719c4073L248
> 
> So it only makes sense to also remove it from the translations schema and all translation files

Thanks for confirming. That makes sense, I'll remove the unused key from the schema and locale files and update the PR accordingly.

server/utils/keytrace.ts
BittuBarnwal7479: > 🙏 Can you please add unit tests for this utility ?

@graphieros I have added unit test for this utility.

BittuBarnwal7479:
Hey @graphieros PR is ready for review. Please take a look whenever you get a chance. thanks!

PR #2628, opened 2026-04-25
fix(ui): incorrect z-index on tooltip

bastiendmt:
### 🧭 Context

Z-index on tooltip is above header. This is especially strange on mobile since the tooltip stay active.

### 📚 Description

I've updated the tooltip z-index as well as the header so it stays behind. 

### Before

https://github.com/user-attachments/assets/5e0ff5ce-bc60-4adc-9513-a377a436714f

### After

https://github.com/user-attachments/assets/e739869b-bb52-480c-8762-31a0ff8fddb7

### DIFF
--- a/app/components/Package/Header.vue
+++ b/app/components/Package/Header.vue
@@ -235,7 +235,7 @@ useShortcuts({
   </header>
   <div
     ref="header"
-    class="w-full bg-bg sticky top-14 z-10 border-b border-border pt-2"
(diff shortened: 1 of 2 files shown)


ghostdevv (dismissed):
could you take a look at the code rabbit comment?

bastiendmt:
> could you take a look at the code rabbit comment?

I've updated it to 40 👍

serhalp (approved):

PR #2635, opened 2026-04-26
feat: add the new package timeline page to the command palette

serhalp:
### 🔗 Linked issue

N/A

### 🧭 Context

The string placeholder shape used by `useCommandPaletteVersionCommands` made callers that already had named route helpers resolve routes to strings and then patch encoded placeholders back into place. This is brittle, especially for routes where vue-router encodes params differently from the URL shape users see.

We have [a documented guideline](https://github.com/npmx-dev/npmx.dev/blob/69351cfb0698622331a180a764e054c48514a0eb/CONTRIBUTING.md#internal-linking) that internal navigation should use route object syntax with named routes instead of strings. This shape was preventing us from respecting our own guideline.

### 📚 Description

Pass a versioned package route builder to `useCommandPaletteVersionCommands` instead of a string with a version placeholder.

> [!NOTE]
> This PR was originally a pure refactor of `useCommandPaletteVersionCommands` and #2636 was stacked on it, but since that was accidentally merged into this, this PR now contains both!


serhalp:
ping ping bump bump 🙏🏼

43081j (approved):

PR #2640, opened 2026-04-27
feat: add compact badge style variant

OrbisK:
### 🔗 Linked issue

fixes #2509

### 🧭 Context

Working on https://github.com/OrbisK/renovate-config-npmx i wanted to use the npmx styled badges, but they were to big.

### 📚 Description

This PR add a third style (compact) to the badges. compact has less x padding and "trimmed" labels

`?style=default`

image

`?style=compact`

image

Alternative naming: `?style=slim`

------------

<sub>Funfact: [fff5849](https://github.com/npmx-dev/npmx.dev/pull/2640/commits/fff5849b0b07be5987a48784c32c67abaadcd418) commit hash is [tomato color](https://colorkit.co/color/ff5849/) in hex </sub>

PR #2641, opened 2026-04-27
feat(ui): add `/badge/compare/` endpoints

OrbisK:
### 🔗 Linked issue

NA

### 🧭 Context

For the [npmx Renovate preset](https://github.com/OrbisK/renovate-config-npmx), I wanted to add some e18e related badges. I think it would be useful to be able to identify e18e changes when reviewing a Renovate PR.

### 📚 Description

This PR adds `/badge/compare/` endpoints. One for single package compare and one for cross package compare.

single package compare:

image

image

cross package compare:

image

image

I am not fully satisfied with the API. Perhaps it could be made a bit cleaner. Feedback welcome! (opening as draft for now)

This PR includes the changes of #2640 

---------------

Disclaimer: I have used claude for docs and to extract the shared utils for compare and normal badges


OrbisK:
@43081j what do you think from an e18e perspective?

@serhalp any thoughts on the compare stuff?

trueberryless:
Hi!

Sorry that we didn't have the time for a reciew yet.

Regarding the high level idea of this PR, I am a bit sceptical: The badge itself doesn't include the information of what is compared against each other, making it a little bit dependent on the context surrounding the badge. Not to say this is different for all the other badges, most of them actually are tide to the package name. But to me it feels like with comparisons, there is just too much information missing from the badge. 

My ideology of badges is that they should be universally recognisable, for example if you see a green Netlify badge, you already know that this stands for some successful deployment.

Here, it seems that those badges just display some specific stats / differences, which IMHO is too _abusive_ of the badge philosophy.

But at this point I'm just yapping. Would love to hear other opinions on this idea 💡

OrbisK:
Basically, I agree with you (especially when it comes to the “misuse” of badges and the missing information).

In reality however, badges have also found a use in areas where “latest” information needs to be displayed without the ability to query/fetch or rebuild it.

Lets stick with the renovation example. 

image

On their own, these badges are completely meaningless. But in context, they are exactly whats needed. If a value changes (e.g., the relative date), the tag remains “up-to-date” without having to update the source (md, in this case). 

Another use case is simply a nicer-looking link, for example to link to the package on npmx in its README. 

image


-----------

Regarding the missing information, I also tried to include the name and version being compared, but it looked wrong when the badge ended up taking up 300% of its original space.

Apart from renovate, I can think of some use cases in the packages readme for comparing your latest version with that of a competitor. For example, I can imagine `ofetch` being compared to `axios` in the docs/readme, and with the badge the information will always be up to date in this context.

PR #2292, opened 2026-03-27
feat: new og images

harlan-zw:
### 🔗 Linked issue

https://github.com/npmx-dev/npmx.dev/pull/1654

### 🧭 Context

The previous PR was rolled back due to some Takumi stability issues. These are now solved afaik and OG Image got a [v6 stable release](https://github.com/nuxt-modules/og-image/releases/tag/v6.0.0) so it's a good time to reintroduce it. (Takumi is v1 beta now so also likely stable very soon)

:warning:  Docus < v5.9.0 uses the old version of og image which causes a version compat issue. Upgrading Docus to v5.9.0 creates a bunch of new issues that seem unrelated to og images, we should investigate that seperately.

### 📚 Description

Same as before, I've resynced the UI to match recent branding changes.

<details>
<summary>sample - check PR files for more</summary>

https://npmx.dev/
og-image-home

https://npmx.dev/accessibility
og-image-accessibility

https://npmx.dev/blog/alpha-release
og-image-blog-alpha-release

https://npmx.dev/package/@nuxt/kit
og-image-package--nuxt-kit

https://npmx.dev/package-docs/ufo/v/1.6.3
og-image-package-docs-ufo-v-1-6-3

</details>

Supersedes #2122 (rebased on latest main with CI fixes).

PR #2647, opened 2026-04-28
chore: use smooth path utility from vue-data-ui

graphieros:
Follow up to #2292

This replaces the smooth path creation function used for og images with the available utility in vue-data-ui.

### DIFF
--- a/app/components/OgImage/Package.takumi.vue
+++ b/app/components/OgImage/Package.takumi.vue
@@ -16,7 +16,7 @@ const KIND_ICONS: Record<string, string> = {
 import type { JsDelivrFileNode } from '#shared/types'
 import { joinURL } from 'ufo'
 import { useCharts } from '~/composables/useCharts'
-import { createSmoothPath as smoothPath } from 'vue-data-ui/utils'
+import { createSmoothPath } from 'vue-data-ui/utils'
 
 const REPO_PROVIDER_ICONS: Record<string, string> = {
   github: 'i-simple-icons:github',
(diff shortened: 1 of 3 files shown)


app/components/OgImage/Package.takumi.vue
```diff
@@ -15,7 +15,8 @@ const KIND_ICONS: Record<string, string> = {
 <script setup lang="ts">
 import type { JsDelivrFileNode } from '#shared/types'
 import { joinURL } from 'ufo'
-import { smoothPath, useCharts } from '~/composables/useCharts'
+import { useCharts } from '~/composables/useCharts'
+import { createSmoothPath as smoothPath } from 'vue-data-ui/utils'
```
shuuji3:
```suggestion
import { createSmoothPath } from 'vue-data-ui/utils'
```

nit: It seems that the usage of this fuction is only here in the project (https://github.com/npmx-dev/npmx.dev/pull/2647/changes#diff-989803a8350f5b506a5bdea17a1237c410d7eee66b2e7015b0bd46378a78e097R243)
Maybe we could refactor to use the original exported name?

shuuji3 (approved):
LGTM 👍🏻

I put one suggestion for small refactoring.

app/components/OgImage/Package.takumi.vue
graphieros: @shuuji3 thank you :) It is cleaner indeed

shuuji3 (approved):

PR #2657, opened 2026-04-29
fix: restore maximumFractionDigits: 0 on sparkline data label formatter

### LINKED ISSUE #2656: bug: weekly downloads label shows decimal digits
### Describe the bug

Weekly downloads data label shows decimals like `1,195,399.72` instead of a rounded integer.

### Root cause

PR #2556 attempted to fix this by adding `Math.ceil` in `applyDataCorrection` — but it also **removed** the formatter-level guard that was actually preventing the issue:

```diff
-const numberFormatter = useNumberFormatter({
-  maximumFractionDigits: 0,
-})
+const numberFormatter = useNumberFormatter()
```

`VueUiSparkline` does its own internal interpolation on the dataset and passes those float values directly to `dataLabel.formatter`. The `Math.ceil` in `applyDataCorrection` only rounds the input dataset — it cannot affect what the component computes internally.

Evidence from the rendered SR table:

```
1138304              ← first point, no interpolation
1195399.7219423861   ← VueUiSparkline's internally computed float
1249433.206396996
...
```

These floats bypass `applyDataCorrection` entirely and land in the formatter, which now has no `maximumFractionDigits` constraint.

### Fix

In `app/components/Package/WeeklyDownloadStats.vue`, either:

**Option A** — restore `maximumFractionDigits: 0` on the formatter's number instance:
```ts
const numberFormatter = useNumberFormatter({ maximumFractionDigits: 0 })
```

**Option B** — round in the formatter callback:
```ts
formatter: ({ value }) => numberFormatter.value.format(Math.round(value))
```

The `Math.ceil` added to `applyDataCorrection` in #2556 is fine to keep for other callers, but it does not fix the sparkline label.

2u841r:
### 🔗 Linked issue

Closes #2656

### 🧭 Context

PR #2556 added `Math.ceil` in `applyDataCorrection` to fix decimal numbers in the weekly downloads label, but also removed `maximumFractionDigits: 0` from the formatter in `WeeklyDownloadStats.vue` — which was actually the effective guard.

`VueUiSparkline` does its own internal interpolation and passes float values directly to `dataLabel.formatter`, bypassing `applyDataCorrection` entirely. Without `maximumFractionDigits: 0`, those floats render as `1,195,399.72` instead of `1,195,400`.

### 📚 Description

Restore `maximumFractionDigits: 0` on the `useNumberFormatter` call used for the sparkline data label. One-line change.

### DIFF
--- a/app/components/Package/WeeklyDownloadStats.vue
+++ b/app/components/Package/WeeklyDownloadStats.vue
@@ -63,7 +63,7 @@ function handleModalTransitioned() {
 }
 
 const { fetchPackageDownloadEvolution } = useCharts()
-const numberFormatter = useNumberFormatter()
+const numberFormatter = useNumberFormatter({ maximumFractionDigits: 0 })
 
 const { accentColors, selectedAccentColor } = useAccentColor()
 


2u841r:
Closing -- this reverts a deliberate decision from the maintainer in #2556. The fix is already in main via #2519 and #2550, pending next release.

Issue #1122, opened 2026-02-07
`VersionSelector` is not accessible for screen readers


knowler:
The [`<VersionSelector>`](https://github.com/npmx-dev/npmx.dev/blob/3cf7fef97cef6be97e8f9ace628c88757a1b0c84/app/components/VersionSelector.vue) component [used in the package code page](https://github.com/npmx-dev/npmx.dev/blob/3cf7fef97cef6be97e8f9ace628c88757a1b0c84/app/pages/package-code/%5B...path%5D.vue#L322-L330) is not accessible for screen readers.

The root of the problem is that [ARIA role `option`](https://www.w3.org/TR/wai-aria-1.2/#option) shouldn’t nest interactive content (or really any semantic content). Its children are expected to be [presentational](https://www.w3.org/TR/wai-aria-1.2/#childrenArePresentational).

The current incorrect structure is like this:

- `role=listbox`
  - `role=option`
    - `role=button`
    - `role=link`
    - (then a list of `role=option` shown when the button is `aria-expanded=true`)

ARIA doesn’t expect there to be anything past the `role=option` (i.e. the contents are meant to be used for its accessible name).

On Windows 11, NVDA (in Firefox and Edge) is able to focus the expand/collapse button and link and even trigger the buttons, however, it cannot trigger the link.

On macOS and iOS, VoiceOver both do not go further than the `role=option` widgets (and do nothing when activated).

**We will need different semantics** which I can research. I suspect it might be a multi-level list of expand/collapse buttons + links that is focus trapped while it’s open. A treeview could be another option, but that has some significant complexity to it and shouldn’t be used lightly. The README table of contents is also using `role=listbox` when it probably shouldn’t, so it would be good to unify on an approach (even if the components are styled different).

I could add some screen caps when I get a change to better demonstrate the issue.

knowler:
To avoid a whole lot of ARIA complexity (e.g. a tree view), this is an example of the structure I’m recommending:

```html
<nav aria-label="version">
	<button popovertarget="version-selector" popovertargetaction="toggle">
		<span>4.3.1</span>
		<span>latest</span>
		<span class="i-lucide:chevron-down" aria-hidden="true"></span>
	</button>
	<ul aria-label="nuxt versions" id="version-selector" popover>
		<li>
			<a href="/package-docs/nuxt/v/4.3.1" aria-current="page">
				<span>4.3.1</span>
				<span>latest</span>
			</a>
[...]

PR #2660, opened 2026-04-30
feat: added feature to copy version number by hovering on the version

### LINKED ISSUE #2646: A quick way to copy package version
### Describe the feature

Similar to a quick way to copy the package name, would be nice to have a quick way to copy the version

Also I think the package version should be more noticeable, currently it is very small and not easily noticeable 

### Additional information

### Final checks

PreethiPantangi:
### 🔗 Linked issue

Resolves #2646 

### 🧭 Context

The feature provides user with the capability to copy the version number of the package by hovering over it. This helps the user to easily grab the data instead of typing it out. 

Adds copy-to-clipboard functionality for package version via hover interaction in the header.

### 📚 Description

Files changed: 
    - Header.vue
    - en.json

Header.vue
    - The VersionSelector component is wrapped with CopyToClipboardButton component to ensure the user is able to copy the version number on hover.

en.json
    - Added the key value pair "copy_version": "Copy version number" to ensure internationalization

No tests added as this change is UI-only and does not affect business logic. Existing functionality remains unchanged.

No documentation changes required

I've used co-pilot to understand some parts of the code before diving into making the feature changes


app/components/Package/Header.vue
```diff
@@ -2,6 +2,7 @@
 import type { RouteLocationRaw } from 'vue-router'
 import type { CommandPaletteContextCommandInput } from '~/types/command-palette'
 import { SCROLL_TO_TOP_THRESHOLD } from '~/composables/useScrollToTop'
+import { useClipboard } from '@vueuse/core'
```
gameroman: iirc, this isn't necessary since it's auto-imported

app/components/Package/Header.vue
```diff
@@ -98,6 +105,17 @@ useCommandPaletteContextCommands(
...
+        group: 'package',
+        label: $t('package.versions.copy_version'),
+        keywords: [props.resolvedVersion],
+        iconClass: 'i-lucide:copy',
+        action: () => {
+          copyPkgVersion()
+          announce($t('command_palette.announcements.copied_to_clipboard'))
+        },
+      },
     ]
```
gameroman: coderabbit's suggestions makes sense here. I would do something like this here

```ts
const commands: CommandPaletteContextCommandInput[] = []

if (packageName.value) {
  commands.push({
    id: 'package-copy-name',
    group: 'package',
    label: $t('package.copy_name'),
    keywords: [packageName.value],
    iconClass: 'i-lucide:copy',
    action: () => {
      copyPkgName()
      announce($t('command_palette.announcements.copied_to_clipboard'))
    },
  })
}

if (props.resolvedVersion) {
  commands.push({
    id: 'package-copy-version',
    group: 'package',
    label: $t('package.versions.copy_version'),
    keywords: [props.resolvedVersion],
    iconClass: 'i-lucide:copy',
    action: () => {
      copyPkgVersion()
      announce($t('command_palette.announcements.copied_to_clipboard'))
    },
  })
}
```

gameroman (commented):
Overall very good, just make sure the CI passes, also left a couple of comments

[👉
gameroman, line 1: This is accidental probably?

Please
gameroman, line 1: This too

PreethiPantangi:
@gameroman I’ve addressed the feedback from the code review. Also, deleted the two junk files created accidentally.

gameroman:
Looks good

One thing to change, it probably should hide the copy button when you press the button to change the version

image

PreethiPantangi:
@gameroman the changes have been implemented

app/components/Package/Header.vue
```diff
@@ -97,8 +120,22 @@ useCommandPaletteContextCommands(
           copyPkgName()
           announce($t('command_palette.announcements.copied_to_clipboard'))
         },
-      },
-    ]
+      })
+    }
```
gameroman, line 124: I think you can do it like this

```ts
    const commands: CommandPaletteContextCommandInput[] = [
      {
        id: 'package-copy-name',
        group: 'package',
        label: $t('package.copy_name'),
        keywords: [packageName.value],
        iconClass: 'i-lucide:copy',
        action: () => {
          copyPkgName()
          announce($t('command_palette.announcements.copied_to_clipboard'))
        },
      }
    ]
```

app/components/Package/Header.vue
```diff
@@ -21,6 +21,21 @@ const packageHeaderHeight = usePackageHeaderHeight()
...
+watch(versionSelectorOpen, isOpen => {
+  if (isOpen) {
+    shouldHideCopyText.value = true
+    if (hideTextTimeout) clearTimeout(hideTextTimeout)
+  } else {
+    hideTextTimeout = setTimeout(() => {
+      shouldHideCopyText.value = false
+    }, 300)
+  }
+})
```
gameroman, line 38: What does the timeout do here?

PreethiPantangi: When we close the version selector, there's a short delay before the copy text comes back, to prevent UI flickering if someone opens and closes the selector quickly (like an accidental click). 300ms keeps it feeling smooth. The 300ms was added to keep it smooth.

gameroman: It feels a little bit weird/unnatural with the delay, as if it's broken

I'm thinking,

Either the delay could be decreased

Or, hide the `Copy version number` button completely, so even when the version picker is closed, it is not shown until you move your mouse away from it and hover again

app/components/Package/Header.vue
knowler, line 1: If I recall, the previous layout of package pages had the version number next to the package name which was a better candidate for a convenience like this. Now, with the version being a part of the version selector — [a component that has significant issues on its own](https://github.com/npmx-dev/npmx.dev/issues/1122)  — this seems like a bit too much is going on. If we want this functionality, I think a dedicated button next to the version number would be better, as it avoids two popover interactions for the same element and the button on hover could be inferred as a tooltip for what the version selector does. That or we should find somewhere else to do this.

PreethiPantangi: I made the change to copy the version number on hover to make sure that it is consistent within the application. @gameroman Should I go ahead and include a copy button next to version number?

PR #2664, opened 2026-05-01
chore(deps): bump @storybook-vue/nuxt to latest

cylewaitforit:
### 🧭 Context

The latest version of `@storybook-vue/nuxt` published to npm is not fully compatible with Storybook 10, Nuxt 4, or Vite 8. 

This PR updates the version to the latest merged PR for that package using pkg.pr.new.

### 📚 Description

     
 Grabbed pkg.pr.new version from latest merged PR for `@storybook-vue/nuxt` and removed the override set to force Storybook to version 10.


pnpm-workspace.yaml
```diff
@@ -18,7 +18,7 @@ catalogs:
     'msw': '^2.13.2'
 
   storybook:
-    '@storybook-vue/nuxt': '9.0.1'
+    '@storybook-vue/nuxt': 'https://pkg.pr.new/@storybook-vue/nuxt@1021'
```
ghostdevv, line 21: my pkg.pr.new game is weak - how do I see where this is going 😅

ghostdevv (changes requested):

pnpm-workspace.yaml
cylewaitforit: The number at the end is the PR number on the repo. So in this case https://github.com/nuxt-modules/storybook/pull/1021#issuecomment-4151109339 .


I picked the latest PR merged to main in that package’s history at the time. https://github.com/nuxt-modules/storybook/commits/main/packages/storybook-addon

Looks like we could probably go with 1051 now if preferred, but either have storybook 10 versions which is the main reason for bumping at this point.

ghostdevv (approved):

PR #2665, opened 2026-05-01
fix: npmx connector Node.js v22 compatibility

### LINKED ISSUE #2645: npmx-connector failed to run via pnpm
### Describe the bug

```
➜  psn-api git:(main) pnpm dlx npmx-connector
Packages: +16
++++++++++++++++
Progress: resolved 21, reused 11, downloaded 5, added 16, done
file:///Users/luna/Library/Caches/pnpm/dlx/165b4bdd1d2a08fbc55efcbb9bfc8f811f2510e6c69ab469f8f5cf61df8e0695/19dd0783035-16ffa/node_modules/.pnpm/npmx-connector@0.9.0/node_modules/npmx-connector/dist/server-C9jFi7Vy.mjs:773
	await using tempDir = await mkdtempDisposable(join(tmpdir(), "npmx-init-"));
	            ^^^^^^^

SyntaxError: Unexpected identifier 'tempDir'
    at compileSourceTextModule (node:internal/modules/esm/utils:344:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:105:18)
    at #translate (node:internal/modules/esm/loader:534:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:581:27)

Node.js v22.16.0
```

### Additional context

_No response_

### Logs

```shell-script

```

DDeenis:
### 🔗 Linked issue

Resolves #2645

### 🧭 Context

`npmx-connector` currently crashes for users trying to run it with Node.js 22, which is still in the maintenance and used by many people. This PR fixes the only incompatible function (`packageInit`).

### 📚 Description

Node v24+ only features used:
- `mkdtempDisposable` - for temp dir creation
- `await using` - to ensure the temp dir cleanup

Replaced with compatible versions:
- `mkdtempDisposable` -> `mkdtemp`
- Automatic disposal with `await using` -> Manual disposal with `rm`

Fix verified on Node v22

~~Tradeoffs: if both the function code and `rm` will throw, only the error from `rm` would be preserved. Should be possible to fix at a cost of complex cleanup logic, but the situation seems unlikely.~~

### DIFF
--- a/cli/src/npm-client.ts
+++ b/cli/src/npm-client.ts
@@ -574,6 +574,9 @@ export async function packageInit(
 
   const tempDirPath = await mkdtemp(join(tmpdir(), 'npmx-init-'))
 
+  let publishResult: NpmExecResult | null = null
+  let publishError: unknown = null
+
   try {
     // Determine access type based on whether it's a scoped package
     const isScoped = name.startsWith('@')
@@ -621,8 +624,29 @@ export async function packageInit(
       logError(result.stderr.split('\n')[0] || 'Command failed')
     }
 
-    return result
-  } finally {
-    await rm(tempDirPath, { recursive: true, force: true })
+    publishResult = result
+  } catch (error) {
+    publishError = error
   }
+
+  try {
+    await rm(tempDirPath, { recursive: true, force: true })
+  } catch (cleanupError) {
+    if (publishError) {
+      // Preserve original error
+      Object.assign(cleanupError as Error, { cause: publishError })
+    }
+
+    throw cleanupError
+  }
+
+  if (publishError) {
+    throw publishError
+  }
+
+  if (!publishResult) {
+    throw new Error('packageInit completed without a publish result')
+  }
+
+  return publishResult
 }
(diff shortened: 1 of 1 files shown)


ghostdevv (changes requested):
holding this while we check we definitely want to downgrade the Node version we support

also this is effectively a revert of #2365

DDeenis:
@ghostdevv okay, this makes more sense. In this case, the issue is that Node 22 users don't get a warning of unsupported engine. Feel free to close this PR in case the downgrade is unwanted.

gameroman:
> @ghostdevv okay, this makes more sense. In this case, the issue is that Node 22 users don't get a warning of unsupported engine. Feel free to close this PR in case the downgrade is unwanted.

Adding a warning is a good idea probably

trueberryless:
@DDeenis Would you mind updating the PR to print a warning instead of downgrading?

And it would probably also be a good idea to specify the minimum version in contribution guidelines: https://github.com/npmx-dev/npmx.dev/blob/4aed48bbedf27958a0c383cf9229e9c08934ddc7/CONTRIBUTING.md?plain=1#L84

PR #2666, opened 2026-05-01
feat: press freedom day noodle

alexdln:
### 🔗 Linked issue

A new noodle dedicated to World Press Freedom Day

The press has been shedding light on events throughout history - reporting, investigating, raising critical issues, and revealing defining moments of our past. Journalists don’t just give a part of themselves; they often risk their lives and their freedom.

In recent years, press freedom has been in decline. And when it falls, it leads to a loss of freedom for everyone - creating a cycle that only deepens over time.

This is a small noodle to once again highlight the importance of journalism. In the background, you can find several key articles from our history, many of which led to significant change

<details>
<summary>Background for nerds</summary>
press-stories

(See all in figma)
</details>

PR #2668, opened 2026-05-02
feat: press freedom day noodle (#2666)

alexdln:
### 🧭 Context

Cherry-pick to the release branch to make the noodle available without full-framed release. The release itself will likely happen in the next few days, but we probably won't do it early just because of the noodle


patak-cat:
Thanks Alex! Let's go with the regular release flow.

PR #2675, opened 2026-05-04
refactor(ui): use new useClipboard with async

MatteoGabriele:
The fix for #2151 has been incorporated into vueuse, and with the release of v14.3.0, we can now remove the additional code and rely on the version included in the library. 

see https://github.com/vueuse/vueuse/issues/5368


graphieros (approved):
Works on my Safari 🌿

ghostdevv (approved):
assuming the code rabbit comment is valid, we can fix it here or create an issue? otherwise LGTM

MatteoGabriele:
> assuming the code rabbit comment is valid, we can fix it here or create an issue? otherwise LGTM

@ghostdevv the copying action "never fails", even if it does. 
We might need a different UX for failure. 🤔 I suppose we could add a small "x" icon when it fails, but I'd like to check with some a11y people what the best strategy is for that too.
If it's fine with you, I would prefer to create a separate PR so we can get this out of the way faster.

shuuji3 (approved):
Works as expected on Firefox/Google Chrome on Linux/macOS/Android ✨ 🧹

PR #2679, opened 2026-05-05
docs(ui): add stories for blog index page

IestynGage:
### 🔗 Linked issue

https://github.com/npmx-dev/npmx.dev/issues/2150

### 🧭 Context

Noticed this issue still had missing storybooks and it was labelled with Help wanted, so I added the storybook.

### 📚 Description

Added the storybook for the blog index. Seems to be working just fine. I looked at the other page storybooks and followed the convention used.

Screenshot 2026-05-05 at 23 39 17


ghostdevv:
@cylewaitforit if you have time would you be able to take a look? no rush or pressure!

app/pages/blog/index.stories.ts
cylewaitforit, line 1: The new story renders fine in the [PR's deployed storybook ](https://698b88d5157d89f1f33a6c21-yloeulncvy.chromatic.com/?path=/story/pages-blog--default).

@ghostdevv ~~I don't have access to check [the Chromatic tests](https://www.chromatic.com/build?appId=698b88d5157d89f1f33a6c21&number=4676) anymore so you may want to check what the visual regression tests capture.~~

It looks like the blogs that are being displayed are live data. So my assumption is that would probably mean the chromatic tests would suddenly be out of date when a new blog post is added.

So while I know both are ultimately wanted, we would need to decide if disabling visual regression tests on this page to start would be a blocker to merging or if increasing the coverage of our documented pages is a fine enough short term goal. 

If documenting to start is the more important goal then we could add the chromatic parameter to disable.

```ts
const meta = {
  component: Blog,
  parameters: {
    layout: 'fullscreen",
    // Disables Chromatic's snapshotting on a component level
    chromatic: { disableSnapshot: true },
  },
  decorators: [pageDecorator],
} satisfies Meta<typeof Blog>;
```
https://www.chromatic.com/docs/disable-snapshots/#with-storybook

We would probably want to then add a follow up issue to mock out the blog posts in order to re-enable snapshots.

cylewaitforit: Update:

I have access to the chromatic tests again and understandably the snapshot has the same live data. So I think my assumption that new blog data would cause the snapshot to become outdated seems correct.

ghostdevv: could you take a look @IestynGage :eyes:

IestynGage: @cylewaitforit Could you clarify what you mean by live data? From your suggestion I thought the blog posts was from a CMS system. But if you see the spurios commit I added where I create a blog post. [The storybook correctly updated](https://www.chromatic.com/test?appId=698b88d5157d89f1f33a6c21&id=69fe2f936a7eef0494c05efd). Kinda suggests to me that the cards are built by SSG. This works only if you restart nuxt/vite.

IDK maybe I'm missing something but it seems like they are built by a vite module in `modules/blog.ts` which creates a object in `.nuxt/blog/posts.ts`. If i'm missing something I'm happy to add the `chromatic: { disableSnapshot: true }`

cylewaitforit: @IestynGage  At the risk of sounding like an LLM "You are totally right". The npmx pages don't have a ton of SSG. So I was not thinking with that SSG part of my brain at all, and fell back to the idea of needing to mock the things that change like we need to for the other stories. But those changes we are mocking against all happen outside of the PRs. 

Since the blog changes would only happen through the PR pipeline where the new chromatic changes can(and should) be reviewed, this is actually a great thing to have a visual regression test on.

app/pages/blog/another-blog.md
cylewaitforit, line 1: Thanks for this test. This helped me think through the blog workflow better. We will need to remember to remove this file before @ghostdevv (or someone else) approves and merges the PR.

app/pages/blog/index.stories.ts
cylewaitforit: For completeness I had started down the path of https://storybook.js.org/docs/writing-stories/mocking-data-and-modules/mocking-modules but I think this default story showing the actual static posts is appropriate.

cylewaitforit (approved):

app/pages/blog/another-blog.md
IestynGage: I've removed said commit instead of reverting it for a better commit history.

app/pages/blog/index.stories.ts
IestynGage: Initially I looked into trying the mock modules. However it didn't seem to work and vite/storybook/nuxt was complaining . When I was trying to figure out why is when I stumbled upon the SSG code. 

Note I suspect storybook `sb.mock()` is intended to mock functions only 🤔  And you need to mock the whole module using the [mock files](https://storybook.js.org/docs/writing-stories/mocking-data-and-modules/mocking-modules#mock-files) to mock a exported module. I never got to the bottom of why what I was doing was incorrect so perhaps I missed something.

ghostdevv (approved):

PR #2694, opened 2026-05-09
fix: skip emoji conversion inside code blocks

### LINKED ISSUE #2364: `:1234:` is converted to emoji when found w/in an IPv6 address

gameroman:
### 🔗 Linked issue

Closes #2364
Closes #2373

### 🧭 Context

| | Screenshot |
| - | - |
| Before | image | 
| After | image |

### 📚 Description

Skips emoji conversion inside code blocks

Thank you to @ffedex (#2373) for the original PR

### DIFF
--- a/shared/utils/emoji.ts
+++ b/shared/utils/emoji.ts
@@ -1,5 +1,5 @@
 // copied from https://github.com/markdown-it/markdown-it-emoji/blob/master/lib/data/full.mjs
-const emojis = {
+const emojis: Record<string, string> = {
   '100': '💯',
   '1234': '🔢',
   'grinning': '😀',
@@ -1905,15 +1905,16 @@ const emojis = {
   'wales': '🏴󠁧󠁢󠁷󠁬󠁳󠁿',
 }
 
-const emojisKeysRegex = new RegExp(
-  Object.keys(emojis)
-    .map(key => `:${key}:`)
-    .join('|'),
-  'g',
-)
-export function convertToEmoji(text: string): string {
-  return text.replace(emojisKeysRegex, match => {
-    const key = match.slice(1, -1) as keyof typeof emojis
-    return emojis[key] || match
-  })
+export function convertToEmoji(html: string): string {  
+  return html.replace(
+    /(<code[\s>][\s\S]*?<\/code>|<pre[\s>][\s\S]*?<\/pre>)|(:[\w+-]+:)/gi,
+    (match, codeBlock: string | undefined, shortcode: string | undefined) => {
+      if (codeBlock) return codeBlock
+      if (shortcode) {
+        const key = shortcode.slice(1, -1)
+        return emojis[key] ?? shortcode 
+      }
+      return match
+    },
+  )
 }
--- a/shared/utils/emoji.ts
+++ b/shared/utils/emoji.ts
@@ -1905,14 +1905,14 @@ const emojis: Record<string, string> = {
   'wales': '🏴󠁧󠁢󠁷󠁬󠁳󠁿',
 }
 
-export function convertToEmoji(html: string): string {  
+export function convertToEmoji(html: string): string {
   return html.replace(
     /(<code[\s>][\s\S]*?<\/code>|<pre[\s>][\s\S]*?<\/pre>)|(:[\w+-]+:)/gi,
     (match, codeBlock: string | undefined, shortcode: string | undefined) => {
       if (codeBlock) return codeBlock
       if (shortcode) {
         const key = shortcode.slice(1, -1)
-        return emojis[key] ?? shortcode 
+        return emojis[key] ?? shortcode
       }
       return match
     },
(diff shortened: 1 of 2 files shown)


shared/utils/emoji.ts
```diff
@@ -1905,15 +1905,16 @@ const emojis = {
...
-    return emojis[key] || match
-  })
+export function convertToEmoji(html: string): string {
+  return html.replace(
+    /(<code[\s>][\s\S]*?<\/code>|<pre[\s>][\s\S]*?<\/pre>)|(:[\w+-]+:)/gi,
+    (match, codeBlock: string | undefined, shortcode: string | undefined) => {
+      if (codeBlock) return codeBlock
+      if (shortcode) {
+        const key = shortcode.slice(1, -1)
+        return emojis[key] ?? shortcode
```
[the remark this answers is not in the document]
gameroman: @coderabbitai, `:SMILE:` shouldn't convert, only `:smile:`

shared/utils/emoji.ts
```diff
@@ -1905,15 +1905,16 @@ const emojis = {
...
-  'g',
-)
-export function convertToEmoji(text: string): string {
-  return text.replace(emojisKeysRegex, match => {
-    const key = match.slice(1, -1) as keyof typeof emojis
-    return emojis[key] || match
-  })
+export function convertToEmoji(html: string): string {
+  return html.replace(
+    /(<code[\s>][\s\S]*?<\/code>|<pre[\s>][\s\S]*?<\/pre>)|(:[\w+-]+:)/gi,
```
ghostdevv, line 1910: is there another way that doesn't involve regex? wonder if elsewhere in the markdown rendering process we can hook into regular text rendering to apply the emojis but other things like code blocks just skip that pathway altogether?

ghostdevv (changes requested):

shared/utils/emoji.ts
gameroman: I'm not really sure if that's possible but would be good to do that if it is

shuuji3: The `convertToEmoji()` is only used in `readme.ts` here: https://github.com/npmx-dev/npmx.dev/blob/c8b60871b82208fa6dae360c21302a4d28184c27/server/utils/readme.ts#L796

Maybe we could replace `:` with `&colon;` via [`sanitizeHtml()`](https://github.com/npmx-dev/npmx.dev/blob/c8b60871b82208fa6dae360c21302a4d28184c27/server/utils/readme.ts#L686) or [`renderer.code`](https://github.com/npmx-dev/npmx.dev/blob/c8b60871b82208fa6dae360c21302a4d28184c27/server/utils/readme.ts#L596-L595) before passing html string to `convertToEmoji()`.

gameroman: > The `convertToEmoji()` is only used in `readme.ts` here:
> 
> https://github.com/npmx-dev/npmx.dev/blob/c8b60871b82208fa6dae360c21302a4d28184c27/server/utils/readme.ts#L796
> 
> Maybe we could replace `:` with `&colon;` via [`sanitizeHtml()`](https://github.com/npmx-dev/npmx.dev/blob/c8b60871b82208fa6dae360c21302a4d28184c27/server/utils/readme.ts#L686) or [`renderer.code`](https://github.com/npmx-dev/npmx.dev/blob/c8b60871b82208fa6dae360c21302a4d28184c27/server/utils/readme.ts#L596-L595) before passing html string to `convertToEmoji()`.

Conditionally based on if it's inside codeblock or not?

shuuji3: Yes, and if I understand correctly, marked library (or sanitizedHtml) should handle that. It should process markdown blocks and when it encounters a `<code>` block, it will pass the content within a code block to `renderer.code` handler.

gameroman: I'm not actually sure how this would work

shuuji3: Since this PR already fixes the problem, maybe we can deliver the current PR as is first, and iterate in the future PR when we find a better solution. What do you think? @ghostdevv

shuuji3 (approved):
I confirmed the fix 👍🏻 

- latest: https://main.npmx.dev/package/ipaddr.js#user-content-shared-properties
- PR preview: https://npmxdev-git-fork-gameroman-fix-emoji-in-codeblocks-npmx.vercel.app/package/ipaddr.js#user-content-shared-properties

ghostdevv (approved):

shared/utils/emoji.ts
ghostdevv: > Since this PR already fixes the problem, maybe we can deliver the current PR as is first, and iterate in the future PR when we find a better solution. What do you think? @ghostdevv

for sure! https://github.com/npmx-dev/npmx.dev/issues/2822

PR #2696, opened 2026-05-09
refactor: useColors composable for charts

graphieros:
This refactors the `useColors` composable to remove code repetition in chart components by placing used css variables directly into the composable.  Also adds a test for the composable.

While I was at it, I also fixed another regression that persisted the state of the zoom when granularity or date ranges are changed in download charts. While keeping the zoom state is necessary when mutating the tooltip position dynamically, it must however be refreshed when the number of datapoints changes. To test this:
- load a downloads chart (modal), switch to daily granularity, the zoom should be reset (meaning: all the range is visible)
- refresh the page, the chart modal opens on the same date range selection and granularity, and the zoom is also reset


serhalp (approved):
Very nice!

PR #2697, opened 2026-05-09
chore: upgrade vite-plus to 0.1.20 and fix surfaced a11y violations

fengmk2:
### 📚 Description

Upgrade vite-plus and related packages from 0.1.16 to 0.1.20 (latest stable), and fix the 14 WCAG AA color-contrast violations that the upgrade exposed in `test/nuxt/a11y.spec.ts`.

The strongest candidate for the behavior change is upstream vitest-dev/vitest#10049 ("fix(browser): spread user server options into browser Vite server in project"), shipped in `vitest@4.1.4`. 
Before that fix, createBrowserServer silently dropped user `server.*` options, so the browser-mode Vite server didn't fully serve UnoCSS-generated CSS and axe defaulted to plain black-on-white. 
After it, styles render correctly and axe surfaces pre-existing contrast issues. vite-plus 0.1.16 bundles vitest 4.1.2 (without the fix); 0.1.20 bundles 4.1.5 (with it). 
The violations are real and pre-date this PR — the upgrade only stops masking them.

Source-level a11y fixes:
- LinkBase: disabled buttons drop `opacity-50 + text-fg/bg-fg (3.69:1)` for `text-fg-muted / bg-fg-muted` tokens (~8:1) and add aria-disabled.
- DiffSidebarPanel + DiffMobileSidebarDrawer: stats and semver badges use `text-{green,red,yellow,purple}-700 dark:text-{...}-400`, matching the existing `Alert.vue` convention.
- DiffSkipBlock: drop redundant `opacity-50; text-fg-muted` on `bg-bg-muted` already meets contrast.
- DiffLine: line-number cell uses `text-fg-subtle (5.5:1+)` instead of `opacity-50` over inherited text (3.69:1).
- main.css: light-mode --accent default `oklch(0.5 0.16 247.27)` so text-accent on `bg-fg/10` reaches ~5:1 (was 4.19); user-set --accent-color overrides remain unaffected.

### DIFF
--- a/app/assets/main.css
+++ b/app/assets/main.css
@@ -95,11 +95,11 @@
   --border-subtle: oklch(0.922 0 0);
   --border-hover: oklch(0.715 0 0);
 
-  --accent: var(--accent-color, oklch(0.53 0.16 247.27));
+  --accent: var(--accent-color, oklch(0.5 0.16 247.27));
   --accent-muted: var(--accent-color, oklch(0.205 0 0));
 
   /* accent colors */
-  --swatch-sky: oklch(0.53 0.16 247.27);
+  --swatch-sky: oklch(0.5 0.16 247.27);
   --swatch-coral: oklch(0.56 0.17 10.75);
   --swatch-amber: oklch(0.58 0.18 46.34);
   --swatch-emerald: oklch(0.51 0.13 162.4);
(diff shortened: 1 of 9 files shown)


serhalp (approved):
LGTM, thank you for this!

PR #2704, opened 2026-05-10
fix(i18n): update ukrainian translations

Sasha125588:
### 🧭 Context

The Ukrainian locale was missing some translation keys compared to the English source locale. Some existing pluralized strings also needed updates to better match Ukrainian plural forms.

### 📚 Description

Updates `uk-UA` translations, including missing keys, plural forms, and the Ukrainian pluralization mapping.

Verification:

- `pnpm vp run i18n:check uk-UA`

### DIFF
--- a/config/i18n.ts
+++ b/config/i18n.ts
@@ -252,7 +252,7 @@ const locales: (LocaleObjectData | (Omit<LocaleObjectData, 'code'> & { code: str
     code: 'uk-UA',
     file: 'uk-UA.json',
     name: 'Українська',
-    pluralRule: createPluralRule('uk-UA', { zero: 0, one: 1, two: 0, few: 2, many: 3, other: 4 }),
+    pluralRule: createPluralRule('uk-UA', { zero: 2, one: 0, two: 1, few: 1, many: 2, other: 3 }),
   },
   /*{
       code: 'ru-RU',
--- a/i18n/locales/uk-UA.json
+++ b/i18n/locales/uk-UA.json
@@ -12,7 +12,7 @@
   "non_affiliation_disclaimer": "не пов'язаний з npm, Inc.",
   "trademark_disclaimer": "npm є зареєстрованою торговою маркою npm, Inc. Цей сайт не пов'язаний з npm, Inc.",
   "footer": {
-    "about": "про проект",
+    "about": "про проєкт",
     "blog": "блог",
     "docs": "документація",
     "source": "код",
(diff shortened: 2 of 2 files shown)


config/i18n.ts
```diff
@@ -252,7 +252,7 @@ const locales: (LocaleObjectData | (Omit<LocaleObjectData, 'code'> & { code: str
     code: 'uk-UA',
     file: 'uk-UA.json',
     name: 'Українська',
-    pluralRule: createPluralRule('uk-UA', { zero: 0, one: 1, two: 0, few: 2, many: 3, other: 4 }),
+    pluralRule: createPluralRule('uk-UA', { zero: 2, one: 0, two: 1, few: 1, many: 2, other: 3 }),
```
ghostdevv, line 255: Could you explain this change? @Sasha125588

ghostdevv (changes requested):

config/i18n.ts
Sasha125588: > Could you explain this change? @Sasha125588

I changed it because the updated Ukrainian plural strings use the order:

`one | few | many | other`

`Intl.PluralRules('uk-UA')` returns `one`, `few`, `many`, or `other` for Ukrainian counts, so those categories should map to indexes `0`, `1`, `2`, and `3`.

`0` returns `many` in Ukrainian (`0 пакетів`), so a separate zero form is not needed. `two` is not returned for `uk-UA` either; I mapped it to the same form as `few` as a fallback.

i18n/locales/uk-UA.json
```diff
@@ -76,6 +83,113 @@
...
+    "diff": {
+      "merge_modified_lines": "Об'єднати змінені рядки",
+      "word_wrap": "Перенесення рядків"
+    },
+    "version": {
+      "label": "{version}"
+    },
+    "status": {
+      "available_in_context": "{context}. Доступна {count} команда | {context}. Доступно {count} команди | {context}. Доступно {count} команд | {context}. Доступно {count} команди",
+      "matching_in_context": "{context}. {count} відповідна команда | {context}. {count} відповідні команди | {context}. {count} відповідних команд | {context}. {count} відповідні команди",
```
ghostdevv:
```suggestion
      "available_in_context": "{context}. Доступна {count} команда | {context}. Доступні {count} команди | {context}. Доступно {count} команд",
      "matching_in_context": "{context}. {count} відповідна команда | {context}. {count} відповідні команди | {context}. {count} відповідних команд",
```

The AI suggested this change - does this make sense?

ghostdevv (changes requested):

i18n/locales/uk-UA.json
Sasha125588: > The AI suggested this change - does this make sense?

Yes, this makes sense and reads better.

For command counts we only need the integer Ukrainian forms. `other` is only relevant for fractional values, which should not happen for command counts.

ghostdevv (approved):

PR #2714, opened 2026-05-11
feat(ui): highlight files commonly shipped unnecessarily in packages

### LINKED ISSUE #2582: Highlight files in package file browser that may be included accidentally/unnecessarily
### Describe the feature

In the package code browser, highlight files which may be accidental/unnecessary inclusions in the released package. This would help prevent package bloat / make it easier to identify.

For example, generally you wouldn't want to ship a `.claude/settings.json` or `.claude/settings.local.json` file in your package. You probably also don't want to ship a `.vscode/launch.json` file, or `.vscode/settings.json`. These files are generally meant for local development.

Maybe also surface this information elsewhere, to show optimization opportunities in the package, but I'm not sure how/where that should be put. It's not necessarily a huge problem to include these files (usually they are small / don't have anything sensitive). And I don't want to end up with people making tons of issues for maintainers like "why do you include this file that's 100 bytes?", but it'd be nice to surface somehow.

Directories like `test/` or `tests/` are generally not meant to be included in the published package, and could be highlighted as well.

For example, the resolve package has a `claude` directory, tests, and `.editorconfig`: https://npmx.dev/package-code/resolve/v/1.22.12

and ts-log v2.x had `.vscode/settings.json` (which actually caused me weird problems with Claude Code's sandbox being unwilling to even `pnpm install` due to files named `settings.json` not being allowed to be created/edited by the sandbox process): https://npmx.dev/package-code/ts-log/v/2.2.7/.vscode

---

List of possible inclusions for the files/directories to check and warn on:

- `.editorconfig`
- `.github/` (except maybe FUNDING.yml? idk if that's something that is used by package managers)
- `.vscode/`
- `.prettierignore`
- `.eslintignore`
- `.gitignore`
- `.gitattributes`
- `test/`, `tests/`, etc.
- `.claude/`
- `tsconfig.json`
- `.eslintrc` / `eslint.config.js` / `eslint.config.mjs` / etc
- `oxlint.config.ts` / `.oxlintrc.json` / `oxfmt.config.ts` / `.oxfmtrc.json`
- `prettierrc.yml` / `.prettierrc` / `prettierrc.json` / etc
- `.node-version`
- `mise.toml`
- `.env`
- `.env.local`
- `.nycrc` / `nyc.json`, etc

### Additional information

### Final checks

mvanhorn:
### 🔗 Linked issue

Closes #2582

### 🧭 Context

The package code browser shows every file shipped in the published tarball, but offers no signal that some of those files are usually not meant to be there. The issue lists a concrete set of patterns -- editor configs, lint/format configs, test directories, `.claude/`, `.github/`, local env files -- that authors usually intend to keep out of their publish.

### 📚 Description

The file tree and the directory listing now flag those patterns with a small `i-lucide:info` indicator and an aria-label. The icon and existing layout aren't changed otherwise, and the predicate is case-sensitive to match npm's POSIX filename semantics.

The matching rules live in a new `app/utils/package-content-hints.ts` so follow-ups (the "surface this elsewhere" half of the issue, or carve-outs like `.github/FUNDING.yml`) have a single place to extend. Unit tests cover the exact filenames, the regex patterns for ESLint/Prettier/oxlint/oxfmt configs, the directory matches, and obvious negatives like `index.js`, `package.json`, `src/`.

i18n: added `code.possibly_unnecessary` to `en.json`, mirrored the English placeholder into every locale file that has a `code` section, and extended `i18n/schema.json` to match.

### Verification

- `pnpm test test/unit/app/utils/package-content-hints.spec.ts`
- `pnpm vp lint`
- `pnpm vp run i18n:check`


i18n/locales/ar-EG.json
```diff
@@ -230,7 +230,8 @@
   },
   "code": {
     "binary_file": "ملف ثنائي",
-    "binary_rendering_warning": "لا يمكن عرض هذا الملف الثنائي في المتصفح"
+    "binary_rendering_warning": "لا يمكن عرض هذا الملف الثنائي في المتصفح",
+    "possibly_unnecessary": "May be unnecessary in a published package"
```
gameroman: From what I remember, English text shouldn't be added to other locales when adding a new key

app/components/Code/FileTree.vue
```diff
@@ -11,6 +12,8 @@ const props = defineProps<{
   depth?: number
 }>()
 
+const { t } = useI18n()
```
graphieros: You can remove this import and use `$t` directly

graphieros (changes requested):

mvanhorn:
Thanks @gameroman and @graphieros, addressed both in 56dbe35:

- Dropped the `possibly_unnecessary` English fallback from all 31 non-en locale files. `en.json` is unchanged; the runtime will fall back to source until contributors add native translations.
- Removed `const { t } = useI18n()` from `FileTree.vue` and switched the four template bindings to `$t('code.possibly_unnecessary')` directly.

Merge conflict against main is the only thing still outstanding; I'll rebase separately.

gameroman:
Looks nice

Here are some more files/dirs ideas that could be added:

- any file/directory that starts with a dot (`.xyz`) as usually they're config files
- any directory that is `__xyz__` - usually tests, etc - probably unintentional


Also we could highlight the file/dir name as yellow, like this for example

{D798DEEA-9826-4AAE-A84A-8E270C7D1C4E}

gameroman:
Also `.travis.yml` probably

And lockfiles like `bun.lock`, `bun.lockb`, `package-lock.json`, `pnpm-lock.yanl`


I'd also probably make text color a bit more dark/orange, something like this maybe

image




Also looks like https://npmx-m4m1wuo4h-npmx.vercel.app/package/eslint-import-resolver-node isn't loading anymore (https://main.npmx.dev/package/eslint-import-resolver-node does)

image

gameroman:
Looks very good overall, I love it

graphieros (changes requested):
Coverage is ok, but I think it could be completed with test cases for:

'.env.development.local'
'.env.test'
'.env.production.local'

Also, only a few pattern extensions are tested. These are not tested:

eslint.config.cjs
eslint.config.mts
eslint.config.cts
.prettierrc.toml
prettier.config.mjs
oxlint.config.cjs
.oxlintrc.yaml
oxfmt.config.mjs
.oxfmtrc.yml

mvanhorn:
@graphieros @gameroman thanks for the reviews. Pushed 33f4ff7 with the changes:

- `__xyz__` directories now match via a regex pattern (catches `__mocks__`, `__snapshots__`, `__fixtures__`, etc. without enumerating each).
- Added three conservative dot-prefixed config patterns: `.babelrc`/`.browserslistrc`/`.stylelintrc`-style `.xrc` and `.xrc.{json,js,yml,...}` files, plus `.x.config.{js,ts,...}`. Negative lookaheads explicitly exclude `.npmrc` since it's sometimes an intentional shipped artifact. Pinned `.npmrc → false` in the unit test so future broadening can't silently break the exclusion.
- Name spans in `FileTree.vue` and `DirectoryListing.vue` now get `text-yellow-600 dark:text-yellow-400` when `isPossiblyUnnecessaryContent` is true, alongside the existing amber info icon.

Didn't include a blanket `^\..+` rule for all dotfiles - too much false-positive risk (.npmrc, .gitkeep, occasional `.well-known` style files). Happy to add more patterns if you have specific ones in mind.

graphieros (approved):
Test coverage now LGTM 🌿

gameroman (approved):
Looks fine, can iterate later 👍

mvanhorn:
Added in dd10ece: test cases for .env.development.local, .env.test, and .env.production.local, plus broadened extension coverage - each config family (eslint, prettier, oxlint/oxfmt, stylelint, tailwind) now iterates its full extension set with matching negative cases so the pattern boundaries are exercised, not just a couple of extensions. The pattern source already matched the .env variants, so no source change was needed. Verified: the spec passes (12/12) at 100% coverage of package-content-hints.

mvanhorn:
Appreciate the merge, @gameroman - hope the shipped-files highlighting nudges some slimmer packages.

PR #2722, opened 2026-05-11
fix: avoid false license-change warning on first release

### LINKED ISSUE #2720: Package page shows license changed warning for new package
### Describe the bug

On my [package](https://npmx.dev/package/vsxtools/v/0.0.1)'s page, there is a warning about a license change since the previous version, but there is no previous version for the package.

### Additional context

I found the bug in https://github.com/npmx-dev/npmx.dev/blob/92cbee76e549ebe92a21cadca1be3786c115c69f/server/api/registry/license-change/%5B...pkg%5D.get.ts#L46.

I might make a PR if I have time.

### Logs

```shell-script

```

stellhub:
## Summary
- avoid reporting a license change when the current package version is the first published release
- keep the existing comparison behavior when a real previous version exists
- add focused unit coverage for the first-release and explicit-first-version cases

## Root cause
The license-change API always compared the current version against `currentVersionIndex - 1`. For a package with only one version, or when requesting the first published version explicitly, that previous index is negative and gets coerced into `UNKNOWN`, which incorrectly produces a "license changed" warning.

## Testing
- `node_modules\\.bin\\lex.CMD build --lexicons lexicons --out shared/types/lexicons --clear`
- `node_modules\\.bin\\vp.CMD test --project unit test/unit/server/api/registry/license-change/pkg.get.spec.ts`

Closes #2720.

### DIFF
--- a/server/api/registry/license-change/tsconfig.json
+++ b/server/api/registry/license-change/tsconfig.json
@@ -1,4 +1,0 @@
-{
-  "extends": "../../../../tsconfig.json",
-  "include": ["./*.ts"]
-}
--- a/server/api/registry/license-change/tsconfig.json
+++ b/server/api/registry/license-change/tsconfig.json
@@ -1,0 +1,6 @@
+{
+  "extends": "../../../../tsconfig.json",
+  "include": [
+    "./*.ts"
+  ]
+}
--- a/server/api/registry/license-change/tsconfig.json
+++ b/server/api/registry/license-change/tsconfig.json
@@ -1,6 +1,4 @@
 {
   "extends": "../../../../tsconfig.json",
-  "include": [
-    "./*.ts"
-  ]
+  "include": ["./*.ts"]
 }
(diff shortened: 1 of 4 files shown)


server/api/registry/license-change/tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
```
gameroman: This is not needed

test/unit/server/api/registry/license-change/tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
```
gameroman: This is not needed

gameroman (changes requested):

test/unit/server/api/registry/license-change/pkg.get.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+      },
+    })
+
+    await expect(handler(fakeEvent)).resolves.toEqual({
+      change: {
+        from: 'MIT',
+        to: 'Apache-2.0',
+      },
+    })
+  })
```
gameroman, line 93: `license` can be an object like
```js
  license: { type: 'MIT' }
```
could you add a test for that, please?

gameroman (changes requested):

ghostdevv:
We got quite a few PRs for this so I'm gonna close this one in favor of #2792 - thanks for your help, and we'll make sure you get credit when we merge :smile:

Issue #2630, opened 2026-04-26
Timeline page improvements


gameroman:
### Describe the feature

- [ ] Lower the threshold for decreased dependencies
I think celebrating improvements is always good so we should reduce the threshold

- [ ] For Install size increase/decrease show "previous vs now"
Currently it says `Install size increased by 41% (1.7 MB)`
What I think it could look like: `Install size increased by 41% (4.2 MB + 1.7 MB = 5.9 MB)` or similar <img width="426" height="43" alt="Image" src="https://github.com/user-attachments/assets/7267a984-78e7-4a4e-b200-a2cc1d8d6e25" />

- [ ] If size increase/decrease is `>= 3x`, show size increased/decreased `3x` instead of `by 200%`

- [ ] Show when a specific version of the packages is deprecated

- [ ] Celebrate 0 dependencies: if package had even 1 dependency and now it has 0, show that on timeline too

- [ ] Add some options to hide pre-release versions, filter by major versions
Currently it can be confusing: <img width="372" height="758" alt="Image" src="https://github.com/user-attachments/assets/aced6a5e-dc2d-4415-80d8-e9580ff2d0fb" />

### Additional information

- [x] Would you be willing to help implement this feature?

### Final checks

- [x] Read the [contribution guide](https://github.com/npmx-dev/npmx.dev/blob/main/CONTRIBUTING.md).
- [x] Check existing [issues](https://github.com/npmx-dev/npmx.dev/issues).

graphieros:
Note regarding future version filtering for the timeline list:
- if a filter is applied upstream to hide pre-release versions, the filter toggle on the chart might become redundant

mootari:
#822 might also be a good candidate for inclusion in the timeline.

PR #2728, opened 2026-05-12
feat: toggle stable versions in timeline chart

### LINKED ISSUE #2725: Timeline filters
Some users surfaced the need for filtering capabilities on the timeline.

For example: https://bsky.app/profile/styfle.dev/post/3mlmxtnoszk2k

Related to #2630

graphieros:
Follow up to #2663 
Resolves #2725 

This adds a toggle above the timeline chart to show only stable versions.
The toggle's state is persisted in the user settings.

When enabled, stable versions are reordered, and all prereleases are filtered out.
When disabled, the chart is the same as before and follows the list.

| Enabled (default)| Disabled (previous rendering, still available)|
|-|-|
|image|image|

Other:
- bump vue-data-ui to latest (just to stay fresh)


app/utils/versions.ts
```diff
@@ -39,6 +39,21 @@ export function parseVersion(version: string): ParsedVersion {
   }
 }
 
+/**
+ * Parse a semver stable version string into its components
+ * @param version - The version string (e.g., "1.2.3" or "1.0.0-beta.1")
+ * @returns Parsed version object with major, minor, patch or null
+ */
+export function parseStableVersion(version: string): Omit<ParsedVersion, 'prerelease'> | null {
+  const match = version.match(/^v?(\d+)\.(\d+)\.(\d+)(?:\+[\w.-]+)?$/)
```
43081j: we import `semver` at the top of this file, so i wonder... could we not use that to parse both this and `parseVersion`? 🤔

graphieros: @43081j awesome!
These regex are now semver's  business

43081j (approved):

styfle:
@graphieros I tried out the preview deployment and it looks like it’s filtering the graph, but it’s not filtering the version list underneath.

graphieros:
> @graphieros I tried out the preview deployment and it looks like it’s filtering the graph, but it’s not filtering the version list underneath.

@styfle Yes, this was the requirement we agreed upon with @43081j 
The toggles only affect the chart (quick win).
Filtering features for the timeline itself might be for a future iteration, as part of the parent #2630 issue

styfle:
@graphieros Another problem I noticed is that the number of versions with stable-only filter is really small.

Compare these two:

## npmx
https://npmx.dev/package-timeline/@vercel/sandbox/v/2.0.1
image

## packagephobia
https://packagephobia.com/result?p=@vercel/sandbox@2.0.1
image

That's because you need to fetch more data, from the list.
This would all be solved by filters for the list itself, so filtered data could be fetched directly. Again, this is part of improvements listed on the parent issue.
Kindly consider continuing this discussion on the discord server, I think it would be better suited than on a merged PR

PR #2733, opened 2026-05-13
fix: show 404 page instead of blank on first visit to non-existent package

### LINKED ISSUE #1602: Sometimes 404 page is not displaying correctly
### Describe the bug

We currently show an empty page if a package (or anything else for that matter) doesn't exist:

https://npmx.dev/package/thisdoesnotexistttt

### Additional context

_No response_

### Logs

```shell-script

```

2u841r:
When you visit a package URL that doesn't exist for the first time, Vercel serves an ISR fallback HTML while SSR runs in the background. On that first load the page was completely blank until you refreshed.

The cause was an `await` on `useResolvedVersion` in the component setup. On the client that suspends the component through Vue's Suspense, and since there is no fallback defined, nothing renders while the fetch is in flight.

The fix skips the await on the client. The server still awaits so the 404 can be thrown synchronously. On the client the existing watch already handles the error case and calls `showError`. Also added `resolvedStatus === 'pending'` to the skeleton condition so a loading state shows while the version fetch is running.

Fixes #1602


gameroman (commented):
This seems to work but looks like a workaround. Is there potentially a better solution?

2u841r:
Making `useResolvedVersion` itself lazy would fix the suspension, but lazy composables don't block SSR so server-side 404 detection breaks.

A route middleware using `$fetch` would also work, but it can't share the cache with the page's `useResolvedVersion` call, so we'd end up fetching `npm.antfu.dev` twice on every package load.

Skipping the `await` on client keeps the server blocking for synchronous 404 detection while letting the client render without suspension. The existing `watch` already handles the client-side error case.

Not 100% confident this is the cleanest fix. Would defer to your judgment if you see something better.

PR #2734, opened 2026-05-13
fix: parse hyphenated language names for code examples on package docs

### LINKED ISSUE #2437: docs feature codeblock hyphenated lang regex

lisiqi1983:
## Summary
- parse `@example` fence languages up to whitespace/backticks so hyphenated Shiki language ids are preserved
- add a regression test for a `glimmer-ts` fenced example

## Testing
- `STORYBOOK=true pnpm vp test --project unit test/unit/server/utils/docs/render.spec.ts`
- `pnpm vp lint server/utils/docs/render.ts test/unit/server/utils/docs/render.spec.ts`
- `pnpm vp fmt --check server/utils/docs/render.ts test/unit/server/utils/docs/render.spec.ts`

Fixes #2437

### DIFF
--- a/server/utils/docs/render.ts
+++ b/server/utils/docs/render.ts
@@ -191,9 +191,9 @@ async function renderJsDocTags(tags: JsDocTag[], symbolLookup: SymbolLookup): Pr
     : null
   const examplePromises = examples.map(async example => {
     if (!example.doc) return ''
-    const langMatch = example.doc.match(/```[ \t]*([^` \t\r\n]*)/)
+    const langMatch = example.doc.match(/```[ \t]*([-\w]+)?/)
     const lang = langMatch?.[1] || 'typescript'
-    const code = example.doc.replace(/```[ \t]*[^` \t\r\n]*[ \t]*(?:\r\n|\r|\n)?/g, '').trim()
+    const code = example.doc.replace(/```[ \t]*[-\w]*[ \t]*(?:\r\n|\r|\n)?/g, '').trim()
     return highlightCodeBlock(code, lang)
   })
 
--- a/server/utils/docs/render.ts
+++ b/server/utils/docs/render.ts
@@ -191,9 +191,9 @@ async function renderJsDocTags(tags: JsDocTag[], symbolLookup: SymbolLookup): Pr
     : null
   const examplePromises = examples.map(async example => {
     if (!example.doc) return ''
-    const langMatch = example.doc.match(/```(\w+)?/)
+    const langMatch = example.doc.match(/```[ \t]*([^` \t\r\n]*)/)
     const lang = langMatch?.[1] || 'typescript'
-    const code = example.doc.replace(/```\w*\n?/g, '').trim()
+    const code = example.doc.replace(/```[ \t]*[^` \t\r\n]*[ \t]*(?:\r\n|\r|\n)?/g, '').trim()
     return highlightCodeBlock(code, lang)
   })
 
--- a/test/unit/server/utils/docs/render.spec.ts
+++ b/test/unit/server/utils/docs/render.spec.ts
@@ -211,6 +211,7 @@ describe('renderDocNodes examples', () => {
     expect(html).toContain('<h4>Example</h4>')
     expect(html).toContain('shiki')
     expect(html).toContain('greeting')
+    expect(html).not.toMatch(/(^|[>\s])-ts([<\s]|$)/)
     expect(html).not.toContain('-ts')
     expect(html).not.toContain('```')
   })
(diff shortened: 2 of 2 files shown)


gameroman:
@coderabbitai full review

gameroman (approved):

server/utils/docs/render.ts
```diff
@@ -191,9 +191,9 @@ async function renderJsDocTags(tags: JsDocTag[], symbolLookup: SymbolLookup): Pr
     : null
   const examplePromises = examples.map(async example => {
     if (!example.doc) return ''
-    const langMatch = example.doc.match(/```(\w+)?/)
+    const langMatch = example.doc.match(/```[ \t]*([^` \t\r\n]*)/)
     const lang = langMatch?.[1] || 'typescript'
-    const code = example.doc.replace(/```\w*\n?/g, '').trim()
+    const code = example.doc.replace(/```[ \t]*[^` \t\r\n]*[ \t]*(?:\r\n|\r|\n)?/g, '').trim()
```
alexdln: Image

It seems to me that the new exp allows extra characters

If the previous one didn't include only the dash, I would suggest adding only it

Image

```ts
([\w\-]+)
```

Or
```ts
([\w\-_]+)
```

alexdln (changes requested):

server/utils/docs/render.ts
lisiqi1983: Thanks, good catch. I narrowed the fence language matching back to the original `\w` set plus `-`, so it no longer accepts arbitrary non-whitespace characters. I also tightened the regression assertion so it only checks for a leaked standalone `-ts` token instead of rejecting any incidental `-ts` substring.

Validated locally:
- `STORYBOOK=true pnpm vp test --project unit test/unit/server/utils/docs/render.spec.ts`
- `pnpm vp lint server/utils/docs/render.ts test/unit/server/utils/docs/render.spec.ts`
- `pnpm vp fmt --check server/utils/docs/render.ts test/unit/server/utils/docs/render.spec.ts`
- `git diff --check -- server/utils/docs/render.ts test/unit/server/utils/docs/render.spec.ts`

test/unit/server/utils/docs/render.spec.ts
```diff
@@ -192,3 +194,24 @@ describe('renderDocNodes ordering', () => {
...
+        },
+      ],
+    })
+
+    const html = await renderDocNodes([symbol], new Map())
+
+    expect(html).toContain('<h4>Example</h4>')
+    expect(html).toContain('shiki')
+    expect(html).toContain('greeting')
+    expect(html).not.toMatch(/(^|[>\s])-ts([<\s]|$)/)
```
gameroman, line 214: You can probably keep both tests? Or that wouldn't work?

```suggestion
    expect(html).not.toMatch(/(^|[>\s])-ts([<\s]|$)/)
    expect(html).not.toContain('-ts')
```

lisiqi1983:
Quick maintenance update: rebased onto latest `main` and adjusted the hyphen character class placement after the CodeRabbit nit. I used `[-\w]` rather than `[\w\-]` because the repository lint rule flags the escaped hyphen as `no-useless-escape`, while putting `-` first still makes it an explicit literal.

Validated locally:
- `STORYBOOK=true pnpm vp test --project unit test/unit/server/utils/docs/render.spec.ts`
- `pnpm vp lint server/utils/docs/render.ts test/unit/server/utils/docs/render.spec.ts`
- `pnpm vp fmt --check server/utils/docs/render.ts test/unit/server/utils/docs/render.spec.ts`
- `git diff --check -- server/utils/docs/render.ts test/unit/server/utils/docs/render.spec.ts`

test/unit/server/utils/docs/render.spec.ts
lisiqi1983: Yep, that works. I kept both assertions now: the narrower regex catches a standalone leaked suffix token, and the broader `not.toContain('-ts')` keeps the original regression guard.

Re-validated locally:
- `STORYBOOK=true pnpm vp test --project unit test/unit/server/utils/docs/render.spec.ts`
- `pnpm vp lint server/utils/docs/render.ts test/unit/server/utils/docs/render.spec.ts`
- `pnpm vp fmt --check server/utils/docs/render.ts test/unit/server/utils/docs/render.spec.ts`
- `git diff --check -- server/utils/docs/render.ts test/unit/server/utils/docs/render.spec.ts`

ghostdevv (approved):

ghostdevv:
@alexdln I think you need to re-review before we can merge :pray:

alexdln (approved):

PR #2736, opened 2026-05-13
chore: update blog post description

alexdln:
### 🧭 Context

I think I accidentally pressed tab with autocomplete and didn't double-check that section... Fixed og:description


patak-cat (approved):
thanks! 🙏🏼

graphieros (approved):

PR #2742, opened 2026-05-13
chore: improve toggle transition on touch devices

### LINKED ISSUE #2540: Switch icons get weird hover style on mobile
### Describe the bug

This is a fairly niche one but when pressing the switches on the settings page the switches keep a sticky hover/active style that darkens its icon.

In the screenshot below I've just interacted with the Relative Date switch.

Image

### Additional context

Chrome Android version: 149.0.7793.0 (Official Build) canary (64-bit) 

I also have dark mode and high contrast enabled on the device as that might help debug the specific state.

### Logs

```shell-script

```

alexdln:
### 🔗 Linked issue

Resolves #2540

### 🧭 Context

A micro-change to improve hover effects. I simply disabled this effect on touch devices (for checked checkbox on hover). Tailwind seems to be able to do this automatically, but it doesn't work for us. Current change should be more than enough for now

### DIFF
--- a/app/components/Settings/Toggle.client.vue
+++ b/app/components/Settings/Toggle.client.vue
@@ -35,62 +35,32 @@ const id = useId()
         : 'grid-template-areas: \'label-text . toggle\''
     "
   >
-    <template v-if="props.reverseOrder">
-      <input
-        role="switch"
-        type="checkbox"
-        :id="id"
-        v-model="checked"
-        class="toggle appearance-none h-6 w-11 rounded-full border border-fg relative shrink-0 bg-fg/50 transition-colors duration-200 ease-in-out checked:bg-fg [@media(hover:hover)]:hover:bg-fg/60 [@media(hover:hover)]:checked:hover:(bg-fg/80 after:opacity-50) focus-visible:(outline-2 outline-accent outline-offset-2) before:(content-[''] absolute h-5 w-5 top-px start-px rounded-full bg-bg transition-transform duration-200 ease-in-out) checked:before:translate-x-[20px] rtl:checked:before:-translate-x-[20px] after:(content-[''] absolute h-3.5 w-3.5 top-[4px] start-[4px] i-lucide:check bg-fg opacity-0 transition-all duration-200 ease-in-out) checked:after:opacity-100 checked:after:translate-x-[20px] rtl:checked:after:-translate-x-[20px]"
-        style="grid-area: toggle"
-      />
-      <TooltipApp
-        v-if="tooltip && label"
-        :text="tooltip"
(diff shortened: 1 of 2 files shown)


app/components/Settings/Toggle.client.vue
gameroman: Same here?

alexdln: Oops, good catch, thanks, done

gameroman: @alexdln It looks like both classes are identical. Could it be simplified/deduplicated somehow to not have to update both at the same time and not forget to?

alexdln: I'm honestly not even sure why there are two variations here... It seems like it could be solved with classes. I'll check it out tomorrow 👀

alexdln: Yes, this was already done via styles before, but it still changed the semantic order. Essentially, it was only for crawlers/readers. And I think it's unnecessary, since even for a11y, it's generally more useful to leave the text at the beginning. _Removed condition_

app/components/Settings/Toggle.client.vue
```diff
@@ -35,62 +35,32 @@ const id = useId()
...
+      style="grid-area: label-text"
+    >
+      {{ label }}
+    </span>
+    <input
+      role="switch"
+      type="checkbox"
+      :id="id"
+      v-model="checked"
+      class="toggle appearance-none h-6 w-11 rounded-full border border-fg relative shrink-0 bg-fg/50 transition-colors duration-200 ease-in-out checked:bg-fg [@media(hover:hover)]:hover:bg-fg/60 [@media(hover:hover)]:checked:hover:(bg-fg/80 after:opacity-50) focus-visible:(outline-2 outline-accent outline-offset-2) before:(content-[''] absolute h-5 w-5 top-px start-px rounded-full bg-bg transition-transform duration-200 ease-in-out) checked:before:translate-x-[20px] rtl:checked:before:-translate-x-[20px] after:(content-[''] absolute h-3.5 w-3.5 top-[4px] start-[4px] i-lucide:check bg-fg opacity-0 transition-all duration-200 ease-in-out) checked:after:opacity-100 checked:after:translate-x-[20px] rtl:checked:after:-translate-x-[20px]"
```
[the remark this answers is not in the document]
alexdln: It works, but it seems like the media itself is broken in chrome [for hybrid devices]. Everything is fine in firefox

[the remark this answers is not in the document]
alexdln: UPD: Oh, I updated chrome and now everything is fixed...

graphieros (approved):
LGTM 🌿

PR #2743, opened 2026-05-14
fix: show 404 for missing package versions

### LINKED ISSUE #1602: Sometimes 404 page is not displaying correctly

mzl2233:
### 🔗 Linked issue

Resolves #1602

### 🧭 Context

Missing packages can resolve through fast-npm-meta as version `0.0.0`, which lets the package page continue rendering with placeholder data instead of reliably surfacing the not-found state.

### 📚 Description

This change treats the `0.0.0` fallback from the resolved-version request as a not-found result and raises a 404 immediately. That keeps missing package URLs, including uncached ones with cache-busting query strings, on the normal 404 error page instead of leaving the user with a blank package page.

### DIFF
--- a/app/composables/npm/useResolvedVersion.ts
+++ b/app/composables/npm/useResolvedVersion.ts
@@ -13,6 +13,11 @@ export function useResolvedVersion(
         ? `https://npm.antfu.dev/${name}@${version}`
         : `https://npm.antfu.dev/${name}`
       const data = await $fetch<ResolvedPackageVersion>(url)
+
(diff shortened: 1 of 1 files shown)


gameroman (changes requested):
This does not solve the bug

PR #2202, opened 2026-03-22
feat: add cursor-pointer to ButtonBase component

thegreatalxx:
### 🔗 Linked issue
resolves #1760

### 🧭 Context
The `ButtonBase` component (`app/components/Button/Base.vue`) lacked an explicit `cursor-pointer` class. Browsers default to `cursor: auto` for `<button>` elements styled with Tailwind/UnoCSS resets, so clickable buttons appeared with the default arrow cursor instead of the expected pointer.

### 📚 Description
Added `cursor-pointer` to the button's static class list so all enabled `ButtonBase` instances show the pointer cursor. The existing `disabled:(cursor-not-allowed)` modifier is preserved and continues to override on disabled buttons.

A regression test (`test/nuxt/components/ButtonBase.spec.ts`) mounts the component and asserts that `cursor-pointer` is present on the enabled button and `cursor-not-allowed` is in the class string for the disabled state.

PR #2752, opened 2026-05-15
feat: set cursor-pointer on buttons

### LINKED ISSUE #1760: Use pointer cursor for interactive elements

aryanpingle:
### 🔗 Linked issue

Partially resolves #1760

### 🧭 Context

[This commit](https://github.com/thegreatalxx/npmx.dev/commit/391d28e215c5b22dfab3812877f3ed8480349a12) removed `cursor: pointer` from `ButtonBase`. The reason being, [the W3C spec says](https://www.w3.org/TR/css-ui-3/#ref-for-valdef-cursor-pointer):

> The cursor is a pointer that indicates a link.

The commit was on Valentine's Day, but it is far from being something I love.

`CONTRIBUTING.md` says:

> **npmx** uses `cursor: pointer` only for links to match users’ everyday experience. For all other interactive elements, including buttons, use the default cursor (_or another appropriate cursor to indicate state_).

And the linked issue states in its conclusion:

> Based on all this, we're returning the pointer cursor for all interactive elements (without any distinction or definition of what a link is and what a button is). This needs to be updated in the contributing guide and updated throughout our site.

There was [a similar PR (#2303)](https://github.com/npmx-dev/npmx.dev/pull/2303) that attempted to do this only for the Download button.
And [another PR (#2202)](https://github.com/npmx-dev/npmx.dev/pull/2303) which was closed for potential spam.

To make things worse, at the moment some buttons and dropdowns arbitrarily have `cursor: pointer` enabled, while more deserving buttons don't:

image

1. Download button - ❌No pointer cursor
2. Package manager dropdown button - ✔`cursor: pointer` AND a fancy mousepress animation
3. Copy button - ❌No pointer cursor
4. Sections dropdown button - ❌No pointer cursor

If we're being inconsistent, let's at least err on the side of a better user experience.

I get that the plan was to adhere to the specification / default behaviour, but consider this: if I cared about being "official" more than the UX, I would be using npmjs.com instead of this awesome site.

### 📚 Description

This PR adds `cursor: pointer` to some of the pure buttons that one is likely to encounter in the first few pages. Normal buttons + the command palette. My philosophy being, "A pointer cursor indicates that something will happen on clicking the element".

I specifically didn't add it to the dropdowns in case that's another point of contention. I just wanna get this shipped soon so I can enjoy it as an end-user.

_W3C this, W3C that. WHO CARES, THE POINTER IS WATCHING. I feel like I'm taking crazy pills, but I'm not because I ran out of crazy pills, I need more._

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -452,7 +452,7 @@ The command palette is a first-class navigation surface. When you add a new user
 
 ### Cursor and navigation
 
-**npmx** uses `cursor: pointer` only for links to match users’ everyday experience. For all other interactive elements, including buttons, use the default cursor (_or another appropriate cursor to indicate state_).
+**npmx** uses `cursor: pointer` for links and buttons to match users’ everyday experience. For all other interactive elements, use the default cursor (_or another appropriate cursor to indicate state_).
 
 > [!NOTE]
 > A link is any element that leads to another content (_go to another page, authorize_)
--- a/app/components/Button/Base.vue
+++ b/app/components/Button/Base.vue
@@ -44,7 +44,7 @@ defineExpose({
 <template>
   <button
     ref="el"
-    class="group gap-x-1 items-center justify-center font-mono border border-border rounded-md transition-all duration-200 disabled:(opacity-40 cursor-not-allowed border-transparent)"
+    class="group gap-x-1 items-center justify-center font-mono border border-border rounded-md transition-all duration-200 cursor-pointer disabled:(opacity-40 cursor-not-allowed border-transparent)"
     :class="{
       'inline-flex': !block,
       'flex': block,
--- a/app/components/Compare/FacetBarChart.vue
+++ b/app/components/Compare/FacetBarChart.vue
@@ -152,6 +152,7 @@ const config = computed<VueUiHorizontalBarConfig>(() => {
           })
         },
       },
+      useCursorPointer: true,
     },
     skeletonDataset: skeletonDataset.value,
     skeletonConfig: {
--- a/app/components/Compare/FacetScatterChart.vue
+++ b/app/components/Compare/FacetScatterChart.vue
@@ -166,6 +166,7 @@ const config = computed<VueUiScatterConfig>(() => {
           })
         },
       },
+      useCursorPointer: true,
     },
     style: {
       backgroundColor: colors.value.bg,
--- a/app/components/InstantSearch.vue
+++ b/app/components/InstantSearch.vue
@@ -29,7 +29,11 @@ onPrehydrate(el => {
           <strong>{{ $t('search.instant_search_on') }}</strong>
         </template>
         <template #action>
-          <button type="button" class="underline" @click="settings.instantSearch = false">
+          <button
+            type="button"
+            class="underline cursor-pointer"
+            @click="settings.instantSearch = false"
+          >
             {{ $t('search.instant_search_turn_off') }}
           </button>
         </template>
@@ -44,7 +48,11 @@ onPrehydrate(el => {
           <strong>{{ $t('search.instant_search_off') }}</strong>
         </template>
         <template #action>
-          <button type="button" class="underline" @click="settings.instantSearch = true">
+          <button
+            type="button"
+            class="underline cursor-pointer"
+            @click="settings.instantSearch = true"
+          >
             {{ $t('search.instant_search_turn_on') }}
           </button>
         </template>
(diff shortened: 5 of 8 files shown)


graphieros (commented):
@aryanpingle 🙏 thank you for your PR, it's been a while indeed since the decision was taken to have the pointer on buttons.

I think the PR should solve all the cases to be merged, to avoid a partial situation which would remain confusing for users. Since this is my opinion, and perhaps other maintainers have a different one, I'm just commenting and not requesting changes.

aryanpingle:
@graphieros I get what you mean about the partial solution problem, in that it may be confusing for users to see some buttons have `default` cursors and `pointer` for others.
That said, this is **already** the case, as in the screenshot I've attached as well as other places on the site. 

---

For a comprehensive fix across all buttons, there are two scenarios:

1. Setting `cursor: pointer` on the `ButtonBase` component - this will affect every single button, and we can simply revert this PR
2. Setting `cursor: pointer` on each button as needed - it will be an extension of this PR (since I've not covered every case)

So from a tech standpoint, it's either easily revertible or a first step.
But from a UX standpoint, it's an improvement that we either have today or sometime down the line.
And as someone who uses the site almost daily, I'd rather have it sooner than later.

Happy to hear any thoughts!

graphieros:
@aryanpingle I feel well aligned with your approach.
@knowler, @alexdln, @MatteoGabriele what do you think ?

MatteoGabriele:
My 2 cents: if we are adding this, let's add it to the main button component and call it a day.

aryanpingle:
Music to my ears :)

I've set it on the `ButtonBase` component, and taken the liberty of updating `CONTRIBUTING.md`. Lmk how it looks.

Edit: And on the instant search button separately, since it's not using `ButtonBase`

graphieros:
@aryanpingle 

Ideally, the change mentioned [here](https://github.com/npmx-dev/npmx.dev/issues/1760#issuecomment-3977563885) for the charts should be part of this PR, but it can be done on a next iteration as well.

aryanpingle:
@graphieros added those changes here, thanks for pointing them out.

I couldn't really test the chart changes as we don't have any clickable chart elements (afaik), but at least any future elements will automatically be cursor-pointer'd

graphieros:
> @graphieros added those changes here, thanks for pointing them out.
> 
> I couldn't really test the chart changes as we don't have any clickable chart elements (afaik), but at least any future elements will automatically be cursor-pointer'd

No problem, I'll check (the code looks good, I'll check manually, it's quick)

graphieros:
@aryanpingle the failing browser test is due to a flaky test unrelated to your changes. It has been notified to its author

graphieros (approved):
LGTM 🌿

gameroman (approved):

PR #2756, opened 2026-05-16
feat: update team list and design on about page

alexdln:
### 📚 Description

Since we added a separate core role, there are multiple roles in one group now. Divided it into two groups and updated the cards and contributors style to make everything more streamlined and neat

<details>
<summary>Screenshot</summary>
image
</details>


ghostdevv (changes requested):
Not sure that "Governance" is the right word for stewards/core as it encompasses all the groups incl maintainers - we could write just "Core" as the title, and keep daniel/matias with the "steward" on their card

alexdln:
@ghostdevv Thanks, good point, done

ghostdevv (approved):

PR #2758, opened 2026-05-16
fix: package comparison broken on org and user pages

### LINKED ISSUE #2746: Package comparison UI is nonfunctional on org and user pages
### Describe the bug

Search pages allow users to select and compare packages. The same selection is possible on org and user pages, but
- on org pages only the "View selected" button is shown, and clicking it has no effect
- on user pages neither the "View selected" button nor the "Compare" button are shown

Steps:
1. Go to https://npmx.dev/org/antfu and select 1 or more packages
2. Observe that clicking "View selected" has no effect
1. Go to https://npmx.dev/~antfu and select 1 or more packages
2. Observe that no buttons appear

|Search|Org|User
|-|-|-
|Image|Image|Image

### Additional context

The easiest short term fix might be to limit package selection to the search and not render checkboxes anywhere else.

MatteoGabriele:
### 🔗 Linked issue

closes #2746

### 🧭 Context

Selection in org and user pages was enabled, but neither an action bar nor the selected packages page was displayed.

### 📚 Description

The action bar and selection state are now correctly rendered on the org page, but on the user page, I chose to omit them entirely. The UI differs significantly, and adding the missing toolbar would have required a major refactor. The user page can also be seen as a different type of page, so I think that's not a bad idea. Let me know what you think.

I implemented a default hidden selection state and activated selection only where necessary. 
Due to the complex component structure and to avoid excessive prop-drilling, I opted to use the provide/inject pattern for easy access to elements deeper in the component tree.

### DIFF
--- a/app/components/Package/Card.vue
+++ b/app/components/Package/Card.vue
@@ -16,6 +16,7 @@ const props = defineProps<{
   searchQuery?: string
 }>()
 
+const { selectable } = usePackageSelectionContext()
 const { isPackageSelected, togglePackageSelection, canSelectMore } = usePackageSelection()
 const isSelected = computed<boolean>(() => {
   return isPackageSelected(props.result.package.name)
@@ -65,6 +66,7 @@ const numberFormatter = useNumberFormatter()
       </component>
 
       <PackageSelectionCheckbox
+        v-if="selectable"
         :package-name="result.package.name"
         :disabled="!canSelectMore && !isSelected"
         :checked="isSelected"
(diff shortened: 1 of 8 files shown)


gameroman (approved):
Looks good overall ✅

Might make sense to add some tests to prevent future regressions

The coderabbit suggestions should be addressed probably too

graphieros (commented):
LGTM 🌿

graphieros (approved):

PR #2763, opened 2026-05-17
feat: strip version info from the search query

Codefoxdev:
See #1626

### 🔗 Linked issue

Partially addresses #1416

### 🧭 Context

Currently, when searching for a package along with version info (e.g. when copying a package name from the output of a package manager) it just returns random results, instead of useful ones. (e.g. https://npmx.dev/search?q=nuxt@3.4.1)

This pr strips the version info from the query to improve search results.

### 📚 Description

This pr implements a new regex query to extract essential date from the search query on the search page, such as package scope (which replaces an existing method), package specifier and (optionally) version info. The query is then reconstructed without version info to support the existing search api. It can currently parse a package name in the following format:
```
@scope/specifier@version
```

An object containing the separate parts of the query is exposed as the `parsedQuery` ref and it is able to be used for other features. (Such as implementing the other part of the linked issue)


Codefoxdev:
I resolved the merge conflict on github but I still need to change something that depends on the upstream changes, how can i now merge these changes? Sorry, but I'm a bit new to this...

I will then also update the linting issue

gameroman:
> I resolved the merge conflict on github but I still need to change something that depends on the upstream changes, how can i now merge these changes? Sorry, but I'm a bit new to this...
> 
> I will then also update the linting issue

Everything looks good, I think you can just fix the lint warning and the coderabbit's suggestion and this is good to go 👍

Codefoxdev:
I still need to update the composable to use the new committedquery, otherwise this will break that behaviour I think?

gameroman:
> I still need to change something that depends on the upstream changes

From what I see your branch is already up-to-date with the main branch, so I think you should already be able to make those changes

Codefoxdev:
My bad, I forgot to update my local repo.
Everything should be resolved now!

gameroman (approved):

PR #2770, opened 2026-05-19
fix: package sub-banner no longer blocks "connect" dropdown

### LINKED ISSUE #2747: "connect" dropdown is partially behind the header
> There's another z-index issue in the code browser that may be related:
> 
> Image 

 _Originally posted by @mootari in [#2700](https://github.com/npmx-dev/npmx.dev/issues/2700#issuecomment-4425603284)_

ptdewey:
### 🔗 Linked issue

Fixes https://github.com/npmx-dev/npmx.dev/issues/2747

### 🧭 Context

The "connect" dropdown menu items were obscured by the banner on the package page:

image

image

### 📚 Description

This change adjusted the z-index of the banner from z-50 to z-40 to ensure the dropdown is drawn above the banner.
As requested in #2747, I also added a new test case to catch regressions of this (it fails w/o 
the change and passes with it).

Please let me know if there are any issues, thanks!

### DIFF
--- a/test/e2e/interactions.spec.ts
+++ b/test/e2e/interactions.spec.ts
@@ -139,6 +139,40 @@ test.describe('Package Page', () => {
       ).toBe(true)
     }
   })
+
+  test('AccountMenu dropdown items are clickable on package pages', async ({ page, goto }) => {
+    await goto('/package/vue', { waitUntil: 'hydration' })
+
+    await page.getByRole('button', { name: /^connect$/i }).click()
+    const items = page.getByRole('menuitem')
+    await expect(items.first()).toBeVisible()
+
+    // Verify each menu item is the topmost element at its own center.
+    for (const item of await items.all()) {
+      const box = await item.boundingBox()
+      if (!box) throw new Error('Menu item has no bounding box')
+      const cx = box.x + box.width / 2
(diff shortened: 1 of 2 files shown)


test/e2e/interactions.spec.ts
```diff
@@ -139,6 +139,40 @@ test.describe('Package Page', () => {
       ).toBe(true)
     }
   })
+
+  test('AccountMenu dropdown items are clickable on package pages', async ({ page, goto }) => {
```
gameroman, line 143: I think you can do something like this since you only need to check the center

```ts
test('AccountMenu dropdown items are clickable on package pages', async ({ page, goto }) => {
  await goto('/package/vue', { waitUntil: 'hydration' })

  await page.getByRole('button', { name: /^connect$/i }).click()
  const items = page.getByRole('menuitem')

  for (const item of await items.all()) {
    await item.click({ trial: true })
  }
})
```

ptdewey: I tried something like that initially, but this version passes w/o the change. I'm not certain why, but I think it might be because the upper and lower edges of the item can be clicked from behind the banner? 

image

gameroman (approved):

PR #2777, opened 2026-05-21
fix: show 404 page instead of blank on first visit to non-existent package

2u841r:
Reopened from #2733.

previous one was mistakenly "closed by deleting the head repository 2 days ago"

### DIFF
--- a/app/pages/package/[[org]]/[name].vue
+++ b/app/pages/package/[[org]]/[name].vue
@@ -30,10 +30,13 @@ onMounted(() => {
 
 const { packageName, requestedVersion } = usePackageRoute()
 
-const { data: resolvedVersion, status: resolvedStatus } = await useResolvedVersion(
-  packageName,
-  requestedVersion,
-)
+// On server: await so 404 can be detected and thrown synchronously below.
+// On client (SPA fallback): skip await to avoid Suspense suspension → white page.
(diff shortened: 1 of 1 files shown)


2u841r:
previous one was mistakenly "closed by deleting the head repository 2 days ago"

2u841r:
The description on #2733 (which this reopens) said the page stayed "completely blank until you refreshed." That's no longer quite accurate - retesting now, including with a genuinely never-queried, made-up package name, it still shows a blank flash (~200-300ms) before the 404 renders, but it resolves on its own without needing a manual refresh. So it's not just previously-queried names responding fast - even fresh, first-time lookups are quick now.

The root cause described there is still accurate and still present: `useResolvedVersion` is unconditionally awaited on the client, which suspends the component through Vue's Suspense with no fallback defined, so nothing renders while that fetch (to an external service, npm.antfu.dev) is in flight. I'm not sure why the flash is shorter than when I originally filed #2733 - possibly the upstream service or its network path just got faster - but the underlying bug is unchanged: any latency on that fetch still blocks rendering with no fallback.

Fix itself is unchanged: skip the await on the client, let the server still await so 404 can throw synchronously, rely on the existing watch for the client error case, and show the skeleton while resolvedStatus is pending.

gameroman:
Looks like this was solved already at some point?

2u841r:
> Looks like this was solved already at some point?

yes

PR #2778, opened 2026-05-21
feat: add nodejs noodle

alexdln:
### 🧭 Context

New noodle 🍜

image


app/components/Tooltip/App.vue
```diff
@@ -12,6 +12,8 @@ const props = defineProps<{
   defer?: boolean
   /** Offset distance in pixels (default: 4) */
   offset?: number
+  /** Additional attributes to be applied to the tooltip element */
+  tooltipAttr?: Record<string, unknown>
```
graphieros, line 16: I think you can replace this prop by using Vue's attrs inheritance system:

```js
defineOptions({
  inheritAttrs: false,
})

const attrs = useAttrs()

// Then
const tooltipAttrs = computed(() => {
  const mergedAttrs: Record<string, unknown> = {
    role: 'tooltip',
    id: tooltipId,
    ...attrs,
  }

  if (props.interactive) {
    mergedAttrs.onMouseenter = show
    mergedAttrs.onMouseleave = hide
  }

  return mergedAttrs
})
```

alexdln: Oof, I can't get this setting [inheritAttrs] to work. As soon as I enable it, props stop passing correctly. And I always see two logos. Either it's broken somehow due to vanillajs logic, or it's just passing incorrectly, or I'm missing something 🫠

graphieros: I should have looked deeper, since there are both the trigger and the tooltip itself. Sorry if I wasted your time 🫢

alexdln: Don't worry, it was fun to try anyway, so thanks ☺️

danielroe (approved):

PR #2784, opened 2026-05-23
feat: stream recently liked packages on the homepage

serhalp:
### 🔗 Linked issue

N/A 😁

### 🧭 Context

The homepage is minimal at the moment. There's an opportunity to use that highly visible real estate to lean into social features while providing value through package discovery. As a bonus, this makes our package like feature more discoverable.

### 📚 Description

https://github.com/user-attachments/assets/e82e7349-b985-4bae-a33a-91873668872b

This adds a Recently Liked Packages feed to the homepage, powered by [Spacedust](https://www.microcosm.blue/) package like events.

The feed loads deferred on the client so the home page load stays non-blocking.

Because the npmx package like throughput is quite low at the moment (<10/day 🥹), just about no one would ever see anything if we truly only showed _live_ events here. Thus, to keep things engaging, we first fetch the 5 most recent likes (from the [UFOs](https://ufos.microcosm.blue/) service) and "replay" them, then start listening to the live stream. (🤞🏼 Perhaps someday we'll be able to remove this seeding!)

### DIFF
--- a/app/components/Landing/RecentlyLiked.vue
+++ b/app/components/Landing/RecentlyLiked.vue
@@ -1,0 +1,268 @@
+<script setup lang="ts">
+import type { RecentPackageLike } from '~/utils/recent-package-likes'
+import { RECENT_PACKAGE_LIKES_LIMIT } from '~/utils/recent-package-likes'
+
+const props = defineProps<{
+  isLoading?: boolean
+  likes: readonly RecentPackageLike[]
+}>()
+
+const LIKE_CARD_CLASSES = [
+  'grid h-full min-h-[7rem] content-start grid-cols-[minmax(0,1fr)_auto]',
+  'gap-x-3 gap-y-1 rounded-md border border-border bg-bg/70 px-3.5 py-3',
+  'text-start text-base text-fg shadow-sm transition-colors duration-200',
+  'hover:border-border-hover hover:bg-bg-elevated focus-within:border-accent',
+]
+
+const SKELETON_CARD_CLASSES = [
+  'grid min-h-[7rem] grid-cols-[minmax(0,1fr)_auto] gap-x-3 gap-y-3',
+  'rounded-md border border-border bg-bg/70 px-3.5 py-3',
+]
+
+const compactNumberFormatter = useCompactNumberFormatter()
+
+const skeletonRows = Array.from({ length: RECENT_PACKAGE_LIKES_LIMIT }, (_, index) => index)
+const hasLiveLike = computed(() => props.likes.some(like => like.origin === 'live'))
+const liveRegionPackageName = shallowRef('')
+const loadingPlaceholderRows = computed(() => {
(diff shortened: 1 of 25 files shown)


serhalp:
😆 🤔 maybe it should deduplicate...

Screenshot 2026-05-31 at 15 24 44

trueberryless:
Do you think we can prevent users from abusing this feature with unliking and liking constantly?

https://github.com/user-attachments/assets/217185ec-fa91-4ff6-a52f-ca8ce4f41c20

Feels like deduplication would fix this to some degree, but if two users do this at the same time, it would alternate. Except we deduplicate across the whole list 🤔

trueberryless:
Also another small commit (maybe a bug?):

How is it possible that notifications come in that have 0 likes? For some background info: I have it open in the background, running for the last 15min, but I think the problem is probably some caching 🤔 

image

Just FYI: Those likes are not done by me, those are actual events from others (feel free to expose yourself here 😅).

app/components/Landing/RecentlyLiked.vue
```diff
@@ -0,0 +1,268 @@
+<script setup lang="ts">
```
ghostdevv, line 1: I wonder if we should replace this arbitary packages list with the recently liked packages so that the home page remains less noisy - then perhaps we move this component to a new page that shows both a leaderboard of likes as well as the recent likes?

Image

ghostdevv (changes requested):

mootari:
> Do you think we can prevent users from abusing this feature with unliking and liking constantly?

A small delay might help: fill a bucket with the last 100 or so events, and only show the first occurrence for each user/package combo, maybe even with a cap on the max events shown per user. That way it becomes fairly difficult to game the system.

... and if that doesn't help, then we [beat them at their own game](https://discord.com/channels/1464542801676206113/1464542802565140506/1504071971711750174) 🙃 
Bildschirmfoto 2026-06-01 um 20 44 44

app/components/Landing/RecentlyLiked.vue
serhalp: I've tried to advocate for removing this list of frameworks several times (on Discord) but haven't been able to reach any kind of consensus about it, so I didn't touch it in this PR 🙈.

ghostdevv: I'm up for removing it, but it could be nicer to instead replace it with recent likes. I love the idea behind the PR, I just don't think that the UI, as it currently is, belongs on the otherwise fairly minimal homepage if that makes sense?

trueberryless: I agree that if we removed the list of frameworks, the minimal look goes missing. But I like the idea of sorting those current links after likes - basically integrating a little leaderboard on the homepage, while keeping it minimal.

But I wonder if there is another nice place where we could showcase the Jetstream of package likes? 🤔

PR #2786, opened 2026-05-24
chore(deps): update knip ✂️ and add enable treatConfigHintsAsErrors

TkDodo:
### 🧭 Context

This PR updates `knip` to the latest version, which surfaced some new findings (unused `export` statements) and enables the `treatConfigHintsAsErrors` config, because `vue-router` was unnecessarily added to knips ignore config.

### 📚 Description

As knip and its plugins keeps getting better, exceptions added to `ignoreDependencies` or `ignore` in the knip config might become unnecessary, but will prevent knip from reporting it if it ever becomes unused. With `treatConfigHintsAsErrors: true`, knip will fail if it finds an unnecessar config, forcing us to delete said config, which is a good thing.

No AI was used to make this contribution.

### DIFF
--- a/cli/src/schemas.ts
+++ b/cli/src/schemas.ts
@@ -95,7 +95,6 @@ export const PermissionSchema = v.picklist(
 
 /**
  * Validates operation types
- * @internal
  */
 export const OperationTypeSchema = v.picklist([
   'org:add-user',
@@ -132,7 +131,6 @@ export const HexTokenSchema = v.pipe(
 
 /**
  * Validates operation ID (16-char hex)
- * @internal
  */
 export const OperationIdSchema = v.pipe(
   v.string(),
--- a/knip.ts
+++ b/knip.ts
@@ -1,6 +1,7 @@
 import type { KnipConfig } from 'knip'
 
 const config: KnipConfig = {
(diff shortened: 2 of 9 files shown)


cli/src/schemas.ts
TkDodo, line 1: those were reported by knip’s “Tag hints” as unnecessary. I removed them but haven’t found a way to make knip error out if there are tag hints.

knip.ts
```diff
@@ -31,9 +31,6 @@ const config: KnipConfig = {
         'vite-plugin-pwa',
         '@vueuse/shared',
 
-        /** Some components import types from here, but installing it directly could lead to a version mismatch */
-        'vue-router',
-
```
TkDodo, line 36: this was flagged by config hints

TkDodo:
you don’t like custom scopes huh 😂

serhalp (approved):
Thank you! 🙌🏼

ghostdevv (approved):

cli/src/schemas.ts
cylewaitforit: I like this idea. I opened https://github.com/webpro-nl/knip/issues/1767

PR #2800, opened 2026-05-26
feat: support npmjs versions redirect

### LINKED ISSUE #2781: Downloads by package version
### Describe the feature

Hello!

I was interested in the number of downloads by version of webpack.  I could find this on npmjs.com here: https://www.npmjs.com/package/webpack?activeTab=versions 

Image

On npmx we see: https://www.npmx.dev/package/webpack?activeTab=versions

Image

So we don't get the individual version breakdown.  It would be nice to get that view.  Perhaps downloads per week could appear inline next to each version here? 

Image

What do you think?

### Additional information

### Final checks

johnnyreilly:
### 🔗 Linked issue

Resolves https://github.com/npmx-dev/npmx.dev/issues/2781

### 🧭 Context

There's a good level of intentional URL compatibility between npmjs.com and npmx.dev. The idea being that if you swapped out the prefix you'd see the same feature on npmx.dev that you see on npmjs.com

This doesn't apply for versions. So if you go to:
https://www.npmjs.com/package/webpack?activeTab=versions

Then go to
https://www.npmx.dev/package/webpack?activeTab=versions

You do not see a versions view. However, a versions view does exist:
https://npmx.dev/package/webpack/versions

### 📚 Description

This PR changes npmx.dev such that https://www.npmx.dev/package/webpack?activeTab=versions would be 301'd to https://npmx.dev/package/webpack/versions 

This means that users swapping the prefix will see the versions view. The URL is a redirect, but that seems appropriate given there isn't a "tab" for versions in npmx

cc @gameroman

### DIFF
--- a/server/middleware/canonical-redirects.global.ts
+++ b/server/middleware/canonical-redirects.global.ts
@@ -38,21 +38,26 @@ const cacheControl = 's-maxage=3600, stale-while-revalidate=36000'
 export default defineEventHandler(async event => {
   const [path = '/', query] = event.path.split('?')
 
-  // /package/name?activeTab=versions → /package/name/versions
-  // /package/@scope/name?activeTab=versions → /package/@scope/name/versions
   if (query) {
     const params = new URLSearchParams(query)
-    if (params.get('activeTab') === 'versions') {
-      const pkgPathMatch = path.match(/^\/package\/((?:@[^/]+\/)?[^/]+)$/)
-      if (pkgPathMatch) {
-        params.delete('activeTab')
-        const remaining = params.toString()
-        setHeader(event, 'cache-control', cacheControl)
-        return sendRedirect(
-          event,
-          `/package/${pkgPathMatch[1]}/versions` + (remaining ? '?' + remaining : ''),
-          301,
-        )
+
+    switch (params.get('activeTab')) {
+      case 'versions': {
+        // /package/name?activeTab=versions → /package/name/versions
+        // /package/@scope/name?activeTab=versions → /package/@scope/name/versions
+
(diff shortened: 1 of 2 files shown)


server/middleware/canonical-redirects.global.ts
```diff
@@ -36,13 +36,32 @@ const pages = [
 const cacheControl = 's-maxage=3600, stale-while-revalidate=36000'
 
 export default defineEventHandler(async event => {
+  const [path = '/', query] = event.path.split('?')
+
+  // /package/name?activeTab=versions → /package/name/versions
+  // /package/@scope/name?activeTab=versions → /package/@scope/name/versions
+  if (query) {
+    const params = new URLSearchParams(query)
+    if (params.get('activeTab') === 'versions') {
```
wojtekmaj: We could turn it into switch(), hence making it trivial to handle more redirects. and got example ?activeTab=code begs for it!

johnnyreilly: Yup - done!

gameroman (approved):

43081j (approved):

johnnyreilly:
I can see this was merged, but it doesn't seem to be active on npmx - is there a long gap between releasing?

gameroman:
Looks like it just doesn't work

Maybe we might need to add `allowQuery` to https://github.com/npmx-dev/npmx.dev/blob/cc715ee5fe56d7e18c7a541927fa80379f07b09f/nuxt.config.ts ?

johnnyreilly:
Oh that looks plausible. It's curious that running local dev doesn't appear to require that? But yeah - could try that.

How would we test it though? (given that with local dev it looks like things work already)

PR #2803, opened 2026-05-26
fix: use local font provider

alexdln:
### 🧭 Context

The Geist Mono font recently broke. Because of this, ligatures (-->, ==, !=, ..., --, etc.) no longer displayed correctly. This has already been fixed - https://github.com/google/fonts/pull/10540 - but due to several other services and caching issues font continues to be broken.

### 📚 Description

I used local fonts, as they are already configured in our repository. This will be more stable and should not impact performance overall, as users connect to our servers and wit ha few locations

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -277,27 +277,13 @@ export default defineNuxtConfig({
       {
         name: 'Geist',
         provider: 'local',
-        weight: 400,
-        preload: true,
-        global: true,
-      },
-      {
-        name: 'Geist',
-        provider: 'local',
-        weights: [500, 600],
+        weights: [400, 500, 600],
         global: true,
       },
       {
         name: 'Geist Mono',
         provider: 'local',
-        weight: 400,
-        preload: true,
-        global: true,
-      },
-      {
-        name: 'Geist Mono',
-        provider: 'local',
-        weight: 500,
+        weight: [400, 500],
         global: true,
       },
       {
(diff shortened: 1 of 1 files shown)


nuxt.config.ts
mootari, line 1: Why ttf, and why not the npm provider?

alexdln: Hmm, I don't see much reasons in adding a package. It seems like the font should remain unchanged (and auto package updates could also break in this case)

But honestly, these fonts simply already existed. We even used them before due to the specifics of previous og generation (_now only for storybook_)

nuxt.config.ts
```diff
@@ -276,14 +276,14 @@ export default defineNuxtConfig({
...
-        weights: ['400', '500', '600'],
-        preload: true,
+        provider: 'local',
+        weights: [400, 500, 600],
         global: true,
       },
       {
         name: 'Geist Mono',
-        weights: ['400', '500'],
-        preload: true,
```
danielroe, line 286: are we deliberately disabling preload?

nuxt.config.ts
```diff
@@ -276,14 +276,14 @@ export default defineNuxtConfig({
...
+        provider: 'local',
+        weight: [400, 500],
```
danielroe:
```suggestion
        weights: [400, 500],
```

nuxt.config.ts
alexdln: image
jsdoc says this is true by default 👀

danielroe (approved):

PR #2804, opened 2026-05-26
refactor: switch to packumeta for trust levels

43081j:
### 🔗 Linked issue

N/A

### 🧭 Context

Unifying trust level scales across ecosystem tools.

### 📚 Description

This switches our trust level logic to use the new `packumeta` package,
so we can share the same scale as other tools in the space (e.g. e18e
tools already use this).

Once staging signals are added to npm, this package will be updated to
expose them as the existing `stagedPublish` level.

NOTE: This does not fix the current problem of staged packages showing
up as a "downgrade".

### DIFF
--- a/app/components/Package/Versions.vue
+++ b/app/components/Package/Versions.vue
@@ -185,7 +185,7 @@ const allTagRows = computed(() => {
         tags,
         trustStatus: versionData?.trustStatus,
         deprecated: versionData?.deprecated,
-      } as VersionDisplay,
+      },
     }))
     .sort((a, b) => compare(b.primaryVersion.version, a.primaryVersion.version))
 })
--- a/app/components/Package/Versions.vue
+++ b/app/components/Package/Versions.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import { compare, validRange } from 'semver'
 import type { RouteLocationRaw } from 'vue-router'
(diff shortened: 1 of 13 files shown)


app/components/Package/Versions.vue
```diff
@@ -182,7 +183,7 @@ const allTagRows = computed(() => {
         version,
         time: props.time[version],
         tags,
-        hasProvenance: versionData?.hasProvenance,
+        trustStatus: versionData?.trustStatus,
         deprecated: versionData?.deprecated,
       } as VersionDisplay,
```
serhalp: could we kill this unsafe type assertion while we're here? this is exactly the type of change that can go wrong because of its presence

serhalp (approved):
LGTM

serhalp:
maybe not a `feat:` if it's a pure refactor / dep change?

43081j:
this now actually fixes the trust scale problem too - through a terrible but working hack...

basically, we now detect something as having used staging if `_id` is the first field in the document. surprisingly, no packuments exist with `_id` at the start in the registry other than those which used staging.

@serhalp its a true hack but im happy with it if you are, and it'll fix the false downgrade errors people are seeing in the timeline.

once npm expose a real signal, we can update the library

EDIT:

**npm are now going to ship the fix**, so im going to merge this and do a follow up once they have.

PR #2805, opened 2026-05-26
fix(ui): keep compare header links inside their grid column

### LINKED ISSUE #2802: fix: comparison header overflows column with long scoped package names
On the compare page, header names like `@agentmemory/agentmemory` render wider than the column and bleed into neighbouring columns. The left side gets hidden behind the sticky label cell, and the right side overlaps the next package header.

Repro: https://npmx.dev/compare?packages=@agentmemory/agentmemory,claude-mem,supermemory,mem0ai&layout=split

Screenshot (`@age` is missing from column 1 and `c` from `claude-mem`):

overlap

Looking at `app/components/Compare/ComparisonGrid.vue`, the `<LinkBase>` inside each header cell has `flex min-w-0 flex-col items-center text-center text-sm` but no `flex-1` / `w-full`. The parent flex uses `justify-center`, so the link sizes to its intrinsic content width and is free to grow past the grid track. The inner `<span class="min-w-0 break-words line-clamp-1">` then has nothing to clamp against, so the text never wraps or truncates.

Happy to send a PR.

rohitg00:
Fixes #2802

On the compare page, package headers like `@agentmemory/agentmemory` ran wider than their grid column and visibly overlapped both the sticky label cell on the left and the next package header on the right.

The header link in `ComparisonGrid.vue` was `flex min-w-0 flex-col` with no `flex-1` / `w-full`, so it sized to its intrinsic content width and escaped the grid track. Adding `flex-1` makes the link fill the available track width, which gives the inner `line-clamp-1` spans a real width to clamp against. The existing `min-w-0` already lets it shrink when the column is narrower than the name.

Repro URL: https://npmx.dev/compare?packages=@agentmemory/agentmemory,claude-mem,supermemory,mem0ai&layout=split

I also added a small regression test that mounts the grid with the four packages from the screenshot above and asserts the long-name header link carries `flex-1` and `min-w-0`.

### DIFF
--- a/test/nuxt/components/compare/ComparisonGrid.spec.ts
+++ b/test/nuxt/components/compare/ComparisonGrid.spec.ts
@@ -60,10 +60,10 @@ describe('ComparisonGrid', () => {
     })
 
     it('constrains the header link to its grid track so long scoped names cannot overflow', async () => {
-      const longName = '@agentmemory/agentmemory@0.9.22'
+      const longName = '@scope/very-long-package-name@1.2.3'
       const component = await mountSuspended(ComparisonGrid, {
         props: {
-          columns: cols(longName, 'claude-mem@13.3.0', 'supermemory@4.21.1', 'mem0ai@3.0.3'),
+          columns: cols(longName, 'pkg-b@1.0.0', 'pkg-c@1.0.0', 'pkg-d@1.0.0'),
         },
       })
 
--- a/test/nuxt/components/compare/ComparisonGrid.spec.ts
+++ b/test/nuxt/components/compare/ComparisonGrid.spec.ts
@@ -58,24 +58,6 @@ describe('ComparisonGrid', () => {
       expect(link.attributes('title')).toBe(longName)
       expect(link.findAll('.truncate')).toHaveLength(2)
     })
-
-    it('constrains the header link to its grid track so long scoped names cannot overflow', async () => {
-      const scopedName = '@scope/very-long-package-name'
-      const scopedVersion = '1.2.3'
-      const component = await mountSuspended(ComparisonGrid, {
-        props: {
-          columns: [
-            { name: scopedName, version: scopedVersion },
-            ...cols('pkg-b@1.0.0', 'pkg-c@1.0.0', 'pkg-d@1.0.0'),
-          ],
-        },
-      })
-
-      const link = component.find(`a[title="${scopedName}@${scopedVersion}"]`)
-      expect(link.exists()).toBe(true)
-      expect(link.classes()).toContain('flex-1')
-      expect(link.classes()).toContain('min-w-0')
-    })
   })
 
   describe('column layout', () => {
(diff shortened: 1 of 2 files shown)


graphieros (changes requested):
@rohitg00 Thank you for your PR :)

From what I can see, long package names still bleed, and the start of the name can also appear cropped in some cases:

Image

I think using ellipsis for package names in the headers can be a good alternative (it is already used in other places in the compare page, inside the charts for example).

rohitg00:
Thanks @graphieros — you're right, the previous fix didn't help with long names. Pushed 8f569907 to switch the name + version spans to `truncate` (single-line ellipsis), same pattern as the chart labels in this page.

Root cause of the cropped start in your screenshot: `break-words` was breaking the scope prefix at the `@`, pushing it onto a second line that `line-clamp-1` then hid, so visually you only saw the tail. `truncate` keeps the start anchored and ends in `…` when it overflows the column.

test/nuxt/components/compare/ComparisonGrid.spec.ts
```diff
@@ -56,7 +56,25 @@ describe('ComparisonGrid', () => {
...
+            ...cols('pkg-b@1.0.0', 'pkg-c@1.0.0', 'pkg-d@1.0.0'),
+          ],
+        },
+      })
+
+      const link = component.find(`a[title="${scopedName}@${scopedVersion}"]`)
+      expect(link.exists()).toBe(true)
+      expect(link.classes()).toContain('flex-1')
+      expect(link.classes()).toContain('min-w-0')
     })
```
gameroman: not sure this test makes sense

gameroman (changes requested):

test/nuxt/components/compare/ComparisonGrid.spec.ts
graphieros: I agree, the css change is enough and does not require coverage in my opinion.

graphieros:
@rohitg00 

I suspect your previous answer (if not this PR) was generated by an agent.
It is important for the npmx team to have a 'people first' approach, and if we are ok with contributors using AI as a tool, the human connection is part of the essence of npmx.

Therefore I would be happy to merge your PR (probably without the test, if you were so kind to consider its removal), should you answer to this message in your own words.

rohitg00:
Hey! @graphieros Yeah, I use Agents a lot. I made this PR with Claude code, but I guided it and faced problem myself when I was comparing different memories with my agentmemory project.

Thanks, happy to help.

graphieros:
@rohitg00 

Thanks, and nice to meet you :)

graphieros (changes requested):
(see remarks regarding the test)

rohitg00:
Agreed @gameroman @graphieros : dropped that test in f29eb4bb. Kept only the title-attribute / render assertion.

graphieros (approved):
LGTM
Thank you for dropping the em dash in that last comment ;)

rohitg00:
> LGTM Thank you for dropping the em dash in that last comment ;)

I hate emdashes.

graphieros:
@gameroman 
You can merge if it's good for you

gameroman (approved):

PR #2816, opened 2026-05-29
refactor: use normalizeLicense util consistently

charpeni:
### 🔗 Linked issue

n/a

### 🧭 Context

n/a

### 📚 Description

This refactor uses the `normalizeLicense` function more consistently.

### DIFF
--- a/server/api/registry/license-change/[...pkg].get.ts
+++ b/server/api/registry/license-change/[...pkg].get.ts
@@ -1,4 +1,4 @@
-import { normalizePackageLicense } from '#shared/utils/npm'
+import { normalizeLicense } from '#shared/utils/npm'
 
 interface LicenseChangeRecord {
   from: string
@@ -44,19 +44,18 @@ export default defineCachedEventHandler(
         version === 'latest' ? versions.length - 1 : versions.findIndex(v => v.version === version)
 
       const previousVersionIndex = currentVersionIndex - 1
-      if (previousVersionIndex < 0) {
-        return { change }
-      }
 
-      const currentLicense =
-        normalizePackageLicense(versions[currentVersionIndex]?.license) ?? 'UNKNOWN'
-      const previousLicense =
-        normalizePackageLicense(versions[previousVersionIndex]?.license) ?? 'UNKNOWN'
+      // Skip when there's no real previous version, else we'd diff against a phantom 'UNKNOWN'.
+      if (currentVersionIndex > 0) {
+        const currentLicense = normalizeLicense(versions[currentVersionIndex]?.license) ?? 'UNKNOWN'
+        const previousLicense =
+          normalizeLicense(versions[previousVersionIndex]?.license) ?? 'UNKNOWN'
 
-      if (currentLicense !== previousLicense) {
-        change = {
-          from: previousLicense,
-          to: currentLicense,
+        if (currentLicense !== previousLicense) {
+          change = {
+            from: previousLicense,
+            to: currentLicense,
+          }
         }
       }
       return { change }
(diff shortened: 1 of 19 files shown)


serhalp (approved):
LGTM, thank you!

test/unit/server/api/registry/license-change/pkg.get.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    )
+
+    await expect(handler(fakeEvent)).resolves.toEqual({
+      change: {
+        from: 'MIT',
+        to: 'ISC',
+      },
+    })
+  })
+})
```
gameroman: `license` can be an object like
```js
  license: { type: 'MIT' }
```
could you add a test for that, please?

gameroman (changes requested):

test/unit/server/api/registry/license-change/pkg.get.spec.ts
charpeni: Added a regression test for the legacy object license shape and normalized license comparisons so `{ type: "MIT" }` compares as `MIT` instead of `[object Object]`. Focused lint, format, and unit tests pass.

server/api/registry/license-change/[...pkg].get.ts
```diff
@@ -3,6 +3,15 @@ interface LicenseChangeRecord {
...
 }
 
+function normalizeLicense(license: unknown): string {
+  if (typeof license === 'string') return license
+  if (license && typeof license === 'object' && 'type' in license) {
+    const { type } = license as { type?: unknown }
+    if (typeof type === 'string') return type
+  }
+  return 'UNKNOWN'
+}
```
Comment (gameroman) [inline server/api/registry/license-change/[...pkg].get.ts]: We have a util for this somewhere already I think

gameroman (changes requested):

server/api/registry/license-change/[...pkg].get.ts
Comment (charpeni) [reply server/api/registry/license-change/[...pkg].get.ts]: That would make sense, I was just looking in the codebase to see if I missed other spots. It seems flaky (or prone to duplicates) to have it in this file.

Comment (gameroman) [reply server/api/registry/license-change/[...pkg].get.ts]: https://github.com/npmx-dev/npmx.dev/blob/b3ba515137c1db7e455b36fa92e040a8a00d1d94/app/composables/npm/usePackage.ts#L6

Comment (charpeni) [reply server/api/registry/license-change/[...pkg].get.ts]: Thanks! I found some places that could use this util along the way: db425d8.

shared/utils/npm.ts
```diff
@@ -16,6 +16,18 @@ export function encodePackageName(name: string): string {
...
+  if (!license) return undefined
+  if (typeof license === 'string') return license
+
+  if (typeof license === 'object' && 'type' in license) {
+    const { type } = license as { type?: unknown }
+    return typeof type === 'string' ? type : undefined
+  }
+
+  return undefined
+}
```
gameroman: Why not what we had before?

```suggestion
export function normalizePackageLicense(license?: PackumentLicense): string | undefined {
  if (!license) return undefined
  if (typeof license === 'string') return license
  if (typeof license.type === 'string') return license.type
  return undefined
}
```

charpeni: I used `unknown` because npm registry data can be malformed or differ from `@npm/types`, and this helper is meant to be defensive at the normalization boundary. That said, matching the previous `PackumentLicense` signature is reasonable if we want the helper to stay type-narrow and mirror the old utility.

f78042b.

gameroman (approved):

ghostdevv:
We got a few PRs for this fix, so we merged all the changes into #2792 and closed the other ones - in this case though we can turn this PR into a refactor of the use of the normalise util :eyes:

ghostdevv (approved):

PR #2826, opened 2026-05-31
feat: pride noodle

alexdln:
### 🧭 Context

New noodle 🎉

### 📚 Description

Preview with active noodle - https://npmx-l2zigx3fh-npmx.vercel.app/

<details>
image
image

</details>


alexdln:
_Oops, working on it_
_UPD: still working..._

alexdln:
Okay, so I'm not a good person since I deleted the test. I added this test in the previous pr, but it's linked to real data, and attempts to mock the data don't really work since the page is static, and the logic is inside prehydrate (that is, it is also impossible to move the logic into a separate function)

I hope we can bring back the test later, but for now it doesn't seem worth wasting time on it (@gameroman, @ghostdevv fyi)

ghostdevv (approved):

PR #2827, opened 2026-05-31
fix(i18n): add singular forms to en strings

trueberryless:
### 🧭 Context

English Translation strings.

### 📚 Description

While translating, I noticed some translation strings in `en` only have plural forms, but the `{count}` could be 1 as well, so we should include the singular form as well. 

I'm not sure if the translation keys have to change, so the translation workflow works correctly...


ghostdevv (changes requested):
~~I can't explain why but `0 dependencies` sounds more correct than `0 dependency` 😂~~ edit: it's weird because it's wrong, and the translation library is way ahead of me on that one xD

ghostdevv (approved):

PR #2828, opened 2026-05-31
fix(i18n): localize size increase/decrease percentage

trueberryless:
### 🧭 Context

Display correctly localized percentage value for size increase / decrease notice.

### 📚 Description

While translating some UI strings, I noticed that the size increase / decrease components show the percentage value hard coded with the `%` sign right after the calculated number.

In some languages however, this is wrong, so I updated the components to use the [`Intl.NumberFormat` API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#percent) (suggestion from Wilco) to get the correct string for all languages.

### 📸 Screenshots

Here you can see screenshots of before and after in English and German (not translated yet, but you see the space in the after image):

|   | Before  | After  |
|---|---|---|
| English  | image  | image  |
| German  | image  | image  |

### DIFF
--- a/app/components/Package/SizeDecrease.vue
+++ b/app/components/Package/SizeDecrease.vue
@@ -10,10 +10,8 @@ const { locale } = useI18n()
 const bytesFormatter = useBytesFormatter()
 const numberFormatter = useNumberFormatter()
 
-const sizePercent = computed(() => {
-  const percentFormatter = new Intl.NumberFormat(locale.value, { style: 'percent' })
-  return percentFormatter.format(props.diff.sizeRatio)
-})
+const percentFormatter = computed(() => new Intl.NumberFormat(locale.value, { style: 'percent' }))
+const sizePercent = computed(() => percentFormatter.value.format(props.diff.sizeRatio))
 const sizeDecreaseAbs = computed(() => Math.abs(props.diff.sizeIncrease))
 const depDecreaseAbs = computed(() => Math.abs(props.diff.depDiff))
 </script>
--- a/app/components/Package/SizeDecrease.vue
+++ b/app/components/Package/SizeDecrease.vue
@@ -5,12 +5,10 @@ const props = defineProps<{
   diff: InstallSizeDiff
 }>()
 
-const { locale } = useI18n()
-
 const bytesFormatter = useBytesFormatter()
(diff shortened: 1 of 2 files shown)


ghostdevv (changes requested):
Could we use one of these instead?:

- https://vue-i18n.intlify.dev/guide/essentials/number#custom-formatting
- https://github.com/npmx-dev/npmx.dev/blob/98be51d4f0b9d507104dbb0c88dca757ede83ec1/app/composables/useNumberFormatter.ts#L1-L5

trueberryless:
> Could we use one of these instead?:
> 
>     * [vue-i18n.intlify.dev/guide/essentials/number#custom-formatting](https://vue-i18n.intlify.dev/guide/essentials/number#custom-formatting)
> 
>     * https://github.com/npmx-dev/npmx.dev/blob/98be51d4f0b9d507104dbb0c88dca757ede83ec1/app/composables/useNumberFormatter.ts#L1-L5

That's so much better, I haven't noticed them, sry.
Did you have [that](https://github.com/npmx-dev/npmx.dev/pull/2828/commits/d9eae7b3d7417a6cd4dc2da7201b2256cdb4fbe4) in mind?

ghostdevv (changes requested):
awesome! one more thing, could you add a before/after screenshot?

trueberryless:
> awesome! one more thing, could you add a before/after screenshot?

Of course, it's a pleasure 👒

gameroman (approved):

ghostdevv (changes requested):
One more thing sorry, in the screenshots you sent the new one has `-100%` - I think it should still be positive?

trueberryless:
> One more thing sorry, in the screenshots you sent the new one has `-100%` - I think it should still be positive?

Oh great catch, sorry I overlooked that. Will take a look later 👍

trueberryless:
Thanks for the detailed reviews again @ghostdevv!
The mistake was pretty obvious after you pointed it out, and the screenshots have been updated 🥳

ghostdevv (approved):

ghostdevv:
ty!

> The mistake was pretty obvious after you pointed it out

ahh easy to miss!

PR #2832, opened 2026-06-01
feat(ui): independent terminal cards with fixed copy button

felipecorredor:
### 🧭 Context

The GET STARTED section showed all commands grouped inside a single shared terminal block. This made it harder to copy individual commands and the copy button was only visible on hover.

### 📚 Description

- Each command (install, dev install, @types, run, create) is now wrapped in its own independent terminal card with the three-dot header and the same dark border/rounded style as the original container
- The copy button is now always visible (removed `opacity-0` / `group-hover` pattern) and includes the `i-lucide:copy` icon alongside the "Copy" label
- Updated the CSS selector from `display: flex` to `display: block` for `[data-pm-cmd]` elements since each card is now a block container, with the inner row using flex

### 📚 Screenshots
Screenshot 2026-05-31 at 8 33 39 PM

### DIFF
--- a/app/components/Terminal/Install.vue
+++ b/app/components/Terminal/Install.vue
@@ -324,7 +324,9 @@ useCommandPaletteContextCommands(
             :title="$t('package.get_started.view_types', { package: typesPackageName })"
           >
             <span class="i-lucide:arrow-right rtl-flip w-3 h-3 align-middle" aria-hidden="true" />
-            <span class="sr-only">View {{ typesPackageName }}</span>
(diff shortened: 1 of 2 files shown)


ghostdevv:
Hey, thanks for the PR! I am however going to tentatively close this as I believe the current behaviour is still the desired one. If you'd like to make the case though, you can chat in https://github.com/npmx-dev/npmx.dev/issues/1771 and we can consider re-opening this PR in the future :pray:

PR #2833, opened 2026-06-01
feat: endpoint to embed downloads chart static svg

### LINKED ISSUE #2811: endpoint to embed downloads chart
### Describe the feature

it'd be super nice for a way to embed the downloads graph in other websites, something like an endpoint that returns the svg (which you already can manually download) would be ideal, since it'd work without loading in any javascript!

currently i use https://npm.chart.dev for the tinyglobby website but that loads in a lot of javascript and i have to use a few css workarounds

download svg option circled with a caption that says endpoint that returns this

### Additional information

### Final checks

graphieros:
Resolves #2811 for the embedding aspect of the issue.

This creates an endpoint that generates a static svg for download charts, from:
- chart modal
- compare page

| page | closed | open |
|------|-------|-------|
| chart modal |image|image|
| compare | image|image|

- dark or light theme can be toggled
- URL can be copied from the input
- small preview image using the same URL

Important note regarding the data:
- data corrections are ignored, because embedded static charts must show the raw data.
- the local zoom state (using the chart's minimap) is ignored. Only the date filters are applied.
- the embed only works for downloads for now.

Refactoring:
- common logic used both in the embed & the client chart is refactored into shared utils

Rendering:
- the rendered SVG will appear with the default font of the browser. Consumers can add styles to a wrapper, the chart will inherit the font-family.

Other:
- vue-data-ui is updated to latest, with support for static svg generation for the VueUiXy component ([release notes](https://github.com/graphieros/vue-data-ui/releases/tag/v3.21.0)).
- VueUiXy is the only component with this feature for now. Should embeds be created for other chart components used on npmx, the library would need to be updated. I'm planning on gradually adding this feature to all mainstream chart types.


graphieros:
TS fails are related to `server/api/registry/timeline/[...pkg].get.ts` which are not modified by this PR 🤨

app/utils/charts.ts
```diff
@@ -799,14 +799,29 @@ export async function copyAltTextForTimelineChart({
...
+
+  if (!documentReference) {
+    return
+  }
+
+  const anchor = documentReference.createElement('a')
+  anchor.href = link
+  anchor.download = filename
+  anchor.click()
+  anchor.remove()
```
ghostdevv: I think there's a util for this somewhere now

app/utils/charts.ts
```diff
@@ -799,14 +799,29 @@ export async function copyAltTextForTimelineChart({
...
+
 // Used in chart context menu callbacks
 // @todo replace with downloadFileLink
 export function loadFile(link: string, filename: string) {
-  const a = document.createElement('a')
-  a.href = link
-  a.download = filename
-  a.click()
-  a.remove()
+  const documentReference = (globalThis as { document?: DownloadDocument }).document
```
ghostdevv: How come?

ghostdevv (changes requested):

app/utils/charts.ts
graphieros: yeah, just stupid belly dancing. I refactored to use the download link utility

alexdln (approved):

pnpm-lock.yaml
```diff
@@ -24276,4 +24276,4 @@ snapshots:
 
   zod@4.3.6: {}
 
-  zwitch@2.0.4: {}
+  zwitch@2.0.4: {}
```
gameroman: what happened to this line? 🤔

app/components/Package/TrendsChart.vue
```diff
@@ -2219,6 +1966,72 @@ watch([selectedGranularity, startDate, endDate], async () => {
     >
       {{ $t('package.trends.loading') }}
     </div>
+
```
gameroman, line 1969: When expanding this, the width of container changes due to scrollbar appearing, which causes some jumps.

Could we add some style (https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/scrollbar-gutter#stable) to the container to prevent it from doing that?

gameroman (changes requested):

pnpm-lock.yaml
```diff
@@ -24277,3 +24277,4 @@ snapshots:
   zod@4.3.6: {}
 
   zwitch@2.0.4: {}
+
```
gameroman:
```suggestion

```

pnpm-lock.yaml
gameroman: There should be a trailing line but empty 😅 

```suggestion
  zwitch@2.0.4: {}

```

gameroman:
It also jumps in height, due to having `max-height: 90vh`, so when expanding it gets the height of `90vh`, and when not expanded, it has less height

I don't know what we can about it, but potentially we could set `height: 100%`, so it's always the height of `90vh`

Or we can just ignore it and leave it as is

graphieros:
> It also jumps in height, due to having `max-height: 90vh`, so when expanding it gets the height of `90vh`, and when not expanded, it has less height
> 
> I don't know what we can about it, but potentially we could set `height: 100%`, so it's always the height of `90vh`
> 
> Or we can just ignore it and leave it as is

@gameroman honestly, I was never happy with the chart modal in the first place. On some smaller screens, the modal opens already with cropped content and scrollbar, which is ugly. I think it should be in a separate page, maybe along future stats page, instead of a modal.

gameroman:
> > It also jumps in height, due to having `max-height: 90vh`, so when expanding it gets the height of `90vh`, and when not expanded, it has less height
> > I don't know what we can about it, but potentially we could set `height: 100%`, so it's always the height of `90vh`
> > Or we can just ignore it and leave it as is
> 
> @gameroman honestly, I was never happy with the chart modal in the first place. On some smaller screens, the modal opens already with cropped content and scrollbar, which is ugly. I think it should be in a separate page, maybe along future stats page, instead of a modal.

Yeah, a separate page would be nice, but that would be a separate PR

This is good to go then 🎉

gameroman (approved):

graphieros:
> This is good to go then 🎉

Thank you for your review :)

ghostdevv (approved):

PR #2842, opened 2026-06-02
fix(ui): render Geist consistently across Firefox and Chrome

### LINKED ISSUE #2841: font: firefox showing a mono font for headers
### Describe the bug

I'm seeing the font on macos firefox as notably thinner than in chrome.

**Firefox**

Image

**Chrome:**

Image

### Additional context

_No response_

### Logs

```shell-script

```

birkskyum:
### 🔗 Linked issue

resolves #2841

### 🧭 Context

Firefox on macOS renders the site's header text (logo, nav, "jump to..." button) noticeably thinner than Chrome. Reproducible on any machine with Geist or Geist Mono installed in `~/Library/Fonts/`, which is pretty common since Vercel ships them with their dev tools.

The cause is the `local()` fallbacks in the auto-generated `@font-face` rules:

    src: local("Geist Mono Regular"), local("Geist Mono"), url("/fonts/GeistMono-Regular.ttf") format(truetype);

Chrome 86+ ignores `local()` for user-installed fonts as an anti-fingerprinting measure, so it falls through to the bundled TTF. Firefox honours it and picks up the local OTF. Same Geist family, but the OTF (CFF outlines) and TTF (TrueType outlines) rasterize differently in Firefox on macOS, and the OTF ends up lighter. Identical CSS, two different fonts actually rendered.

### 📚 Description

Sets `fonts.experimental.disableLocalFallbacks: true` in `nuxt.config.ts`. With it on, `@nuxt/fonts` stops emitting `local(...)` in the generated `nuxt-fonts-global.css`, so both browsers always load the bundled TTFs from `public/fonts/`.

Verified by checking the generated CSS no longer contains `local(...)` entries and that Firefox now matches Chrome after a hard reload.

**Before ( macos w/ firefox ) - npmx.dev**
Screenshot 2026-06-02 at 23 40 00

**After ( macos w/ firefox ) - localhost**
Screenshot 2026-06-02 at 23 39 56

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -273,6 +273,9 @@ export default defineNuxtConfig({
     providers: {
       fontshare: false,
     },
+    experimental: {
+      disableLocalFallbacks: true,
+    },
     families: [
       {
         name: 'Geist',


serhalp:
🤔 @alexdln can you take a look at this one? I thought we fixed this with https://github.com/npmx-dev/npmx.dev/pull/2803?

alexdln:
Previously, nuxt loaded fonts from the internet during build time, and in my PR, we changed it to use fonts stored in our repository.

This PR changes what nuxt-fonts generates - it no longer uses fonts already installed on the user's system, but always loads them from our assets.

On the one hand, we guarantee that the font will be exactly what we want, but on the other, it will depend only on our assets.

Perhaps there's something else going on.

In general, I think fontFallback is about a fallback font (like in [nuxt/fontaine](https://nuxt.com/modules/fontaine#how-it-works)), not a priority font (as it is [in nuxt/fonts](https://fonts.nuxt.com/advanced#how-it-works)). It seems like there might be an issue in the module itself. @danielroe wdyt?

_On Windows everything seems to work without problems_

Otherwise lgtm

alexdln (approved):
Overall, it's lgtm, but I suggest waiting for another approval (_see comment above_)

trueberryless:
Interestingly, for me the current alpha deployment looks the same in Firefox (top) and Chrome (bottom):

image

I assume that is because I don't have the Geist font family installed locally, so it always picks `url("/fonts/GeistMono-Regular.ttf") format(truetype)`?

ghostdevv (changes requested):
Is this just an issue with the 400 font weight? The tagline renders correctly in your screenshot, but the other text doesn't 🤔 This is also surely a bug in firefox? What version are you testing with?

birkskyum:
Firefox version 151.0.3 (aarch64) for macOS

ghostdevv (approved):
I couldn't find anything on this bug so this seems like the solution unfortunately

Issue #2785, opened 2026-05-24
Translation Status Page doesn’t load


TkDodo:
### Describe the bug

On the page https://main.npmx.dev/translation-status I’m seeing a never ending skeleton loader.

The GET request to https://main.npmx.dev/lunaria/status.json shows a 404 error.

So we are:
- missing error handling, never replacing the skeleton with and error
- missing the `status.json` file somehow?

### Additional context

I wanted to contribute some german translations and saw that this page doesn’t work.

Also I just saw that it works for https://npmx.dev/translation-status but the links and contribution guide point to https://main.npmx.dev/translation-status

PR #2848, opened 2026-06-03
chore: print more lunaria generation logs

alexdln:
### 🔗 Linked issue

Related #2785

### 🧭 Context

The error only in mentioned issue occurs only in canary. We have logic that outputs logs that can help understand what exactly happened, but these logs are only printed in release, not canary. To understand what exactly happened, enabling logs in canary would help

I'm almost sure that the problem is because lunaria, for some reason, can't process git data in the main branch (it definitely didn't generate the file and it definitely crashed here, but without logs it's hard to understand why)

Overall, I can't find a reason why we should ignore logs, so I want us to display them

### DIFF
--- a/modules/lunaria.ts
+++ b/modules/lunaria.ts
@@ -31,13 +31,16 @@ export default defineNuxtModule({
         try {
           execSync('node --experimental-transform-types ./lunaria/lunaria.ts', {
             cwd: nuxt.options.rootDir,
+            stdio: 'inherit',
           })
         } catch (e) {
           // Always throw in local dev.
           // In CI, only throw if building for production.
           const { env } = await getEnv(!isCI)
-          if (env === 'dev' || env === 'release') {
+          if (env === 'dev' || env === 'release' || env === 'canary') {
             throw e
+          } else {
+            console.error(e)
           }
         }
       })


trueberryless (approved):
I hope this doesn't cost Daniel too much 😅

alexdln:
@trueberryless It should be one line for success and probably ~10 lines for failure. So it's nothing compared to the volume of logs from 100k users 😅

trueberryless:
> @trueberryless It should be one line for success and probably ~10 lines for failure. So it's nothing compared to the volume of logs from 100k users 😅

Ah true, Lunaria only runs once at build, right? So it's not per request. I confused it.

alexdln:
Yes, and generates static file - /lunaria/status.json

ghostdevv (approved):

ghostdevv:
@alexdln I think this is failing the canary deployment - was that intentional? https://github.com/npmx-dev/npmx.dev/actions/runs/27038588422/job/79808765455

alexdln:
@ghostdevv, I was hoping that the build itself wouldn't crash. And that there would be more logs...
But now it's been confirmed for absolutely sure that the reason why translations aren't in canary - because this file isn't working

PR #2857, opened 2026-06-05
fix: use monospace font in embed chart

graphieros:
Follow up to #2833

Embedded charts currently have the default browser font, which does not convey the npmx design language.
This adds a style tag inside the svg to set the font-family of text elements to monospace.

### DIFF
--- a/server/utils/embed-downloads-svg.ts
+++ b/server/utils/embed-downloads-svg.ts
@@ -334,6 +334,7 @@ export async function createDownloadsSvgResponse(query: QueryParameters): Promis
       })
 
       return `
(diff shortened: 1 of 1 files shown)


ghostdevv:
<table>
<tr>
 <td>Before:</td>
 <td>After:</td>
<tr>
 <td>image</td>
 <td>image</td>
</table>

(my monospace font is [Comic Mono](https://dtinth.github.io/comic-mono-font/))

ghostdevv (approved):

ghostdevv:
We could also probably load the Geist font in the SVG

PR #2860, opened 2026-06-06
chore: enable fetch-depth for canary build

alexdln:
### 🔗 Linked issue

Related #2848, #2785

### 🧭 Context

From the logs, the problem of missed translations in canary is definitely in the lunaria/lunaria file, but nothing much has come up. So, as suggested in discord earlier, I'm trying to configure fetch depth. It seems even more likely that it's related to git

### DIFF
--- a/.github/workflows/deploy-canary.yml
+++ b/.github/workflows/deploy-canary.yml
@@ -20,6 +20,7 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
+          fetch-depth: 2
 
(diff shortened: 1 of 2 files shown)


ghostdevv (approved):

alexdln:
@ghostdevv It didn't help to fix the error too 🤔 But at least now there is no process errors

ghostdevv:
> @ghostdevv It didn't help to fix the error too 🤔 But at least now there is no process errors

It does appear to be the same underlying error?

> Error:  fatal: not a git repository (or any parent up to mount point /vercel)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
Error:  Command failed: node --experimental-transform-types ./lunaria/lunaria.ts

alexdln:
Yes, now at least there is a full trace, but it turned out to be useless

image

PR #2870, opened 2026-06-07
fix: data label collisions in compare downloads chart

### LINKED ISSUE #2868: Overlapping last value data labels in compare page downloads chart
Overlapping labels can occur when multiple series share roughly the same Y value:

Image

Proposed fix:
- if collisions are detected, distribute labels vertically, with a marker connecting the last point to its label

Quick mock up from @alfonsusac:

Image

graphieros:
Resolves #2868 

Add labels collision detection in compare downloads chart and if so:
- distributes evenly the labels vertically
- connects them to the last point with an elbowed marker
- additional padding-right is used for certain locales for which abbreviated units or number format require more space

image

The fix is also applied to the embed version.


shuuji3 (approved):
LGTM! I was about to ask if you can dynamically adjust labels when I see the overlapped labels in the original screenshot 😄

PR #2872, opened 2026-06-07
feat: stats page

alexdln:
### 🧭 Context

Adding a page with all the statistics - a table with key data and graphs of downloads and likes


graphieros (commented):
This is awesome work 😍 !

A few remarks, just pasting the ones I left on discord:

- I recommend you merge main into your branch, you might have conflicts after the changes I made on the display of last values data labels #2874
- I think the side bar should be limited in width & versions ellipsed as on the main page (try react package)
- facet, granularity, date range for the downloads chart should be persisted in the url
- settings of the versions chart are correctly persisted in url, however changing the settings should maintain the scroll position (?)
- The 'contributors' facet seems to be missing from the options in the facet select of the trends chart
- when charts are loading, the 'loading...' indicator is not always present (to replicate: load the page, change the facet to 'Likes' => CLS)
- the padding.right value of the versions chart (VueUiXy config) can probably be increased so both charts are better aligned in terms of perceived width (this is a suggestion, you be judge^^)
- I saw no regression on the compare page trends chart :)

alexdln:
@graphieros ah, I put too much faith in the ready-made components 😅
Thanks for the great test, 👀

graphieros (approved):
Great, I looked quickly and it seems to me like everything is fixed.🌿
Don't trust my approval 100% though, as I can't check thoroughly right now.

alexdln:
@graphieros there's still a problem with CLS sometimes on charts loading, but it seems like it can happen everywhere in the current format, and a more complicated fix is ​​needed (I think I will skip it for now). The rest should be fixed now

graphieros:
> @graphieros there's still a problem with CLS sometimes on charts loading, but it seems like it can happen everywhere in the current format, and a more complicated fix is ​​needed (I think I will skip it for now). The rest should be fixed now

Yes, this can be fixed later, I agree.
I'm really happy to see this new stats page finally come to life ⭐️

Issue #2871, opened 2026-06-07
Missing "Open changelog" shortcut in Keyboard Shortcuts search


BittuBarnwal7479:
### Describe the feature

Describe the feature
The package page shows a `-` keyboard shortcut for opening the changelog, but this shortcut is not listed on the Keyboard shortcuts search section.

<img width="1560" height="826" alt="Image" src="https://github.com/user-attachments/assets/7da7611e-8e19-4832-81e0-75ee39c15a19" />

<img width="1761" height="977" alt="Image" src="https://github.com/user-attachments/assets/89498d23-c2b5-4d8f-9391-282b27cd3e44" />

Expected Behavior

The keyboard shortcuts in search section should include:

`-` → Open changelog

This keeps the search keyboard shortcuts in sync with the available keyboard  shortcuts search in the UI.

### Additional information

- [x] Would you be willing to help implement this feature?

### Final checks

- [x] Read the [contribution guide](https://github.com/npmx-dev/npmx.dev/blob/main/CONTRIBUTING.md).
- [x] Check existing [issues](https://github.com/npmx-dev/npmx.dev/issues).

PR #2873, opened 2026-06-07
fix: add changelog shortcut to the keyboard shortcuts menu

BittuBarnwal7479:
### 🔗 Linked issue

#2871 

### 🧭 Context

The `-` shortcut for opening the changelog tab on package is shown in the UI but missing from the Keyboard Shortcuts search section.

### 📚 Description

The package page shows a `-` keyboard shortcut for opening the changelog, but this shortcut is not listed on the Keyboard shortcuts search section.


ghostdevv (changes requested):
I assume you used AI to do the translations? Usually we prefer it if you only translate the languages you're familiar with as it's easier for people to add missing translations and fix broken ones (plus gives others a good way to contribute). Otherwise LGTM

BittuBarnwal7479:
> I assume you used AI to do the translations? Usually we prefer it if you only translate the languages you're familiar with as it's easier for people to add missing translations and fix broken ones (plus gives others a good way to contribute). Otherwise LGTM

Yes, I used AI to help with the translations. The change was very small, just replacing a few English text.

I agree that native speakers are best suited for translation work, and I appreciate you pointing it out. I'll keep that in mind for future translation contributions. 
Thanks!

i18n/locales/nl.json
```diff
@@ -39,6 +39,7 @@
     "settings": "Open instellingen",
     "compare": "Open vergelijken",
     "compare_from_package": "Vergelijking openen (vooraf ingevuld met het huidige pakket)",
+    "changelog": "Open changelog",
```
WilcoSp: For Dutch this should be "Open wijzigingenoverzicht"

Also for languages you don't know I would suggest leaving them empty instead of "{open} changelog" as it can be that some languages will localise it differently compared to translating directly.

WilcoSp (changes requested):
Looking good, but like what ghostdev said about the translations, at least in Dutch I localise it differently

i18n/locales/nl.json
BittuBarnwal7479: Thanks!
I agree, I'll keep that in mind for future translation contributions.

WilcoSp (approved):
at least for Dutch looking good now, further reviewing I'll leave up to ghostdevv.

I'll also share this pr in the i18n channel in discord, maybe some from there will look at the translations

gameroman (changes requested):
I'd recommend to not adding translations for the languages you are not familiar with and wait until someone who does adds them later

BittuBarnwal7479:
> I'd recommend to not adding translations for the languages you are not familiar with and wait until someone who does adds them later

I've removed the unverified translations and kept the change focused on the original fix.

ghostdevv (approved):

gameroman (approved):

PR #2875, opened 2026-06-07
fix: remove ambiguous community wording from replacement suggestions

TheAlexLichter:
### 🔗 Linked issue

None (simple change)

### 🧭 Context

"The community" feels quite ambiguous/unclear at the moment, and two links (leading to the same main site) in a single warning draw focus away from the important part.

### 📚 Description

This PR slightly changes the wording, removes the "community" part, uses passive voice and makes the statement les ambiguous.


danielroe (approved):
this seems good to me, with the link doing the heavy lifting to point users to e18e

patak-cat (approved):
I also think this is a better way to word the warnings. Nice one!

gameroman (changes requested):
iirc, locales for other languages shouldn't be touched

i18n/locales/nl.json
```diff
@@ -400,12 +400,11 @@
       "example": "Voorbeeld:",
       "native": "Dit kan vervangen worden door {replacement}, Beschikbaar sinds node {nodeVersion}.",
       "native_no_version": "Dit pakket kan vervangen worden door {replacement}.",
-      "simple": "De {community} heeft dit pakket als overbodig gemarkeerd, met het advies: {replacement}.",
-      "documented": "De {community} heeft aangegeven dat er voor dit pakket beter presterende alternatieven zijn.",
+      "simple": "Dit pakket is als overbodig gemarkeerd, met het advies: {replacement}.",
+      "documented": "Dit pakket is gemarkeerd als pakket met beter presterende alternatieven.",
```
WilcoSp:
```suggestion
      "documented": "Dit pakket is gemarkeerd omdat er beter presterende alternatieven beschikbaar.",
```
"documented" doesn't read/speak naturally in Dutch with the double "pakket", ^ this will improve it.

for "simple" it's already good when reading/speaking in Dutch

WilcoSp (changes requested):
For Dutch "documented" can be improved.

43081j (approved):

WilcoSp (approved):

ghostdevv:
> iirc, locales for other languages shouldn't be touched

+1 it's better to remove translations that need changes rather than use AI :pray:

ghostdevv (changes requested):
~~The only other thing is that now there won't be any link to e18e on non-documented replacements~~ edit: there already isn't

ghostdevv (approved):

gameroman (approved):

PR #2883, opened 2026-06-09
fix(ui): use clearer icon for indicating outdated deps

typed-sigterm:
### 🔗 Linked issue

N/A for tiny fix.

### 🧭 Context

`lucide:circle-alert` is too serious, users may think it means security vulnerabilities. 

### 📚 Description

Change icon to `lucide:circle-fading-arrow-up`.

<table>
<tr>
 <td>Before:</td>
 <td>After:</td>
<tr>
 <td>image</td>
 <td>image</td>
</table>


typed-sigterm:
Also, outdated deps should not be red. It's a harmless *option* of maintainers, unlike security vulnerabilities.

gameroman (commented):
Could you add before-after screenshots, please

typed-sigterm:
Added in description

gameroman:
image

I like this, but I think it looks a bit unreadable at small size?

Do you think we could do the arrow without circle? Or maybe you have some other better ideas

Maybe `clock-arrow-up` or `circle-plus`?

I want to know what others think too

alfonsusac:
I made several mock-ups trying various options especially: `clock-arrow-up` and `square-arrow-up`. We've discussed this in [Discord](https://discord.com/channels/1464542801676206113/1483508842988961992/1516447469905379470) and decided that `arrow-up` fits well. We avoided border to not make it similar to an "alert" badge (the one with shield and checkmark) and it also has easily distinguishable overall shape (arrow). 

Mock-up for `arrow-up`
image

trueberryless:
I agree that the current icons and colours might be misleading. When the feature was released I was a little confused myself.

I really like the icon you chose, I think the simple arrow is a too... well simple. But I agree we probably want more opinions on this.

MatteoGabriele:
My 2 cents: I think the best one would be "i-lucide:history", but this icon renders way too small for all those details, so if we want something that brings attention without overwhelming, I think a simple "i-lucide:triangle" could just work.

Screenshot 2026-06-16 at 22 46 36

typed-sigterm:
Perhaps it's also unnecessary to set error colors on deps with newer major versions

43081j:
I'd go with what Matteo suggested, but would leave the colouring for a separate PR.

though my opinion on that is that we should still colour majors vs minors differently, maybe just not red if people are concerned about that being too strong of a signal

trueberryless:
Looks like a simple arrow up the winner of our little discussion on Discord:



Now we have to decide: Discord vs GitHub 😆

gameroman:
I actually like the triangle too

typed-sigterm:
Let's vote at #2930 and wait for 7 days :) It'll be great if anyone would like to post this in the Discord maintainers channel.

mootari:
Fwiw, I think a triangle is too abstract because you have to learn its meaning before you can read it, while you can arguably intuit the meaning of an up-arrow.

MatteoGabriele:
> Fwiw, I think a triangle is too abstract because you have to learn its meaning before you can read it, while you can arguably intuit the meaning of an up-arrow.

I'm not sure I can connect "outdated" with an upward arrow. A triangle is a common warning symbol, and it is designed to get your attention.

gameroman:
Looks like the arrow is the most supported one. Lets do that then

app/components/Package/Dependencies.vue
```diff
@@ -142,7 +142,7 @@ const numberFormatter = useNumberFormatter()
                 class="inline-flex items-center justify-center p-2 -m-2"
                 :aria-label="getOutdatedTooltip(outdatedDeps[dep], $t)"
               >
-                <span class="i-lucide:circle-alert w-3 h-3" aria-hidden="true" />
+                <span class="i-lucide:circle-fading-arrow-up w-3 h-3" aria-hidden="true" />
```
gameroman:
```suggestion
                <span class="i-lucide:arrow-up w-3 h-3" aria-hidden="true" />
```

gameroman (approved):
Good 👍

Also update the PR description with new icons, please

trueberryless (approved):

ghostdevv (approved):

ghostdevv:
> Also update the PR description with new icons, please

done it for em

PR #2889, opened 2026-06-11
ci: block compromised npm packages with Socket Firewall Free

serhalp:
### 🔗 Linked issue

N/A

### 🧭 Context

See https://docs.socket.dev/docs/socket-firewall-free.

Using `sfw pnpm i` instead of `pnpm i` prevents any known malicious package versions from being installed at all, and will cause the install to fail if any are encountered.

As we've seen these last few months, the npm ecosystem is unfortunately frequently victim to supply chain attacks, and this is a simple way to add a layer of protection against such. Socket has built a reputation for detecting these malicious packages within ~5 mins of them being published to the npm registry, so this is a pretty reliable solution for the ~3-36 hour period before these get unpublished by npm.

Socket Firewall Free is 100% free to use, and does not require a licence, account, or attribution:
https://github.com/SocketDev/sfw-free/blob/164c32dcbe3bbebcc0b091fd6a31ff93883545b2/README.md#license.

Originally, this wasn't going to be possible with vite-plus `vp install`, but I chatted with the maintainers and they added built-in support for this in v1.12.0:
https://github.com/voidzero-dev/setup-vp/pull/72.

### 📚 Description

Bump to v1.12.0 of the action and use `sfw: true` to opt in to Socket Firewall Free.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -33,11 +33,11 @@ jobs:
           cache: true
           sfw: true
 
-      - name: 🟧 Install pnpm globally
-        run: vp install -g pnpm
-
       - name: 🧪 Run Chromatic Visual and Accessibility Tests
         uses: chromaui/action@8a2b82547aef5a3efc8ec3c7905f4ab09a76ed0b # v16.1.0
+        with:
+          buildCommand: vp run build-storybook
+          outputDir: storybook-static
         env:
           CHROMATIC_BRANCH: ${{ github.event.pull_request.head.ref || github.ref_name }}
           CHROMATIC_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -27,10 +27,11 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
           persist-credentials: false
 
-      - uses: voidzero-dev/setup-vp@ca1c46663915d6c1042ae23bd39ab85718bfb0fa # v1.10.0
+      - uses: voidzero-dev/setup-vp@2dec1e33f4ab2c6d5bce1b0c4607961bb1a3f7a1 # v1.12.0
         with:
           node-version: lts/*
           cache: true
+          sfw: true
 
       - name: 🟧 Install pnpm globally
         run: vp install -g pnpm
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,13 +30,13 @@ jobs:
         with:
           persist-credentials: false
 
-      - uses: voidzero-dev/setup-vp@ca1c46663915d6c1042ae23bd39ab85718bfb0fa # v1.10.0
+      - uses: voidzero-dev/setup-vp@2dec1e33f4ab2c6d5bce1b0c4607961bb1a3f7a1 # v1.12.0
         with:
           node-version: lts/*
-          run-install: false
-
-      - name: 📦 Install dependencies (root only, no scripts)
-        run: vp install --filter . --ignore-scripts
+          sfw: true
+          # root only, no scripts
+          run-install: |
+            - args: ['--filter', '.', '--ignore-scripts']
 
       - name: 🔠 Lint project
         run: vp run lint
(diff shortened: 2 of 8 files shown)


.github/workflows/chromatic.yml
```diff
@@ -27,16 +27,17 @@ jobs:
...
           persist-credentials: false
 
-      - uses: voidzero-dev/setup-vp@ca1c46663915d6c1042ae23bd39ab85718bfb0fa # v1.10.0
+      - uses: voidzero-dev/setup-vp@2dec1e33f4ab2c6d5bce1b0c4607961bb1a3f7a1 # v1.12.0
         with:
           node-version: lts/*
           cache: true
-
-      - name: 🟧 Install pnpm globally
-        run: vp install -g pnpm
```
serhalp, line 36: This was a workaround for chromatic needing access to the package manager that it detects we use (pnpm), but _that's_ just needed so it can detect our command to build storybook, so we can just specify that command instead!

.github/workflows/chromatic.yml
```diff
@@ -27,16 +27,17 @@ jobs:
...
+          sfw: true
 
       - name: 🧪 Run Chromatic Visual and Accessibility Tests
         uses: chromaui/action@8a2b82547aef5a3efc8ec3c7905f4ab09a76ed0b # v16.1.0
+        with:
+          buildCommand: vp run build-storybook
+          outputDir: storybook-static
```
serhalp, line 40: required when you pass `buildCommand`. this value is the default.

trueberryless: Isn't the default a temporary directory?
In docs, I can only see that `"storybook-static"` is default for [TurboSnap](https://www.chromatic.com/docs/turbosnap/setup/#turbosnap-with-prebuilt-storybook-using-webpack), but it says temporary dir for CLI and therefore GitHub action: https://www.chromatic.com/docs/configure/#outputdir

But I guess we have to set it to something since it is [indeed required](https://www.chromatic.com/docs/configure/#buildcommand), and since our [`.gitignore` already has this entry](https://github.com/npmx-dev/npmx.dev/blob/1c48d0c6f21d597f6ec6e090fbab7d9434b56bdb/.gitignore#L52), I approve ✅

.github/workflows/ci.yml
```diff
@@ -30,13 +30,13 @@ jobs:
...
         with:
           node-version: lts/*
-          run-install: false
-
-      - name: 📦 Install dependencies (root only, no scripts)
-        run: vp install --filter . --ignore-scripts
+          sfw: true
+          # root only, no scripts
+          run-install: |
+            - args: ['--filter', '.', '--ignore-scripts']
```
trueberryless, line 39: Don't we need the `cwd` option here, or does it fallback to `.`?

```suggestion
            - cwd: .
              args: ['--filter', '.', '--ignore-scripts']
```

Applies further down and in `release-tag.yml` below as well…

trueberryless (commented):
I really like this security change!

Technically, support for sfw was added in [v1.11.0](https://github.com/voidzero-dev/setup-vp/releases/tag/v1.11.0), but I would also definitely use v1.12.0 here. Also fun details: I like how the voidzero action and sfw-free versions are currently in sync (both v1.12.0). Coincidence? I think it's a signal from shai hulud. 😆 

Generally, the changes look good to me, I just left small thoughts 💭

.github/workflows/chromatic.yml
serhalp: Oh, heh, it's an example provided, not the default 🤦🏼.

.github/workflows/ci.yml
serhalp: Yeah, defaults to the workspace root: https://github.com/voidzero-dev/setup-vp#inputs

.github/workflows/deploy-canary.yml
```diff
@@ -22,12 +22,17 @@ jobs:
...
+      - uses: voidzero-dev/setup-vp@2dec1e33f4ab2c6d5bce1b0c4607961bb1a3f7a1 # v1.12.0
         with:
           node-version: lts/*
           run-install: false
 
-      - run: vp install -g vercel
+      - uses: SocketDev/action@ba6de6cc0565af1f42295590380973573297e31f # v1.3.2
+        with:
+          mode: firewall-free
+          firewall-version: v1.12.0
```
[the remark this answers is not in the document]
serhalp: whut: https://github.com/SocketDev/action#socket-firewall-free

trueberryless (approved):
LGTM, Shai-Hulud-proofed!

PR #2892, opened 2026-06-12
fix: misconfigured robots.txt

### LINKED ISSUE #2891: robots.txt is not valid
### Describe the bug

I ran a [Pagespeed Insights Lighthouse test](https://pagespeed.web.dev/analysis/https-npmx-dev-package-vue/d1tdr4ez3b?form_factor=mobile) on a random npmx page, and it reported that the robots.txt is malformed. This allows bots not specifically banned to scan pages they should not be able to and verify it by seeing pages that should be disallowed are currently indexed by bots. 

For example, a package code page like: https://npmx.dev/package-code/srilanka-api/v/1.0.6 is currently indexed by Google, and according to manual testing from [robots.txt Validator and Testing Tool](https://technicalseo.com/tools/robots-txt/), it is allowed by non-ai crawlers like Google Bot even though it should be banned according to line

https://github.com/npmx-dev/npmx.dev/blob/1ef8d3c19af1274e829cdac7c31104d80fff7570/public/robots.txt#L29

The fix  is adding the following code to the top of the robots.txt
```
User-agent: *
```

### Additional context

_No response_

### Logs

```shell-script

```

MaxwellCohen:
### 🔗 Linked issue

Resolves: #2891

### 🧭 Context

The robots.txt file is invalid because the default settings are missing a user agent, so they cannot be read.

Added a default user agent and comments to robots.txt

### 📚 Description

Problem: pages that should not be crawled according to robots.txt are being crawled, resulting in SEO dilution and extra bandwidth spent on invalid crawling 

Before: a page that should be blocked is able to be crawled

image

After: the page is blocked 
image
image

I leveraged Nuxi to learn about Nuxt robots.txt and Composer 2.5 to find and validate robots.txt.

I will gladly make any changes needed, and thanks for a great resource!

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,3 +1,7 @@
+# Default Settings
+User-agent: *
+
+# Pages that bots are allowed to visit
 Allow: /$
 Allow: /about$
 Allow: /accessibility$
@@ -11,9 +15,6 @@ Allow: /_og/d/*
 Allow: /opensearch.xml$
 
 Disallow: /
-
-# User-agent: *
-# Allow: /
 
 # Search pages: infinite query-param combinations
 Disallow: /search


mootari:
This would probably be a good opportunity to remove the redundant Disallow rules?

MaxwellCohen:
I can remove the redundant rules. I would like someone on the core or maintainer teams to request/sign off because I want changes to be as focused as possible, and my main goal is to fix the malformed robots.txt that is giving incorrect access to bots



I had Composer 2.5 LLM run a report for me on redundant fields, and this is what it came up with, and all items seem reasonable to me :

editing to remove direct quote of AI response for human-written summary. Thanks to a suggestion from @trueberryless 

1. Remove the disallow rules after `Disallow: /` covering lines 19–44 because `Disallow: /.` covers everything that is not explicitly allowed. Thus, delete - I think this is a good idea, so we do not reveal extra paths that we do not want bots/people to see.  
2.  `/opensearch.xml' is both allowed and disallowed (line 15, 41). I think this should be allowed. There feels like there is a story around this that I would like to understand before taking any action.
3. Specific AI classifier blocks are duplicates and can be simplified to just list out the crawlers and blocks once. I am neutral on this and can see value in having it both ways.
  
If I should implement any of these items, please let me know, and I can do it.

trueberryless:
Please note that while using LLMs is allowed, we are aiming to interact with human responses. I refer to our [contributing guidelines](https://github.com/npmx-dev/npmx.dev/blob/main/CONTRIBUTING.md#using-ai). Ideally this means that you learn from any LLM responses, understand the context and write responses with your own words. 

If something is still unclear, you can always ask us. We also have a [Discord server](https://chat.npmx.dev) if you want to chat and learn with us 😉

MaxwellCohen:
@trueberryless I'm sorry about directly quoting llm in my previous response instead of summarizing it. I will have a 100% written human summary  of the LLM report tonight. I miss interpreted the LLM policy to allow for quoting when disclosured, my bad.


My goal of the summary was to identify optional improvements for easy decisions (that could be considered outside the scope of this pr). I will make sure to not do this in the future.

ghostdevv (approved):

ghostdevv:
Let's merge this and we can do a follow up :pray:

PR #2896, opened 2026-06-12
ci: remove `edited` type trigger from `enforce-release-source`

trueberryless:
### 🧭 Context

CI on main is (seemingly randomly) failing because the [`enforce-release-source` action](https://github.com/npmx-dev/npmx.dev/blob/1c48d0c6f21d597f6ec6e090fbab7d9434b56bdb/.github/workflows/enforce-release-source.yml) gets cancelled. This PR aims to fix this issue.

### 📚 Description

To my understanding, the problem is cause because merge queues quickly add commits to main, which causes two things:
1. the commit from main lands in the automatically generated next release PR, which causes the `enforce-release-source` action to trigger because of the **`synchronize` type**
2. the [`release-pr` workflow](https://github.com/npmx-dev/npmx.dev/blob/1c48d0c6f21d597f6ec6e090fbab7d9434b56bdb/.github/workflows/release-pr.yml) updates the PR description, which causes the `enforce-release-source` action to trigger a second time within a short time frame because of the **`edited` type**

This results in two workflow runs getting triggered and one getting cancelled because both events trigger quickly after each other. Therefore, removing the `edited` type _should_ resolve the issue, as it removes one of the concurrent runs. Since an edit of the PR description does not include target branch changes, removing the `edited` type would be my solution. I also explicitly set `cancel-in-progress` to `false` to avoid future confusion.

So I _think_ this cancellation happens seemingly random because workflows just do not take the same time always, resulting in the cancellation sometimes happening because of timing reasons, and sometimes not.

This is just how I understood it after debugging for the last 2 hours. I would really appreciate another four eyes on my deduction.


trueberryless:
Well, that was unexpected: https://github.com/npmx-dev/npmx.dev/actions/runs/27443844535/job/81124080867#step:3:1

🤣

ghostdevv:
How confident are you with this? @trueberryless

trueberryless:
> How confident are you with this? @trueberryless 

70% because it has been some time since I investigated this. Back then probably 90% 😅

ghostdevv:
> > How confident are you with this? @trueberryless
> 
> 70% because it has been some time since I investigated this. Back the probably 90% 😅

Do you think we should still merge this? Or should we give it some more time to figure it out?

trueberryless:
> Do you think we should still merge this? Or should we give it some more time to figure it out?

I'll revisit my ideas and will let you know 👍

alexdln:
Looks like now we can let it check and then change the branch right before merging (and without the "edited" trigger, this check will not restart)...

ghostdevv:
> Looks like now we can let it check and then change the branch right before merging (and without the "edited" trigger, this check will not restart)...

wdym?

alexdln:
The edited trigger allows the workflow to restart whenever a change occurs - whether to the name, description, or base branch.
If we remove it - the workflow may not restart automatically after the base branch changes (_and remain in the "passed" state, making this check bypassable_)

trueberryless:
> The edited trigger allows the workflow to restart whenever a change occurs - whether to the name, description, or base branch. If we remove it - the workflow may not restart automatically after the base branch changes (_and remain in the "passed" state, making this check bypassable_)

Okay, but do we need the `synchronise` type then? Because a new commit can't change the base branch 🤔

alexdln:
> Okay, but do we need the synchronise type then? Because a new commit can't change the base branch

It seems like yes, it can be removed 🤔

serhalp (approved):

PR #2909, opened 2026-06-14
ci: fix in invalid `firewall-version` in deploy-canary.yml

serhalp:
### 🔗 Linked issue

N/A

### 🧭 Context

Follow-up to #2889.

The deploy-canary workflow (only runs on `main`) failed on merge: https://github.com/npmx-dev/npmx.dev/actions/runs/27503413651/job/81290268327.

The version needs to be specified without the `v` prefix.

This isn't well documented, but you can see it here: https://github.com/SocketDev/action/blob/ba6de6cc0565af1f42295590380973573297e31f/src/tools/firewall.js#L47.

### 📚 Description

Remove `v` prefix


trueberryless (approved):
Sorry for not catching that.
Thanks for the quick fix.

PR #2918, opened 2026-06-15
fix: disable transitions on dots when resizing zoom range

### LINKED ISSUE #2916: Changing the width of the visible range in Downloads charts only animates points
### Describe the bug

When you move the selected range of a Downloads chart, both points and lines are animated towards their new position. However, when changing the size of the domain lines no longer animate, causing points to move independently (when visible).

https://github.com/user-attachments/assets/90a689af-cbfa-44d7-8be1-e0c8b46fbf31

### Additional context

_No response_

### Logs

```shell-script

```

graphieros:
Resolves #2916 

- Update vue-data-ui to latest with a fix to disable dot transitions when resizing the zoom range ([release notes](https://github.com/graphieros/vue-data-ui/releases))
- Impact on the timeline chart

**Notes on why the lines and dots animate only when moving the zoom range, and why lines do not when moving a single handle:**
_Line paths and dots are transitioned when their size remain the same (for example when sliding a range selection without changing its length). When the range changes, a new svg path is computed, therefore cannot be transitioned from a previous state. However, since dot circles are reused, when changing the range size, remaining ones were transitioned to their new position. The fix therefore consists in preventing this transition for circles in the case of a range size change._

### DIFF
--- a/app/components/Package/TimelineChart.vue
+++ b/app/components/Package/TimelineChart.vue
@@ -914,6 +914,10 @@ const indexSelection = computed(() => {
   transition: all 0.5s var(--super-ease-out) !important;
 }
 
+:deep(.vdui-shape-no-transition) {
+  transition: none !important;
+}
+
 @media (prefers-reduced-motion: reduce) {
   ::deep(.vue-data-ui-component .serie_line_0 path),
   .svg-element-transition,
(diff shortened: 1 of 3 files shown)


MatteoGabriele (approved):
works like a charm! ✨

PR #2926, opened 2026-06-17
fix: set frame-ancestors 'none' to prevent embedding

danielroe:
### 🧭 Context

disables embedding using [frame-ancestors](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/frame-ancestors)

### 📚 Description

~~noticed that it was possible to embed npmx.dev in a slidev presentation~~ - this aims to harden things by adding an additional csp

**edit** - this was based on an out-of-date experience! https://github.com/npmx-dev/npmx.dev/pull/2075 already fixes this, but unsure if we still want to add frame-ancestors as boot-and-braces?


ghostdevv (approved):

43081j:
Probably still nice to have I think 👍

43081j (approved):

PR #2913, opened 2026-06-15
fix(docs): respect dark theme in badge style dropdown

BittuBarnwal7479:
### 🔗 Linked issue

#2885 

### 📚 Description

The Badge Style dropdown on the Badges page does not fully respect the active dark theme. When opened in dark mode, the dropdown menu appears with light styling, causing inconsistent appearance and reduced readability. The dropdown should match the active theme for a consistent user experience.

PR #2929, opened 2026-06-17
fix(ui): resolve root-relative README markdown links to repo blob URLs

### LINKED ISSUE #2928: root-relative markdown links resolve to npmx host
### Describe the bug

On package readme.ts page, root-relative markdown links such as `/CONTRIBUTING.md` are resolved as local npmx routes instead of repository files.

For example, on:
https://npmx.dev/package/astro page

Clicking a readme.ts link to `/CONTRIBUTING.md` opens:
https://npmx.dev/CONTRIBUTING.md

Instead, it should open the package repository’s rendered markdown file, for example:
https://github.com/withastro/astro/blob/HEAD/CONTRIBUTING.md

### Additional context

**Steps to Reproduce**

1. Open `https://npmx.dev/package/astro`
2. Find a README link that points to `/CONTRIBUTING.md`
3. Click the link

https://github.com/user-attachments/assets/c32faa01-12c9-4f50-901e-cebd72df1f6e

### Logs

```shell-script

```

BittuBarnwal7479:
### 🔗 Linked issue

Closes #2928 

### 🧭 Context

Steps to Reproduce

1. Open https://npmx.dev/package/astro
2. Find a README link that points to /CONTRIBUTING.md
3. Click the link

https://github.com/user-attachments/assets/66da3d63-5f49-4495-aeb6-f4faabb73f9b

### 📚 Description

On package readme.ts page, root-relative markdown links such as /CONTRIBUTING.md are resolved as local npmx routes instead of repository files.


test/unit/server/utils/readme.spec.ts
```diff
@@ -249,6 +249,36 @@ describe('Markdown File URL Resolution', () => {
...
+      )
+    })
+
+    it('keeps root-relative non-markdown links local', async () => {
+      const repoInfo = createRepoInfo()
+      const markdown = `[Package](/package/test-pkg)`
+      const result = await renderReadmeHtml(markdown, 'test-pkg', repoInfo)
+
+      expect(result.html).toContain('href="/package/test-pkg"')
+    })
```
gameroman, line 337: Is there a reason why we want this behaviour for non-markdown links?

Why not something similar to what is done for this?

https://github.com/npmx-dev/npmx.dev/blob/46e7c59a1d67daef018106f516d2e2b29f550453/test/unit/server/utils/readme.spec.ts#L219-L227

BittuBarnwal7479: Good catch. I updated this to resolve root-relative non-markdown links via `rawBaseUrl`, consistent with existing relative-link handling. Markdown files still resolve via `blobBaseUrl`.

Local npmx routes (`/package`, `/org`, `/search`, etc.) are preserved separately, and I added regression tests covering both behaviors.

gameroman: Is there a reason / place where preserving Local npmx routes would be useful instead of them resolving to rawBaseUrl too?

BittuBarnwal7479: yes. the main case is npmjs links that the README renderer intentionally converts to local npmx routes. For example, `https://www.npmjs.com/package/test-pkg` becomes `/package/test-pkg`.

If we treated every root-relative path as a repo file, that converted route would incorrectly become `rawBaseUrl/package/test-pkg`.

So the updated logic preserves known npmx routes separately, while root-relative repo files like `/CONTRIBUTING.md` and `/assets/logo.png` resolve to `blobBaseUrl` / `rawBaseUrl`.

gameroman: Hmm, is there no way to differentiate a `/package` that was because of `https://www.npmjs.com/package/test-pkg` from a `/package` that someone wrote in their readme?

BittuBarnwal7479: good point. i will update this to preserve npmjs-originated links via an internal marker rather than path matching. README-authored root-relative paths now resolve normally against the repository. Added regression tests for both cases.

BittuBarnwal7479:
@coderabbit review again.

gameroman (approved):

server/utils/readme.ts
```diff
@@ -269,6 +269,7 @@ const reservedPathsNpmJs = [
 const npmJsHosts = new Set(['www.npmjs.com', 'npmjs.com', 'www.npmjs.org', 'npmjs.org'])
 
 const USER_CONTENT_PREFIX = 'user-content-'
+const LOCAL_NPMX_REDIRECT_PREFIX = '$npmx-local:'
```
ghostdevv: is this because `resolveUrl` is being called on the same link twice?

docs/app/components/BadgeGeneratorParameters.vue
```diff
@@ -253,9 +253,11 @@ const copyToClipboard = async () => {
...
         <select
           v-model="badgeStyle"
-          class="min-w-30 px-3 py-1.5 rounded-lg border border-gray-200 dark:border-white/10 bg-white dark:bg-black/20 text-xs outline-none cursor-pointer hover:border-emerald-500 transition-colors"
+          class="min-w-30 px-3 py-1.5 rounded-lg border border-gray-200 dark:border-white/10 bg-white dark:bg-black/20 text-gray-900 dark:text-gray-100 text-xs outline-none cursor-pointer hover:border-emerald-500 transition-colors"
         >
-          <option v-for="s in styles" :key="s" :value="s">{{ s }}</option>
+          <option v-for="s in styles" :key="s" :value="s" class="dark:bg-gray-900">
+            {{ s }}
+          </option>
         </select>
```
ghostdevv, line 261: can you remove the changes here as they're in #2913

ghostdevv (changes requested):

server/utils/readme.ts
BittuBarnwal7479: yep. it’s needed because `resolveUrl` processes the link twice, so we may lose its origin.

BittuBarnwal7479:
sorry for the force push. 
i accidentally merged and pushed changes from my un-updated main branch. i’ve reset the branch back to the intended commit and removed the unrelated changes.

server/utils/readme.ts
WilcoSp: I do think it is, for changelog I prefix relative links for npmx with `$` to prevent the resolveUrl/processUrl turning them into links for the git repo

BittuBarnwal7479:
@ghostdevv  Could you please review and merge this PR before new conflicts arise? thanks!

Edit: I initially used Codex to merge it, but that failed. I've now resolved the conflicts manually :-)

gameroman:
> @ghostdevv Could you please review and merge this PR before new conflicts arise? thanks!

@BittuBarnwal7479 a maintainer will review your PR whenever they get time. Telling a maintainer to merge a PR will not make that happen faster

BittuBarnwal7479:
> > @ghostdevv Could you please review and merge this PR before new conflicts arise? thanks!
> 
> @BittuBarnwal7479 a maintainer will review your PR whenever they get time. Telling a maintainer to merge a PR will not make that happen faster

Understood, thanks for the clarification. I won't ping maintainers for merges going forward.

docs/app/components/BadgeGeneratorParameters.vue
trueberryless: Resolved in [`734f947` (this PR)](https://github.com/npmx-dev/npmx.dev/pull/2929/commits/734f9478bc16244612c52f57dd4c48b7e83df6fa)

trueberryless:
Hi @BittuBarnwal7479 👋 

There seems to be an unresolved thread where you have not answered yet: https://github.com/npmx-dev/npmx.dev/pull/2929#discussion_r3486933080. Could you please explain your thoughts on that and verify Willow's assumptions?

Also, I am not sure if this was addressed yet: https://github.com/npmx-dev/npmx.dev/pull/2929#discussion_r3430487917. The force commit makes it a little hard to follow. Can you make sure that all threads are resolved? That would make reviewing much easier 👍

BittuBarnwal7479:
Yes @WilcoSp , that is the reason.
`resolveUrl()` can see the same link twice: once during markdown rendering, then again during sanitization of the generated `<a>`. For npmjs links we intentionally rewrite `https://www.npmjs.com/package/foo` to a local npmx route like `/package/foo`.

without an internal marker, the second pass cannot tell whether `/package/foo` was authored in the README or came from an npmjs rewrite. That means it could incorrectly resolve the local npmx route as a repository-root file.

The marker is only internal and is stripped before output. Authored root-relative README links like `/CONTRIBUTING.md` still resolve to the repository blob URL.

Btw sorry for the late reply.

BittuBarnwal7479:
I’ve addressed all reviewer comments from my side. If anything still needs clarification or another change, please let me know and I’ll update it. Thanks!

ghostdevv (approved):
I pushed [`cc1accc` (this PR)](https://github.com/npmx-dev/npmx.dev/pull/2929/changes/cc1accc1d517df79e8db5c13272f48692cdc4bda) to align closer with the changelog renderer, but it should behave the same

https://github.com/npmx-dev/npmx.dev/blob/cc1accc1d517df79e8db5c13272f48692cdc4bda/server/utils/changelog/markdown.ts#L129-L130

gameroman (approved):
Seems to work 👍

server/utils/readme.ts
```diff
@@ -176,14 +176,18 @@ function toUserContentHash(value: string): string {
   return `#${withUserContentPrefix(value)}`
 }
 
+function isMarkdownFileUrl(url: string): boolean {
+  return /\.(?:md|markdown)$/i.test(url.split('?')[0]?.split('#')[0] ?? '')
+}
+
```
WilcoSp, line 182: the `isMarkdownFileUrl` function could be moved to mdkit because changelog also needs to check whether an url is for markdown or not

WilcoSp (approved):
looks good, I did have a comment for the `isMarkdownFileUrl` function but it's not a stopper for me.

server/utils/readme.ts
ghostdevv: Would also be good to use `URL.parse` too, something like `URL.parse(url)?.pathname.endsWith...`

server/utils/mdKit.ts
```diff
@@ -217,6 +217,8 @@ export function getHeadingSlugSource(text: string): string {
 }
 
 const htmlAnchorRe = /<a(\s[^>]*?)href=(["'])([^"']*)\2([^>]*)>([\s\S]*?)<\/a>/gi
+const anchorTokenOpenRe = /^<a(?:\s.+)?\/?>$/i
```
trueberryless, line 220: I think this would miss a link element that has multiple newlines, e.g.:

```html
<a 
  href="link"
  class="button">
```

We could add the `s` option to also include newlines in the `.` expression to solve this issue:

```suggestion
const anchorTokenOpenRe = /^<a(?:\s.+)?\/?>$/is
```

trueberryless (commented):
Overall looks good to me, just found a potential issue with the regex.

PR #2933, opened 2026-06-19
feat(i18n): add Korean translations

yeonjulee1005:
### 🧭 Context

Adds Korean (ko-KR) translations and registers the locale.

### 📚 Description

Added a new `ko-KR.json` locale and updated `config/i18n.ts` accordingly.

### DIFF
--- a/config/i18n.ts
+++ b/config/i18n.ts
@@ -306,12 +306,12 @@ const locales: (LocaleObjectData | (Omit<LocaleObjectData, 'code'> & { code: str
       code: 'gl-ES',
       file: 'gl-ES.json',
       name: 'Galego',
-    },
-    {
-      code: 'ko-KR',
-      file: 'ko-KR.json',
-      name: '한국어',
     },*/
+  {
+    code: 'ko-KR',
+    file: 'ko-KR.json',
+    name: '한국어',
+  },
   {
     code: 'id-ID',
     file: 'id-ID.json',
--- a/i18n/locales/ko-KR.json
+++ b/i18n/locales/ko-KR.json
@@ -1,0 +1,1788 @@
+{
+  "$schema": "../schema.json",
+  "seo": {
+    "home": {
+      "title": "npmx - npm 레지스트리를 위한 패키지 브라우저",
+      "description": "npm 레지스트리를 위한 빠르고 현대적인 브라우저입니다. 현대적인 인터페이스로 패키지를 검색하고, 둘러보고, 탐색하세요."
+    }
+  },
+  "built_at": "빌드 {0}",
+  "alt_logo": "npmx 로고",
+  "tagline": "npm 레지스트리를 위한 빠르고 현대적인 브라우저",
+  "non_affiliation_disclaimer": "npm, Inc.와 관련이 없습니다.",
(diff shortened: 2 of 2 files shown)


graphieros (approved):
Thank you! 🌿

I do not speak Korean, and could only check with translation tools, and LGTM.

yeonjulee1005:
@graphieros Thank you for your review! 😉
I've translated it into the language frequently used by Korean developers~!
I'll continue to contribute whenever needed!
how can I become a maintainer?

kimkanu:
Hi! Sorry to bother, but I also have created a pull request #2862  2 weeks ago with the same feature. Could you check that out as well?

yeonjulee1005:
@kimkanu 
Oh, I see there was a previous PR submitted. My apologies for missing that.

However, after reviewing the PR you mentioned, it seems to include a `koreanModifiers` converter for handling Korean particles `(e.g., 은/는, 이/가)`.
I believe this may introduce additional complexity and increase the number of maintenance points in the codebase.

In our case, I think it would be better for native Korean developers to participate directly in the translation process instead.

@graphieros , I would also appreciate your input on this matter.

graphieros:
> @graphieros , I would also appreciate your input on this matter.

@yeonjulee1005, as a non Korean speaker, I can only trust you find an agreement on an approach that is most maintainable and accurate.

kimkanu:
@yeonjulee1005 , thank you for your considerate comment.

Personally, as another native speaker of Korean language, I think choosing the correct *josa* is one of the most important feature of Korean translation. I find it very distracting if there's something wrong with *josa*, or the page is using something like "은(는)" to include both cases. Not only that they seem imperfect and not caring Korean readers, but also it could be distracting for screen reader users because the pronunciations are quite different -- though I'm not a screen reader user, so I'm not 100% sure on this point.

I agree on the point that using bunch of custom modifiers makes the code complicated, especially compared to other languages. And my code has a lot of heuristics which are incorrect under a handful of circumstances.

I'm really open to find a good sweet spot between complexity and correctness. :)

yeonjulee1005:
@kimkanu 

I totally agree with your point on _josa_ it's genuinely `one of the trickiest parts of Korean`, and getting it wrong breaks the reading flow immediately for native speakers.

That said, I think the most pragmatic sweet spot here might be simpler than any heuristic: leave Korean-specific grammar decisions to native Korean-speaking contributors.
_Josa_ correctness is hard to automate well, and the edge cases are numerous enough that even clever heuristics will occasionally produce something that sounds off.

So rather than trying to solve it in code, it might be worth setting a contribution guideline that Korean translations should be reviewed or contributed by native speakers, similar to how some projects handle RTL languages or complex scripts.

Happy to help improve the Korean translations directly if that's useful! 😊

ghostdevv:
Where should we go from here @yeonjulee1005 @kimkanu - none of the maintainers speak Korean as far as I'm aware so we need some pointers about how best to merge these changes. We should pick one PR to merge, and copy across any changes necessary - I'll also make sure both you two get credit on the merge :smile:

graphieros:
#3024 is now another candidate

PR #2956, opened 2026-06-24
perf: use lightning css transformer

TheAlexLichter:
### 🧭 Context

Faster builds with what Vite 8 will use as the default.

### 📚 Description

This branch enables Vite’s Lightning CSS transformer and makes UnoCSS directive transforms run in `pre` to expand directives before Lightning CSS. It also fixes one invalid `::deep(...)` -> `:deep(...)` reference.

On my machine, this reduces build time from 9.16s to 8.57s (~6.4% faster).

### DIFF
--- a/app/components/Package/TimelineChart.vue
+++ b/app/components/Package/TimelineChart.vue
@@ -919,7 +919,7 @@ const indexSelection = computed(() => {
 }
 
 @media (prefers-reduced-motion: reduce) {
-  ::deep(.vue-data-ui-component .serie_line_0 path),
+  :deep(.vue-data-ui-component .serie_line_0 path),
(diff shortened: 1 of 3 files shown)


app/components/Package/TimelineChart.vue
```diff
@@ -919,7 +919,7 @@ const indexSelection = computed(() => {
 }
 
 @media (prefers-reduced-motion: reduce) {
-  ::deep(.vue-data-ui-component .serie_line_0 path),
+  :deep(.vue-data-ui-component .serie_line_0 path),
```
graphieros, line 922: omg thank you ^^

graphieros (approved):
😎🌿

PR #2957, opened 2026-06-24
fix(i18n): add missing Dutch translations

larsdouweschuitema:
### 🔗 Linked issue

No open issue related to Dutch translations.

### 🧭 Context

Dutch language contains incorrect and missing translations.

### 📚 Description

All translations were reviewed and manually tested in a local environment.

- Some keys were incorrectly translated in a sentence. For example, "affiliëerd" should be "geaffilieerd"
- Some keys were wrongly typed. For example, "geregisteerd" should be "geregistreerd"
- Some keys were using English term. For example, "trademark" instead of "handelsmerk"
- Mix of informal and formal pronouns (u/jij) was used. Majority of existing translations uses formal style and so I opted for a consistent formal style of pronouns.
- Mixed pronouns were used (u/uw, je/jij). The majority of usage is formal, so I opted for the formal pronouns usage
- Translated missing keys listed in the "translation status" for the Dutch language
- Some translations (across multiple languages in fact) use "ChatGPT-style" title capatalization, where each word starts with a capital letter. Since styles are mixed, I removed the "ChatGPT-style" capatalization for some keys
- English translations referred to "the repository", while Dutch translation used "een repository". I aligned the Dutch translation with the English version for consistency.
- Replaced "Wijzigingenoverzicht" with "Changelog" because I believe it is more clear to the target audience of this app. Although it is an English term, it is commonly used in technical discussions within the Netherlands. This is subjective change and chan be reverted if preferred.
- I translated "stats" to "statistieken" where appropriate. However, I kept "stats" in the navigation menu because space is more limited and the term is common to the target audience (even if it's a non-Dutch term). This is also a subjective choice and can be adjusted if necessary.


i18n/locales/nl.json
```diff
@@ -154,7 +160,8 @@
       "diff": "Verschillen",
       "compare": "Vergelijk dit pakket",
       "download": "Download tarball",
-      "changelog": "Wijzigingenoverzicht"
+      "changelog": "Changelog",
```
WilcoSp:
```suggestion
      "changelog": "Wijzigingen",
```

I would rather keep it Dutch, but I do think removing "overzicht" is better because the current ui does imply it being an "overzicht" (overview).

In general I do want to keep English terms to a minimum and only have them when the Dutch translation doesn't have the same context when applied in npm(x)
It's for example why I had chosen "dependency" instead of "afhankelijkheid" because the context to me is different.

i18n/locales/nl.json
```diff
@@ -450,8 +485,9 @@
       "fund": "Steunen",
       "compare": "vergelijk",
       "timeline": "tijdlijn",
+      "stats": "stats",
       "compare_this_package": "Vergelijk dit pakket",
-      "changelog": "wijzigingenoverzicht"
+      "changelog": "changelog"
```
WilcoSp:
```suggestion
      "changelog": "Wijzigingen",
```

i18n/locales/nl.json
```diff
@@ -1182,6 +1226,9 @@
     },
     "team": {
       "title": "Team",
+      "core": "Core",
```
WilcoSp:
```suggestion
      "core": "Kern",
```

This can be translated to "Kern" instead of "core"

i18n/locales/nl.json
```diff
@@ -1182,6 +1226,9 @@
...
+      "maintainers": "Onderhouders",
+      "role_core": "core",
```
WilcoSp:
```suggestion
      "role_core": "kernlid",
```

this could be translated either to "kern" (core) or "kernlid" (core member)

WilcoSp (changes requested):
In general good changes, only for "Changelog" & "Core" I do have suggested changes as I do think it's better to keep the English terms to a minimum.

also nice that you caught the informal translations as it should've been formal, at my work we use informal which makes it that when translating that I sometimes translate to informal instead of formal.

larsdouweschuitema:
> In general good changes, only for "Changelog" & "Core" I do have suggested changes as I do think it's better to keep the English terms to a minimum.
> 
> also nice that you caught the informal translations as it should've been formal, at my work we use informal which makes it that when translating that I sometimes translate to informal instead of formal.

Thanks for your review! I agree with all suggestions and resolved them. With regards to keeping Dutch at a minimum, I don't have a strong preference, but I can understand!

ghostdevv (approved):

ghostdevv:
> [Lars Douwe Schuitema](https://github.com/larsdouweschuitema) requested a review from [WilcoSp](https://github.com/WilcoSp)

Looks like there weren't many more changes since last review so we can merge this and fix anything in a follow up :smile: 

> With regards to keeping Dutch at a minimum, I don't have a strong preference, but I can understand!

English*

i18n/locales/nl.json
```diff
@@ -229,6 +237,31 @@
...
+    "title": "noodles",
+    "meta_description": "Elke noodle die ooit op npmx verscheen — de seizoenslogo's, easter eggs en de verhalen erachter.",
+    "latest": "Laatste noodles",
+    "what_is": "Wat zijn noodles",
+    "what_is_body": "Noodles zijn de speelse varianten van het npmx-logo die we op de homepage tonen bij releases, feestdagen, evenementen en andere momenten die de moeite waard zijn om te vieren. Zie ze als Google Doodles, maar dan voor npm-packages.",
+    "empty": "Nog geen noodles.",
+    "load_more": "Laad nog {count}",
+    "dates": "Actieve datums",
+    "shipped_in": "Uitgebracht in",
+    "credits": "Credits",
```
WilcoSp, line 250: nl: In de volgende pr misschien dit veranderen naar "Ontworpen door" i.p.v. "Credits"

en: in the next pr maybe replace this with "Ontworpen door" instead of "Credits"

WilcoSp (commented):
Yeah looks good in general, did find out that the casing for "Wijzigingen" is correct compared to the other tabs and I've a comment for "credits".
Further it looks all good

Image

Also maybe an idea could be to have a monthly cycle with updating nl.json to where a draft pr is created every month and then in the last week of the month it's opened for review.

this could allow while a pr is in draft to update it in smaller chunks instead of needing to update all new additions in a short amount of time, it could also allow for discussion in cases of being unsure about something.

WilcoSp (approved):

ghostdevv:
> Also maybe an idea could be to have a monthly cycle with updating nl.json to where a draft pr is created every month and then in the last week of the month it's opened for review.
> 
> this could allow while a pr is in draft to update it in smaller chunks instead of needing to update all new additions in a short amount of time, it could also allow for discussion in cases of being unsure about something.

I generally try to reduce open PRs in favour of quicker more focused ones as there is a lot of activity on the repo and I don't want them to go stale or have duplicates. That being said, if you want to do it go for it, we do have that stale bot now anyway - also, I'm happy to tag you for reviews on Dutch related PRs we get if you like!

since I don't speak Dutch the best I can do is run it through a translator to make sure there's nothing obviously wrong, but it's always great if we have native speakers that we know (we can't tell if it's been translated, or clanked) to make sure we don't have mistakes - currently there's no process for tagging people for this though

PR #2958, opened 2026-06-24
perf: lazy load Bluesky video player

TheAlexLichter:
### 🔗 Linked issue

None (but perf is always good!)

### 📚 Description

This PR lazy-loads the Bluesky video player so non-video Bluesky embeds no longer pull in the full HLS player chunk if they dont need it.

Optionally, a `media-hint="video"` preload hint for known video posts can be set, which starts loading the video player earlier while still rendering only when post.embed?.playlist is present. This is optional but recommended when we know the content.

During tests, HLS split into its own (now optional) chunk at about 152.2 KB gzip, with the Bluesky embed chunk staying small at about 2.1 KB gzip.

### DIFF
--- a/app/components/global/BlueskyPostEmbed.client.vue
+++ b/app/components/global/BlueskyPostEmbed.client.vue
@@ -10,7 +10,13 @@ const props = defineProps<{
   uri?: string
   /** Bluesky URL of the post, e.g. https://bsky.app/profile/handle/post/rkey */
   url?: string
+  /** Optional preload hint. Rendering still depends on the fetched embed metadata. */
+  mediaHint?: 'video'
 }>()
+
+if (props.mediaHint === 'video') {
+  void import('~/components/VideoPlayer.vue')
+}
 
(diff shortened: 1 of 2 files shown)


app/components/global/BlueskyPostEmbed.client.vue
```diff
@@ -10,8 +10,14 @@ const props = defineProps<{
   uri?: string
   /** Bluesky URL of the post, e.g. https://bsky.app/profile/handle/post/rkey */
   url?: string
+  /** Optional preload hint. Rendering still depends on the fetched embed metadata. */
+  mediaHint?: 'video'
 }>()
 
+if (props.mediaHint === 'video') {
+  void import('~/components/VideoPlayer.vue')
```
TheAlexLichter, line 18: didnt use `preloadComponent` because this only works with globally registered components.

alexdln (approved):
Ah, looks great, thank you 🤍🚀

PR #2959, opened 2026-06-24
perf: import module replacements util directly

TheAlexLichter:
### 📚  Description

We don't use the module replacements manifest, only the util.
As @gameroman pointed out in https://github.com/e18e/module-replacements/pull/751, we can already use the dist alias to avoid pulling in the manifest and saving 21 kB.

This PR does exactly that.

Note: If https://github.com/e18e/module-replacements/pull/751 lands I suggest to use the "official alias".


gameroman (approved):

ghostdevv:
I'll close this for now since we'll get https://github.com/e18e/module-replacements/pull/753 soon™ and I can confirm that the fix works here :tada:

PR #2964, opened 2026-06-26
feat: render docs for multi-entry packages

### LINKED ISSUE #1479: api docs don't generate when package is missing default export
npmx seems to not recognize my types, the project isn't using typescript source files, but it does expose .d.ts types and exposes them in the package.exports

https://www.npmx.dev/package/@thepassle/app-tools

Image
Image
Image

maraisr:
### 🔗 Linked issue

resolves #1479

### 🧭 Context

Some packages don't ship a single root entry. My library [`tctx`](https://github.com/maraisr/tctx) is one of them. It only exposes submodule exports:

```jsonc
"exports": {
  "./traceparent": { "types": "./traceparent.d.mts", "default": "./traceparent.mjs" },
  "./tracestate": { "types": "./tracestate.d.mts", "default": "./tracestate.mjs" },
  "./package.json": "./package.json"
}
```

Because the docs pipeline assumes a single root (`.` / `index`) entry, packages like this fall through and render nothing on the Docs tab. You can see it live today on [npmx.dev/package-docs/tctx](https://npmx.dev/package-docs/tctx/v/0.2.6):

empty docs for tctx

### 📚 Description

This PR is really my attempt at enhancing the docs generator to resolve and render **multiple entry points** instead of assuming one root module.

I can now see my docs rendering correctly! 🎉 

With this, `tctx` now has real docs. one group per submodule (`traceparent`, `tracestate`):

tctx Docs tab with this PR — grouped submodule docs

It also handles the common case of a package that exports **a root *and* submodules** together, eg `unstorage` (`.` + `./server`):

unstorage Docs tab. root + ./server grouped

#### Scope & follow-ups

I deliberately kept this slice small so it's easy to reason about:

- **Wildcard submodules** (e.g. `"./foo/*"`) aren't solved yet. I'd love to iterate toward that, but wanted to land something useful first rather than block on the harder cases.

And a small **sneak peek** of what I've got working locally for a follow-up PR, rendering `@module` docs too:

@module description + example rendered above an entry's symbols

#### A note on landing this

I genuinely don't know what the right rollout or shape here is, and I'm very open to feedback on the approach, the entry resolution, and especially the grouped-docs design (headings, ordering, how multi-entry should *feel*). Happy to chat it through and iterate; please push back wherever it makes sense. 🙏


danielroe:
this is nice!

I wonder if we could distinguish these 'files' headings a bit, or make it more obvious that it's a subpath export:

image

e.g. `tctx/tracestate` as the heading?

maraisr:
Hey thanks for taking a look! ✨ I'll have another play, but have you got ideas? I didn't really like what [jsr did with indentation](http://jsr.io/@mr/tracecontext/doc/all_symbols), but perhaps simply some top margin there is enough to distinguish? I also trimmed the `"./"`. Which maybe we can put back to better indicate that its a "submodule". Thoughts, ideas?

maraisr:
What about something like this @danielroe?

CleanShot 2026-06-26 at 23 16 35@2x

Think the text mono on the headings is a little whimsical and fits better with the vibe of the site? Thoughts? 

Added this too, so that it ellipsis, and on the left. Thinking long submodules like `./foo/bar/baz`, the `baz` part is possibly a little more important. But we can play with it iteratively, like maybe we prune things like `./foo/bar/baz/index → ...bar/baz`. Lets do that in a follow-up though, thoughts?

CleanShot 2026-06-26 at 23 37 17@2x

danielroe:
yes, I think that looks better ... it doesn't seem to show up in the preview yet though, if you've pushed it

maraisr:
I've definitely pushed. The latest preview deployment shows up for me 😅

danielroe (approved):
nice work ❤️ 

... and sorry for the long delay!

PR #2967, opened 2026-06-27
feat: sponsors

alexdln:
### 🧭 Context

A dedicated page for sponsors. This is useful for both current sponsors and those just considering it. It's a public and always-open source of important information


i18n/locales/en.json
```diff
@@ -1268,6 +1269,69 @@
...
+        "name": "Silver",
+        "price": "$500/month",
+        "benefit_1": "We display your logo on the sponsors and about pages.",
+        "benefit_2": "We announce you on our socials."
+      },
+      "gold": {
+        "name": "Gold",
+        "price": "$1,000/month",
+        "description": "Everything in Silver, plus:",
+        "benefit_1": "We display your name in the footer on every page.",
```
patak-cat: I think for now we can say that gold will get the logo displayed as Gold sponsor in /about and the home footer. And same for Silver. So same as we merged in main.

i18n/locales/en.json
```diff
@@ -1268,6 +1269,69 @@
...
+        "benefit_2": "We announce you in the next release."
+      },
+      "platinum": {
+        "name": "Platinum",
+        "price": "Custom",
+        "description": "Everything in Gold, plus:",
+        "benefit_1": "We display your logo at npmx meetups and talks related to npmx, where event organizers allow it.",
+        "benefit_2": "We publish a dedicated article when sponsorship begins, plus follow-up articles after 6 and 12 months. We're also happy to help you write an article for your own blog and share any info you need."
```
patak-cat: I would say something like "Let's talk" as each of these bigger sponsoring relationship may be unique in what we do. I think we should display the logos of every sponsor (also silver and gold, smaller) if we show them in a slide in a talk (but it is good to be flexible and don't commit to always show them as it depends on the conference or meetup).

i18n/locales/en.json
```diff
@@ -1268,6 +1269,64 @@
...
+        },
+        "visitors": {
+          "description": "unique monthly visitors"
+        },
+        "stars": {
+          "title": "stars"
+        },
+        "community": {
+          "title": "fast-growing",
+          "description": "We have one of the fastest-growing communities in the ecosystem"
```
patak-cat: Let's reword this so it is clear the project is a community, instead of having one.
```suggestion
          "description": "We are one of the fastest-growing communities in the ecosystem"
```

alexdln: Great point, thanks ☺️

patak-cat (approved):

PR #2977, opened 2026-06-29
chore: add more Wikipedia references for noodles

shuuji3:
### 🔗 Linked issue

N/A

### 🧭 Context

Some noodle pages are missing references.

### 📚 Description

- Added more Wikipedia references to the Pride Month, International Transgender Day of Visibility, and Artemis II pages.
- Also added ` - Wikipedia` to the end of the links as it wasn't clear where the link would bring me until checking the URL.

(Personally, [International Transgender Day of Visibility](https://en.wikipedia.org/wiki/International_Transgender_Day_of_Visibility) article is a memorable article, which is the first LGBTQ+ related article I translated from English Wikipedia into Japanese several years ago 🙂)


app/noodles.ts
```diff
@@ -40,6 +40,12 @@ const entries: Noodle[] = [
     authors: [ALEX],
```
alexdln: Oh, can I ask you to correct this one too 🙃

```suggestion
    authors: [ALEX, ALFON],
```

alexdln (approved):
Wow, this is cool, thank you! 🏯

app/noodles.ts
shuuji3: Sure!

PR #2979, opened 2026-07-01
feat: add dependency size chart to timeline

### LINKED ISSUE #1005: Bundlephobia-like package size history chart
Display (maybe in a popup kinda like downloads history) package size history chart, kind of like on Bundlephobia:

Image

Take advantage of cached package/install size information we already have. Display gray bars - again, similarly to Bundlephobia - for versions we didn't calculate yet and calculate them on demand on click, in order not to flood the server with requests.

43081j:
### 🔗 Linked issue

Closes #1005

### 🧭 Context

The timeline charts currently show the overall package size, but showing the segmented dependency sizes would be useful too.

### 📚 Description

Does what it says on the tin! Shows a nice new 3rd chart that has the segmented install size (i.e. each significant dependency is a segment in the bar).


graphieros (approved):
;)

gameroman (approved):
Love it!

The animation when toggling "stable only" is a bit weird/laggy though

app/components/Package/TimelineChart.vue
```diff
@@ -173,6 +208,99 @@ const e18eGradientColors = [
...
+// Creates the segments for the dependency size chart. Basically, each
+// significantly sized dependency is a segment, the package itself is a segment,
+// and the rest is a segment ("Other").
+const dependencySegments = computed<DependencySegment[]>(() => {
+  const data = orderedConvertedData.value
+  const reference = data.at(-1)
+  if (!reference) return []
+
+  const topNames = reference.dependencies.slice(0, DEP_SEGMENT_COUNT).map(dep => dep.name)
+  const topNameSet = new Set(topNames)
```
gameroman, line 237: What does `reference` do?

graphieros:
> Love it!
> 
> The animation when toggling "stable only" is a bit weird/laggy though

I used the same ease-out function as for the transitions in the line charts. Maybe they can be slightly faster for bars, as bars are... heavy by definition.

graphieros:
> > The animation when toggling "stable only" is a bit weird/laggy though
@gameroman wdyt now ?

gameroman:
> > The animation when toggling "stable only" is a bit weird/laggy though
>
> @gameroman wdyt now ?

Still feels kinda off. Maybe we can try without animation at all?

app/components/Package/TimelineChart.vue
graphieros: the last version in the data is taken as reference to pick which dependencies are part of the segmentation throughout the timeline

graphieros:
> Still feels kinda off. Maybe we can try without animation at all?

I'd prefer to keep the transition as it is now, users will often toggle legend items, and transitions are nice in this case

app/components/Package/TimelineChart.vue
```diff
@@ -1,5 +1,4 @@
 <script setup lang="ts">
```
ghostdevv, line 1: we really need to start splitting up some of this 😄

graphieros: I can work on improvements

graphieros: @ghostdevv I extracted 210 lines (it ain't much but it's honest work)

graphieros:
I believe this PR will also close #1005

phaux:
This doesn't fix the linked issue. See my comment there.

PR #2982, opened 2026-07-01
feat(ui): premium app features — installability, WCO, share, badging

tomayac:
## Summary
- Enable PWA installability (`@vite-pwa/nuxt`) with a manifest id and generated install-UI screenshots.
- Add Web Share API support on package pages and a Window Controls Overlay + app shortcuts layout.
- Add the App Badging API to surface new-likes counts on the app icon.
- Replace the custom toggle component with the native `<input type="checkbox" switch>` polyfill, kept live-synced with the accent color/theme.
- Keep `theme-color` in sync with `--bg` so the WCO title-bar strip matches the header.

## Test plan
- [x] `pnpm test:unit` — 1652/1652 passing
- [x] `pnpm run test:types` — passing (fixed 2 pre-existing type errors in the switch polyfill plugin)
- [x] `vp lint` — 0 errors (3 pre-existing style warnings, unrelated to correctness)
- [x] `pnpm run build` — succeeds


gameroman (commented):
This probably needs a discussion first

tomayac:
> This probably needs a discussion first

(I was chatting about this with @danielroe.)

tomayac:
## Installed PWA with Window Controls Overlay  enabled

Screenshot 2026-07-01 at 18 38 42

## Installed PWA with Window Controls Overlay  disabled

Screenshot 2026-07-01 at 18 40 12

## Share button for packages

Screenshot 2026-07-01 at 18 38 58

## App icon shortcuts

Screenshot 2026-07-01 at 18 40 01

## Install prompt with screenshots

Screenshot 2026-07-01 at 18 41 30

## Accent colors for switches 

(Uses native switch on Safari with support for dragging the toggle, polyfill on other browsers)

Screenshot 2026-07-01 at 18 44 51

app/components/Button/Base.vue
```diff
@@ -41,6 +41,7 @@ const keyboardShortcutsEnabled = useKeyboardShortcuts()
 
 defineExpose({
   focus: () => el.value?.focus(),
+  click: () => el.value?.click(),
```
ghostdevv: iirc vue would auto expose this

pnpm-workspace.yaml
```diff
@@ -40,6 +40,7 @@ overrides:
   sharp: 0.34.5
   vite: npm:@voidzero-dev/vite-plus-core@0.1.20
   vitest: npm:@voidzero-dev/vite-plus-test@0.1.20
+  vue: 3.5.39
```
ghostdevv, line 55: how come?

scripts/generate-pwa-screenshots.ts
```diff
@@ -0,0 +1,206 @@
...
+  } catch (err) {
+    server.kill('SIGTERM')
+    throw err
+  }
+
+  console.log('Preview server ready.\n')
+  return { process: server, url }
+}
+
+async function main(): Promise<void> {
```
ghostdevv: you can use top level await so you don't need a `main` fn

scripts/generate-pwa-screenshots.ts
```diff
@@ -0,0 +1,206 @@
+#!/usr/bin/env node
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 npmx contributors
```
ghostdevv: this isn't necessary

app/plugins/input-switch-polyfill.client.ts
```diff
@@ -0,0 +1,26 @@
+export default defineNuxtPlugin(async () => {
```
ghostdevv: We already have switches styled with `role="switch"` - what's the benefit of polyfiling this?

ghostdevv (changes requested):
There are a couple distinct, but related, changes in this PR - could you split them up into multiple PRs?

app/components/Button/Base.vue
tomayac: Verified against Vue's own source (`@vue/runtime-core`, `getComponentPublicInstance`): once a component calls `defineExpose()`, the proxy a parent's template ref sees only contains what's explicitly listed, plus Vue's `$`-prefixed magic properties (`$el`, `$props`, `$emit`, etc., from `publicPropertiesMap`) — there's no fallback to arbitrary DOM methods like `.click()`.

```js
// @vue/runtime-core
function getComponentPublicInstance(instance) {
  if (instance.exposed) {
    return instance.exposeProxy || (instance.exposeProxy = new Proxy(proxyRefs(markRaw(instance.exposed)), {
      get(target, key) {
        if (key in target) return target[key]
        else if (key in publicPropertiesMap) return publicPropertiesMap[key](instance)
      },
      ...
```

So without this line, `buttonRef.value?.click()` in `ShareButton.client.vue` (used for the `v` keyboard shortcut, so the share sheet anchors at the button rather than the cursor) would just be `undefined`. It's consistent with the existing `focus`/`getBoundingClientRect` entries above it, which already manually forward DOM methods the same way rather than relying on `$el` — keeping `click` explicit here matches that pattern. Happy to switch to `buttonRef.value?.$el?.click()` at the call site instead if you'd prefer avoiding the extra expose entry, though that trades one explicit line for a slightly fuzzier `$el` cast.

pnpm-workspace.yaml
tomayac: Resolving a pnpm-lock.yaml merge conflict caused pnpm to resolve two different vue versions (3.5.34 via @nuxt/test-utils, 3.5.39 elsewhere), despite dedupePeers already being set. Two Vue instances in one app is a real hazard since reactivity tracking is per-instance. Pinning forces a single resolved copy.

scripts/generate-pwa-screenshots.ts
tomayac: Fixed in 71eedc52

app/plugins/input-switch-polyfill.client.ts
tomayac: The reason to use the `switch` attribute with the polyfill is that on [WebKit you get the real switch control](https://webkit.org/blog/15054/an-html-switch-control/) (and the polyfill isn't even loaded), and the polyfill supports all the features the real control supports, like the accent color and also dragging the thumb to change the state of the toggle, which CSS-based solutions don't support.

scripts/generate-pwa-screenshots.ts
tomayac: Fixed in 209bac62

tomayac:
Per the review feedback about scope, I've split this PR:

- Web Share button → #2997
- Input switch polyfill → #2998

This PR is now focused on PWA installability, WCO, and app badging.

tomayac:
CC: @userquin per his request `:-)`

userquin:
I have no idea what's doing Google with Chrome, from time to time devtools panels stop working, a few months ago, all pwa icons broken, now it is time to break Application > Manifest:

http error 429 when opening manifest at application tab at devtools

empty manifest panel at application tab

webmanifest works when opening the link in a new tab

tomayac:
This is Vercel's staging preview: it fails with HTTP error 429, not DevTools' fault. The workaround is to right-click the request, selecting **Override content** ([instructions](https://developer.chrome.com/docs/devtools/overrides)):

Screenshot 2026-08-28 at 10 21 11

Then you paste the manifest contents and the app starts working. I have no idea why Vercel would choke on this one request, but it's definitely on their side.

userquin:
Uhmmm, we need to enable a flag at pwa conf, I will check the content type, nitro wont add it

userquin:
This one https://github.com/vite-pwa/nuxt/blob/main/src/types.ts#L65

userquin:
> This is Vercel's staging preview: it fails with HTTP error 429, not DevTools' fault.

upps, vercel using content disposition and content-security-policy

@danielroe is there a way to disable this?, I'm going to enable `registerWebManifestInRouteRules` and check it...

image

userquin:
Initial review:
- missing base and scope at pwa options, looks like nuxt using base `./` and the sw won't install when installing from a page !== '/' => we need to add webmanifest, sw and dev-sw at `server/middleware/canonical-redirects.global.ts` (at dev not working)
- it seems pwa module breaking vue and nuxt module augmentations
- all pwa shortcut icons using the same icon, we need to use custom ones
- we need to enable some pwa options
- enable devOptions via some env var: check how we did it at elk.zone
- change pwa icons per environment: `pwaAssets` option must be removed, pwa assets generated with pwa assets generator CLI ~~iirc I did it using unshift at nitro public dirs (I need to check it), but not working on local dev => pwa icons should be green (public-dev) instead white ones (public); it is not working at npmx the favicon.ico resolved from public folder, the build-env module needs a review (maybe nitro issue 🤔?)~~
- figure out the best strategy for offline support: Thomas removed the navigation fallback and the html files from the precaching ~~right now it is not working (I think enabling pwa options should fix it)~~
- move strategy to custom SW (`injectManifest` strategy)
- include api calls at denylist
- add network first handler for SSR pages with custom runtime caching with max entries
  - this way offline will work if cached
  - we need to check if we can change the max entries at runtime (SW): if so, we can include a new setting (maybe using indexedDB or broadcast channel)
  - we need to include storage quota checks and iOS can be a pain (50MB only and can deleted the storage if the app not being used for a few days without any advise)

<details>
<summary>New PWA Options</summary>

```ts
    registerWebManifestInRouteRules: true,
    experimental: {
      enableWorkboxPayloadQueryParams: true,
    },
    devOptions: {
      enabled: true, // ==> set it to false when pushing here <==
      type: 'module',
      suppressWarnings: true,
    },
```

new pwa optionas at nuxt config file
</details>

Maybe we can use the new nuxt and unplugin pwa packages via file protocol using local tgz files (I still need to release initial alpha version, but I need to do some internal deep change before releasing), I have tested it with nuxt 3, 4 and 5:
- SW code splitting: no more SW barrel
- build dual ESM/Classic SW
- Vite 8 dual SW parallel build (workbox fork now detects Vite via Vite+ correctly)
- use Nuxt virtual modules at SW: `import manifest from '#app-manifest'`
- use Nuxt path alias at pwa options and external pwa config file: custom SW path and custom external PWA config file location (`swSrc: '~/sw.ts'` at pwa config file and `pwa: { path: '~~/pwa.config.ts' }` at nuxt config file)
- I'm also working to include some new i18n and color scheme stuff at webmanifest:
  - [color_scheme_dark member](https://w3c.github.io/manifest/#color_scheme_dark-member)
  - [*_localized members](https://w3c.github.io/manifest/#x_localized-members)

userquin:
Here some screenshots with the new PWA stuff:

<details>
<summary>ESM SW</summary>

ESM service worker on chrome on local build
</details>

<details>
<summary>Vite PWA dual SW build</summary>

nuxt build showing the parallel build
</details>

<details>
<summary>External PWA configuration</summary>

nuxt pwa option with just the path entry for the pwa configuration
</details>

<details>
<summary>Parallel SW precaching</summary>

service worker configuration to download precache assets in parallel: 5
</details>

userquin:
We may need to disable a few things at pr preview, will require exporting `isPreview` from `config/env.ts` (`/_v/scripts.js`  also with 429 in this PR):
- disable sw parallel precaching: maybe also for production (review also what's should be precached)
- disable prefetching links (I need to check if enabled)
- disable payloadExtraction (it is enabled by default)

/cc @danielroe

PR #1879, opened 2026-03-03
fix(blog): resolve blog author avatars at build time

danielroe:
### 📚 Description

this is a slight perf improvement, hopefully!

PR #2994, opened 2026-07-04
fix: standard site sync

ghostdevv:
Looks like after #1879 (probably) when the blog post avatars were done at build time, the schema that was being used in the standard site sync code for the blog post frontmatter validation wasn't updated, so it's been failing to sync. I ran this fix so the standard site is working in production, and this should keep it working for future posts.

### DIFF
--- a/modules/standard-site-sync.ts
+++ b/modules/standard-site-sync.ts
@@ -4,7 +4,7 @@ import { glob } from 'node:fs/promises'
 import { join } from 'node:path'
 import { defineNuxtModule, useNuxt, createResolver } from 'nuxt/kit'
 import { safeParse } from 'valibot'
-import { BlogPostSchema, type BlogPostFrontmatter } from '#shared/schemas/blog'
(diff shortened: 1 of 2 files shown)


ghostdevv:
thanks to Francisco for noticing: https://bsky.app/profile/frabarz.cl/post/3mpquscaios24

modules/standard-site-sync.ts
```diff
@@ -205,7 +206,7 @@ const syncFile = async (
     ).toISOString()
   }
 
-  const result = safeParse(BlogPostSchema, normalizedFrontmatter)
+  const result = safeParse(RawBlogPostSchema, normalizedFrontmatter)
```
ghostdevv, line 209: (this is the only fix, the other changes are because I can't help myself but tidy up a tiny bit)

trueberryless (approved):
Looks good to me 🚀

Thanks for the quick fix!

PR #2996, opened 2026-07-04
fix: sync org page filters and columns with URL

### LINKED ISSUE #2818: bug: org page filters and selected columns state not reflected in URL
### Describe the bug

npmx intends for state to be reflected in the URL so that links can be shared and bookmarked. On the org page, only the sort option is reflected in the URL; selected filters and selected columns are not.

### Additional context

Example: https://npmx.dev/org/nuxt?sort=updated-desc. Try changing filters and/or columns and watch the URL.

### Logs

```shell-script

```

iamnitishpattar:
### 🔗 Linked issue

resolves #2818

### 🎯 Context

Currently, on the organization page, the selected filters (like download range, security, updated within) and the customized table columns are not synced to the URL. This prevents users from sharing links that preserve their specific view state.

This PR updates the `[org].vue` page to fully synchronize these states with the URL parameters.

### 📝 Description

- **State Initialization:** Added logic to parse `route.query.columns` on page load to initialize the visibility of the table columns if they are present in the URL.
- **Initial Filters:** Passed the `downloadRange`, `security`, and `updatedWithin` URL parameters into `useStructuredFilters` as `initialFilters` to ensure the filter chips are correctly restored on load.
- **URL Syncing:** Updated the debounced `updateUrl` function and the Vue `watch` array to track changes to these filters and the visible columns. 
- Clean URLs are maintained by only appending these parameters when their values differ from the defaults (e.g., ignoring `downloadRange: 'any'`).


ghostdevv (changes requested):
hey, thanks for the PR! apologies for the delay in the review

Do you think we could use the `usePermalink` composable like Alec did in https://github.com/npmx-dev/npmx.dev/pull/3051/changes#diff-85891e2ad98a9054da32712794223bc59f763c418d7cb0fa54b3e564d98ead3cR195-R197?

PR #2998, opened 2026-07-04
feat(ui): replace custom toggle with native input switch polyfill

tomayac:
## Summary
- Replaces the hand-rolled toggle styling in `Settings/Toggle.client.vue` with the native `<input type="checkbox" switch>` element.
- Adds `input-switch-polyfill` for browsers that don't yet support the native `switch` attribute, loaded only when needed, kept live-synced with the accent color and color-mode via a MutationObserver.

image

Split out of #2982 per review feedback to keep that PR focused on PWA/WCO/badging.

## Test plan
- [x] `pnpm test:unit` — 1673/1673 passing
- [x] `pnpm run test:types` — passing
- [x] `vp lint` — 0 errors
- [x] `knip` — clean


tomayac:
CC: @userquin per his request `:-)`

PR #3000, opened 2026-07-05
fix: check the file tree before the timeline reports removed types

### LINKED ISSUE #2791: Incorrect label in the timeline "Typescript types removed"
### Describe the bug

For the package [`@blocz/react-responsive`](https://npmx.dev/package/@blocz/react-responsive/v/5.0.0), I merged for the 4.0.1 release (that was later unpublished from NPM so depending on what you see in npmx it might be the 5.0.0) this PR: https://github.com/bloczjs/react-responsive/pull/144 that removes `types: <>; exports: { types: <> }` in the `package.json`

Image

Image

But when you check the files, it's because instead of having 1 declaration file, I now generate 1 per file (see `.d.cts` & `.d.mts`):

Image

I don't fully know if this is because the top-level `types` was removed, or the one in `exports: { types }`, but I think that according to TS, this new config is fine (TS auto looks for .d.ts alongside files). And even in npmx, there is the badge "✔️ Types included"

Image

So I'm reporting this just in case

### Additional context

_No response_

### Logs

```shell-script

```

thribhuvan003:
The timeline computed hasTypes from packument fields only, so packages that ship declaration files next to their entry points (without a types field) showed a false "TypeScript types removed" event.

Now when a version would produce that event, the endpoint fetches its file tree and re-checks it with the same detectTypesStatus the package page uses, capped at 5 lookups per request so a timeline page never fans out. Added unit tests for the implicit-declarations case, a genuine removal, and a failed file tree lookup.

fixes #2791


ghostdevv (changes requested):
There seems to be a util we can re-use `analyzePackage`, for example:

https://github.com/npmx-dev/npmx.dev/blob/32e324adedf19b618c32de689c1753ef36e24ecf/server/api/registry/analysis/%5B...pkg%5D.get.ts#L36-L42

(we don't need to pass `createPackage` here)

ghostdevv:
> ah nice, missed that util — switched to fetchPackageWithTypesAndFiles + analyzePackage, much cleaner. thanks!

This change doesn't seem to have gotten added, or maybe it got removed when you dealt with the conflicts? Could you also check out our [AI usage policy](https://github.com/npmx-dev/npmx.dev/blob/main/CONTRIBUTING.md#using-ai) 😅

thribhuvan003:
hey, it's there — lines 105-109, fetchPackageWithTypesAndFiles + analyzePackage like you suggested. no new commits have landed since jul 7 (fb2a3fe), so nothing got dropped — might just be the diff showing stale? happy to rebase onto main if that'd refresh it. and read through the AI Usage policy.

gameroman:
Closing this as this does not seem to follow our contributions guidelines

If anyone else would like to open a fix for this issue, please feel free to do that

PR #3009, opened 2026-07-05
refactor: use atcute oauth client

ghostdevv:
### 🧭 Context

Updating `@atproto/oauth-client-node` breaks stuff, so I thought instead of trying to figure that out we could switch fully to atcute libraries - this being the first step of that


server/utils/atproto/oauth-session-store.ts
```diff
@@ -17,16 +17,20 @@ export class OAuthSessionStore implements NodeSavedSessionStore {
     return `sessions:${did}`
   }
 
-  async get(key: string): Promise<NodeSavedSession | undefined> {
-    let session = await this.cache.get<NodeSavedSession>(this.createStorageKey(key))
+  async get(key: string): Promise<StoredSession | undefined> {
+    const session = await this.cache.get<StoredSession>(this.createStorageKey(key))
```
fatfingers23, line 21: Are the types the same here from @atproto to @atcute? Do we want to do some kind of translate layer or just have everyone login again?

fatfingers23 (commented):
Looks good to me! Only thing I had to flag was migration of OAuth session store

server/utils/atproto/oauth-session-store.ts
ghostdevv: I'm happy to make everyone login again xD the sessions are already pretty short, and it reduces the amount of complexity and code we have to remove later

PR #3018, opened 2026-07-07
fix(ui): consistent terminal box height across package managers

### LINKED ISSUE #3016: Package manager command box height inconsistent when last PM in list is selected
### Describe the bug

The install/execute command terminal boxes (Terminal/Install.vue, Terminal/Execute.vue) render all package manager variants at once and toggle visibility with CSS based on a data-pm attribute, rather than conditionally rendering only the selected one. The container uses space-y-1 for spacing between rows.

space-y-1's underlying CSS rule excludes the structurally-last DOM child from getting margin-bottom (:not(:last-child)). Since all variants are always in the DOM (just hidden via display:none), whichever package manager happens to be last in the packageManagers array loses that margin whenever it's the one actually selected/visible - making its command box render about 4px shorter than every other package manager's box.

Confirmed by inspecting the rendered box in devtools on /package/starlight-theme-mdbook (local build off main): every other PM shows the same box height, but whichever PM is last in the packageManagers array is consistently about 4px shorter. This isn't tied to a specific package manager - it'll affect whichever one ends up last in the array, so the exact PM affected may differ between main and what's live in production. Exact pixel values will vary by browser/viewport/zoom too, but the relative gap (~4px, matching space-y-1's spacing unit) is consistent.

### Additional context

Affects both app/components/Terminal/Install.vue and app/components/Terminal/Execute.vue.

kpab:
### 🔗 Linked issue

Fixes #3016

### 🧭 Context

Every package manager's command variant stays in the DOM and is shown/hidden with CSS. The container used `space-y-1`, which only adds spacing to items that aren't the last child. The last PM in the list (`nub`) is always the structurally-last child, so when you select it, it loses its bottom margin and its command box renders ~4px shorter than the others.

### 📚 Description

Swapped `space-y-1` for `flex flex-col gap-1` on the command containers in `Terminal/Install.vue` and `Terminal/Execute.vue`. `gap` only spaces items that are actually rendered, so the hidden variants no longer affect the height and every PM's box is the same size. Visible spacing is unchanged.

Checked with `pnpm lint`, `pnpm test:types` and `pnpm test` (all passing). No new test — the bug lives in the generated CSS layout, which Vitest doesn't render.

### 📸 Screenshots

Before: non-last package managers rendered slightly taller than the last package manager (`nub`).

| npm | nub |
| --- | --- |
| Before npm terminal height | Before nub terminal height |

After: `npm` and `nub` render with the same terminal box height.

| npm | nub |
| --- | --- |
| After npm terminal height | After nub terminal height |

### DIFF
--- a/app/components/Terminal/Execute.vue
+++ b/app/components/Terminal/Execute.vue
@@ -51,7 +51,8 @@ const copyExecuteCommand = () => copyExecute(getFullExecuteCommand())
         <span class="w-2.5 h-2.5 rounded-full bg-fg-subtle" />
         <span class="w-2.5 h-2.5 rounded-full bg-fg-subtle" />
       </div>
-      <div class="px-3 pt-2 pb-3 sm:px-4 sm:pt-3 sm:pb-4 space-y-1">
+      <!-- flex gap (not space-y) so hidden PM variants never affect spacing -->
+      <div class="px-3 pt-2 pb-3 sm:px-4 sm:pt-3 sm:pb-4 flex flex-col gap-1">
         <!-- Execute command - render all PM variants, CSS controls visibility -->
         <div
           v-for="pm in packageManagers"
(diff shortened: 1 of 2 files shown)


ghostdevv:
thank you for your PR! a fix for this was already merged unfortunately, but I'm sure there'll be something else you can work on if you like!

PR #3022, opened 2026-07-08
feat: add "create new account" button to PDS page

### LINKED ISSUE #3021: PDS page should have a "Create new account" button
### Describe the feature

The current [`/pds`](https://npmx.dev/pds) page only has a link to PDS MOOver. To create a new account (rather than migrating) a user needs to open the "Connect" menu, select "Connect to Atmosphere", then click the button in the modal. There should be a button on the actual PDS page, enabling people to create an account with no extra navigation.

### Additional information

### Final checks

ascorbic:
### 🔗 Linked issue

#3021 

### 🧭 Context

The current [/pds](https://npmx.dev/pds) page only has a link to PDS MOOver, with no way to directly create a new account. To create a new account (rather than migrating) a user needs to open the "Connect" menu, select "Connect to Atmosphere", then click the button in the modal. 

### 📚 Description

This PR adds a button alongside the PDS Moover one that calls `authRedirect` in the same way as the equivalent in the modal.


ghostdevv (approved):
ty! I switched the new translation string out for the one that we use on the "Connect to Atmosphere" modal as they're identical - we can split them again later if it makes sense 🙏

ThisIsMissEm:
Keep in mind that prompt of create isn't part of the OAuth spec yet for at protocol, so it'll probably only work with the reference implementation (you can discover this via the AS metadata)

ghostdevv:
> Keep in mind that prompt of create isn't part of the OAuth spec yet for at protocol, so it'll probably only work with the reference implementation (you can discover this via the AS metadata)

good to know! if we switch out the underlying PDS impl in the future we can always change the logic to match

PR #3037, opened 2026-07-13
fix: convert emoji in markdown text tokens

### LINKED ISSUE #2822: Skip emoji conversion without regex

Semak12345:
Closes #2822

Moves emoji conversion into the marked text renderer instead of scanning rendered HTML for code tags. Inline code, fenced code, and escaped shortcodes now bypass conversion naturally. Shortcode-based heading anchors remain unchanged.

Tested with:
- 109 targeted unit tests
- formatter and lint
- full workspace typecheck

### DIFF
--- a/server/utils/changelog/markdown.ts
+++ b/server/utils/changelog/markdown.ts
@@ -14,11 +14,11 @@ import {
   createMarkedHeadingExtension,
   renderToRawHtml,
   sanitizeRawHTML,
+  emojiText,
 } from '../mdKit'
 import { slugify } from '#shared/utils/html'
 import { Marked } from 'marked'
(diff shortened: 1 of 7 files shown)


gameroman (commented):
Please use the pull request template

gameroman (changes requested):
Please follow the PR template

PR #3051, opened 2026-07-15
feat: persist timeline chart metric in URL

### LINKED ISSUE #3049: Store timeline page state in the URL
### Describe the feature

Currently `https://npmx.dev/package-timeline/*` page stores the selected tab (`Install Size`, `Dependencies`, `Dependency Size`) state only in the current session

It should store it the URL instead

### Additional information

### Final checks

graphieros:
### 🔗 Linked issue

Resolves #3049

### 🧭 Context

Timeline page: the selected timeline metric should be persisted in the URL.

### 📚 Description

- Set the active tab from the permalink
- Some issues related to the fact that the stackbar chart (dependency size chart) was previously never loaded as the first chart, are fixed:

  - the chart internally has an autosize feature that adapts the chart area depending on the axis labels: on load, this auto-sizing  was visible
  - the chart showed its built-in skeleton loader, which was also visible until the chart computed the data set. The ideal solution in general is to set the skeleton to the same types of params as the final chart, but not in this case, since the params are unpredictable (package version names have various lengths, number of data points can vary a lot, making CLS unavoidable during the skeleton -> final render transition).
  
Both of these issues are fixed by setting the opacity of the chart to 0 while waiting for its internals to stabilize before bumping its opacity back to 1. It's not perfect, a proper solution might be required in the chart library itself, to signal when a chart is stable, but it seems to do the job for now.

The aspect ratio of the ClientOnly's fallback is now also adapted based on the metric, since both chart components have different aspect ratios.

## Other: 
Bump vue-data-ui to latest, with css transition improvements on scale labels
- [3.22.12](https://github.com/graphieros/vue-data-ui/releases/tag/v3.22.12)
- [3.22.13](https://github.com/graphieros/vue-data-ui/releases/tag/v3.22.13)

### DIFF
--- a/app/components/Package/TimelineChart.vue
+++ b/app/components/Package/TimelineChart.vue
@@ -49,7 +49,6 @@ const props = defineProps<{
   timelineEntries: TimelineVersion[]
   selectedVersion: string | null
   loading: boolean
-  permalink?: boolean
 }>()
 
 const { settings } = useSettings()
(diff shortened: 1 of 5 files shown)


shuuji3 (approved):
LGTM 👍 Great attention to detail.

PR #3055, opened 2026-07-17
fix: world Emoji Day logo styles & annotations

jviide:
### 🧭 Context

Interacting with the World Emoji Day logo is a difficult on mobile platforms, due to pointer movement getting interpreted as panning, and the emoji particles getting selected.

The logo is also missing accessibility roles and labeling.

### 📚 Description

This pull request sets `touch-action: pinch-zoom` for the emoji particle container element, which allows pointer movement that would otherwise get interpreted as panning to interact with the logo instead.

The emoji particles also get `user-select: none` and `pointer-events: none` styles.

The pointer event handlers are marked as passive, as they'll never call `event.preventDefault()`.

Additionally, this pull request adds `role="img"` and an `aria-label` to the logo's root element, and `aria-hidden="true"` to its immediate children.

### DIFF
--- a/app/components/Noodle/EmojiDay/Logo.vue
+++ b/app/components/Noodle/EmojiDay/Logo.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="relative flex w-full">
+  <div class="relative flex w-full" role="img" :aria-label="$t('alt_logo')">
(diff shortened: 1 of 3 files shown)


graphieros (approved):
🌿

MatteoGabriele (commented):
works really well on my phone now ❤️

PR #3061, opened 2026-07-19
refactor: replace semver with verkit

sxzz:
### 🔗 Linked issue

Related to https://github.com/e18e/ecosystem-issues/issues/277

### 🧭 Context

npmx is an ESM Nuxt application that uses semantic-version operations in both client and server code. `verkit` provides a pure ESM, zero-runtime-dependency implementation with first-party TypeScript declarations, so it fits the project while removing the CommonJS `semver` package and its separate type package.

### 📚 Description

- replace `semver` and `@types/semver` with `verkit@0.1.2`
- migrate version parsing, comparison, range, prerelease, difference, and satisfying-version calls to the verkit APIs
- preserve nullable parsing behavior with `tryParse`
- update Vite dependency optimization and regenerate the pnpm lockfile

There should be no user-facing behavior change to version sorting, filtering, dependency resolution, or package analysis.

### ✅ Validation

- `pnpm lint:fix`
- `pnpm test:types`
- 154 focused unit tests
- 95 focused Nuxt tests
- `pnpm build`


danielroe (approved):
❤️

PR #3068, opened 2026-07-22
fix: support shorthand git provider URLs

gameroman:
### 🔗 Linked issue

None

### 🧭 Context

Previously shorthand git provider URLs were not supported,
which was breaking https://main.npmx.dev/package/smol-toml/v/1.7.0 for example

### 📚 Description

Adds support for shorthand git provider URLs

Now https://npmx-7wjuoush4-npmx.vercel.app/package/smol-toml/v/1.7.0 works

### DIFF
--- a/shared/utils/git-providers.ts
+++ b/shared/utils/git-providers.ts
@@ -271,13 +271,34 @@ const providers: ProviderConfig[] = [
   },
 ]
 
+const SHORTHAND_PROVIDERS = {
+  github: 'github.com',
+  gitlab: 'gitlab.com',
+  bitbucket: 'bitbucket.org',
+  codeberg: 'codeberg.org',
+  gitee: 'gitee.com',
+  sourcehut: 'git.sr.ht',
+} satisfies Partial<Record<ProviderId, string>>
+
 /**
  * Normalize various git URL formats to a standard HTTPS URL.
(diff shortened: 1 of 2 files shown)


shared/utils/git-providers.ts
```diff
@@ -271,13 +271,34 @@ const providers: ProviderConfig[] = [
   },
 ]
 
+const SHORTHAND_PROVIDERS = {
+  github: 'github.com',
+  gitlab: 'gitlab.com',
+  bitbucket: 'bitbucket.org',
+  codeberg: 'codeberg.org',
+  gitee: 'gitee.com',
+  sourcehut: 'git.sr.ht',
```
trueberryless, line 280: Are there other providers we need to consider?
Like Tangled, Gitea Cloud, Forgejo?

trueberryless (commented):
Overall looks go to me, just left a small thought.

shared/utils/git-providers.ts
gameroman: I think we can add them 👍

gameroman: I'm not sure if there are any other

If we find an npm package that uses it, we can add later

trueberryless (approved):
Amazing, thanks for the quick update 🚀

PR #3079, opened 2026-07-25
feat: tools

alexdln:
### 🧭 Context

Added a new section - tools. We've become a service with big data, features and capabilities. I think the logical next step is to start adding development tools for all package-related issues. 
One of these is the ability to quickly analyze your project's dependencies - to understand their status, relevance, and alternatives. So I added the deps-stats page.

### 📚 Description

image

<details>
<summary>
Additional screenshot
</summary>

image

image

</details>


server/api/registry/direct-deps-health.post.ts
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * POST /api/registry/direct-deps-health
+ *
+ * Resolve declared dependency ranges and report which packages are vulnerable
+ * or deprecated at the resolved version (direct packages only — no tree walk).
+ */
+export default defineEventHandler(async (event): Promise<DirectDependencyHealthResult> => {
+  const body = v.parse(DirectDepsHealthBodySchema, await readBody(event))
+  return await analyzeDirectDependencyHealth(body.dependencies)
+})
```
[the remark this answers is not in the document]
alexdln: Share more about the source of the proposals

graphieros (commented):
This is really awesome :)
I wonder if we could somehow make the charts smaller so everything could fit without having to scroll. Maybe just reducing the charts height config could work, which would require to add a height prop (or a more general options object). Right now a chartHeight computed (which I just noticed is a bit legacy since we don't use TrendsChart.vue inside modals anymore), could be based on this height prop, defaulting to decent values for mobile & desktop. (?)

Also the tools page should be easier to find, it's so good^^

Other things:
- the uploaded package.json is not persisted when navigating (for example show more stats and back)

shared/utils/trends-chart.ts
```diff
@@ -437,15 +438,24 @@ export function buildTrendsChartConfig(
...
-            showOnlyAtModulo: options.dates.length > XAXIS_LABELS_MOD_THRESHOLD,
-            modulo: Math.max(1, Math.round(options.dates.length / XAXIS_LABELS_MOD_THRESHOLD)),
             values: options.dates,
             datetimeFormatter: {
               enable: true,
               locale: options.locale,
               useUTC: true,
               options: getTrendsDatetimeFormatterOptions(options.selectedGranularity),
             },
+            ...(options.compactXAxisLabels
```
graphieros: You can simplify this by setting `showOnlyFirstAndLast: options.compactXAxisLabels`, and leave `showOnlyAtModulo` as it is since `showOnlyFirstAndLast` takes precedence when true.

graphieros (approved):
LGTM 🌿
I left a tiny remark to simplify the configuration.

app/components/Package/TrendsChart.vue
```diff
@@ -54,10 +54,14 @@ const props = withDefaults(
...
     permalink?: boolean
     defaultRange?: 'auto' | '52-weeks'
+    hideControls?: boolean
+    compactXAxisLabels?: boolean
   }>(),
   {
     defaultRange: 'auto',
     permalink: false,
+    hideControls: false,
+    compactXAxisLabels: false,
```
graphieros: boolean props are false by default, so adding them here is not necessary

alexdln: autocomplete made me too lazy 😅
thank you :)

shared/utils/trends-chart.ts
```diff
@@ -445,17 +445,9 @@ export function buildTrendsChartConfig(
...
-                  showOnlyFirstAndLast: true,
-                }
-              : {
-                  showOnlyAtModulo: options.dates.length > XAXIS_LABELS_MOD_THRESHOLD,
-                  modulo: Math.max(
-                    1,
-                    Math.round(options.dates.length / XAXIS_LABELS_MOD_THRESHOLD),
-                  ),
-                }),
+            showOnlyFirstAndLast: true,
```
graphieros: I'm sorry I meant

`showOnlyFirstAndLast: options.compactXAxisLabels`,

because just true will make it so on all charts

alexdln: Oh goddesses, thank you so much for catching it, I accidentally just deleted the condition and left it as true 🫠

PR #123, opened 2026-01-26
feat: add support for markdown extension in readme files

idleberg:
Both GitHub and npm CLI support the `.markdown` extension for README files. It's the file-extension endorsed by John Gruber, the inventor of Markdown, but we all know that almost no one uses it. However, since it's a simple change, I thought I'd open a PR. I was unsure regarding the casing; not all file names support all variants. JSDelivr allows them as well, see https://www.jsdelivr.com/package/npm/json-stable-stringify?tab=files&version=1.0.1

PR #3083, opened 2026-07-25
fix(i18n): added missing RU translations

nikelborm:
### 🔗 Linked issue

> Please ensure there is an open issue and mention its number. For example, "resolves #123"

Does it have to be done even for such small changes?

### 🧭 Context

Some RU translations were missing

### 📚 Description

PR adds missing RU translations.

### DIFF
--- a/i18n/locales/ru-RU.json
+++ b/i18n/locales/ru-RU.json
@@ -1092,7 +1092,7 @@
     "file_path": "Путь к файлу",
     "binary_file": "Бинарный файл",
     "binary_rendering_warning": "Предпросмотр для типа «{contentType}» не поддерживается.",
-    "possibly_unnecessary": "Публикация этих файлов может быть бессмысленна"
+    "possibly_unnecessary": "Публикация этих файлов возможно ошибочна или бессмысленна"
   },
   "badges": {
     "provenance": {
--- a/i18n/locales/ru-RU.json
+++ b/i18n/locales/ru-RU.json
@@ -191,6 +191,7 @@
       "theme_changed": "Тема изменена на {theme}.",
       "accent_color_changed": "Акцентный цвет изменён на {color}.",
       "background_theme_changed": "Оттенок фона изменён на {theme}.",
+      "foreground_theme_changed": "Основной цвет изменён на {theme}.",
       "download_started": "Скачивание архива {package}.",
       "copied_to_clipboard": "Скопировано в буфер обмена.",
       "npm_disconnected": "npm CLI отключён.",
@@ -261,7 +262,8 @@
     "back_to_archive": "Назад ко всем нудлам",
     "missing": {
       "title": "Этот нудл так и не выбрался из тарелки.",
-      "body": "У нас нет нудла «{slug}» в меню. Возможно, он ещё готовится, или его никогда не было. В любом случае — назад в архив."
+      "body": "У нас нет нудла «{slug}» в меню. Возможно, он ещё готовится, или его никогда не было. В любом случае — назад в архив.",
+      "empty_bowl": "Тарелка пуста"
     }
   },
   "settings": {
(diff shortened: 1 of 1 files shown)


i18n/locales/ru-RU.json
```diff
@@ -191,6 +191,7 @@
       "theme_changed": "Тема изменена на {theme}.",
       "accent_color_changed": "Акцентный цвет изменён на {color}.",
       "background_theme_changed": "Оттенок фона изменён на {theme}.",
+      "foreground_theme_changed": "Основной цвет изменён на {theme}.",
```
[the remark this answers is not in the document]
nikelborm: I considered "Цвет текста": looked through the sources and observed that it wasn't applied to text exclusively, and for that reason, rejected it. I also considered "Цвет переднего плана", but that's too chunky, too bureaucratic, and nobody says it like that.

i18n/locales/ru-RU.json
```diff
@@ -1082,7 +1091,8 @@
     },
     "file_path": "Путь к файлу",
     "binary_file": "Бинарный файл",
-    "binary_rendering_warning": "Предпросмотр для типа «{contentType}» не поддерживается."
+    "binary_rendering_warning": "Предпросмотр для типа «{contentType}» не поддерживается.",
+    "possibly_unnecessary": "Публикация этих файлов может быть бессмысленна"
```
gameroman: I'd maybe call it more something like "ненужной", rather than "бессмысленна"

nikelborm: Is this blocking?

nikelborm: I chose "может быть бессмысленна" ("may not make sense") instead of "может быть ненужной" ("may not be needed") as something that I felt better reflected the semantic meaning that I interpreted in the UI.

nikelborm: For example:
People keep publishing their package.json files with an unfiltered `"scripts"` object that sometimes can be the majority part of the file. The drive to trim it would not be the fact that nobody has a need to see the original contents of that field, but the fact that it is actively useless (nobody would even be able to reasonably inspect and utilitize the "dev", "lint", "format", and "publish" scripts from the package that's in the downloaded "node_modules" already) and harmful to the registries' bandwith.

nikelborm: Yeah, I just thought more about that. Probably an even better wording would be the one suggesting and explaining that every single download causes bits of harm by growing NPM's cloud services bill and putting useless strain on end-users' storage, etc that accumulate over time and increase with the package's popularity. I'm not familiar with other languages to do that in all places, but I can do it in Russian and English

nikelborm: "A penny saved is a penny earned" kind of thing

gameroman: > Is this blocking?

No, just thinking there might be a better wording for it

This is the context it appears in, so maybe there is a better wording we could use here

image

nikelborm: Yeah, I saw this too, on a different package here:
https://npmxdev-git-fork-nikelborm-ru-i18n-missing-npmx.vercel.app/package-code/@evadev/playground-demo/v/1.0.1

gameroman: Another option could be "ошибочно"

maybe something like 

```suggestion
    "possibly_unnecessary": "Публикация этих файлов возможно ошибочна или бессмысленна"
```

nikelborm: > Probably an even better wording would be the one suggesting and explaining that every single download causes bits of harm by growing NPM's cloud services bill and putting useless strain on end-users' storage, etc that accumulate over time and increase with the package's popularity.

Maybe instead of just a title, it can be turned into a full-blown tooltip. So that we don't have to look for the most dense with meaning and the most precise wording, and just explain the whole value proposition and consequences of either choice in full

gameroman (approved):
Looks good I think

i18n/locales/ru-RU.json
gameroman: I was thinking about that, it could be a note on the package page itself

It would be something like "This package includes 49kB of files that may be unnecessary"

image

I think we should open a separate issue for that

gameroman: #3084

PR #3087, opened 2026-07-28
fix: use `arg_<index>` for anonymous doc parameters

KazariAI:
`@deno/doc` omits `name` for parameters destructured directly in parameter lists, causing package docs to render `undefined` as the parameter name. This adds positional fallback names and regression coverage.

The fallback uses `arg_<index>` because [TypeScript uses the same form for unlabeled tuple elements](https://github.com/microsoft/TypeScript/blob/v5.9.3/src/compiler/checker.ts#L38299), while the zero-based index keeps multiple anonymous parameters distinct.

---

*This pull request was created with assistance from a code agent.*


gameroman (approved):

ghostdevv (approved):

ghostdevv:
<table>
<tr>
 <td>Before:</td>
 <td>After:</td>
<tr>
 <td>image</td>
 <td>image</td>
</table>

PR #3088, opened 2026-07-28
chore: adopt `typescript-native-bridge`

KazariAI:
Adjusts the Nuxt app config augmentation and cached fetch wrapper to avoid two tsgo diagnostic differences exposed by the migration.

---

*This pull request was created with assistance from a code agent.*

### DIFF
--- a/modules/build-env.ts
+++ b/modules/build-env.ts
@@ -63,7 +63,7 @@ export default defineNuxtModule({
 })
 
 declare module '@nuxt/schema' {
-  interface AppConfig {
+  interface CustomAppConfig {
     env: BuildInfo['env']
     buildInfo: BuildInfo
   }
--- a/modules/runtime/server/cache.ts
+++ b/modules/runtime/server/cache.ts
@@ -897,7 +897,7 @@ export default defineNitroPlugin(nitroApp => {
   nitroApp.hooks.hook('request', event => {
     event.context.cachedFetch = async (url: string, options?: any) => {
       return {
-        data: await globalThis.$fetch(url, options),
+        data: await fetchWrapper(url, options),
         isStale: false,
         cachedAt: null,
       }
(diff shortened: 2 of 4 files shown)


ghostdevv (changes requested):
Could you give some more context around this issue?

KazariEX:
`typescript-native-bridge` makes it possible to integrate TS7 type checking into projects built on the TS6 API. This PR primarily aims to speed up CI and the agent development cycle. It also helps bring TNB into real-world use, where broader adoption can uncover more issues and generate valuable feedback.

modules/runtime/server/cache.ts
```diff
@@ -897,7 +897,7 @@ export default defineNitroPlugin(nitroApp => {
   nitroApp.hooks.hook('request', event => {
     event.context.cachedFetch = async (url: string, options?: any) => {
       return {
-        data: await globalThis.$fetch(url, options),
+        data: await fetchWrapper(url, options),
```
ghostdevv, line 900: this seems like it's a behavior change?

modules/build-env.ts
```diff
@@ -63,7 +63,7 @@ export default defineNuxtModule({
 })
 
 declare module '@nuxt/schema' {
-  interface AppConfig {
+  interface CustomAppConfig {
```
ghostdevv, line 66: How come?

ghostdevv (changes requested):

modules/runtime/server/cache.ts
KazariEX: `fetchWrapper` has already been assigned to `globalThis.$fetch` above. Using `globalThis.$fetch` here causes the TS2321 error: "Excessive stack depth comparing types ..."

modules/build-env.ts
KazariEX: The `AppConfig` interface merges in a literal object inferred from the runtime value, where `prNumber` is always inferred as `null`. Therefore, overriding the type of `buildInfo` directly on `AppConfig` causes the TS2430 error: "Interface 'AppConfig' incorrectly extends interface 'MergedAppConfig<...>'."

ghostdevv (approved):

PR #3089, opened 2026-07-28
fix(ui): prevent facet label wrapping to fix info icon alignment

### LINKED ISSUE #3063: Incorrect alignment of info icon on compage packages page
### Describe the bug

The small info badge renders right next to single-paragraph text, but has a lot of spacing when it's multi-paragraph. From my perspective it would look less jarring if it always renders the same (probably next to the text). Maybe it's even possible to get rid of the word-wrapping, it seems like there's enough space (or almost at least), and then the rows have similar height too. 

A screenshot from the npmx compare packages page showing a comparison between the express and hono packages

### Additional context

Seems like a first good issue to me.

coder-tejas:
### 🧭 Linked issue
resolves #3063

### 🧭 Context
On the compare packages page, the info icon next to each facet label (e.g. "GITHUB STARS", "GITHUB FORKS", "GITHUB ISSUES") only sat close to the text when the label fit on one line. Longer labels wrapped to two lines inside the fixed-width label column, and since the row used `items-center`, the icon centered itself between the two lines instead of next to the text — creating an inconsistent, jarring gap compared to single-line labels like "DEPRECATED?".

### 📚 Description
The wrapping was caused by `tracking-wider` (extra letter-spacing on the uppercase label) pushing labels like "GITHUB STARS" past the 140px label column width. Rather than trying to re-align the icon around variable wrapped text, I removed the wrapping at the source:

- Changed `tracking-wider` → `tracking-wide` to claw back a few px of width per label
- Added `whitespace-nowrap` so labels always render on a single line

Checked the longest label ("GITHUB ISSUES") against the label column width to confirm it doesn't overflow with the reduced tracking.

BEFORE:
image
AFTER:
image

### DIFF
--- a/app/components/Compare/FacetRow.vue
+++ b/app/components/Compare/FacetRow.vue
@@ -91,7 +91,9 @@ function isCellLoading(index: number): boolean {
     <div
       class="comparison-label relative bg-bg flex items-center gap-1.5 px-4 py-3 border-b border-border"
     >
-      <span class="text-xs text-fg-muted uppercase tracking-wider">{{ label }}</span>
+      <span class="text-xs text-fg-muted uppercase tracking-wide whitespace-nowrap">{{
+        label
+      }}</span>
       <TooltipApp v-if="description" :text="description" position="top">
         <span class="i-lucide:info w-3 h-3 text-fg-subtle cursor-help" aria-hidden="true" />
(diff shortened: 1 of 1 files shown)


graphieros (approved):
LGTM 🌿

PR #3091, opened 2026-07-28
fix: clear modal input and error state on close

iiio2:
Both the Local Connector and Atmosphere modals kept their form state after being dismissed. If a connection attempt failed, closing the modal and reopening it showed the previously entered token/handle still populated, along with the stale error message from the last attempt. Now, both modals now reset their form state when the dialog closes.

Before: 
when we write on input fieldn closes it and opens it again, we can still see the input and error message. 

before

After: 
Now we cannot see. 

after

### DIFF
--- a/app/components/Header/AuthModal.client.vue
+++ b/app/components/Header/AuthModal.client.vue
@@ -42,6 +42,12 @@ async function handleLogin() {
   }
 }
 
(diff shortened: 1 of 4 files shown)


ghostdevv (approved):

iiio2:
Thanks @ghostdevv . :blush:

ghostdevv:
np!

PR #3102, opened 2026-07-31
fix: detect implicit types in package timeline

### LINKED ISSUE #2791: Incorrect label in the timeline "Typescript types removed"

anilloutombam:
### 🔗 Linked issue

Fixes #2791 

### 🧭 Context

The timeline incorrectly showed removed TypeScript types when declarations still existed as `.d.mts` files beside `.mjs` entry points.

### 📚 Description

I updated the timeline to check the published package files when types appear to be removed. I also added a regression test for declarations stored beside the package entry points.

### DIFF
--- a/server/api/registry/timeline/[...pkg].get.ts
+++ b/server/api/registry/timeline/[...pkg].get.ts
@@ -1,5 +1,6 @@
+import { enrichTimelineVersionTypes } from '#server/utils/package-timeline'
 import { normalizeLicense } from '#shared/utils/npm'
-import { analyzePackage, hasBuiltInTypes } from '~~/shared/utils/package-analysis'
+import { hasBuiltInTypes } from '~~/shared/utils/package-analysis'
 
 const DEFAULT_LIMIT = 25
 
@@ -84,40 +85,7 @@ export default defineCachedEventHandler(
         .sort((a, b) => Date.parse(b.time) - Date.parse(a.time))
       const visibleVersions = allVersions.slice(offset, offset + limit)
 
-      // File-aware detection is limited to potential removal events: metadata-untyped
-      // versions with an older typed release. Checking every untyped version would
-      // require additional registry and file-tree requests.
-      const possibleTypeRemovals = visibleVersions
-        .filter(version => {
-          if (version.hasTypes) return false
-
-          const versionIndex = allVersions.indexOf(version)
-          return allVersions
-            .slice(versionIndex + 1)
-            .some(previousVersion => previousVersion.hasTypes)
-        })
-        .map(async version => {
-          try {
-            const { pkg, typesPackage, files } = await fetchPackageWithTypesAndFiles(
-              packageName,
-              version.version,
-            )
-
-            const analysis = analyzePackage(pkg, {
-              typesPackage,
-              files,
-            })
-
-            if (analysis.types.kind === 'included') {
-              version.hasTypes = true
-            }
-          } catch {
-            // Preserve the metadata-only result when the file list is unavailable.
-          }
-        })
-
-      await Promise.all(possibleTypeRemovals)
-
+      await enrichTimelineVersionTypes(packageName, allVersions, visibleVersions)
       return {
         versions: visibleVersions,
         total: allVersions.length,
--- a/server/api/registry/timeline/[...pkg].get.ts
+++ b/server/api/registry/timeline/[...pkg].get.ts
@@ -85,10 +85,10 @@ export default defineCachedEventHandler(
       const visibleVersions = allVersions.slice(offset, offset + limit)
 
       const possibleTypeRemovals = visibleVersions.filter(version => {
-        const versionIndex = allVersions.indexOf(version)
-        const previousVersion = allVersions[versionIndex + 1]
+        if (version.hasTypes) return false
 
-        return !version.hasTypes && previousVersion?.hasTypes
+        const versionIndex = allVersions.indexOf(version)
+        return allVersions.slice(versionIndex + 1).some(previousVersion => previousVersion.hasTypes)
       })
 
       await Promise.all(
--- a/server/api/registry/timeline/[...pkg].get.ts
+++ b/server/api/registry/timeline/[...pkg].get.ts
@@ -82,17 +82,15 @@ export default defineCachedEventHandler(
           }
         })
         .sort((a, b) => Date.parse(b.time) - Date.parse(a.time))
+
       const visibleVersions = allVersions.slice(offset, offset + limit)
 
-      const possibleTypeRemovals = visibleVersions.filter(version => {
-        if (version.hasTypes) return false
-
-        const versionIndex = allVersions.indexOf(version)
-        return allVersions.slice(versionIndex + 1).some(previousVersion => previousVersion.hasTypes)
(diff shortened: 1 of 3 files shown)


server/api/registry/timeline/[...pkg].get.ts
```diff
@@ -82,9 +82,39 @@ export default defineCachedEventHandler(
...
         })
         .sort((a, b) => Date.parse(b.time) - Date.parse(a.time))
+      const visibleVersions = allVersions.slice(offset, offset + limit)
+
+      const possibleTypeRemovals = visibleVersions.filter(version => {
+        const versionIndex = allVersions.indexOf(version)
+        const previousVersion = allVersions[versionIndex + 1]
+
+        return !version.hasTypes && previousVersion?.hasTypes
+      })
```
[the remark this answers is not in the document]
Comment (anilloutombam) [reply server/api/registry/timeline/[...pkg].get.ts]: Fixed in `bdeddc4`. The timeline now checks every untyped visible version that has an older typed version, rather than only the version directly after the metadata transition. I also extended the regression test with a third version and verified that both implicit-type versions are detected.

test/unit/server/api/registry/timeline/pkg.get.spec.ts
```diff
@@ -227,6 +229,41 @@ describe('timeline API', () => {
...
+        main: './dist/index.mjs',
+      },
+      files: new Set(['dist/index.mjs', 'dist/index.d.mts']),
+    })
+
+    const result = await handler(fakeEvent)
+
+    expect(fetchPackageWithTypesAndFilesMock).toHaveBeenCalledWith('my-pkg', '2.0.0')
+    expect(result.versions[0]!.hasTypes).toBe(true)
+    expect(result.versions[1]!.hasTypes).toBe(true)
```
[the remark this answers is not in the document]
anilloutombam: Fixed in `bdeddc4`. I added a length assertion and changed the indexed version access to optional chaining, so the test no longer uses non-null assertions.

server/api/registry/timeline/[...pkg].get.ts
```diff
@@ -83,8 +83,37 @@ export default defineCachedEventHandler(
         })
         .sort((a, b) => Date.parse(b.time) - Date.parse(a.time))
 
+      const visibleVersions = allVersions.slice(offset, offset + limit)
+
+      const possibleTypeRemovals = visibleVersions
+        .filter((version, index) => {
+          if (version.hasTypes) return false
+          return allVersions.slice(index + 1).some(previousVersion => previousVersion.hasTypes)
```
Comment (ghostdevv) [inline server/api/registry/timeline/[...pkg].get.ts]: Could you explain the thinking here a bit? I guess it's so that you don't need to fetch a bunch of data for every package to check the types? Do you think it may miss packages that add types in weird ways? I wonder if instead we should run the fetch logic in the main `allVersions` loop if `hasBuiltInTypes` fails 🤔 I guess that would be less efficient... maybe you could add a comment explaining the rationale?

ghostdevv (changes requested):

server/api/registry/timeline/[...pkg].get.ts
Comment (anilloutombam) [reply server/api/registry/timeline/[...pkg].get.ts]: Yes, that was the thinking. The file-aware check requires extra package and file-tree requests, so I limited it to metadata-untyped versions that have an older typed release. Those are the versions that could otherwise produce a false “types removed” event.

This could miss a package that has always shipped types implicitly, or adds them without using `types` or `typings`. Detecting those would require checking every metadata-untyped version, which is broader than this removal-event fix and would add more requests.

I added a comment explaining the tradeoff. I’m happy to broaden the check if you think the extra requests are acceptable.

server/api/registry/timeline/[...pkg].get.ts
```diff
@@ -82,9 +82,44 @@ export default defineCachedEventHandler(
...
+      const visibleVersions = allVersions.slice(offset, offset + limit)
+
+      // File-aware detection is limited to potential removal events: metadata-untyped
+      // versions with an older typed release. Checking every untyped version would
+      // require additional registry and file-tree requests.
+      const possibleTypeRemovals = visibleVersions
+        .filter(version => {
+          if (version.hasTypes) return false
+
+          const versionIndex = allVersions.indexOf(version)
```
Comment (ghostdevv) [inline server/api/registry/timeline/[...pkg].get.ts]: I changed this to use the `index` from filter earlier, any reason that was undone?

ghostdevv (changes requested):

server/api/registry/timeline/[...pkg].get.ts
Comment (ghostdevv) [reply server/api/registry/timeline/[...pkg].get.ts]: Let's see what others think

server/api/registry/timeline/[...pkg].get.ts
Comment (anilloutombam) [reply server/api/registry/timeline/[...pkg].get.ts]: Yeah, that was intentional. `visibleVersions.filter()` gives us a page-relative index, but we’re slicing `allVersions`. That breaks as soon as the offset is non-zero. Using `allVersions.indexOf(version)` gives us the version’s actual position in the full timeline.

server/api/registry/timeline/[...pkg].get.ts
```diff
@@ -82,9 +82,44 @@ export default defineCachedEventHandler(
...
+            })
+
+            if (analysis.types.kind === 'included') {
+              version.hasTypes = true
+            }
+          } catch {
+            // Preserve the metadata-only result when the file list is unavailable.
+          }
+        })
+
```
Comment (gameroman) [inline server/api/registry/timeline/[...pkg].get.ts]: Should we maybe extract this into a separate file where we have some utils?

Comment (anilloutombam) [reply server/api/registry/timeline/[...pkg].get.ts]: Yeah, that makes sense. I’ll extract this into a server-side timeline utility so the endpoint stays focused on fetching, pagination, and building the response.

Comment (anilloutombam) [reply server/api/registry/timeline/[...pkg].get.ts]: Moved the file-aware type detection into `server/utils/package-timeline.ts`. The endpoint now delegates the enrichment step to that utility, and the existing tests still pass.

gameroman (approved):
Looks like it works 👍 

<details><summary>Before</summary>
<p>

Image

</p>
</details> 


<details><summary>After</summary>
<p>

Image
</p>
</details>

anilloutombam:
> Looks like it works 👍
> 
> Before
> Image
> 
> After
> Image

Thanks for checking and confirming!

anilloutombam:
@ghostdevv just following up on the needs discussion label. Is there anything you’d like to align on or change here?

PR #3111, opened 2026-08-02
fix: correctly handle native package dependencies size and counts

### LINKED ISSUE #3110: Optional dependencies for native packages are counted incorrectly
### Describe the bug

Example: [rolldown](https://main.npmx.dev/package/rolldown)

Actual:

2 direct, 10 total

Image

Expected:

2 direct, 2 total

### Additional context

While it is correct to count optional dependencies as dependencies in general,
for native packages they should not be treated as an actual dependency

### Logs

```shell-script

```

jibin7jose:
### 🔗 Linked issue

Resolves #3110

### 🧭 Context

Optional dependencies for native packages (specifically WASM bindings, like `@rolldown/binding-wasm32-wasi`) were being incorrectly tracked as regular dependencies. This resulted in massively inflated install size calculations and incorrect total transitive dependency counts (e.g., rolldown reporting 10 total dependencies and +14 MB in size instead of 2 total dependencies).

This PR fixes the issue by ensuring native bindings are correctly categorized and excluded from standard dependency metrics.

### 📚 Description

*   **Dependency Resolver (`server/utils/dependency-resolver.ts`)**:
    *   Added detection logic for WASM bindings (by matching `-wasm32-`, `-wasm32`, or `wasm32-wasi`) to correctly categorize them as native packages.
    *   Ensured that the `isNative` flag correctly trickles down and is inherited by all transitive children of a native package.
*   **Install Size Calculator (`server/utils/install-size.ts`)**:
    *   Updated the install size calculation to correctly exclude the individual file sizes of all native optional bindings from the total install size, accurately reflecting the true footprint of the package.
*   **Dependency Analysis (`server/utils/dependency-analysis.ts`)**:
    *   Since the `isNative` flag now propagates correctly, transitive children of native packages are effectively filtered out of the `totalPackages` count.
*   **Testing**:
    *   Validated that `rolldown` now correctly reports 2 total dependencies and an install size of ~895 KB.
    *   Verified that the "Significant size increase" banner no longer incorrectly appears for packages relying on WASM optional bindings.
    *   Added unit tests to cover these scenarios.

### DIFF
--- a/server/utils/dependency-analysis.ts
+++ b/server/utils/dependency-analysis.ts
@@ -325,13 +325,28 @@ export const analyzeDependencyTree = defineCachedFunction(
     })
 
     // Aggregate total counts
-    const totalCounts = { total: 0, critical: 0, high: 0, moderate: 0, low: 0 }
-    for (const pkg of vulnerablePackages) {
-      totalCounts.total += pkg.counts.total
-      totalCounts.critical += pkg.counts.critical
-      totalCounts.high += pkg.counts.high
-      totalCounts.moderate += pkg.counts.moderate
-      totalCounts.low += pkg.counts.low
+    // Count how many native optional packages were included, to exclude them from the total
+    let nativeOptionalCount = 0
+    for (const pkg of resolved.values()) {
+      if (pkg.optional && pkg.isNative) {
+        nativeOptionalCount++
+      }
+    }
+
+    const result: VulnerabilityTreeResult = {
+      package: name,
+      version,
+      vulnerablePackages,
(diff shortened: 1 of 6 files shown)


gameroman (changes requested):
Please add tests and follow the PR template

jibin7jose:
> Please add tests and follow the PR template

ok

gameroman (changes requested):
Please follow the PR template

jibin7jose:
Updated the PR description to follow the template

PR #3120, opened 2026-08-03
fix: show fallback for oversized package files

### LINKED ISSUE #2732: Code browser never loads UI for large files, swallows errors
### Describe the bug

When trying to serve a large file the `/api/registry/file` endpoint returns a [413 response](https://npmx.dev/api/registry/file/@types/vscode/v/1.118.0/index.d.ts) (`PAYLOAD_TOO_LARGE`, see #2332). Due to a lack of explicit error handling the UI remains in a perpetual loading state and fails to render remaining elements like the "Open raw file" button.

Steps:
1. Go to https://npmx.dev/package-code/@types/vscode/v/1.118.0/package.json
2. From the file list select `index.d.ts` (~720KB)
3. Observe that no error is displayed in the UI and that the "Open raw file" button is never rendered

### Additional context

While #2332 is related it is arguably more focused on the backend side. I've filed this issue separately to track the error handling in the UI.

### Logs

```shell-script

```

anilloutombam:
### 🔗 Linked issue

Fixes #2732

### 🧭 Context

The package code browser remained in a loading state when the selected file exceeded the supported size limit. The page already had a fallback for oversized files, but it was never rendered because the skipped content request remained in the idle state and was still treated as loading.

The issue can be reproduced on the current production site:

https://npmx.dev/package-code/@types/vscode/v/1.118.0/index.d.ts

The updated behaviour can be verified on the PR preview:

https://npmx-h3s403v0g-npmx.vercel.app/package-code/@types/vscode/v/1.118.0/index.d.ts

### 📚 Description

I updated the loading-state logic so an oversized file no longer leaves the page waiting for a request that will not run. The existing large-file warning and “Open raw file” action are now displayed as intended.

I also added a regression test that covers the oversized-file path, verifies that the fallback is rendered, and confirms that the application does not request the file content.

### DIFF
--- a/app/pages/package-code/[[org]]/[packageName]/v/[version]/[...filePath].vue
+++ b/app/pages/package-code/[[org]]/[packageName]/v/[version]/[...filePath].vue
@@ -184,6 +184,8 @@ const isLoading = computed<boolean>(() => {
     return treeStatus.value !== 'success' && treeStatus.value !== 'error'
   }
 
+  if (isFileTooLarge.value) return false
+
   return !fileStatus.value || fileStatus.value === 'pending' || fileStatus.value === 'idle'
 })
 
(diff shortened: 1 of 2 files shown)


ghostdevv (approved):
ty! The only issue I see currently is that it doesn't seem to work when navigating to it - e.g. if I go from `README.md` to `index.d.ts` it will show me this

Image

And then when I refresh I correctly get

Image

We can fix this in a follow up since it also happens on production

PR #3144, opened 2026-08-05
fix(ui): make diff page layout responsive

BittuBarnwal7479:
### 🔗 Linked issue

N/A

### 📚 Description

- The compare page was full-screen while the code page uses a container layout, so both pages looked inconsistent.
- This updates the compare page to use the same responsive layout as the code page. No logic changes.
- Nothing was functionally broken before this, this is mainly a consistency and responsive layout improvement.

Before: 
image

After:
image

### DIFF
--- a/app/components/Diff/ViewerPanel.vue
+++ b/app/components/Diff/ViewerPanel.vue
@@ -93,7 +93,7 @@ function getCodeUrl(version: string): string {
   return `/package-code/${props.packageName}/v/${version}/${props.file.path}`
 }
 
-const { toggleCodeContainer } = useCodeContainer()
+const { codeContainerFull, toggleCodeContainer } = useCodeContainer()
 
 const { announce } = useCommandPalette()
 
@@ -365,6 +365,7 @@ useCommandPaletteContextCommands(
             class="px-3 max-xl:hidden"
             classicon="i-lucide:unfold-horizontal [.container-full_&]:i-lucide:fold-horizontal"
             :aria-label="$t('code.toggle_container')"
+            :aria-pressed="codeContainerFull"
             @click="toggleCodeContainer()"
           />
         </TooltipApp>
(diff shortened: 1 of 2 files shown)


ghostdevv (changes requested):
What about it broke?

BittuBarnwal7479:
> What about it broke?

The diff page was using a full-screen layout, which broke at narrow widths and caused the content to overflow. This fix makes the layout responsive and aligns it with the code page.
Nothing was functionally broken.

ghostdevv:
> > What about it broke?
> 
> The compare page was using a full-screen layout, which broke at narrow widths and caused the content to overflow. This fix makes the layout responsive and aligns it with the code page. Nothing was functionally broken.

Could you share a video of an overflow happening? I wasn't able to reproduce it earlier

BittuBarnwal7479:
Ah, just to clarify my previous comment: by “overflow” i didn't mean actual horizontal overflow. I meant that the empty-state message “Select a file from the sidebar to view its diff” was not properly centered at narrower widths, which made the page look a bit confusing with the black space. Sorry for the confusion.

BittuBarnwal7479:
Example:
npmx.dev/diff/solid-js/v/2.0.0-beta.31...1.9.14

image

ghostdevv:
Ah, I believe this is intentional. I don't see why you couldn't add something like we have on the code browser page though

image

BittuBarnwal7479:
> Ah, I believe this is intentional. I don't see why you couldn't add something like we have on the code browser page though
> 
> image

Yep, you’re right. I think we should add the same behavior on the Diff tab as well, `Toggle code container width`. Am I right?

ghostdevv:
> > Ah, I believe this is intentional. I don't see why you couldn't add something like we have on the code browser page though
> > image
> 
> Yep, you’re right. I think we should add the same behavior on the Diff tab as well, `Toggle code container width`. Am I right?

yea, that's what I said :sweat_smile: /lh

BittuBarnwal7479:
https://github.com/user-attachments/assets/9d1c7c3b-5a25-4a69-89fb-dbf0ad3656d8

PR #3147, opened 2026-08-05
ci: improve workflows

luc122c:
### 🔗 Linked issue

Related: #2971

### 🧭 Context

A few improvements to the GitHub Actions workflows, following the same pass I've been doing across other repos.

#### `ubuntu-slim` runner

Use `ubuntu-slim` for the lightweight jobs — `ci.yml` (`i18n`), `stale.yml`, `welcome-open.yml`, `mirror-tangled.yml`, and `deploy-canary.yml`. The [runner readme](https://github.com/actions/runner-images/blob/main/images/ubuntu-slim/ubuntu-slim-Readme.md) provides more info about the tools and the [GA blog post](https://github.blog/changelog/2026-01-22-1-vcpu-linux-runner-now-generally-available-in-github-actions/) gives some potential use cases.

The heavier jobs (full dependency install + local build) stay on `ubuntu-24.04-arm`: `lint` (the oxc-parser-based linter needs more than the 5 GB RAM on `ubuntu-slim`), `types`, `unit`, `test`, `browser`, `knip`, `autofix`, and `zizmor` (which runs as a Docker-container action and needs a Docker-capable runner). `a11y` stays on `ubuntu-latest` for the documented [Lighthouse reason](https://github.com/GoogleChrome/lighthouse/discussions/16834). `deploy-canary` and `mirror-tangled` only install a CLI / do git ops, so the build that matters for canary happens remotely on Vercel.

#### `parallel` keyword

Run the two independent Codecov uploads (test results + coverage) in parallel in the `unit` and `test` jobs, using the new native [`parallel` keyword](https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel/).

#### Restrict preview/comment workflows to the main repo

Gate `chromatic.yml` and `lunaria.yml` behind `if: github.repository == 'npmx-dev/npmx.dev'` so fork PRs don't consume Chromatic builds or trigger `pull_request_target` comments in the org repo. The other release/deploy/mirror/comment workflows already had this guard.


.github/workflows/mirror-tangled.yml
```diff
@@ -18,7 +18,7 @@ jobs:
   mirror:
     name: 🕸️ Mirror to Tangled
     if: ${{ github.repository == 'npmx-dev/npmx.dev' }}
-    runs-on: ubuntu-24.04-arm
+    runs-on: ubuntu-slim
```
ghostdevv, line 21: How come?

ghostdevv (changes requested):

.github/workflows/mirror-tangled.yml
luc122c: The runner only gets 1vcpu and 5GB RAM which should be enough to clone the repo. It also has less tools but [still includes Git and OpenSSL tools](https://github.com/actions/runner-images/blob/main/images/ubuntu-slim/ubuntu-slim-Readme.md). It also runs in a container instead of a full VM so it's faster to provision. 

The only reason I see not to use it would be if the resources aren't enough and it's taking a long time to push.

Screenshot 2026-08-06 at 7 29 09 pm
https://docs.github.com/en/actions/reference/runners/github-hosted-runners

ghostdevv: The reason we used arm seems to be given here: https://github.com/npmx-dev/npmx.dev/pull/829#issue-3890580726 - not sure how that factors in

luc122c: Yeah, if anything it should be faster than the arm runners as they still run in a full VM. The only reason I could see them being slower is CPU/RAM bottleneck which is why I've only swapped small jobs, not the ones with full installs.

serhalp (approved):
Makes sense to me, thank you!

ghostdevv:
Changes sound good to me, just wanna try and follow up on why the changes to the runners were originally made - will get back to you soon hopefully :pray:

.github/workflows/lunaria.yml
trueberryless: Let's update this to a working version (Lunaria got some updates recently, so the other hash is not valid anymore): 

```suggestion
        uses: lunariajs/action@e5ab09c3c8353fa80786d8c2149873ba3978a3a7 # v0.2.0
```

.github/workflows/lunaria.yml
```diff
@@ -36,4 +37,4 @@ jobs:
           sfw: true
 
       - name: Generate Lunaria Overview
-        uses: lunariajs/action@4911ad0736d1e3b20af4cb70f5079aea2327ed8e # astro-docs
+        uses: lunariajs/action@e5ab09c3c8353fa80786d8c2149873ba3978a3a7 # v0.2.0
```
trueberryless, line 40: Not sure why Zizmor is not able to map the SHA to the tag (maybe it's an annotated tag 🤔)...

```suggestion
        # zizmor: ignore[unpinned-uses]
        uses: lunariajs/action@e5ab09c3c8353fa80786d8c2149873ba3978a3a7 # v0.2.0
```

trueberryless: Let me guess.... The comment is not in the range of the audit report 🙃 
[That's why it's still failing](https://github.com/npmx-dev/npmx.dev/actions/runs/35469006282/job/105966435498?pr=3147)

trueberryless (approved):
Thanks for the PR, @luc122c 

Everything looks good to me, I just adapted the Zizmor ignore rule to point to the updated line. I'll wait for @ghostdevv to approve and merge 🚀

ghostdevv (approved):

Issue #3140, opened 2026-08-05
Add `pkg.pr.new` support


gameroman:
### Describe the feature

For example if there's a `pkg.pr.new` release like `https://pkg.pr.new/knip@1853`, npmx would allow seeing it's information via `https://npmx.dev/pkg.pr.new/knip@1853`

### Additional information

- [ ] Would you be willing to help implement this feature?

### Final checks

- [x] Read the [contribution guide](https://github.com/npmx-dev/npmx.dev/blob/main/CONTRIBUTING.md).
- [x] Check existing [issues](https://github.com/npmx-dev/npmx.dev/issues).

PR #3157, opened 2026-08-07
feat(ui): add pkg.pr.new preview releases link

AmirSa12:
### 🔗 Linked issue

related to #3140 

### 🧭 Context

https://discord.com/channels/1464542801676206113/1508756434937188372

### 📚 Description

I added the necessary headers to the HEAD response in pkg.pr.new so we can detect whether a package has preview releases. if a package turns to have them, we render that "preview releases" button at the bottom of the "versions" section, as shown in the below screenshot.

cleanshot

cc @danielroe


gameroman (commented):
I was thinking more of adding a new route similar to `/pacakage/...` which would allow to view all the info about a published preview package on npmx

AmirSa12:
> I was thinking more of adding a new route similar to `/pacakage/...` which would allow to view all the info about a published preview package on npmx

That also sounds great! but apparently its been discussed before with Daniel (https://discord.com/channels/1464542801676206113/1508756434937188372/1518531069190738040)
> I think we could have a 'preview releases' or something like that, which displays at the bottom of the 'Versions' tab

IDK if that has changed or not.

gameroman:
I'd say these are 2 related features but different and could be both implemented

One for just detecting and linking to `pkg.pr.new`, which is what the Discord discussion was bout

And later we could add an actual route on npmx to view them, where this would later link instead of to `pkg.pr.new` directly, which is what my issue is about

ghostdevv:
How can I test this? I tried going to the page for knip but I can't see something like in the screenshot

AmirSa12:
> How can I test this? I tried going to the page for knip but I can't see something like in the screenshot

Hey @ghostdevv, thanks for reviewing.
Can you try again now? it was a caching matter.

gameroman:
> > How can I test this? I tried going to the page for knip but I can't see something like in the screenshot
> 
> Hey @ghostdevv, thanks for reviewing. Can you try again now? it was a caching matter.

Thank you! I see it now

image

image

ghostdevv:
One idea that came to me re: design is that it could be like the playgrounds UI

image

PR #3169, opened 2026-08-11
fix: prevent modal close button from crowding header text

iiio2:
Fixed the close button in the modal header sitting flush against the title/subtitle text.

The header is a flex row with the text block and close button as siblings, but had no gap between them and no width constraint on the text. On narrow viewports the subtitle wraps and expands until it butts right into the button.

before:
before-1

after: 

correct


ghostdevv (approved):

iiio2:
Thanks.

PR #3170, opened 2026-08-12
fix: only show changelog skeleton when actually loading

### LINKED ISSUE #3168: Package Changelog skeleton sometimes not hiding
### Describe the bug

https://npmx.dev/package-changelog/fcy it happens on this package but doesn't on https://npmx.dev/package-changelog/in-container

### Additional context

_No response_

### Logs

```shell-script

```

jibin7jose:
### 🔗 Linked issue

Resolves #3168

### 🧭 Context

Previously, the `ChangelogSkeleton` was conditionally rendered using `v-if="!changelog && !changelogError"`. This caused the skeleton to remain visible permanently for packages that successfully returned from the API but simply didn't have any changelogs (like the `fcy` package).

This PR resolves the issue by extracting the `pending` state from `usePackageChangelog` and using it to conditionally render both the skeleton and the fallback error messages.

### 📚 Description

- Extracted `pending: changelogPending` state from `usePackageChangelog`.
- Updated the skeleton loader to only display while the fetch is still pending (`v-if="changelogPending && !changelog && !changelogError"`).
- Wrapped the fallback/error messages in a `<template v-else-if="!changelogPending">` so they don't render simultaneously while the skeleton is loading.

**Testing:**
Tested against packages with no changelogs (e.g., `fcy`) to confirm the skeleton gracefully hides and properly shows the "Sorry, this package does not publish changelogs..." text without overlap.

### DIFF
--- a/app/pages/package-changelog/[[org]]/[name].vue
+++ b/app/pages/package-changelog/[[org]]/[name].vue
@@ -45,7 +45,11 @@ const latestVersion = computed(() => {
 })
 
 // getting info
-const { data: changelog, error: changelogError } = usePackageChangelog(packageName, version)
+const {
+  data: changelog,
+  error: changelogError,
+  pending: changelogPending,
+} = usePackageChangelog(packageName, version)
 
 const providerIcon = useProviderIcon(() => changelog.value?.provider)
 const viewOnProvider = useViewOnGitProvider(() => changelog.value?.provider)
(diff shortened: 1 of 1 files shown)


gameroman (dismissed):
Please follow the PR template

jibin7jose:
Updated the PR description to follow the template

WilcoSp (approved):
good fix, I've tested this with fcy, Vue and Nuxt and so far I can test it doesn't cause any unexpected changes in ux & behaviour

ghostdevv (approved):

PR #3179, opened 2026-08-14
chore(deps): renovate/nuxt additions

mootari:
### 🔗 Linked issue

PR against #3094

### 📚 Description

Changes:
- Moves a nuxt patch intended for Storybook into a storybook plugin, allowing e2e tests to succeed again
- Disables an e2e test that cannot pass because the expected meta tag doesn't currently get added in test environments


mootari:
Looks like the removed nuxt patch was meant to make Storybook happy. This will require a little more work... 🤔

mootari:
Storybook in dev on main is already pretty fragile. Depending on how you navigate between components you may end up with missing locale configs or "illegal invocation" errors.

I did see a few seemingly new CSP errors in this branch but haven't been able to repro them reliably. Likely less of an issue with SSR anyway.

danielroe (approved):

PR #3188, opened 2026-08-17
feat: add "Security Sources" concept; add Socket in addition to OSV

ljharb:
### 🔗 Linked issue

No issue yet; I discussed this with @danielroe a few weeks ago on a call.

### 🧭 Context / 📚 Description

This PR (which I'm happy to split into multiple PRs - the commits are already designed for that) adds the concept of "security sources" to pair with the existing "data sources" concept, and makes OSV (the only current security source) into one. It then adds [Socket](https://socket.dev) as an additional security source. (Disclosure: Socket is my employer).

It's designed to follow the npmx philosophy of giving users control and choices by allowing them to see maximal security information, and if desired, hide any security source they wish. Additional security sources can be easily added in the future - Socket competitors won't have any less opportunity to display their data.

Additional disclosure: I used Claude Code pretty heavily on this PR, with a *lot* of guidance, adversarial LLM review, and also lots of manual human review throughout the process.

This is my first contribution to npmx, so please be gentle, and I'm more than happy to make any requested changes (including submitting this as multiple PRs, if commit-by-commit is too much for anyone).


app/composables/useSettings.ts
```diff
@@ -231,6 +249,55 @@ export function useSearchProvider() {
...
+    socket: !!runtimeConfig.public.socketConfigured,
+  }))
+
+  const effectiveSources = computed<Record<SecuritySourceId, boolean>>(() => {
+    const result = {} as Record<SecuritySourceId, boolean>
+    for (const source of SECURITY_SOURCE_IDS) {
+      result[source] = enabledSources.value[source] && sourceAvailability.value[source]
+    }
+    return result
+  })
```
ljharb, line 289: Good catch on the latent hole. `effectiveSources` iterates `SECURITY_SOURCE_IDS` but reads from the persisted object, and since `mergeDefaults` only shallow-merges, a source id added later would be `undefined` for existing visitors and leak into a boolean-typed slot. No current bug (osv/socket are both defaulted), but I've added the nullish fallback to `DEFAULT_SECURITY_SOURCES` so a newly introduced source defers to its default rather than `undefined`.

playwright.config.ts
```diff
@@ -19,6 +19,16 @@ export default defineConfig<ConfigOptions>({
...
+    env: {
+      ...process.env,
+      // Dummy Socket credentials so the Socket security source is active in
+      // e2e; actual API calls are intercepted by the fixture plugin
+      // (modules/runtime/server/cache.ts). The public flag must be set
+      // explicitly because the build runs without the credentials.
+      NUXT_SOCKET_API_KEY: 'npmx-test-fixture-key',
+      NUXT_SOCKET_ORG_SLUG: 'npmx-test-fixtures',
+      NUXT_PUBLIC_SOCKET_CONFIGURED: 'true',
+    },
```
ljharb, line 31: Good catch that the Socket source is server-configured (`getSocketConfig` gates on `NUXT_SOCKET_API_KEY`/`ORG_SLUG`), so the merged-sources fixture test does depend on those vars. In practice the reuse risk is narrow: the webServer command is `TEST=1 vp run preview --port 5678`, and `reuseExistingServer` only reuses a server already bound to :5678. A normal `nuxt dev` server is a different port and doesn't load the fixture plugin (`TEST` unset), so it's never reused. The only way to reuse a misconfigured server is to manually start `start:playwright:webserver` yourself without the dummy Socket env, which surfaces as a single obvious failing assertion rather than a silent bug. Forcing `reuseExistingServer: false` would slow every local run, break when :5678 is already held, and diverge from the `!process.env.CI` convention we use elsewhere, so I'm leaving it as-is.

server/utils/dependency-analysis.ts
```diff
@@ -346,16 +323,28 @@ export const analyzeDependencyTree = defineCachedFunction(
...
+    // Results degraded by a transient source failure (outage, quota
+    // exhaustion) are only served from cache briefly, so a blip doesn't
+    // strip findings from the cache for a whole hour after recovery
+    validate: entry => {
+      // a custom validate replaces nitro's default entry.value !== undefined guard
+      const result = entry.value
+      if (!result) return false
+      if (!hasTransientSourceFailure(result.sourceStatus)) return true
+      return Date.now() - (entry.mtime ?? 0) < CACHE_MAX_AGE_FIVE_MINUTES * 1000
+    },
```
ljharb, line 348: Good catch that this is blocking rather than background — confirmed against nitropack 2.13.4: the SWR early-return is gated on `validate(entry) !== false`, so a degraded entry past the 5-min window does await a recompute on the request path rather than serving stale in the background. Two things make it much less severe than "every request past 5 min," though: `pending[key]` dedups concurrent recomputes, and each recompute resets `mtime` and re-stores, so the blocking recurs only about once per 5 min per key under steady traffic. On the suggested remedies: `validate` is a pure function of the stored entry (value + mtime only) evaluated *before* recompute, so "keep valid while a recompute is in flight" isn't expressible in Nitro's model. And "re-serve until a retry interval elapses" is in fact what the code already does — `mtime` is the last-attempt timestamp and the 5-min check is the retry floor; the only residual is Nitro's blocking boundary refresh, which `validate` can't change. The only way to get non-blocking early refresh would be to drop `maxAge` to ~5 min globally, multiplying OSV/Socket load for healthy entries. Deferring as an accepted, bounded tradeoff.

server/utils/socket.ts
```diff
@@ -0,0 +1,350 @@
...
+  }
+
+  const artifacts: SocketArtifact[] = []
+  for (const line of text.split('\n')) {
+    const trimmed = line.trim()
+    if (!trimmed) continue
+    artifacts.push(JSON.parse(trimmed))
+  }
+  return artifacts
+}
```
ljharb, line 268: Good catch — `parseArtifacts` runs inside `querySocketForTree`'s try, so one malformed/truncated NDJSON line currently throws and drops the entire chunk (up to 1024 packages) even though the other lines parsed fine. I've wrapped both the per-line NDJSON parse and the array-branch parse in try/catch: malformed lines are skipped and the successfully parsed artifacts are returned. Added a regression test asserting a valid artifact followed by a malformed line keeps the valid one.

server/utils/socket.ts
```diff
@@ -0,0 +1,350 @@
...
+          headers: {
+            authorization: `Bearer ${config.apiKey}`,
+            accept: 'application/x-ndjson',
+          },
+          body: {
+            components: chunk.map(pkg => ({ purl: toNpmPurl(pkg.name, pkg.version) })),
+          },
+          responseType: 'text',
+        },
+      )
```
ljharb, line 430: Good catch, and it matches the existing convention in the repo (`jsr.ts`, `docs/client.ts`, `npm-homepage.ts` all set timeouts on third-party fetches). Added a 10s timeout via a shared constant to all three OSV `$fetch` calls (batch/details in the OSV-client commit, count in the could-not-check commit) and to the Socket purl request. All four already catch/degrade gracefully on failure, so the timeout just converts a hang into the existing failed-source path.

ljharb:
(claude's review feedback responses)

ghostdevv (changes requested):
thanks for the PR! The abstraction of sources sounds good, it's similar to a mockup someone did in the Discord recently for data like this (security, module replacements, runtime compat data, etc)

> Additional disclosure: I used Claude Code pretty heavily on this PR, with a lot of guidance, adversarial LLM review, and also lots of manual human review throughout the process.

no worries, thanks for letting us know! LLMs are allowed, we have a simple policy here: https://github.com/npmx-dev/npmx.dev/blob/bf230d7142e14709468aa5ed4ccf26c1900f25bf/CONTRIBUTING.md#using-ai

> (including submitting this as multiple PRs, if commit-by-commit is too much for anyone).

The PR would likely be merged faster split up, purely as then aspects that involve design or discussion with the team can be talked about separately and won't block it as a whole. That being said, if it doesn't make sense to split up or merge different parts at different times then it can remain as one no problem

ljharb:
I'm more interested in correctness than speed, so whatever results in the best outcome is my preference :-)

test/nuxt/composables/use-security-sources-available.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+    previousSettings = localStorage.getItem('npmx-settings')
+    localStorage.removeItem('npmx-settings')
+  })
+
+  afterEach(() => {
+    useRuntimeConfig().public.socketConfigured = previous
+    if (previousSettings === null) {
+      localStorage.removeItem('npmx-settings')
+    } else {
+      localStorage.setItem('npmx-settings', previousSettings)
```
ljharb, line 26: Fixed. `beforeEach` now captures `previousSettings = localStorage.getItem('npmx-settings')` and `afterEach` restores it (removing the key when there was none), so this test can no longer clobber the persisted settings later tests rely on in the shared browser context.

43081j:
we're approaching this backwards to normal: a PR before a discussion.

to help us understand the feature better, can you:

- briefly explain what you're trying to solve here
- show some screenshots or preview URL links to packages which show the various notices
- briefly explain the configurable part of this (the "security sources" and how a user can configure them or why. we should avoid making user options too noisy so we should discuss this part)

as with all agentic PRs, it is oversized. but we can solve that by splitting it up into incremental changes once there's better understanding of the intent and UX here

danielroe:
trying this out with a token, a couple of searches and we are out of usage:

image

it also doesn't display any data:

SCR-20260818-nbem

are there additional scopes I should include? I've only enabled `alerts:list` and `alerts:trend`

ljharb:
@43081j daniel and i already had the discussion :-) the goal is to add additional security data sources, and to surface socket's supply chain alerts on package pages. The configuration options are just checkboxes - OSV on or off, and socket on or off.

ljharb:
@danielroe the docs in the PR specify the `packages:list` scope; can you try that?

re the quote, that's unfortunately expected - the endpoint bills a flat 100 units per request (regardless of package count, up to 1024), and a default org token is 500/hour, so it's only ~5 package scans an hour. Each package page scans the whole transitive tree in one request. To keep it bounded npmx already caches results for an hour per package@version (re-opening a package is free) and trips a 10-minute circuit breaker on a 429/403, degrading to OSV-only instead of hammering the API - so "could not scan" is that back-off once the scope/quota check fails. For the public deployment we'd point it at a token provisioned with real-traffic headroom.

danielroe:
yes, that scope did the trick! here's a screen recording to help anyone else looking at the PR:

https://github.com/user-attachments/assets/06ed3e52-444e-4ead-955d-3b618be8cae6

ljharb:
I'm making some changes to improve the caching/quota story here, and will explore socket-side changes as well.

server/utils/socket.ts
```diff
@@ -0,0 +1,457 @@
...
+  const { hits, misses } = await readPurlCache(packages)
+  const scan = emptyScan('ok')
+  for (const [key, findings] of hits) applyFindings(scan, key, findings)
+
+  if (misses.length === 0) return scan
+  if (Date.now() < socketUnavailableUntil) {
+    // cached findings still stand, but the misses can't be refreshed right now
+    scan.status = hits.size > 0 ? 'partial' : 'unavailable'
+    return scan
+  }
```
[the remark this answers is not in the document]
ljharb: Good catch, and worse than it looked: querySocketForTree runs inside an unguarded Promise.all with the OSV scan, so a cache storage error would've rejected the whole dependency analysis, taking OSV down too. Fixed by guarding both cache calls inside the helpers: a read failure now degrades to treating every purl as a miss (Socket is queried directly), and a write failure is logged and swallowed so already-normalized findings aren't dropped - both via console.warn, matching the other Socket failure paths. Added regression tests that reject getItems/setItems and assert the scan still returns a status with its findings.

43081j:
> @43081j daniel and i already had the discussion :-) 

I saw. Now we can have it here with the team 👍

Just context and examples goes a long way. Daniel has done it for you now but worth noting for future reference. 

Generally the change seems sane. Would be good to split it up into a less monolithic change, though. 

How we show these notices should be pretty abstract and unrelated to where it came from, as that is just an implementation detail. But being able to toggle sources seems fair enough

ljharb:
Totally happy to split up the PR - there's 5 commits and each one can land independently, although in a practical sense, aff3d5accfd19cefdd07ec9cd6a08d3f6da42657 and 60af5e60873ec52253ba4a8d7427340da5af09a9 can def land on their own, but 910f5c9a9ad4159ffec789171e44bc35235691f1 is a bit confusing without a second source, which 860fc76f2edc1edc63828fddb61b2c8a23ec3700 provides. d9da7f54b825584e76ce4dbf5904bf45877781c7 can also land on its own, after everything else. So what I really see is 5 commits across 4 PRs - I can remove the last 4 commits from here, and put up a second PR - after those two have landed, I can put up a third (with 2 commits), and after that, a fourth with the 5th commit.

Give me explicit direction one way or the other, and I'll make it happen!

ljharb:
Any updates on next steps?

PR #3192, opened 2026-08-20
feat(ui): support sorting with semver in timeline

antfu:
**This PR is implemented with the help of agents**.

### 🧭 Context

For packages like [`nuxt`](https://npmx.dev/package-timeline/nuxt/v/4.5.2), we maintain multiple versions at the same time, so the current publish-time-based sorting in the timeline is not really much referencable as:

CleanShot 2026-08-20 at 10 31 32@2x

We care more about the change from `4.5.1 -> 4.5.2` instead of `3.21.10 -> 4.5.2`

This PR is trying to improve that by adding the "semver" sorting algorithm, which would make it render like: 

CleanShot 2026-08-20 at 10 35 55@2x

### 📚 Description

- I keep the default sorting as "publish-time" to keep the original behaviour (tho I would actually recommend having "semver" as the default)
- The sorting would happen on the server side
- Selection would be persisted in query string `?sort=semver`
- As a side-effect, the "Stable only" option is moved from localStorage to query string `?stable-only=true` as well.
- Add a smart detection: if the selected version is an unstable version, the "Stable only" option would be off by default.


graphieros (commented):
This is awesome!

The UI is not updated in these cases:
- loading the page with the default sort by publishing time and switching to semver from the input.
- loading the page with `sort=semver`, switching to publishing time from the input works well, then switching back to semver from the input does not update the UI

(Tested with nuxt)

graphieros (commented):
The same issues as described in my previous comment remain after the latest changes.

antfu:
Figured it out, it was because of the previous data `timelineEntries` not get cleared on sort changes. Should be good now. 

(with this, maybe we should put "stable-only" also on server side?)

graphieros:
> Figured it out, it was because of the previous data `timelineEntries` not get cleared on sort changes. Should be good now.
> 
> (with this, maybe we should put "stable-only" also on server side?)

Stable-only on the servers-side is great ^^

I'm still seeing the same issues I described unfortunately... (tested on the preview, with nuxt)

Also, when more data is loaded through the _Load more_ button, toggling the sort resets the list, and we lose the additional entries. This is probably unrelated to the present PR, as it was already a problem when reloading the page, but now that additional loaded data can be lost when changing the sort, perhaps it should be persisted ?

nuxt.config.ts
```diff
@@ -126,7 +126,7 @@ export default defineNuxtConfig({
       isr: {
         expiration: 300,
         passQuery: true,
-        allowQuery: ['offset', 'limit'],
+        allowQuery: ['offset', 'limit', 'sort', 'stable-only'],
```
antfu, line 129: Now the mystery is finally resolved - no wonder I don't have this locally. 🤣

antfu:
I think it's ready now

graphieros (approved):
Yes, the switch works fine now :)

gameroman (commented):
Should we make `semver` the default for "Sort By"?

graphieros:
> Should we make `semver` the default for "Sort By"?

I think it would be sound to do so.

antfu:
I would love to have it by default (was thinking to do so in the following PR), but if we think that would be the better default I will update this pr directly

gameroman (approved):

Issue #1411, opened 2026-02-12
bug: wrong `unknown` types in package api docs


shi-gg:
I found several issues with the api docs of packages showing `unknown`, even though, the types aren't actually `unknown`.

<img width="2041" height="781" alt="Image" src="https://github.com/user-attachments/assets/902dcbe4-da6f-4203-93ee-df1a60330735" />

I found the following cases that show the wrong types:
- [event listener](https://npmx.dev/package-docs/linkdave/v/0.0.2#interface-LinkDaveClient) - [code](https://npmx.dev/package-code/linkdave/v/0.0.2/dist%2Fclient.d.ts#L12)
- [interface with enum keys](https://npmx.dev/package-docs/linkdave/v/0.0.2#interface-ManagerEvents) - [code](https://npmx.dev/package-code/linkdave/v/0.0.2/dist%2Ftypes.d.ts#L237)
- [type alias with many objects](https://npmx.dev/package-docs/linkdave/v/0.0.2#typeAlias-ClientMessage) - [code](https://npmx.dev/package-code/linkdave/v/0.0.2/dist%2Ftypes.d.ts#L74)
- [arrow function type alias](https://npmx.dev/package-docs/linkdave/v/0.0.2#typeAlias-SendToShardFn) - [code](https://npmx.dev/package-code/linkdave/v/0.0.2/dist%2Fclient.d.ts#L6)

also, [`Pick<>` is broken](https://npmx.dev/package-docs/linkdave/v/0.0.2#typeAlias-RawVoiceServerUpdate) - [code](https://npmx.dev/package-code/linkdave/v/0.0.2/dist%2Fplayer.d.ts#L15)

PR #3200, opened 2026-08-23
fix(docs): format complex tsType kinds instead of rendering unknown

### LINKED ISSUE #3154: docs for types give unknown, unknown
### Describe the bug

https://npmx.dev/package-docs/trslate/v/1.6.4

even if the index.d.ts correctly has them

### Additional context

_No response_

### Logs

```shell-script

```

aniruddhaadak80:
Resolves: #3154

## Problem

On package docs pages, exported symbols whose types are built from intersection / conditional / mapped / tuple constructs rendered as the literal string `unknown` — e.g. https://npmx.dev/package-docs/trslate/v/1.6.4 showed:

```html
constructor(schema: T, arg_1: unknown)
type SKey<T> = unknown[unknown]
```

even though the package's `index.d.ts` contains proper types.

## Root cause

`formatType()` in `server/utils/docs/format.ts` implemented recursive formatters for only 10 of the ~21 `TsType` kinds that `@deno/doc@0.189.1` emits. Unhandled kinds fell back to `type.repr`, but deno_doc returns an **empty string** for `repr` on structured types (intersections, mapped/conditional types, type literals — visible in our own fixtures), so the final fallback produced `unknown`.

This is the same failure mode previously fixed for #1411 by adding formatters (`fnOrConstructor`, `typeLiteral`, `indexedAccess`, `typeOperator`) — this PR extends the same pattern to the remaining kinds.

## Fix

- Added formatters for: `intersection`, `tuple`, `parenthesized`, `rest`, `optional`, `typeQuery`, `conditional`, `mapped`, `importType`, `infer`, `typePredicate`.
- Extended the literal formatter to handle `bigInt` and template literals.
- Arrays now parenthesize union/intersection/conditional/function-typed element types (`(A | B)[]` instead of invalid `A | B[]`); intersections parenthesize union/conditional/function members.
- Extended the hand-rolled `TsType` interface in `shared/types/deno-doc.ts` with the missing fields, matching the flat serde shape of @deno/doc 0.189.1 (`js/types.d.ts`).

With the fix, the trslate signatures render as:

```
constructor(schema: T, args: K & T[length]): Translation<T>
```

## Tests

Added a regression suite in `test/unit/server/utils/docs/format.spec.ts` mirroring trslate's real d.ts shapes:

```
Test Files  1 passed (1)
     Tests  14 passed (14)
```

(8 new tests covering intersections, conditionals, mapped types, indexed access over mapped/intersection members, tuples + rest, typeof queries, import types, infer, type predicates, bigInt/template literals, and an end-to-end function signature.)

### DIFF
--- a/server/utils/docs/format.ts
+++ b/server/utils/docs/format.ts
@@ -90,41 +90,40 @@ export function formatType(type?: TsType): string {
 const TYPE_FORMATTERS: Partial<Record<TsType['kind'], (type: TsType) => string>> = {
   keyword: type => type.keyword || '',
   literal: type => {
-    if (!type.literal)
-      return ''
-    if (type.literal.kind === 'string')
-      return `"${type.literal.string}"`
-    if (type.literal.kind === 'number')
-      return String(type.literal.number)
-    if (type.literal.kind === 'boolean')
-      return String(type.literal.boolean)
-    if (type.literal.kind === 'bigInt')
-      return `${type.literal.string}n`
+    if (!type.literal) return ''
+    if (type.literal.kind === 'string') return `"${type.literal.string}"`
+    if (type.literal.kind === 'number') return String(type.literal.number)
+    if (type.literal.kind === 'boolean') return String(type.literal.boolean)
+    if (type.literal.kind === 'bigInt') return `${type.literal.string}n`
     if (type.literal.kind === 'template')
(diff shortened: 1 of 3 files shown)


gameroman:
Please read our contribution guide fully

https://contributing.npmx.dev/

PR #3203, opened 2026-08-24
feat(ui): paginate the timeline around the selected version

antfu:
**This PR is implemented with help of agents**.

### 🧭 Context

Continue #3192

Right now, the timeline feature ignores the currently selected version. for example, for Nuxt v3.0.0 (https://npmx.dev/package-timeline/nuxt/v/3.0.0), we have:

CleanShot 2026-08-24 at 14 58 15@2x

which only list the latest versions, that have no direct relationships with `v3.0.0`. "Load more" is also not feasible as `v3.0.0` is too long ago to navigate.

This PR implemented a two-way "load more" feature, by default to navigate the range that contains the selected version:

CleanShot 2026-08-24 at 15 01 06@2x

### DIFF
--- a/app/pages/package-timeline/[[org]]/[packageName].vue
+++ b/app/pages/package-timeline/[[org]]/[packageName].vue
@@ -75,35 +75,48 @@ const sort = usePermalink<TimelineSort>('sort', 'semver')
 // server-side so pagination totals and pages already exclude pre-releases.
 const stableOnly = useTimelineStableOnly()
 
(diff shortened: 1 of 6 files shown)


graphieros (commented):
LGTM overall.
Just one small reserve, about the scroll into view when using the "Load newer versions" button; it seems a bit harsh for the user to have to find where they were in the timeline before the click (if this makes sense?).

PR #3208, opened 2026-08-26
fix: update vite-plus to 0.3.0

BittuBarnwal7479:
### 🔗 Linked issue

Resolves #<issue-number>

### 🧭 Context

The Nuxt browser tests fail on Windows when loading a route under `~[username]`.

The issue comes from the version of Vite used by the current `vite-plus` release. `vite-plus@0.2.9` uses Vite `8.2.1`, which has the old Windows path handling.

### 📚 Description

This PR updates `vite-plus` from `0.2.9` to `0.3.0`.

The newer version includes Vite `8.2.2`, which fixes the Windows path issue with `~` in paths.

The affected test was:

`test/nuxt/composables/use-command-palette-commands.spec.ts`

Specifically:

`adds npm account commands and disconnect support when npm is connected`

Before this update, the test failed while loading:

`app/pages/~[username]/orgs.vue`

After updating `vite-plus`, the test passes without needing the `server.fs.strict: false` workaround.

### Testing

- Reproduced the failure on Windows with `vite-plus@0.2.9`
- Updated to `vite-plus@0.3.0`
- Verified the affected test passes
- Verified the Nuxt test suite passes

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -34,14 +34,14 @@ catalogs:
       version: 10.1.1
   vite-plus:
     '@vitest/browser-playwright':
-      specifier: 4.1.10
-      version: 4.1.10
+      specifier: 4.1.11
+      version: 4.1.11
     '@vitest/coverage-v8':
-      specifier: 4.1.10
-      version: 4.1.10
+      specifier: 4.1.11
+      version: 4.1.11
     vite-plus:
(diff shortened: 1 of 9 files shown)


BittuBarnwal7479:
Duplicate : [3205](https://github.com/npmx-dev/npmx.dev/pull/3205)

PR #3214, opened 2026-08-28
feat(ui): add "dependencies" tab and integrate Health Insights into Search results

### LINKED ISSUE #29: add package dependencies tab with grouped view

Mopsgamer:
### 🔗 Linked issue

Resolves #29

### 🧭 Context

The dependencies section in the main tab's right column was too crowded and basic. To better track package health, it needs its own dedicated tab with sorting, filtering, and health insights.

### 📚 Description

This PR introduces a new Dependencies tab <kbd>y</kbd> for packages at `/package-deps/{{package}}/v/{{version}}` (also accessible via `/package-deps/{{package}}`) and removes the old dependencies section from the main tab's right column.

Added health to the Search page.

All tooltips for icons are accessible on mobile devices.

Dependencies Cards

Dependencies Table

image

image

image

---

Stats

---

Updated Search Cards glob

Updated Search Cards request

Updated Search Cards uuid spinner

Updated Search Table

Updated Search Table spinners

Dependencies provide Health Insights section, toolbar and the list of dependencies, similar to Search page Cards/Table.

Health Insights are visible on Dependencies and Stats tab pages. They are also clickable "OR" filters; every insight has 2 numbers: urgent (prod and bundled dependencies) and total (all - urgent + dev, peer, optional). They show updatable, vulnerable, deprecated and replaceable dependencies. Insights' header element provides this information in the `(i)` tooltip.

#### Disadvantages

- Cards and Rows do not provide "latest" links (they lead to a specific version). The only exception is the version component, which opens a tooltip without any redirection.


gameroman:
Instead of a new tab I'd maybe do it the same way the versions page is done?

https://main.npmx.dev/package/nuxt/versions

Mopsgamer:
> Instead of a new tab I'd maybe do it the same way the versions page is done?
> 
> https://main.npmx.dev/package/nuxt/versions

We have the "versions" drop-down on every page, but for dependencies we don't have anything. @gameroman

graphieros (changes requested):
Looks great!
A horizontal scroll is present on the insights section.
The main table also shows a horizontal scroll, perhaps there is a way for this page to use more of the available horizontal real estate ?

Mopsgamer:
Cards/Table switcher is causing hydration problems (or whatever causes them). The issues are with table rows: they get the green-text style applied for ranges but the class definition is missing (`/package-deps/nuxt/v/3.0.0?section=devDependencies`), and skeletons show the correct package name but link to the wrong package.

Also, both search page and dependencies show cards for a split second before switching to table when refreshing with tables enabled.

graphieros (changes requested):
The layout for the insights is better now :)
You might want to address the code rabbit feedbacks.

app/components/Dependencies/List.vue
```diff
@@ -0,0 +1,115 @@
...
+    })
+}
+
+function focusElement(el: HTMLElement) {
+  el.focus()
+  el.scrollIntoView({ block: 'nearest', behavior: 'smooth' })
+}
+
+function focusSearchInput() {
+  const searchInput = document.querySelector<HTMLInputElement>('input[type="search"], #deps-filter')
```
graphieros: This selector looks a bit fragile. Would it be possible to provide the focus method directly from the component where the input is located, and inject it here in this file ?

graphieros (dismissed):
It's a very cool feature:)
Just a suggestion: having insights possibly function as additional filters for the list would make it easier to locate related items, instead of scrolling.

app/components/Dependencies/List.vue
Mopsgamer: https://github.com/npmx-dev/npmx.dev/issues/3216

Mopsgamer:
Enough :)

Mopsgamer:
Bad hydration
image

no dependencies on first load
image

also this
image

and missing protocol parsing functionality
image

graphieros (changes requested):
Noticed some inconsistent behavior with the insights (here with Nuxt):

Image

- Some metrics are clickable but should not (patch behind 0 / 3)
- Some metrics are invalid (replacements available 1 / 0, should be 1 / 1)

Other remarks:
- The insights act as additional filters, and only do so on the dependencies page, which is great. I would recommend to add extra information in the i tooltip of the dependency health block when on the dependencies page to indicate that metrics can act as additional filters, otherwise the feature might be too confidential.

Mopsgamer:
Dependency list has problems with RTL languages (cards and table).

Mopsgamer:
Updated the description with the newest screenshots. It is good to go in my opinion, unless there are HTML spec, code style, or performance issues you want to address.

PR #3220, opened 2026-08-31
perf: prevent source maps from being resolved in package docs

### LINKED ISSUE #2739: getDocNodes() handles sourceMappingURL poorly, causes bad requests to esm.sh
### Describe the bug

`@deno/doc/doc` also passes `sourceMappingURL` specifiers to the resolver. If these are relative then they won't be prefixed with ".", which causes the `createResolver` callback to treat them as absolute paths ([see source](https://github.com/npmx-dev/npmx.dev/blob/c2c2754a7760378e0ce6d4df8ab9930d1a45c9a6/server/utils/docs/client.ts#L145-L149)).

Example (https://esm.sh/effect-web-midi@0.2.1/dist-types/index.d.ts):
~~~
export * from './src/index.d.ts';
//# sourceMappingURL=index.d.ts.map

~~~
causes a request to be made to https://esm.sh/index.d.ts.map which results in a 400 response.

### Additional context

I haven't investigated how the result of `getDocNodes` gets processed. If source maps have no value, then it may be desirable to simply skip any specifier that ends in `.map`.

### Logs

```shell-script

```

golden-fox07:
### 🔗 Linked issue

Resolves: #2739

### 🧭 Context

- `index.d.ts.map` was getting treated as a package name because it didnt start with .
- This was causing a lot of failed requests to esm.sh  (143 out of 440)

### 📚 Description

- `.map` references are now skipped by the resolver
- I checked if resolving them normally would work but esm.sh doesn't serve the `.d.ts.map` files anyway
- Added tests for the source map case
- The docs output stays the same and the unnecessary requests are gone

### Open Question
 I put the check in the resolver since thats where the misclassification happens. It relies on `createLoader` ignoring non http(s) urls though
 - would you prefer an explicit `.map` skip in the loader too?


trueberryless (approved):
Thank you for the awesome PR, @golden-fox07!

The code looks really solid overall and I can approve your changes. I took the liberty to also add check to the loader to make sure edge-cases like e.g. `https://esm.sh/something.map` do not get returned by the resolver and then fetched by the loader and I added some tests for it as well!

Now this looks ready to me and I'll wait for another maintainer to review the PR 👍

ghostdevv (commented):
How come we're attempting to fetch the sourcemaps in the first place? Is this something that `@deno/doc` is doing, and if so can we disable it? Are there places in which we do want to fetch the sourcemap(s)? Just wanna make sure we're fixing it in the appropriate place - if so then LGTM!

trueberryless:
> How come we're attempting to fetch the sourcemaps in the first place? Is this something that `@deno/doc` is doing, and if so can we disable it? Are there places in which we do want to fetch the sourcemap(s)? Just wanna make sure we're fixing it in the appropriate place - if so then LGTM!

`deno_graph` (which is used by `@deno/doc`) needs it to provide accurate stack traces in the module graph. See [this issue](https://redirect.github.com/denoland/deno_graph/issues/146). And there is [no option to disable this](https://docs.rs/deno_graph/latest/deno_graph/source/trait.Loader.html), so yes, this is the right place to handle this 👍

ghostdevv (approved):

Issue #2010, opened 2026-03-09
Aliased dependencies are not linked correctly


ericcornelissen:
### Describe the bug

The npmx interface links to the wrong package when a package specifies a dependency through an alias (e.g., `"foo": "npm:bar@1.2.3"` in package.json) in the "Dependencies" section of the sidebar. See for example `string-width-cjs   npm:string-wi...` in the preview below:

<img width="1031" height="741" alt="Image" src="https://github.com/user-attachments/assets/001d78d9-b348-4166-b465-406cdf1ea5c9" />

which currently links to <https://npmx.dev/package/string-width-cjs> but should link to <https://npmx.dev/package/string-width> per <https://npmx.dev/package-code/@isaacs/cliui/v/8.0.2/package.json#L53>.

### Additional context

Example:

- For <https://npmx.dev/package/@isaacs/cliui/v/8.0.2> the `*-cjs` dependencies link to the wrong package and should instead link to the aliased package.
- Compare that to <https://www.npmjs.com/package/@isaacs/cliui/v/8.0.2?activeTab=dependencies> which only links to the correct dependencies[^1].

[^1]: this appears to be a quite recent change because I definitely remember them linking to the alias-named package not too long ago. See also <https://snyk.io/blog/exploring-extensions-of-dependency-confusion-attacks-via-npm-package-aliasing/>.

### Logs

```shell-script
n/a
```

MatteoGabriele:
Correct me if I'm wrong, but it should also be flagged as 4 versions behind, right? 
If string-width v5 shows a 3 major versions behind warning, then string-width-cjs, pointing to string-width v4, should indicate 4 major versions behind.

ericcornelissen:
You are correct @MatteoGabriele. My guess for why it currently doesn't show that is because it uses the full alias string as the version and probably gets confused. So rather than not flagging it, it just doesn't know the version.

To be more explicit w.r.t. this than the original report, beyond linking to the right package it should also show (and link to, etc.) the version being used rather than the full alias string.

maxchang3:
Quick note: in [`Dependencies.vue`](https://github.com/npmx-dev/npmx.dev/blob/main/app/components/Package/Dependencies.vue) and related files, dependency values are currently assumed to always be version ranges. In reality, according to [`npm-package-arg`](https://github.com/npm/npm-package-arg), a dependency can be:

* a Git repo
* a tag / a specific version / a version range
* a local file or directory 
* a remote URL
[...]

PR #3234, opened 2026-09-08
fix: resolve npm aliases in dependency version links

fengmk2:
The `vite` version link on `vite-plus@0.3.1` returns `404` because `packageRoute()` treats the full `npm:` alias as a version.

`packageRoute()` now uses `parsePackageSpec()` to extract the target package and version. The link opens `/package/@voidzero-dev/vite-plus-core/v/0.3.1`. The fix covers regular, peer, and optional dependency version links.

Regression tests cover scoped and unscoped aliases, ranges, tags, and aliases without versions. All 56 focused tests, lint, formatting, and type checks pass.

Related to #2010. This draft overlaps with #2192 and limits the change to version-link routing.


fengmk2:
duplicate of #2192

PR #3238, opened 2026-09-09
feat: iojs and nodejs merge noodle

alexdln:
### 🧭 Context

14 September 2015 - IO.js and Node.js merge
[Node.js Foundation Combines Node.js and io.js Into Single Codebase](https://nodejs.org/en/blog/announcements/foundation-v4-announce)

image
image


graphieros (approved):
Love it ❤️

PR #3248, opened 2026-09-13
feat: draft menu redesign

alexdln:
### 🧭 Context

A mostly generated PR with the idea of ​​a header redesign - the header feels overloaded to me now. For now, this is a rough draft with some minor refactors. If the idea seems viable, I'll make a new one with a clean implementation

image
image


alexdln:
_We can't really decide on the format yet, so it's hanging in stale for now. I'm thinking about a more universal solution_

PR #3252, opened 2026-09-14
fix: focus the search input through the owning component instead of a DOM query

### LINKED ISSUE #3216: Fragile SearchBox selector queries
Here is original function:
https://github.com/npmx-dev/npmx.dev/blob/16988d8d1b1d79bccfa4a1c5c7b21a440ddd324f/app/pages/search.vue#L483-L488

And this one:
https://github.com/npmx-dev/npmx.dev/blob/16988d8d1b1d79bccfa4a1c5c7b21a440ddd324f/app/app.vue#L67-L74

They should use SearchBox's (already exposed) `useTemplateRef('searchBoxRef').focus` method.

Ref: https://github.com/npmx-dev/npmx.dev/pull/3214#discussion_r3889350009

Usage examples:

https://github.com/npmx-dev/npmx.dev/blob/16988d8d1b1d79bccfa4a1c5c7b21a440ddd324f/app/components/AppHeader.vue#L177

https://github.com/npmx-dev/npmx.dev/blob/16988d8d1b1d79bccfa4a1c5c7b21a440ddd324f/app/components/AppHeader.vue#L187

Chirag6722:
resolves #3216

### Context

The `/` shortcut in `app.vue` and the ArrowUp handler in `pages/search.vue` found the search input with a `document.querySelector`. That input lives in two other components, so a markup change there would silently break both shortcuts.

### Description

Added a small `useSearchInputFocus` composable. `app.vue` provides it, `Header/SearchBox.vue` and `pages/index.vue` register their own input, and the two handlers call the injected `focus()` instead of querying the DOM. The header box returns `false` when its input is not rendered (homepage), so the homepage input is used there. If nothing is registered, `/` still falls back to `router.push({ name: 'search' })`.

Behaviour is unchanged. The existing mobile quirk (header input hidden by CSS, so focus does nothing) is left as is; happy to follow up on it.

Tests: 7 cases in `test/nuxt/composables/use-search-input-focus.spec.ts` plus one e2e case for `/` on the homepage. They fail if I remove the registrations or the cleanup. Lint, fmt, nuxt tests and the search e2e tests pass locally; `test:types` I could only verify on CI.


graphieros (changes requested):
@Chirag6722 
As you know, if you read our [contribution guide](https://contributing.npmx.dev/), the use of AI to assist with coding is perfectly acceptable, as long as it is human driven.
However, using your own voice is required when communicating with npmx contributors.
Please consider writing the PR description with your own words. It can be in your native language if it is more practical.

Thank you^^

PR #3257, opened 2026-09-16
fix: sync org page filters and columns to the url

### LINKED ISSUE #2818: bug: org page filters and selected columns state not reflected in URL

szymonchudy:
### 🔗 Linked issue

Fixes #2818

### 🧭 Context

Org page views are supposed to be shareable. Sort already landed in the URL, but filters and selected columns did not, so copying the address bar lost that state.

### 📚 Description

On the org page only:

- Search scope, download range, security, and updated-within go in the query (`scope`, `downloads`, `security`, `updated`). Missing or invalid values use the code default. Nothing here is read from `localStorage`.
- Visible columns go in `columns` (`name` is always visible, so it is not listed on purpose). If that param is present, this visit uses it and does not write `localStorage`. If it's missing, we use localStorage (or the default columns).
- A query param is only written when the value is not the default. `/org/nuxt` still means the default view. Changing a control back to default drops its param.
- View mode, page size, and pagination stay out of the URL (not part of the issue)
- Shared filter/preference composables are unchanged (search uses them).

Side note: I didn't touch Vue since early v2, so feel free to point out non-idiomatic solutions. That said, I mostly replicated existing patterns. 

### Preview

https://github.com/user-attachments/assets/738bab67-d4f8-4a5f-b171-5aefd657ee4a


graphieros (approved):
Very cool 😎

PR #3258, opened 2026-09-17
fix: align package selection checkbox

szymonchudy:
### 🧭 Context

While working on #3257 I've noticed that the package selection checkbox appears slightly vertically misaligned when it becomes visible on row hover in the organization package list.

#### Before
CleanShot 2026-09-17 at 06 34 43@2x

#### After
CleanShot 2026-09-17 at 06 38 16@2x

### 📚 Description

Use flex alignment on the checkbox label so the native checkbox is centered within its label, rather than following the browser's default inline text baseline.


graphieros (approved):
LGTM 🌿

PR #3262, opened 2026-09-18
feat: install-block redesign and update

alexdln:
### 🧭 Context

Another experiment is updating the style of the installation block. There's too much talk these days about how much clutter we have on the page, especially for this block, so I'm trying to reduce its size

Also fixed a hydration error when a package manager is selected. The error was related to a degradation of the original behavior - https://github.com/npmx-dev/npmx.dev/pull/393/changes

image
image


alexdln:
_finally_

PR #3264, opened 2026-09-18
feat: use `--before` to show old versions in package timeline

nicolo-ribaudo:
### 🔗 Linked issue

None -- we briefly discussed this in Discord, I can create one if needed.

### 🧭 Context

Today we released a new version of Babel (8.0.6) that halved the package size. I went to npmx.dev's timeline view to get my daily dopamine dose, and unfortunately I saw that... Babel 8.0.6 was almost the same size as 8.0.5 :/

image

The reason is that `@babel/core@8.0.5` depends on multiple other Babel packages; so it was now in the stats pulling the 8.0.6 version of them making it look like as if 8.0.5 was already small.

### 📚 Description

This PR introduces a behavior similar to pnpm's `--before`, making sure that a package's computed size reflects the latest state of the registry _before that the next version was released_. This is different from taking a snapshot when a given version is released, so that the shown size reflects what was the _actual_ total size when that old version was still the current one.

A couple notes:
- I also kept the old logic of _not_ using `--before` and instead resolving to whatever are the latest dep versions in range. I think this is important for the "Install Size" information in a package's main page, this:
  image
  that's because even if I'm looking at an older version, that number is mostly useful to answer the question "how big would it be if I were to add this version now to my dependencies?". On hover it then  also shows the `--before`-frozen size:
  image
- Given that I was keeping that logic around anyway, I left a "frozen history" toggle in the timeline view (enabled by default). When it's disabled it restore the old behavior, showing how big an old version is if I were to install it today: 
  image
  image
- The "Package size reduced!" and "Package size increased!" banners use the _historical_ size, since that's useful for people that are updating their dependency and thus are comparing the size at the time they installed the old one with the size they would get now by installing the updated one.

----

I used an LLM to help with this; all the production code has been cleaned up / rewritten by me, but the tests are in large part auto-generated.


nicolo-ribaudo:
It's working for me locally, but not in the PR preview. Any idea of what I might need to look at to figure out why?

PR #2975, opened 2026-06-29
fix: passthrough of query

johnnyreilly:
### 🔗 Linked issue

The hope is that this fix will make the change implemented in https://github.com/npmx-dev/npmx.dev/pull/2800 work when npmx is deployed. That fix worked locally, but not when live.

### 🧭 Context

For reasons that I don't fully understand (still a nuxt noob) the query options were not being passed through.

### 📚 Description

This PR changes the nuxt config in a way that I *think* will mean it passes through what we need and things should start to work. I hope.  I don't know how to test this though.  Working locally isn't enough I know; is there a way to push this to a preview environment or similar?  Dunno.  cc @gameroman

PR #3269, opened 2026-09-20
feat: code redirects - more npmjs.com URL parity

johnnyreilly:
### 🔗 Linked issue

This PR is related to https://github.com/npmx-dev/npmx.dev/issues/2781 - it's a follow on

### 🧭 Context

Consider these pull requests: https://github.com/npmx-dev/npmx.dev/pull/2800 and its follow on https://github.com/npmx-dev/npmx.dev/pull/2975 (as I missed something in the initial PR)

These implemented versions redirects for npmx.dev to bring parity with npmjs.com URLs. This PR does a similar piece of work - this time regarding code view instead.  

If you go to this URL you will see the code view of a package, eg: https://www.npmjs.com/package/webpack?activeTab=code

But if you go the npmx equivalent you will **not**: https://www.npmx.dev/package/webpack?activeTab=code

### 📚 Description

As with the versions view, we would like to 302 and redirect. 

So https://www.npmx.dev/package/webpack?activeTab=code should redirect to
https://npmx.dev/package-code/webpack/v/5.111.1 where 5.111.1 is the latest version of a package.  

This part is harder than the previous approach which did not require knowing the latest package version.  Fortunately the middleware has something in place that seems to support this.

You can validate this works by going to the preview deployment here: https://npmx-43vih5w43-npmx.vercel.app/package/webpack?activeTab=code which redirects to https://npmx-43vih5w43-npmx.vercel.app/package-code/webpack/v/5.111.1

image

Likewise https://npmx-h96tfboc4-npmx.vercel.app/package/webpack/v/4.47.0?activeTab=code redirects to https://npmx-h96tfboc4-npmx.vercel.app/package-code/webpack/v/4.47.0

image


43081j:
I wonder if it should be a 301 or a 308? Since it is permanently redirected rather than temporarily redirected. 

Do you know any reason we should treat it as temporary?

johnnyreilly:
I had forgotten about 308!  

That said, having had a quick read I'll confess to finding the differences between [301 Moved Permanently](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/301) and [308 Permanent Redirect](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/308) smallish.  As I understand it 301's can turn a POST into a GET and 308s will not.  So 308s seem nicer, though I'm not sure that particular benefit makes much difference for this use case. 

For the sake of "I don't think I've ever used that status code" if nothing else, I've switched!

43081j:
yup as far as i understand, 301 rewrites and 308 redirects (i.e. in 308 your method is carried over)

doesn't make much difference here other than semantics i guess

johnnyreilly:
I thank you for the status code education though - I feel I'm forever encountering nuances!

PR #3276, opened 2026-09-21
fix: support link tag for pipe syntax

kazupon:
### 🔗 Linked issue

N/A

### 🧭 Context

`parseJsDocLinks` (server/utils/docs/text.ts) parses `{@link} `tags using a regex that only splits target and label on whitespace:

```js
/\{@link\s+([^\s}]+)(?:\s+([^}]+))?\}/g
```

This doesn't account for the TSDoc/TypeDoc `{@link target | label}` pipe syntax https://tsdoc.org/pages/tags/link/, which many packages use in their JSDoc comments. When a comment uses this form, the | character is captured as part of the label and leaks into the rendered link text

- e.g. `{@link Command | entry command}` renders as | entry command instead of entry command.
- reproduction: https://npmx.dev/package-docs/gunshi/v/0.37.3

### 📚 Description

Updated the `{@link}` parsing regex in `parseJsDocLinks` to treat `|` as an explicit label delimiter, in addition to the existing whitespace-only form:

```diff
- result.replace(/\{@link\s+([^\s}]+)(?:\s+([^}]+))?\}/g, (_, target, label) => {
+ result.replace(/\{@link\s+([^\s|}]+)(?:\s+(?:\|\s*)?([^}]+))?\}/g, (_, target, label) => {
```

The target capture group now excludes | so it isn't swallowed into the target name.
Backward compatible: `{@link target}`, `{@link target label}` (space-only), and `{@link target | label}` (pipe) all continue to work as before/expected.


kazupon:
ops, I'll update unit test cases.

